PR #320, opened 2014-11-11
Refactor config methods

aledsage:
Creates a BrooklynObject.config(), and deprecates the crazy number of inconsistent config accessor/setter methods we have on Entity/Location/Policy/Enricher.

Note that `AbstractEntity`, `AbstractLocation` and `AbstractEntityAdjuct` all still have their own implementation of config. Until we delete the additional deprecated access/setter methods, it will be difficult to rationalise this down into a single implementation within `AbstractBrooklynObject`. I suggest we do that as soon as we begin work on 0.8.0-SNAPSHOT.

Please merge https://github.com/apache/incubator-brooklyn/pull/319 first (for the groovy version upgrade to 2.3.4). Without that, it fails to compile with `BrooklynObject.config()` returning `ConfigurationSupport` and `BrooklynObjectInternal.config()` returning the sub-class `ConfigurationSupportInternal`.

When reviewing, please focus on the two commits `Add BrooklynObject.config()` and `Deprecate *Config methods in Location/Enricher/Policy`. The other commits contain a massive number of boring changes to make things use the new `config().set(key, val)` rather than the old `setConfig(key, val)` etc.


aledsage:
@grkvlt I've incorporated the review comments - could you look again please?

sjcorbett:
@aledsage Could you resolve the merge conflicts?

aledsage:
@grkvlt @sjcorbett @ahgittin - I've finally rebased this, so is good for review and hopefully merge. Could you take a look please?

p.s. by changing the `Configurable` interface, I suspect it will impact  the Clocker downstream project that makes use of it. @grkvlt will you handle that once this is merged please?

sjcorbett:
I've been through f05a6db9540c7a6a339838f0e31acf3d1ee72810 and dce1384731def10e42328849e2f1b795bad0042a and have no further comments to add to @grkvlt's from before but for two minor documentation points.

aledsage:
I've incorporated comments from @sjcorbett and then rebased against master.

Only comment not addressed is from @grkvlt about `<T> T set(ConfigKey<T> key, T val)`: "Should we support setting an Object value instead and doing type coercion if possible?". Let's look at that in a separate PR if we choose to do it. I'm not sure yet either way.

Will merge now.

PR #328, opened 2014-11-14
Swift object store - don't use application/directory markers

neykov:
Softlayer object store doesn't like the content type - can't delete the folders even from its own UI.

The PR is meant to trigger a discussion on the problem more than for merging right away.

### DIFF
--- a/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsUtil.java
+++ b/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsUtil.java
@@ -282,7 +282,8 @@ public class JcloudsUtil implements JcloudsLocationConfig {
 
         ContextBuilder contextBuilder = ContextBuilder.newBuilder(provider).credentials(identity, credential);
         contextBuilder.modules(MutableList.copyOf(JcloudsUtil.getCommonModules())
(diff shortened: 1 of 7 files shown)


sjcorbett:
@neykov Was there any discussion? May be worth writing to the mailing list instead.

aledsage:
@neykov I agree with @sjcorbett - is this a bug we should raise with jclouds? We can fix it in our code base if necessary (and look to get it into jclouds longer term), but would be good to have a bug report in jclouds if that is appropriate.

neykov:
It's not a bug in jclouds code. More like an implementation inconsistency in SL's objectstore.

If I remember correctly calling setContentType(directory) would make it impossible to delete the folder even through SL's UI. I need to look into the failing tests before getting this further.

ahgittin:
close this PR for now?  possibly needs to be taken up with SL to understand what the recommended approach is, and then implement that either here or in jclouds.

neykov:
For persistence we don't need folder markers. Instead of using explicit folder objects, just drop them altogether.

neykov:
Looks like SL fixed their implementation as now it's possible to delete the folders and download change.log & master. This makes the changes in this PR obsolete.

PR #351, opened 2014-11-20
Adds entitlement check to state checks

nakomis:
Resolves BROOKLYN-82


ahgittin:
looks fine to me.  a test would be nice but it's so small i don't think we should hold it up.  merging.

PR #356, opened 2014-11-24
Adds boolean ConfigKey.isNonNull() method

sjcorbett:
Use to enforce non-null values for particular config keys.

Validity checks are made in `AbstractEntity.setConfigInternal`, so `setConfig(required, null)` should fail, and in `InternalEntityFactory` after setting config values on a newly-created entity, so forgetting to set a value on a required key should also fail. 

The `InternalEntityFactory` check will only work for `ConfigKeys` defined on the interfaces implemented by the entity. I think the combination of the two checks is sufficient.

### DIFF
--- a/core/src/main/java/brooklyn/entity/basic/AbstractEntity.java
+++ b/core/src/main/java/brooklyn/entity/basic/AbstractEntity.java
@@ -949,7 +949,9 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
 
     @SuppressWarnings("unchecked")
     private <T> T setConfigInternal(ConfigKey<T> key, Object val) {
-        if (!inConstruction && getManagementSupport().isDeployed()) {
+        if (key.isNonNull() && val == null) {
+            throw new IllegalArgumentException("Null value for required key " + key);
+        } else if (!inConstruction && getManagementSupport().isDeployed()) {
             // previously we threw, then warned, but it is still quite common;
             // so long as callers don't expect miracles, it should be fine.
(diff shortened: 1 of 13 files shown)


ahgittin:
@sjcorbett i still think it is too confusing having `nonNull()` verified at deployment time for entities.  people routinely set config early in the start sequence but after deployment. if the `EntityFactory` check were remove i think we'd be fine, and we are clear that the meaning of `nonNull` is that a `get(key)` call never returns null.

i'm okay with merging it once the deploy-time check is removed.  (that could be refactored to a utility method available on demand.)

however if you have bandwidth or there is not a pressing need for this now my preference would still be to apply a `ConfigValidation` scheme instead of this impl, and to do this now rather than introduce the api proposed here which i think we'll deprecate.  a `ConfigValidation` scheme could have a `NON_NULL_ON_DEPLOY` or `NON_NULL_ON_USE` and we can cleanly do deploy-time checks.  it could then be expanded with things like `MATCHES_REGEX` and `IN_RANGE` etc.

this could follow the `ConfigInheritance` pattern in https://github.com/apache/incubator-brooklyn/pull/483 .

(with apologies for the delay in replying, and for the merge conflicts that are now present)

sjcorbett:
Thanks for your thoughts. I got as far as a draft to the dev list previously. I'm keen to keep the deploy-time check in some form - it's annoying to wait 15 minutes for your cloud to provision an instance and only to then fail because you forgot to set some config that's required by `customize`. I'll close this and work towards the `ConfigValidation` idea.

PR #362, opened 2014-11-26
Fix - legacy catalog items versioning

neykov:
* Correctly load java classes from the catalog (legacy mechanism), taking into account versioning precedence.
* Tag the resulting entities with the catalogItemId used for loading.
* Fix the catalog link in rest response.
* Fix the UI link to catalog items for applications.


neykov:
Moved `getPathFromVersionedId` to `brooklyn.rest.util.WebResourceUtils` and rebased.

PR #366, opened 2014-11-27
Interface to be used by CAMP, TOSCA, etc. application description implementors

neykov:
* The implementors will register themselves with brooklyn.
* To instantiate an app from an unknown spec type all registered providers will be tried until success.

Implementations for CAMP & TOSCA to come as separate PRs.


sjcorbett:
@neykov Anything more to add to this? Are there any implementations?

neykov:
Implemented in https://github.com/apache/incubator-brooklyn/pull/436 with modifications to suit the code requirements.

PR #370, opened 2014-12-01
Skip dry run on node upgrade.

neykov:
Add parameter to choose whether to launch a dry run node or to upgrade directly.
Allow extending classes to customize the effector parameters by overriding a method.

Depends on https://github.com/apache/incubator-brooklyn/pull/413

### DIFF
--- a/software/base/src/main/java/brooklyn/entity/brooklynnode/effector/BrooklynNodeUpgradeEffectorBody.java
+++ b/software/base/src/main/java/brooklyn/entity/brooklynnode/effector/BrooklynNodeUpgradeEffectorBody.java
@@ -18,7 +18,9 @@
  */
 package brooklyn.entity.brooklynnode.effector;
 
+import java.util.Collection;
 import java.util.Map;
+import java.util.NoSuchElementException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -26,6 +28,7 @@ import org.slf4j.LoggerFactory;
 import brooklyn.config.ConfigKey;
 import brooklyn.entity.Effector;
 import brooklyn.entity.Entity;
+import brooklyn.entity.basic.ConfigKeys;
 import brooklyn.entity.basic.Entities;
 import brooklyn.entity.basic.EntityInternal;
 import brooklyn.entity.basic.EntityTasks;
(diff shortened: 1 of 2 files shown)


neykov:
Failure looks unrelated, local tests pass. Rebased to force a rebuild.

neykov:
Forcing another update, local tests passing again - what's with the failures...

neykov:
Failure could be caused by https://github.com/apache/incubator-brooklyn/pull/413, investigating...

neykov:
Test failure should be fixed in https://github.com/apache/incubator-brooklyn/pull/482 (not included in this PR). Rebasing again to kick another build.

software/base/src/main/java/brooklyn/entity/brooklynnode/effector/BrooklynNodeUpgradeEffectorBody.java
```diff
@@ -132,44 +193,34 @@ public Void call(ConfigBag parametersO) {
...
+    }
 
-        DynamicTasks.waitForLast();
-        Entities.unmanage(dryRunChild);
+    protected EntitySpec<? extends BrooklynNode> createDryRunSpec() {
+        return EntitySpec.create(BrooklynNode.class);
+    }
 
-        return null;
+    protected SshMachineLocation getSshMachineLocation() {
```
ahgittin: remove this method; you can use `Locations.findUniqueSshMachineLocation(entity().getLocations()).get()`

software/base/src/main/java/brooklyn/entity/brooklynnode/effector/BrooklynNodeUpgradeEffectorBody.java
```diff
@@ -132,44 +193,34 @@ public Void call(ConfigBag parametersO) {
...
-                String runDir = entity().getAttribute(SoftwareProcess.RUN_DIR);
-                String bkDir = Urls.mergePaths(runDir, "..", Urls.getBasename(runDir)+"-backups", dryRunNodeUid);
-                String dryRunDir = Preconditions.checkNotNull(dryRunChild.getAttribute(SoftwareProcess.RUN_DIR));
-                log.debug(this+" storing backup of previous version in "+bkDir);
-                DynamicTasks.queue(SshEffectorTasks.ssh(
-                    "cd "+runDir,
-                    "mkdir -p "+bkDir,
-                    "mv * "+bkDir,
-                    "cd "+dryRunDir,
-                    "mv * "+runDir
```
ahgittin: looks like we lose the two lines above?  i think that's fine, but wanted to check.

it means the `runDir` contents will have to be repopulated from the (already downloaded) `installDir`, costing perhaps a few extra seconds of downtime.  should be easy enough to add these two statements back in the case where dryRun has just been done if we wanted to.  but it does feel a bit dodgy.  no strong feelings...

ahgittin:
LGTM, minor comments, then can merge

software/base/src/main/java/brooklyn/entity/brooklynnode/effector/BrooklynNodeUpgradeEffectorBody.java
neykov: Removed it altogether, couldn't find any usages (even in subclasses).

software/base/src/main/java/brooklyn/entity/brooklynnode/effector/BrooklynNodeUpgradeEffectorBody.java
neykov: If I remember correctly, this was done on purpose to force the customize step and regenerate the local-brooklyn.properties. Definitely worth a comment.

neykov:
Addressed comments.

ahgittin:
i did wonder whether that method was used at all!  makes sense about the customize step.  merging.

PR #383, opened 2014-12-10
Improve recursion checks for catalog items

neykov:
* `catalog.createSpec` now passes the symbolicName of the item so resolving to items with the same symbolicName already in the catalogue is prevented (fallback to java types only).
* the recursion check was using the type as is, thus making it possible to recurse into items with the same symbolicName where version is explicitly specified. Explicitly forbid catalog items referencing same symbolicName items, even if with another version.
* stop catalog items whose symbolicName matches the java type they export leaking their bundles to callers.

### DIFF
--- a/core/src/main/java/brooklyn/camp/brooklyn/api/AssemblyTemplateSpecInstantiator.java
+++ b/core/src/main/java/brooklyn/camp/brooklyn/api/AssemblyTemplateSpecInstantiator.java
@@ -21,11 +21,16 @@ package brooklyn.camp.brooklyn.api;
 import io.brooklyn.camp.CampPlatform;
 import io.brooklyn.camp.spi.AssemblyTemplate;
 import io.brooklyn.camp.spi.instantiate.AssemblyTemplateInstantiator;
+
+import java.util.Set;
+
 import brooklyn.entity.proxying.EntitySpec;
 import brooklyn.management.classloading.BrooklynClassLoadingContext;
 
 public interface AssemblyTemplateSpecInstantiator extends AssemblyTemplateInstantiator {
 
     EntitySpec<?> createSpec(AssemblyTemplate template, CampPlatform platform, BrooklynClassLoadingContext loader, boolean autoUnwrapIfAppropriate);
+    EntitySpec<?> createNestedSpec(AssemblyTemplate template, CampPlatform platform, BrooklynClassLoadingContext itemLoader, Set<String> encounteredCatalogTypes);
+
     
 }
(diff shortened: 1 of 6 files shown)


sjcorbett:
@neykov Can you resolve the merge conflicts?

neykov:
@sjcorbett Rebased on latest master, conflicts are resolved.

ahgittin:
this sounds reasonable; however it looks like this only applies at the level of the `AssemblyTemplate` aka `Application`.  will we have the same problem when resolving a `Component` aka `Entity`, e.g. when referenced inside an app, or using "Add Child" in the GUI ?

also i'm surprised that it is safe to remove so much of `BrooklynComponentTemplateResolver. tryLoadEntityClass()` -- are you sure about that?

might be worth a few more comments in here to explain things.  (ideally we'd do a major refactor so we don't need comments but i don't think that's a small job...)

neykov:
Presently all instantiations of CAMP YAML are done at the AssemblyTemplate level, no? Even "Add Child" creates an app and then strips it to get the entity. The same `encounteredCatalogTypes` set is used for the all components in the YAML so it shouldn't matter how deep a recursive reference is. As for instantiating entitites within Java code, then this is not applicable as the fix is for breaking catalog item cycles. It's highly probable that I dind't fully understand your comments, so feel free to correct my missunderstanding :).

The change improves how standalone catalog items are instantiated. This only applies when creating an `EntitySpec` from the catalog items directly (when adding them to the catalog to check the app type, when listing the catalog items from REST). When instantiating catalog items as part of an application template the `encounteredCatalogTypes` is already properly initialized.

Creating a spec out of a catalog item directly was treating the catalog yaml as an application, starting any exclusion in `encounteredCatalogTypes` one level below them. If the user was creating catalog items with `id` matching the `javaType` and there was already a catalog item with the same `symbolicName` in the catalog (i.e. when updating catalog items) then we would get the following structure
    * new catalog item
      * old catalog item
        * java type resolved from old catalog item

With the change in place we get:
    * new catalog item
      * java type resolved from new catalog item

The difference is that now we initialize `encounteredCatalogTypes` in `BasicBrooklynCatalog` forcing the code to ignore existing catalog items with the same `id`.


---

Re `tryLoadEntityClass`. With the following in the catalog:
```
brooklyn.catalog:
  id: sample.java.Type
  libraries:
  - http://simple.jar
services: 
  - type: sample.java.Type
```

and then updaing the catalog with

```
brooklyn.catalog:
  id: sample.java.Type
  libraries:
  - http://invalid-url-or-libraries-missing.jar
services: 
  - type: sample.java.Type
```

Where the second catalog item doesn't have any `libraries` or they don't have `sample.java.Type` in them - resolving `sample.java.Type` in the second catalog item would be done with the `OsgiBrooklynClassLoadingContext` of the first item, so it would successfully resolve the java class.

ahgittin:
okay - i think that makes sense.  thanks for the detailed explanation.

PR #389, opened 2014-12-12
BrooklynNode - unmanage on stop, fix abort logic

neykov:
* Allow entities to be able to abort their stop effector. If pre-stop is aborted the machine will still be released.
* Unmanage BrooklynNode on stop

WARNING: Turns out that postStop wasn't wired up, it was never called. Nevertheless some entities (i.e. CouchDBNode, NginxController, JavaWebAppSoftwareProcess) override it which might introduce unwanted behaviour. Shall I remove the override in existing entities?

### DIFF
--- a/software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
+++ b/software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
@@ -111,7 +111,15 @@ public class BrooklynNodeImpl extends SoftwareProcessImpl implements BrooklynNod
     protected void preStart() {
         ServiceNotUpLogic.clearNotUpIndicator(this, SHUTDOWN.getName());
     }
-    
+
+    @Override
+    protected void preStopConfirmCustom() {
+        super.preStopConfirmCustom();
+        if (Boolean.TRUE.equals(getAttribute(BrooklynNode.WEB_CONSOLE_ACCESSIBLE))) {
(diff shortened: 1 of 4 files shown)


software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
```diff
@@ -111,7 +118,15 @@ public void init() {
     protected void preStart() {
         ServiceNotUpLogic.clearNotUpIndicator(this, SHUTDOWN.getName());
     }
-    
+
+    @Override
+    protected void preStopConfirmCustom() {
+        super.preStopConfirmCustom();
+        if (Boolean.TRUE.equals(getAttribute(BrooklynNode.WEB_CONSOLE_ACCESSIBLE))) {
```
sjcorbett: I think @grkvlt means:
```
isSet(AttributeSensor<Boolean>(true)) --> true
isSet(AttributeSensor<Boolean>(false)) --> false
isSet(AttributeSensor<Boolean>(null)) --> false
```

sjcorbett:
The changes here look ok. The Nginx entity's `postStop` method does nothing useful. The other three all do:
```
        setAttribute(REQUESTS_PER_SECOND_LAST, 0D);
        setAttribute(REQUESTS_PER_SECOND_IN_WINDOW, 0D);
```
I'd be inclined to keep them.

@grkvlt Any more comments?

PR #394, opened 2014-12-15
Return null SubscriptionContext when management stopped

grkvlt:
Check for null conext when calling `SubscriptionTracker` methods and allow unsubscribes with null context (harmless) but error otherwise.

### DIFF
--- a/api/src/main/java/brooklyn/management/SubscriptionContext.java
+++ b/api/src/main/java/brooklyn/management/SubscriptionContext.java
@@ -31,31 +31,34 @@ import brooklyn.event.SensorEventListener;
  * This is the context through which an {@link Entity} can manage its subscriptions.
  */
 public interface SubscriptionContext {
+
     /**
      * As {@link SubscriptionManager#subscribe(Map, Entity, Sensor, SensorEventListener)} with default subscription parameters for this context
      */
     <T> SubscriptionHandle subscribe(Map<String, Object> flags, Entity producer, Sensor<T> sensor, SensorEventListener<? super T> listener);
- 
+
     /** @see #subscribe(Map, Entity, Sensor, SensorEventListener) */
     <T> SubscriptionHandle subscribe(Entity producer, Sensor<T> sensor, SensorEventListener<? super T> listener);
-    
+
     /** @see #subscribe(Map, Entity, Sensor, SensorEventListener) */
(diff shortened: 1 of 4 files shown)


ahgittin:
+1 @aledsage -- `NonDeploymentSubscriptionContext` would be simple to write and would mean most of the null checks introduced here aren't needed.  follow the pattern of the other `NonDeployment*` classes.

ahgittin:
@grkvlt bump

grkvlt:
@ahgittin I'll change this to use a `NonDeploymentSubscriptionContext `

grkvlt:
Question for @ahgittin or @aledsage - should I implement the same pattern as other `NonDeploment`_Blah_ classes, saving the inital management context like this:

```Java
    @Override
    public <T> SubscriptionHandle subscribe(Entity producer, Sensor<T> sensor, SensorEventListener<? super T> listener) {
        if (isInitialManagementContextReal()) {
            return initialManagementContext.getSubscriptionContext(subscriber).subscribe(producer, sensor, listener);
        } else throw nonDeploymentContextError();
    }
```

Also, what about a `NonDeploymentSubscriptionManager` to replace the `QueueingSubscriptionManager` as well?

grkvlt:
@ahgittin If there are nofurther issues with this right now, I'll merge?

ahgittin:
With this isn't it likely that when an entity is stopped someone might get the subscription manager populated with a non-deployment context where the `initialManagementContext` is the original context?  And so calls to `subscribe` and `unsubscribe` will do something when we don't want them to?

Granted the code here has grown to be unwieldy -- the original intention was to allow a queuing context before the node is managed, which comes to life on management.  But removing fail-fast behaviour doesn't seem like a good idea -- apart from for `unsubscribe` where it seems fine.

Does it make sense to change `isInitialManagementContextReal()` to `isInitialManagementContextUsable()` and in there put a check `if (mode==NonDeploymentManagementContextMode.MANAGEMENT_STOPPED) return false;` ?

And I think you're right, we should also guard `getSubscriptionManger()` with a `STOPPED` check, throwing or returning a non-deployment.

ahgittin:
@grkvlt Any thoughts on the above?  We should update or close this PR!

grkvlt:
@ahgittin I had forgotten about this, but the intent is still useful. I suggest not closing this yet, until I have had a look at the code again, to see if it can be updated and made mergeable.

grkvlt:
Closing this, to revisit later if necessary

PR #403, opened 2014-12-17
support URL for the keystore supplied to brooklyn web server

ahgittin:
(since we do call it keystore url!)

### DIFF
--- a/usage/launcher/src/main/java/brooklyn/launcher/BrooklynWebServer.java
+++ b/usage/launcher/src/main/java/brooklyn/launcher/BrooklynWebServer.java
@@ -157,6 +157,9 @@ public class BrooklynWebServer {
     private String sslCertificate;
 
     @SetFromFlag
+    private String keystoreUrl;
+
+    @SetFromFlag @Deprecated /** @deprecated use keystoreUrl */
     private String keystorePath;
 
     @SetFromFlag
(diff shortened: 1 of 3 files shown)


usage/launcher/src/main/java/brooklyn/launcher/BrooklynWebServer.java
```diff
@@ -364,19 +367,28 @@ public synchronized void start() throws Exception {
...
             SslContextFactory sslContextFactory = new SslContextFactory();
 
-            if (keystorePath==null) keystorePath = managementContext.getConfig().getConfig(BrooklynWebConfig.KEYSTORE_URL);
+            // allow webconsole keystore & related properties to be set in brooklyn.properties
+            if (Strings.isNonBlank(keystorePath)) {
+                if (keystoreUrl==null) {
+                    log.warn("Deprecated 'keystorePath' used; callers should use 'keystoreUrl'");
+                    keystoreUrl = keystorePath;
+                } else if (!keystoreUrl.equals(keystorePath)) {
+                    log.warn("Deprecated 'keystorePath' supplied with different value than 'keystoreUrl', preferring the latter: "+
```
sjcorbett: If both `keystoreUrl` and `keystorePath` are given I think it would be better to throw rather than to log and ignore the latter. The warning message will be missed, it will not be obvious which property will be used when reading property files and we will end up with confused users.

aledsage:
Agree with @sjcorbett - better to throw exception during startup. Other than that, looks good.

ahgittin:
disagree.  if someone sets `path` and `url` to `X` and someone elase composes with that blueprint it should be sufficient for them to set just `url` to `Y` and have that take effect.  they shouldn't have to also go and set the deprecated parameter.

this is what warnings are for -- use and abuse of deprecated parameters, when there is clearly a dominant setting.

okay to merge?

sjcorbett:
What is the entry point for the scenario you describe? Deploying a `BrooklynNode`?

ahgittin:
@sjcorbett `BrooklynNode` could be one way, or a custom launcher.  unlikely to be a real issue, if you still feel it should fail fast i'll change it.

sjcorbett:
I guess it boils down to whether we want users to get exceptions now or at some point in the future when the deprecated property is removed. I'd prefer the former but am not going to fuss if you know of users it would affect immediately.

ahgittin:
they wouldn't get exceptions in the future would they?  it's just that if they supply different `url` and `path` values i say `url` should trump with a warning now, you say it should fail now.  in the future when `path` is removed `url` will trump with no warning or exception, `path` will just be ignored.

this is a general case of we deprecate property `A` and replace it with property `B`.  what happens if `A` and `B` are supplied and disagree.  failing in this case seems harsh it means anyone interacting with a legacy library which sets `A` would themselves have to use `A` in order to make it work.  better they just use `B` and get a warning about `A` being used by a legacy library.

sjcorbett:
You're right with regard to behaviour. I think I'd just confused myself earlier.

Not sure I follow your second paragraph. Either way, my original suggestion had standalone-server startup in mind. It's not that harsh throwing in that case because it'll happen within seconds of running `brooklyn launch`. I agree that throwing is less desirable if a `BrooklynNode` entity is launched as part of a larger deployment. I defer to you based on your opinion of how likely the latter case is.

ahgittin:
k thx - merging

PR #411, opened 2014-12-19
MongoDB and Node fixes

sjcorbett:
Various fixes and improvements for MongoDB entities. Generally involves `BrooklynAccessUtils` and subnet addresses.

Also ensures the NPM modules are on `NODE_PATH` when running Node entitites.


neykov:
I am getting the following error in the logs when the tests try to start mongodb:
```
Please make at least 422MB available in /tmp/brooklyn-svet/apps/htTbOmQm/entities/MongoDBServer_QQirTLu2/data/journal or use --smallfiles
```

getting it with the following command:
```
mvn test -PIntegration -Dtest=MongoDbBlueprintTest#testMongoSingleServer
```

Before starting the tests I have 600mb free on the drive, 257mb after the test fails. smallfiles = true is already set in the mongodb.conf file.

sjcorbett:
That's a pain. I'll have a look. Can you check that the test passes on master for you?

nakomis:
Changes look sensible. I ran the tests with `mvn test -Dtest=MongoDbBlueprintTest -PIntegration` and they passed. LGTM

neykov:
False warning, the problem above is caused only  by the insufficient disk space. I can confirm that smallfiles is used.

software/webapp/src/main/java/brooklyn/entity/webapp/nodejs/NodeJsWebAppSshDriver.java
```diff
@@ -152,6 +152,8 @@ public void launch() {
             appCommandLine = appCommand + " " + appFile;
         }
 
+        // Ensure global NPM modules are on Node's path.
+        commands.add("export NODE_PATH=\"$NODE_PATH:$(npm root -g)\"");
```
neykov: Looks like it's bad practice to rely on global modules, but not much to do if it's needed by real-world apps.
https://docs.npmjs.com/misc/faq#i-installed-something-globally-but-i-can-t-require-it

neykov:
LGTM.

software/webapp/src/main/java/brooklyn/entity/webapp/nodejs/NodeJsWebAppSshDriver.java
sjcorbett: I think we could rework the Node entity to install packages locally. Something for the future.

sjcorbett:
@nakomis @neykov Thanks.

PR #413, opened 2014-12-19
Release the underlying machine when stopping BrooklynNode

neykov:
The shutdown effector was setting a STOP state which prevented the machine from being released.
Changes:
  * Convert the shutdown effector as a utility functionality - just call the remote shutdown endpoint with the passed parameters
  * Store effector parameters in the EffectorCallTag, methods calls to fetch them. Passing the parameters along the call chain would cause backwards incompatible changes.
  * Pass stop parameters to the shutdown effector so that it's functionality can be fine tuned by calling stop directly
  * Don't unamange the node if the machine is not released.
  * Add logic to wait for the graceful shutdown of the process before killing it forcibly (the java process will take some time to shut down cleanly after closing the rest connection)

Depends on https://github.com/apache/incubator-brooklyn/pull/389.


neykov:
@sjcorbett Thank you for the feedback, renamed `updateEffectorParameters` to `setEffectorParameters`

PR #414, opened 2014-12-22
Entitlements

ahgittin:
add per-user entitlements and documentation for entitlements


core/src/main/java/brooklyn/management/entitlement/PerUserEntitlementManager.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public void addUser(String user, EntitlementManager managerForThisUser) {
+        perUserManagers.put(user, managerForThisUser);
+    }
+
+    @Override
+    public <T> boolean isEntitled(EntitlementContext context, EntitlementClass<T> entitlementClass, T entitlementClassArgument) {
+        EntitlementManager entitlementInEffect = null;
+        if (context==null || context.user()==null) {
+            // no user means it is running as an internal process, always has root
```
aledsage: I'd prefer if we could use some special context to indicate this, rather than null. Can we intercept earlier, e.g. in `Entitlements.getEntitlementContext()` then don't return null.

The reason is that treating null as super-user feels risky as the behaviour if there is a programming error which would otherwise cause a NPE!

core/src/main/java/brooklyn/management/entitlement/PerUserEntitlementManager.java
```diff
@@ -0,0 +1,97 @@
...
+        log.info(getClass().getSimpleName()+" created with "+perUserManagers.size()+" user"+Strings.s(perUserManagers)+" and "
+            + "default "+defaultManager+" (users: "+perUserManagers+")");
+    }
+    
+    public PerUserEntitlementManager(EntitlementManager defaultManager) {
+        this.defaultManager = Preconditions.checkNotNull(defaultManager);
+    }
+
+    public void addUser(String user, EntitlementManager managerForThisUser) {
+        perUserManagers.put(user, managerForThisUser);
```
aledsage: Is it sensible to accept nulls, or should we `put(checkNotNull(user, "user"), checkNotNull(managerForThisUser, "managerForThisUser"))`?

core/src/test/java/brooklyn/management/entitlement/AcmeEntitlementManagerTest.java
```diff
@@ -21,132 +21,38 @@
...
-    @AfterMethod(alwaysRun=true)
-    public void tearDown() {
-        Entitlements.clearEntitlementContext();
-        if (app != null) Entities.destroyAll(app.getManagementContext());
-        if (mgmt != null) Entities.destroyAll(mgmt);
-        app = null;
-        mgmt = null;
+    @Test
+    public void testAnonUserHasAllPermissions() {
+        checkUserHasAllPermissions("anon");
```
aledsage: Why does "anon" have all permissions? I don't see "anon" defined in `AcmeEntitlementManager`.

core/src/test/java/brooklyn/management/entitlement/AcmeEntitlementManagerTestFixture.java
```diff
@@ -0,0 +1,157 @@
...
+import brooklyn.util.exceptions.Exceptions;
+
+public abstract class AcmeEntitlementManagerTestFixture {
+
+    ManagementContext mgmt;
+    Application app;
+    ConfigBag configBag;
+    
+    public void setup(ConfigBag cfg) {
+        mgmt = new LocalManagementContextForTests(BrooklynProperties.Factory.newEmpty().addFrom(cfg));
```
aledsage: Why not do this in `init` (i.e. in the `@BeforeMethod`)?

aledsage:
Looks good; only very minor comments - happy for this to be merged once you've looked through those.

core/src/main/java/brooklyn/management/entitlement/PerUserEntitlementManager.java
ahgittin: Agree, but that's a major change, beyond the scope of this PR

core/src/main/java/brooklyn/management/entitlement/PerUserEntitlementManager.java
ahgittin: done

core/src/test/java/brooklyn/management/entitlement/AcmeEntitlementManagerTest.java
ahgittin: `root` was the default. have changed `anon` to `otherAuthorizedUser` so it's less surprising

core/src/test/java/brooklyn/management/entitlement/AcmeEntitlementManagerTestFixture.java
ahgittin: not all tests need an `app`, and i think this way the config order is clearer

ahgittin:
addressed comments.  however there are conflicts with the docs restructuring, so will hang file until #458 is in (which has a placeholder for the docs mentioned here)

PR #424, opened 2014-12-23
Fixes to make_javadoc.sh script

rdowner:
Three changes:

1. `BROOKLYN_VERSION` annotation to ensure that 0.7.0-SNAPSHOT reference is updated.
2. Javadoc output spews out a lot of noise to stdout and errors/warnings to stderr, but use of `tee` saves only stdout noise and not the useful error messages - changed to send errors to stdout, and what was stdout goes to /dev/null.
3. Abort if the required brooklyn-all-dependencies jar file has not been built.

### DIFF
--- a/docs/_build/make-javadoc.sh
+++ b/docs/_build/make-javadoc.sh
@@ -16,11 +16,19 @@ else
   export SOURCE_PATHS=$BROOKLYN_JAVADOC_SOURCE_PATHS
 fi
 
+mkdir -p target
 rm -rf target/$JAVADOC_TARGET1_SUBPATH/
 
 export DATESTAMP=`date "+%Y-%m-%d"`
 # BROOKLYN_VERSION_BELOW
 export BROOKLYN_JAVADOC_CLASSPATH=../../usage/all/target/brooklyn-all-0.7.0-SNAPSHOT-with-dependencies.jar
+
+if [ \! -f ${BROOKLYN_JAVADOC_CLASSPATH} ]; then
+  echo "Expected to find ${BROOKLYN_JAVADOC_CLASSPATH}"
+  echo "Please run a full Maven build in the project root"
+  exit 1
+fi
+
 echo "building javadoc at $DATESTAMP from:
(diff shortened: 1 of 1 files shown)


docs/_build/make-javadoc.sh
```diff
@@ -19,16 +19,24 @@ export DATESTAMP=`date "+%Y-%m-%d"`
...
   -d target/apidocs/ \
   -subpackages "org.apache.brooklyn:io.brooklyn:brooklyn" \
-  -classpath ../../usage/all/target/brooklyn-all-0.7.0-SNAPSHOT-with-dependencies.jar \
+  -classpath "${deps_jar}" \
   -doctitle "Apache Brooklyn" \
   -windowtitle "Apache Brooklyn" \
   -header "Apache Brooklyn" \
   -footer '<b>Apache Brooklyn - Multi-Cloud Application Management</b> <br/> <a href="http://brooklyn.io/" target="_top">brooklyn.io</a>. Apache License. &copy; '$DATESTAMP'.' \
- | tee target/javadoc.log
+2>&1 1>/dev/null | tee target/javadoc.log
```
sjcorbett: Can you also make sure `target` exists? If it doesn't the output isn't redirected as expected and the subsequent `grep` isn't grepping anything.

rdowner:
Rebased against master (resolves merge conflict) and addressed @sjcorbett's comment

ahgittin:
will use and test this along with #438 before pushing #439 and #440

ahgittin:
lgtm, merging (very similar to something i pushed to master, but have reverted in favour of this; this is better anyway with a check that it has been built!)

ahgittin:
meh there are merge conflicts so this might have to be closed manually -- will see

ahgittin:
ha ha the merge conflicts are because i reverted and meanwhile you merged.  mea maxima culpa.  i will sort it all out.

ahgittin:
looks like asfgit did sort it out :)

PR #425, opened 2014-12-23
Fixes SERVICE_UP timeout error in tests

neykov:
Fixes "Caused by: java.lang.IllegalStateException: Timeout waiting for SERVICE_UP from MyServiceImpl" frequently sabotaging the windows CI build.
MyServiceImpl is setting SERVICE_UP manually while the enrichers toggle it in parallel in response to the SERVICE_STATE_ACTUAL changes. Sometimes the enricher will trigger in the window between MyServiceImpl setting the value to true and calling waitForServiceUp, leading to the timeout error.


sjcorbett:
The test failure is repeatable locally. @neykov for your convenience the test is ScriptHelperTest.testCheckRunningForcesInessential.

neykov:
Fixed the test - no need to check for SERVICE_UP as it's a derivative of SERVICE_PROCESS_IS_RUNNING.

PR #427, opened 2014-12-24
Couchbase SyncGateway should support cluster of ONE node

fatuhoku:
`IS_IN_CLUSTER` [is ONLY set when a server is explicitly added into a cluster](https://github.com/apache/incubator-brooklyn/blob/59208f56a264d85d68ccc51bc9f2acf457dc1107/software/nosql/src/main/java/brooklyn/entity/nosql/couchbase/CouchbaseClusterImpl.java#L461-L462).

The original logic checks for 'IS_IN_CLUSTER'; this means that the sole primary node of a single-node cluster will never be eligible to be used with sync gateway (even though it actually is...!).

We extend the boolean condition to check for this condition.

### DIFF
--- a/software/nosql/src/main/java/brooklyn/entity/nosql/couchbase/CouchbaseSyncGatewaySshDriver.java
+++ b/software/nosql/src/main/java/brooklyn/entity/nosql/couchbase/CouchbaseSyncGatewaySshDriver.java
@@ -92,10 +92,10 @@ public class CouchbaseSyncGatewaySshDriver extends AbstractSoftwareProcessSshDri
 
                 @Override
                 public boolean apply(@Nullable Entity entity) {
-                    if (entity instanceof CouchbaseNode && Boolean.TRUE.equals(entity.getAttribute(CouchbaseNode.IS_IN_CLUSTER))) {
-                        return true;
-                    }
-                    return false;
+                    // Must either be recognised by a cluster as added, or be the primary node in a cluster of ONE.
+                    return entity instanceof CouchbaseNode
+                      && (Boolean.TRUE.equals(entity.getAttribute(entity.IS_IN_CLUSTER))
+                      || Boolean.TRUE.equals(entity.getAttribute(entity.IS_PRIMARY_NODE)));
                 }
             });
             if (cbClusterNode.isPresent()) {
@@ -172,5 +172,4 @@ public class CouchbaseSyncGatewaySshDriver extends AbstractSoftwareProcessSshDri
             return arch + fileExtension;
         }
     }
-
 }


software/nosql/src/main/java/brooklyn/entity/nosql/couchbase/CouchbaseSyncGatewaySshDriver.java
```diff
@@ -92,10 +92,10 @@ public void launch() {
...
                 @Override
                 public boolean apply(@Nullable Entity entity) {
-                    if (entity instanceof CouchbaseNode && Boolean.TRUE.equals(entity.getAttribute(CouchbaseNode.IS_IN_CLUSTER))) {
-                        return true;
-                    }
-                    return false;
+                    // Must either be recognised by a cluster as added, or be the primary node in a cluster of ONE.
+                    return entity instanceof CouchbaseNode
+                      && (Boolean.TRUE.equals(entity.getAttribute(entity.IS_IN_CLUSTER))
+                      || Boolean.TRUE.equals(entity.getAttribute(entity.IS_PRIMARY_NODE)));
```
sjcorbett: Replace the 'entity' in `entity.IS_IN_CLUSTER` and `entity.IS_PRIMARY_NODE` with `CouchbaseNode`, otherwise this will not compile.

sjcorbett:
@fatuhoku Thanks for your contribution. Have you committed to an Apache project before? If you haven't you will need to fill out a [contributor's license agreement](http://www.apache.org/licenses/#clas) before we can merge any changes. It's a fairly lightweight thing.

ahgittin:
@fatuhoku was there previously a case where the `entity.IS_PRIMARY_NODE` but not `entity.IS_IN_CLUSTER` ?  that sounds wrong.

just in case it was right, i've merged your change (fixing the minor compile error).

if you could complete a CLA as per @sjcorbett that would be appreciated.  it's not strictly needed for things this small but easier if it is there.

BTW check out update web site on brooklyn.io -- i think you'll find it friendlier than it was

ahgittin:
@fatuhoku aha i now see the case where `IS_IN_CLUSTER` isn't being set.  thanks for your code highlight when you opened the PR.  i'l open a new PR to fix that, maybe you could look at it?

ahgittin:
@fatuhoku should be fixed in https://github.com/apache/incubator-brooklyn/pull/488

fatuhoku:
@ahgittin Hi thanks so much for the fix!  I'm not really using Couchbase at the moment so please feel free to merge it as long as everything makes logical sense.

PR #431, opened 2014-12-31
Public site fixes

neykov:
* Update links
  * Offset anchors so they are visible below the fixed header

There is still some problem with the breadcrumbs (i.e. https://brooklyn.incubator.apache.org/v/latest/use/examples/webcluster/index.html), though locally they work fine, so couldn't troubleshoot it - could it be just a problem with the last generation?


sjcorbett:
Looks good. @asfbot failure unrelated: `org.apache.maven.InternalErrorException: Internal error: java.lang.IllegalStateException: Invalid object ID 28 iota=55`

PR #436, opened 2015-01-07
[WIP] Pluggable YAML parsers

neykov:
Rebases https://github.com/apache/incubator-brooklyn/pull/100 against master (as of beginning of December). The updates are needed for the TOSCA implementation.


neykov:
Failure unrelated, merge conflict resolving needed anyway.

grkvlt:
# Question

Is there a way to extend this with a mechanism to enable pluggable _extensions_ to YAML parsers. For example, I am happy with using the CAMP parser, but I want to allow the YAML files to be able to include service types like `docker:`_imageId_`:`_tag_ instead of Java class names. The parsing would be part of an external library, in this instance [Clocker](http://clocker.io/), since the entities created for these service types would be defined there, and core Brooklyn has no way of referencing the classes driectly.

neykov:
This should be easy to achieve using the same approach (ServiceLoader), by replacing the hard-coded resolvers in `io.brooklyn.camp.brooklyn.spi.creation.BrooklynComponentTemplateResolver.Factory.computeResolverType`. It's unrelated to this PR though as it's on another level of abstraction.

andreaturli:
@ahgittin do you think we should try to rebase this PR? I guess for things like TOSCA may really be the way to go? Any suggestions?

neykov:
Closing in favour of https://github.com/apache/incubator-brooklyn/pull/760.

PR #438, opened 2015-01-07
Documentation link check improvements

sjcorbett:
Swaps the link-checker gem for html-proofer, which can check [quite a lot more](https://github.com/gjtorikian/html-proofer#whats-tested). Incorporated the tool into `build.sh` and added option to skip with `--skip-tests`.

Also replaces all uses of "brooklyn" with "Brooklyn".

### DIFF
--- a/docs/Gemfile
+++ b/docs/Gemfile
@@ -5,5 +5,5 @@ source 'https://rubygems.org'
 
 gem 'jekyll'
 gem 'therubyracer'
-gem 'link-checker'
+gem 'html-proofer'
 gem 'json'
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -1,9 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    anemone (0.7.2)
-      nokogiri (>= 1.3.0)
-      robotex (>= 1.0.0)
+    addressable (2.3.6)
     blankslate (2.1.2.4)
     celluloid (0.16.0)
       timers (~> 4.0.0)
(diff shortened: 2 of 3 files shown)


sjcorbett:
Replaces brooklyn with Brooklyn comment no longer valid as subsequent merged pull requests cause a whole lot of conflicts.

ahgittin:
awesome!  let me try this before i update the web site (tonight)

sjcorbett:
Just a warning that you'll get _a lot_ of warnings! It's worth considering whether we want the tool enabled by default.

ahgittin:
thanks for the warning

i like the idea of this

unfortunately it has lead (for me) to ruby dependency hell; `bundle install` seemed to work, but then running gave:

    dyld: lazy symbol binding failed: Symbol not found: _rb_funcall2

i then tried the instructions [here](http://judykat.com/ken-judy/force-bundler-rebuild-ruby-rails-gemset/) to cleanly reinstall the gems but now installing nokogiri is giving:

    libxml2 is missing.  Please locate mkmf.log to investigate how it is failing.

i'm testing tips [here](http://www.nokogiri.org/tutorials/installing_nokogiri.html) and [here](http://stackoverflow.com/questions/5528839/why-does-installing-nokogiri-on-mac-os-fail-with-libiconv-is-missing#comment30773476_6163052) but if you have any quick tips let me know

ahgittin:
fixed with `bundle config build.nokogiri --use-system-libraries` as per http://www.nokogiri.org/tutorials/installing_nokogiri.html

ahgittin:
this is *so* useful!

it would be good to configure it to ignore some "problems":
* `anchor has no href attribute` (that's allowed, if you want to link to it, isn't it?)
* `External link http://127.0.0.1:8081 failed: got a time out` (ha!)
* links to other versions (we could work around that by creating mocks)

but it's all just informational (doesn't fail the build) so we can merge as is i think

i'll do a bunch more cleanup based on the output of this (in a separate PR); it will make the docs setup much nicer -- great addition.

PR #439, opened 2015-01-07
Docs restructuring

ahgittin:
better structure of user guide and integration with main website

it's far from perfect, and note there is still some good content not included, under `_extras`, but i'm thinking this is now finally *better* than the docs currently online and we should merge without waiting too much longer

(we should of course flesh out the content a whole lot more, but with a decent structure and simpler menus i think this gives us the foundation for folks to improve it piece by piece)

i think we should then update online: 
* website build
* v-snapshot guide
* v-0.7.0-M2 guide
* v-latest guide

For the latter two I think we should backport docs changes to 0.7.0-M2 (in an M2-docs branch maybe?).

WDYT @richardcloudsoft @robertgmoss @sjcorbett ?


robertgmoss:
A few comments about specific sections

Getting Started
 * I like that the blueprints, catalog, and policies are now separate menu items.
 * The stuff about SSH, Installation, Launch Brooklyn, Configuring a Location could be in its own page so that blueprints is purely about launching applications from blueprints.
 * The monitoring and managing applications section of blueprints could be extracted to its own page.
 * Catalog doesn't explain what it is.
 * Catalog top level tab on web console should be explained 
 * Policies page doesn't really tell the user how to do anything e.g attach a policy.  
 * An 'explore the web console' page would be useful.

YAML
 * Deploying YAML blueprints page looks like it should be merged with getting started blueprints.

Java
  * There is a bit of an oddity in the menu.

rdowner:
@ahgittin I built the main website as if for publication. It seems that the menu bar items that cross-link to the user guide are linking to "/guide/..." instead of "/v/..."

ahgittin:
meh okay that's a bug, i'll fix it

ahgittin:
i see what's happening, when we exclude items from build we don't do url rewriting on those targets.  should be an easy fix.

ahgittin:
@rdowner good spot.  should be fixed now.  /cc @robertgmoss

rdowner:
@ahgittin, just tried building this - it looked fine. I then uploaded it to live - but the [user guide is a bit malformed](https://brooklyn.incubator.apache.org/v/0.7.0-SNAPSHOT/)...

ahgittin:
ouch, what's going on?

ahgittin:
aha:

         Build Warning: Layout 'website-normal' requested in guide/dev/index.md does not exist.

only when you do `_build/build.sh guide-latest`.  i will fix.

BTW a good way to debug this is to do `_build/serve-site.sh` after doing the above command

ahgittin:
aha looks like `exclude: /website` has the effect of excluding `_layouts/website-*.html`, fix imminent

ahgittin:
fixed.  @robertgmoss is addressing a few comments i had on #440, and after that unless he finds anything on this we will test and push the full monty, unless you @rdowner have any other comments.

(obviously if we find anything in the push-test we will address)

rdowner:
Thanks @ahgittin, I check this now. Happy to wait for #440 if you think that's best.

ahgittin:
great, merging this and #440 following successful testing of both with @robertgmoss

PR #441, opened 2015-01-10
Docs

ahgittin:
bunch more clean up -- going to push this to be live (as live site is not entirely working), but here for review


ahgittin:
/cc @richardcloudsoft @robertgmoss @sjcorbett 

the link checker is fantastic!

ahgittin:
I have already pushed this to brooklyn.io (as what was there was broken) so you can review that -- everything seems healthy.  Not perfect but this is the jumping off point for incremental improvements.

Note the style changes (larger font etc).  We should get a designer to tweak it further but I think it feels friendlier already.

/cc @richardcloudsoft @aledsage @duncanjw

rdowner:
Looks good to me - ready to merge - @ahgittin I will leave you to merge this if that's OK, so that what you merge matches what you have already published!

Guide and website are looking really good now!

PR #442, opened 2015-01-13
Catalog libs - fail if user specified symbolicName and version don't match bundle

neykov:
Bundles are always matched by their symbolicName and version, not by the user specified alternatives, so allowing them to diverge will lead to unexpected behaviour. So far the logic was to fail only if the bundle is already installed which was inconsistent with first-time install when it would pass.


sjcorbett:
Looks good.

aledsage:
Looks good to me too; merging.

PR #443, opened 2015-01-13
Split "community" to create new "developers" section

rdowner:
Creates a new top-level "developers" section of the website; moves some community content into it. Revamps organisation and some content in both sections.


ahgittin:
excellent!  i was thinking we needed something like this.

not sure if "devs" is worth a top-level link, or if it's nicer under "community", but this is a step in the right direction for sure

merging

i have some other changes so i suggest we hang fire on updating the live site for a few days...

PR #444, opened 2015-01-13
lots of tweaks to docs

ahgittin:
almost all cosmetic, but in a major way:
* new fonts
* right menu floats
* some pages have sections which show up in menus with auto-advance
* site map updated
* external links shown as such in the menus (with the menu rejigged a bit)
* all headings are links automatically, with link selector support a la github
* (octicons and font-awesome libraries added to achieve the above)


rdowner:
Lots of great stuff in here - looking very good!

A couple of minor points.

* [ ] The `link-external` icon to me has the connotation that the link opens in a new window. Possibly this would be a good thing to do anyway. Could I suggest that either we use `link-external` and have external links open in a different window, or use the ordinary `link` icon and have them open in the current window?
* [ ] I'm a bit concerned that we now have three webdings fonts - Bootstrap's bundled one, Octicons, and FontAwesome. Do we need all three - could we drop one or two of them? What's the "weight" of a page on a clean cache now?
* [ ] I'm not happy with the change to the warning text on the developer's page. We must not be promoting the casual use of development code (because it's not released) so we must warn away casual users. Changing the header from Danger to Caution is OK, but I believe that we *do* need the warnings about potential data loss because that *is* a risk of running SNAPSHOT code, however rare it may be.

A further minor point is to ask if you are also testing these changes on non-Retina displays? I am usually using a non-Retina monitor and the recent change to the Lato font looked OK but not brilliant. This morning I am using a Retina display and Lato looks brilliant so I can see why it was chosen! Lato is acceptable so I am not proposing to change it, but just wanted to make sure that you are also checking out non-Retina rendering. (On a related note, I think we must (**groan**) check out rendering on IE and other browsers, too...)

ahgittin:
from IRC discussion, i think we've agreed:

* font/js/bloat - keep an eye on this, have an optimization task as required, removing things and/or running an optimizer or using CDN sources
* dev warning text - the main thing is to caveat binary downloads; several items should be done
* other browsers + displays:  absolutely needs testing

will merge this now as discussed, and test on other browsers once i've done other changes

as for warning text, idea is:

* even softer caution for "dev" page entry
* strong original danger message on downloads page for snapshot items (e.g. using `{% if snapshot %}` markdown)
* strong caution message, perhaps collapsible div, when looking at index of old (pre-apache) versions (in `meta/versions.md`)
* reinstate the warning message that appears when looking at a user guide for something other than `/v/latest` or `/guide` (looks like i accidentally removed this?)
* modify that warning message to use stronger wording when looking at a snapshot version

rdowner:
+1. Good to merge.

PR #445, opened 2015-01-14
Jsgui only checks that the value of the location select is a location.

sjcorbett:


aledsage:
Thanks @sjcorbett - looks good; merging.

sjcorbett:
Ah, I didn't intend for 359bd1c to be merged. It does not work at all. Shall I revert it?

PR #446, opened 2015-01-14
Load resources from entities' catalog item loader.

neykov:


usage/camp/src/test/java/io/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityTest.java
```diff
@@ -362,6 +364,20 @@ public void testCreateSpecFromCatalogItem() {
...
+        String id = "resource.test";
+        addCatalogOSGiEntity(id, SIMPLE_ENTITY_TYPE);
+        String yaml =
+                "services: \n" +
+                "  - serviceType: "+ver(id);
+        Entity app = createAndStartApplication(yaml);
+        Entity simpleEntity = Iterables.getOnlyElement(app.getChildren());
+        InputStream icon = new ResourceUtils(simpleEntity).getResourceFromUrl("classpath:/brooklyn/osgi/tests/icon.gif");
+        assertTrue(icon != null);
+        icon.close();
```
sjcorbett: Close in a finally block?

neykov: Not really needed as there's nothing to throw between opening and closing it. `getResourceFromUrl` will throw if the resource is not found, but I've added the additional `assertTrue(icon != null);` just to illustrate intent.

sjcorbett: Ah, I was thrown by the illustration of intent.

neykov: :) Well the assert will throw only on null, so the close won't be needed in this case.

sjcorbett:
My comment was irrelevant so looks good. Merging.

PR #447, opened 2015-01-14
Standalone console dependency - redo

neykov:
Alternative approach to exclude brooklyn.war for projects which customize jsgui.


sjcorbett:
Confirmed that `brooklyn.war` is present again with a dependency on `brooklyn-all`.

PR #449, opened 2015-01-14
Fix/whitespace rest api

aledsage:
White space changes only - replaces the two-space indents with 4 space indents in the rest-api and rest-server java code.


sjcorbett:
Thanks. This has bugged me for _ages_.

PR #450, opened 2015-01-15
docs - three minor liquid language things which break in v.next of liquid

ahgittin:
discovered when testing incremental regen in https://github.com/jekyll/jekyll/pull/3116 --
means changes get picked up in 1s instead of 10s, using the next version of jekyll
(unfortunately not yet gemmed)


ahgittin:
closing -- this is included in #453 now

PR #453, opened 2015-01-15
more docs improvements

ahgittin:
add a nifty "blueprint tour" under learning, improve cross-browser functionality, and a few other bits and pieces


rdowner:
Looking very good, as always! Good to merge. Do you want me to merge it and update the website?

I'd avoid the use of colloquialisms, in-jokes etc. - like "this message will self-destruct in 3 seconds". There's a few of them in the docs (wordplay like "When the RAT bites"). I don't want to be seen to be policing fun but they will be confusing to those for whom English is not a first language and/or did not participate in US/UK pop culture, and generally take away from the very polished and professional appearance the website has gained recently.

ahgittin:
thanks - gonna change the javadoc to have only the month, so that we don't update every javadoc file on every update in the future, and fix a link andrea found

i'll replace the "self-destruct" ref with a "this message will go away in 5 seconds" ... personally i think the brooklyn site should have a bit of personality if possible but we'd have to agree on it

ahgittin:
have made the changes in #458 - merging this

PR #454, opened 2015-01-15
Fix StreamGobblerTest under Windows

neykov:


utils/common/src/test/java/brooklyn/util/stream/StreamGobblerTest.java
```diff
@@ -41,9 +43,9 @@ public void testGobbleStream() throws Exception {
         StreamGobbler gobbler = new StreamGobbler(stream, out, null);
         gobbler.start();
         try {
-            gobbler.join(10*1000);
+            gobbler.join(1000*1000);
```
aledsage: That's a very long wait? Can we make it more like 30 seconds, if that is what windows takes?  Should we make this in the `Integration` group if it's going to take many seconds?

aledsage:
oops; thanks for fixing @neykov - what do you think about the length of timeout, and whether we should put this into Integration so it doesn't make a normal build take even longer?

utils/common/src/test/java/brooklyn/util/stream/StreamGobblerTest.java
neykov: Oh, this slipped through, it was just to ease debugging. Reverted.

aledsage:
Merging now.

PR #455, opened 2015-01-16
Cobertura to use maven plugin rather than ant-run

duncangrant:
PLEASE DON'T MERGE YET
Seems to lose ability to output xml but does give us an aggregated report.

Also when run it seems to compile twice (presumably once without and then once with instrumentation) and you get warnings of the form:
WARNING: Duplicate name in Manifest: Manifest-Version.                                                                                  Ensure that the manifest does not have duplicate entries, and                                                                           
that blank lines separate individual sections in both your                                                                              
manifest and in the META-INF/MANIFEST.MF entry in the jar file.                                                                         

You seem to also get these warnings if you try to do an incremental build so I don't know if they're important.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,7 +113,7 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
         <!-- Testing -->
-        <cobertura.plugin.version>2.5.2</cobertura.plugin.version>
+        <cobertura.plugin.version>2.6</cobertura.plugin.version>
         <surefire.version>2.18</surefire.version>
         <plantuml.version>6121</plantuml.version>
         <ant.version>1.8.4</ant.version>
@@ -200,7 +200,6 @@
         <jsr311-api.version>1.1.1</jsr311-api.version>
         <maxmind.version>0.8.1</maxmind.version>
         <jna.version>4.0.0</jna.version>
-        <coverage.target>${working.dir}</coverage.target>
 
         <!-- Release -->
         <!-- no passphrase by default, so we can do automated deploy builds;
@@ -823,11 +822,6 @@
                     <groupId>org.codehaus.mojo</groupId>
                     <artifactId>build-helper-maven-plugin</artifactId>
                     <version>1.7</version>
-                </plugin>
-                <plugin>
-                    <groupId>org.codehaus.mojo</groupId>
-                    <artifactId>cobertura</artifactId>
-                    <version>${cobertura.plugin.version}</version>
                 </plugin>
                 <plugin>
                     <groupId>com.google.code.maven-replacer-plugin</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -824,11 +823,11 @@
                     <artifactId>build-helper-maven-plugin</artifactId>
                     <version>1.7</version>
                 </plugin>
-                <plugin>
+                <!-- <plugin>
```
neykov: Remove if not used.

usage/rest-server/pom.xml
```diff
@@ -246,7 +246,12 @@
             </resource>
         </resources>
         <plugins>
-
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <compilerId>javac</compilerId>
```
neykov: What's the reason for overriding the compiler here?

neykov:
I tried running the tests with `mvn clean install -PCoverage` and get the following error:
```
[ERROR] Failed to execute goal org.codehaus.mojo:cobertura-maven-plugin:2.6:instrument (default) on project brooklyn-parent: Execution default of goal org.codehaus.mojo:cobertura-maven-plugin:2.6:instrument failed: Plugin org.codehaus.mojo:cobertura-maven-plugin:2.6 or one of its dependencies could not be resolved: Could not find artifact com.sun:tools:jar:0 at specified path /System/Library/Java/JavaVirtualMachines/1.6.0.jdk/Contents/Home/../lib/tools.jar -> [Help 1]
```

duncangrant:
What java version are you using?  I'm using oracle jdk 1.7 - Alasdair thinks it might be to do with this.

neykov:
I am using Apple's Java 1.6.
Turns out that apple's build doesn't have tools.jar, it's merged with classes.jar. Cobertura depends on it so it fails.

Here's a related issue https://jira.codehaus.org/browse/MCOBERTURA-180.

Possible (temporary) solution: http://permalink.gmane.org/gmane.comp.java.maven-plugins.mojo.user/4320

duncangrant:
The suggested fix is quite messy.  Perhaps it would be better to just accept that we can't run coverage on a mac running 1.6?

neykov:
It's fine by me.

ahgittin:
dropping java1.6 support coverage sounds fine

this runs fine even without #534 (fixed problem with antrun cobertura); it must be using a different directory structure.

however not only does this compile twice (which is understandable at least in `/src/main` as we need the instrumented and non-instrumented versions of classes), this **runs** all the tests twice which seems bad; acceptable but a major slow-down.

do we have a way to run the coverage check with both unit and integration tests activated?  the latter is important especially for packages such as `ssh`.

@duncangrant do you suggest we merge this now (with the faults which have been noted here) or close this and you reopen it when ready?

ahgittin:
meant also to say:  it works well for me, on os x.  no `output.xml` as duncan g says.  i don't know if that matters (or indeed if the current antrun-based integration generates any such file).

duncangrant:
The current ant-run based integration does produce output.xml.  This is useful for integration with jenkins and if we want to produce reports from the output.  For this reason I am going to close rather than merge until I can investigate the issue.

I hadn't notice it running the tests twice before.  According to @grkvlt this is because cobertura runs the tests once to confirm they pass and then instruments them and runs them again.

I haven't ever managed to get the integration tests to run on my machine so I still don't know what the coverage will look like for those.  Presumably it is possible to get both unit tests and integration.

ahgittin:
if tests running twice is the normal thing then fine

would be nice to move to this "proper" build but no urgency, i support closing it, but no forgetting about it!

PR #457, opened 2015-01-19
Rest client authenticates

sjcorbett:
The `BrooklynApi` class in the rest-client module can authenticate to Brooklyn servers.


grkvlt:
Seems fine, will merge

PR #458, opened 2015-01-19
docs - blueprint tour, features, locations/ops tidy-up

ahgittin:


ahgittin:
@richardcloudsoft @aledsage this is good to review

learn-more, get-start, and ops are now all quasi-respectable i think.  and process optimizations (esp around javadocs).

i'm pushing these to the live web site -- but *not* yet the `/v/latest`.  you can check new ops structure and get start guide out in snapshots under docs.  will backport and apply to M2 when you're all happy with this.

ahgittin:
have fixed panel margin as well as htmlproof, and fixed *all* the dodgy links hurrah!  (including some in our download)

@richardcloudsoft has suggested commit-then-review after a quick look so i am doing that, and will push to m2 site and master

ahgittin:
note one extra commit https://github.com/apache/incubator-brooklyn/commit/974e7139f189798c669f1da53692588ff5b274fb manually pushed to master, of things discovered when actually pushing

PR #460, opened 2015-01-21
JcloudsLocation set arbitrary template options

rdowner:
Adds new config key for JcloudsLocation, templateOptions, that takes a Map<String,String> for setting arbitrary template options. For each entry, searches for a method with the same name as the key, with a single parameter that TypeCoercions supports. The method is then invoked with the value. If such a method cannot be found, then a warning is printed to the log.

The config key's value can be expressed as a YAML fragment in blueprints or in brooklyn.properties:

```properties
brooklyn.location.named.softlayer-ams01.templateOptions={ domainName: 'example.com' }
```

```yaml
location:
  jclouds:softlayer:ams01:
    templateOptions:
      domainName: 'example.com'
```

### DIFF
--- a/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
@@ -30,6 +30,9 @@ import static org.jclouds.scriptbuilder.domain.Statements.exec;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
 import java.security.KeyPair;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -48,6 +51,7 @@ import java.util.regex.Pattern;
 
 import javax.annotation.Nullable;
 
+import com.google.common.reflect.TypeToken;
 import org.jclouds.abiquo.compute.options.AbiquoTemplateOptions;
 import org.jclouds.cloudstack.compute.options.CloudStackTemplateOptions;
 import org.jclouds.compute.ComputeService;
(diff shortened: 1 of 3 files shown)


sjcorbett:
This looks great. Is it testable?

rdowner:
Added a test as suggested. Rebased against latest master.

rdowner:
Test failure is unrelated. Retrying...

andreaturli:
@rdowner that is great! nice and simple

A couple of questions:
1. with this feature, do the other SUPPORTED_TEMPLATE_OPTIONS_PROPERTIES still need to be there?
2. how the yaml infers the custom templateOptions?

rdowner:
Hi @andreaturli,

1 - I don't think there's a need to do that since it's still tidier in blueprints to refer to the Brooklyn-supported properties as before. The purpose of this option is to allow the selection of highly cloud-specific template options, without requiring changes to Brooklyn, and especially having to get Brooklyn to introduce a dependency on the relevant jclouds driver to set options.

2 - Something like:

```yaml
location:
  jclouds:aws:us-east-1:
    templateOptions:
      iamInstanceProfileName: myprofile
```

andreaturli:
@rdowner thanks!
1. makes sense to me
2. I think my question is more "how brooklyn infers that `instanceProfileName` is part of `AWSEC2TemplateOptions` and not `TemplateOptions`?"

rdowner:
This time the test failure is genuine. The test is supposed to be a live test so should not be run here. I'll investigate.

@andreaturli, the reflections API takes care of the `AWSEC2TemplateOptions`/`TemplateOptions` problem. The code is not interested in what type the class is - it searches all the methods available on the class for one with a name that matches and a signature that is appropriate.

richardcloudsoft:
Pushed a fix for the failing test. If the next test run passes, am I good to merge this @sjcorbett @andreaturli ?

andreaturli:
lgtm, if @sjcorbett is happy as well, please merge it

sjcorbett:
Fire away.

ahgittin:
great addition

ahgittin:
could this take `Map<String,Object>` instead of `<String,String>` ?

i think it might actually work as is (ignoring the generics) but it could be handy to support lists and maps where template options expect those, where the input is yaml. /cc @alasdairhodge

rdowner:
@ahgittin @alasdairhodge I don't see why not. Presumably it could be done in a way that keeps the helpful TypeCoercions working?

PR #461, opened 2015-01-21
Improve change-version.sh script

neykov:
* Remove the usage of pcregrep which is not installed by default on OS X or Linux.
* Support for passing the version marker as an argument.
* Simplify the code, no need for loops.


sjcorbett:
Looks good.

PR #463, opened 2015-01-22
DETECT_MACHINE_DETAILS SshMachineLocation config

rdowner:
Adds new configkey for controlling automatic detection of machine details. Defaults to true (matching previous behaviour); setting to false stops SshMachineLocation opening up an SSH channel and running commands. Useful in situations where the application requires strict control over the use of SSH.


ahgittin:
looks good - go ahead and merge

PR #464, opened 2015-01-22
Add deploy repo

neykov:
Based (but doesn't depend) on https://github.com/apache/incubator-brooklyn/pull/461

Only pom.xml changed here.

### DIFF
--- a/usage/downstream-parent/pom.xml
+++ b/usage/downstream-parent/pom.xml
@@ -578,6 +578,28 @@
       </distributionManagement>
     </profile>
 
+    <profile>
+      <id>cloudsoft-artifactory-repo</id>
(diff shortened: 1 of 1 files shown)


neykov:
Perhaps we should go the other way and remove all non-apache repos?

richardcloudsoft:
I'd be in favour of removing the non-Apache repos because that's the only place the Apache Brooklyn will deploy to. If anyone (*cough*Cloudsoft*cough*) want to deploy elsewhere, it should be done with a fork that has different group IDs to avoid stepping on Apache's trademark.

neykov:
Closed in favor of https://github.com/apache/incubator-brooklyn/pull/490.

PR #465, opened 2015-01-22
Jclouds tidy

ahgittin:
refactor to share code for credentials inferencing; then expand this to:
* validate keys, including passphrase (fail fast if bad passphrase)
* support add'l public keys to be authorized
* support add'l first-boot commands as part of the template options script


ahgittin:
closing and re-opening to kick off another build.  it's passing for me locally.

core/src/main/java/brooklyn/location/basic/LocationConfigUtils.java
```diff
@@ -20,39 +20,355 @@
...
+     * {@link LocationConfigKeys#PRIVATE_KEY_DATA} and {@link LocationConfigKeys#PRIVATE_KEY_FILE}
+     * (defaulting to the private key file + ".pub"). 
+     **/
+    public static OsCredential getOsCredential(ConfigBag config) {
+        return OsCredential.newInstance(config);
+    }
     
+    /** Convenience class for holding private/public keys and passwords, inferring from config keys.
+     * See {@link LocationConfigUtils#getOsCredential(ConfigBag)}. */
+    public static class OsCredential {
```
aledsage: This code would be cleaner if there was a separate builder, and then OsCredential was immutable. With that pattern, we wouldn't need calls to `infer()` in each method, or the `dirty()`. We'd also have a clearer separation of which fields are for configuring it, and which are about the actual state (and thus behaviour of the constructed `OsCredential`).

It would also hide the many builder methods from users of the `OsCredential`, simplifying things for such programmers.

core/src/main/java/brooklyn/location/basic/LocationConfigUtils.java
```diff
@@ -20,39 +20,355 @@
...
+            KeyPair privateKey = null;
+            
+            if (Strings.isBlank(privateKeyData)) {
+                // look up private key files
+                String privateKeyFiles = null;
+                boolean privateKeyFilesExplicitlySet = config.containsKey(LocationConfigKeys.PRIVATE_KEY_FILE);
+                if (privateKeyFilesExplicitlySet || (tryDefaultKeys && password==null)) 
+                    privateKeyFiles = config.get(LocationConfigKeys.PRIVATE_KEY_FILE);
+                if (Strings.isNonBlank(privateKeyFiles)) {
+                    Iterator<String> fi = Arrays.asList(privateKeyFiles.split(File.pathSeparator)).iterator();
```
aledsage: Maybe agree, but am hesitant for us to have different formats for the contents of brooklyn.properties on Windows and Unix.

But the value of this field will look very different on linux and windows, given that it normally contains file paths.

I see we use `getResourceAsString(file)` which means it would actually work with a URL (except for `:` having been used as a separator). I don't think we need to support URLs; the config key's description is clear that they are files.

aledsage: Ah, I see you write that in a comment directly below this!

core/src/main/java/brooklyn/location/basic/LocationConfigUtils.java
```diff
@@ -20,39 +20,355 @@
...
+        private void addWarning(String msg, String debugExtension) {
+            log.debug(msg+(debugExtension==null ? "" : "; "+debugExtension));
+            warningMessages.add(msg);
+        }
+
+        @Override
+        public String toString() {
+            return getClass().getSimpleName()+"["+
+                (Strings.isNonBlank(publicKeyData) ? publicKeyData : "public-key")+";"+
+                (Strings.isNonBlank(privateKeyData) ? "private-key" : "private-key")+","+
```
aledsage: `toString` looks wrong: will say "private-key" whether or not there is privateKeyData, and will say "public-key" if there is no publicKeyData.

core/src/main/java/brooklyn/location/basic/LocationConfigUtils.java
ahgittin: yeah - the format is legacy, and bad

core/src/main/java/brooklyn/location/basic/LocationConfigUtils.java
ahgittin: yeah - probably worth refactoring.  are you okay with me marking it `@Beta` for now, and we can do that later?

core/src/main/java/brooklyn/location/basic/LocationConfigUtils.java
ahgittin: oops - yes

core/src/main/java/brooklyn/util/crypto/FluentKeySigner.java
```diff
@@ -42,7 +44,9 @@
  * X509v3CertificateBuilder drags in an add'l dependency (bcmail) and is harder to use. */
 @SuppressWarnings("deprecation")
```
aledsage: I know it's not your code but couldn't help commenting...

I'd prefer us not to use a blanket `@SuppressWarnings("deprecation")` on the class. It means we don't get told what we're using that is actually deprecated, which we should fix (unless we're deliberately doing it for backwards compatibility reasons with multiple versions of a library?!).

core/src/main/java/brooklyn/util/crypto/FluentKeySigner.java
```diff
@@ -42,7 +44,9 @@
...
 public class FluentKeySigner {
-    
+
+    static { Security.addProvider(new BouncyCastleProvider()); }
```
aledsage: I'm uncomfortable with doing this here. Feels like we should set up these JVM-wide provides at the start, as part of setup. My main concern with doing it here is that it could impact other stuff running in the JVM a while after Brooklyn has started, causing weird behaviour.

It feels like the same argument for why it's a bad idea to call `System.setProperty("security....")` in the middle of some class - thus impacting what other seemingly unrelated code in the JVM will subsequently do.

My other concern is that it doesn't give a way to see easily see all the dependencies we're using (i.e. what we're injecting into the JVM).

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
```diff
@@ -19,182 +19,149 @@
...
 
 /**
  * Utility methods for generating and working with keys
  */
-public class SecureKeys {
+public class SecureKeys extends SecureKeysWithoutBouncyCastle {
 
+    private static final Logger log = LoggerFactory.getLogger(SecureKeys.class);
+    
     static { Security.addProvider(new BouncyCastleProvider()); }
```
aledsage: Given the class extends `SecureKeysWithoutBouncyCastle`, it's surprising to see this static code in it.

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
```diff
@@ -19,182 +19,149 @@
...
-        try {
-            keyPairGenerator = KeyPairGenerator.getInstance(algorithm);
-        } catch (NoSuchAlgorithmException e) {
-            throw Exceptions.propagate(e);
-        }
-        keyPairGenerator.initialize(bits);
-        return keyPairGenerator;
+    public static class PassphraseProblem extends IllegalStateException {
+        private static final long serialVersionUID = -3382824813899223447L;
+        public PassphraseProblem() { super("Passphrase problem with this key"); }
```
aledsage: Exceptions should pretty much always just have two constructors: one that takes a message, and one that takes a message+cause. The no-arg constructor is not used, and let's keep it that way by deleting it.

core/src/main/java/brooklyn/util/crypto/FluentKeySigner.java
ahgittin: I share the concern but I don't like the idea of mandating a specific entry point as it would mean we can't easily run test classes on their own.

Maybe the thing to do is have a `BrooklynInitialization` class with statics `initBouncyCastle()` and `initAll()` which we can invoke as and where needed?  Start-sequence-wise it wouldn't be any difference but at least the startup logic would be in one place for reference.

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
```diff
@@ -19,182 +19,149 @@
...
+            if (object==null) {
+                throw new IllegalStateException("PEM parsing failed: missing or invalid data");
+            } else if (object instanceof PEMEncryptedKeyPair) {
+                if (passphrase==null) throw new PassphraseProblem("passphrase required");
+                try {
+                    PEMDecryptorProvider decProv = new JcePEMDecryptorProviderBuilder().build(passphrase.toCharArray());
+                    kp = converter.getKeyPair(((PEMEncryptedKeyPair) object).decryptKeyPair(decProv));
+                } catch (Exception e) {
+                    Exceptions.propagateIfFatal(e);
+                    throw new PassphraseProblem("wrong passphrase");
```
aledsage: I'd pass in the cause as well. Is there any more specific exception we can catch here? Feels risky that we could get other exceptions and swallow them, incorrectly reporting it as a "wrong passphrase".

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
ahgittin: `extends` used here in the "builds-on" sense, not the "has all properties of the parent" sense

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
ahgittin: ? -- most exceptions have a no-arg constructor

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
```diff
@@ -19,182 +19,149 @@
...
+                    public char[] getPassword() {
+                        return passphrase!=null ? passphrase.toCharArray() : new char[0];
+                    }
+                });
+                KeyPair result = (KeyPair) pr.readObject();
+                pr.close();
+                if (result==null)
+                    throw Exceptions.propagate(e);
+                
+                log.warn("PEMParser failed when PEMReader succeeded, with "+result+"; had: "+e);
```
aledsage: Can we include the word "deprecated" in the message? e.g. `"PEMParser failed when deprecated PEMReader succeeded, with "+result+"; had: "+e`.
Perhaps we should log the exception at debug as well - with just the first line and no stacktrace, we mind find it impossible to fix some kinds of exceptions.

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
```diff
@@ -19,182 +19,149 @@
...
+                PrivateKey privKey = converter.getPrivateKey((PrivateKeyInfo) object);
+                kp = new KeyPair(null, privKey);
+            } else {
+                throw new IllegalStateException("PEM parser support missing for: "+object);
+            }
+
+            return kp;
 
-    /** returns keystore of default type read from given source */
-    public static KeyStore newKeyStore(InputStream source, String passphrase) {
```
aledsage: Unfortunately these methods (e.g. `newKeyStore`, `getTrustManager`, `getCanonicalImpl` etc) weren't marked as `@Beta` and neither was the class. Are we definitely ok with just deleting them rather than deprecating them?

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
```diff
@@ -19,182 +19,149 @@
...
-
-            KeyStore store2 = KeyStore.getInstance(KeyStore.getDefaultType());
-            store2.load(new ByteArrayInputStream(out.toByteArray()), "".toCharArray());
-            return (X509Certificate) store2.getCertificate("to-canonical");
-        } catch (Exception e) {
-            throw Exceptions.propagate(e);
-        }
+    /** because KeyPair.equals is not implemented :( */
+    public static boolean equal(KeyPair k1, KeyPair k2) {
+        return Objects.equal(k2.getPrivate(), k1.getPrivate()) && Objects.equal(k2.getPrivate(), k1.getPrivate());
```
aledsage: Looks wrong. Did you mean for one of the `Objects.equal` to compare the `getPublic()`?

core/src/test/java/brooklyn/util/crypto/SecureKeysAndSignerTest.java
```diff
@@ -93,6 +95,40 @@ public void testReadRsaKey() throws Exception {
...
+    public void testReadRsaPublicKeyAsPemFails() throws Exception {
+        // should fail; see next test
+        SecureKeys.readPem(ResourceUtils.create(this).getResourceFromUrl("classpath://brooklyn/util/crypto/sample_rsa.pem.pub"), null);
+    }
+    
+    @Test
+    public void testReadRsaPublicKeyAsAuthKeysWorks() throws Exception {
+        PublicKey key = AuthorizedKeysParser.decodePublicKey(
+            ResourceUtils.create(this).getResourceAsString("classpath://brooklyn/util/crypto/sample_rsa.pem.pub"));
+        KeyPair fromPem = SecureKeys.readPem(ResourceUtils.create(this).getResourceFromUrl("classpath://brooklyn/util/crypto/sample_rsa.pem"), null);
```
aledsage: Not particularly related to this PR, but on first impressions I'm confused why `brooklyn/util/crypto/sample_rsa.pem` is excluded from RAT check. In the root `pom.xml`, it has `<exclude>**/*.pem.pub</exclude>` but nothing for `*.pem`.

core/src/main/java/brooklyn/util/crypto/FluentKeySigner.java
ahgittin: i'll refactor to use the fully qualified class name everywhere, that way we can suppress deprecations on a per-line basis not class-wide

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -752,6 +762,17 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
...
                 }
+                
+                List<String> extraKeyUrlsToAuth = setup.get(EXTRA_PUBLIC_KEYS_TO_AUTH);
+                if (extraKeyUrlsToAuth!=null && !extraKeyUrlsToAuth.isEmpty()) {
+                    List<String> extraKeyDataToAuth = MutableList.of();
+                    for (String keyUrl: extraKeyUrlsToAuth) {
+                        extraKeyDataToAuth.add(ResourceUtils.create().getResourceAsString(keyUrl));
+                    }
+                    sshMachineLocation.execCommands("Authorizing ssh keys", 
+                        MutableList.of(new AuthorizeRSAPublicKeys(extraKeyDataToAuth).render(org.jclouds.scriptbuilder.domain.OsFamily.UNIX)));
```
aledsage: Personal preference: I prefer us using `ImmutableList` everywhere that we expect no nulls and expect not to modify it. In this case, a null is going to be a programming error / bug somewhere so `ImmutableList` feels better.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1305,118 +1326,148 @@ public UserCreation(LoginCredentials creds, List<Statement> statements) {
...
-                LOG.info("Not creating user {}, and not setting its password or authorizing keys", user);
-                
-                if (privateKeyData != null) {
-                    loginCreds = LoginCredentials.builder().user(user).privateKey(privateKeyData).build();
-                } else if (explicitPassword != null) {
-                    loginCreds = LoginCredentials.builder().user(user).password(password).build();
+                LOG.info("Not creating user {}, and not installing its password or authorizing keys (assuming it exists)", user);
+
+                if (credential.isUsingPassword()) {
+                    createdUserCreds = LoginCredentials.builder().user(user).password(credential.get()).build();
```
aledsage: I'd prefer `credential.getPassword()` rather than `credential.get()` - it's much more explicit, rather than relying on `.get()` returning password because `isUsingPassword()`. Same for line below.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1305,118 +1326,148 @@ public UserCreation(LoginCredentials creds, List<Statement> statements) {
...
-            }
-            if (publicKeyData!=null) {
-                statements.add(new AuthorizeRSAPublicKeys("~"+ROOT_USERNAME+"/.ssh", ImmutableList.of(publicKeyData)));
-                if (privateKeyData != null) {
-                    loginCreds = LoginCredentials.builder().user(user).privateKey(privateKeyData).build();
+        } else {
+            String pubKey = credential.getPublicKeyData();
+            String privKey = credential.getPrivateKeyData();
+            
+            if (credential.get()==null) {
```
aledsage: This is confusing code initially - hard to guess what `credential.get()` does, compared to `credential.getXyz()` for getting public key data, private key data, etc.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1305,118 +1326,148 @@ public UserCreation(LoginCredentials creds, List<Statement> statements) {
...
+                        + "Create ~/.ssh/id_rsa or "
+                        + "set "+PRIVATE_KEY_FILE.getName()+" / "+PRIVATE_KEY_PASSPHRASE.getName()+" / "+PASSWORD.getName()+" "
+                        + "as appropriate for this location if you wish to be able to log in without Brooklyn.");
                 }
+                KeyPair newKeyPair = SecureKeys.newKeyPair();
+                pubKey = SecureKeys.toPub(newKeyPair);
+                privKey = SecureKeys.toPem(newKeyPair);
+                LOG.debug("Brooklyn key being created for "+user+" at new machine "+this+" is:\n"+privKey);
             }
+            // ensure credential is not used any more, as we have extracted al useful info
```
aledsage: Typo: "al" instead of "all".

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1305,118 +1326,148 @@ public UserCreation(LoginCredentials creds, List<Statement> statements) {
...
         String loginUser = groovyTruth(explicitLoginUser) ? explicitLoginUser : (image != null && image.getDefaultCredentials() != null) ? image.getDefaultCredentials().identity : null;
         Boolean dontCreateUser = config.get(DONT_CREATE_USER);
         Boolean grantUserSudo = config.get(GRANT_USER_SUDO);
-        String publicKeyData = LocationConfigUtils.getPublicKeyData(config);
-        String privateKeyData = LocationConfigUtils.getPrivateKeyData(config);
-        String explicitPassword = config.get(PASSWORD);
-        String password = groovyTruth(explicitPassword) ? explicitPassword : Identifiers.makeRandomId(12);
+        OsCredential credential = LocationConfigUtils.getOsCredential(config);
+        credential.checkNoErrors().logAnyWarnings();
+        String passwordToSet = Strings.isNonBlank(credential.getPassword()) ? credential.getPassword() : Identifiers.makeRandomId(12);
```
aledsage: If Brooklyn makes up a new password, then should we log that? Looks like it's not logged anywhere currently. Most important to log it if the auto-generated password is the only subsequent login approach.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1305,118 +1326,148 @@ public UserCreation(LoginCredentials creds, List<Statement> statements) {
...
-                    .loginPassword(password);
-
-            if (publicKeyData!=null) {
-                adminBuilder.authorizeAdminPublicKey(true).adminPublicKey(publicKeyData);
+                    .grantSudoToAdminUser(groovyTruth(grantUserSudo));
+                    // login user's password now set to something random
+                    // (could explicitly set with loginPassword)
+            
+            Boolean useKey = null;
+            if (passwordToSet!=null) {
```
aledsage: `passwordToSet` can presumably never be null at this point - we'll have auto-generated a random password if one was not supplied.

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
ahgittin: but agree it's no use here, so removing it

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
ahgittin: done, and we log it (debug) when catching

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
ahgittin: +1 adding the word `deprecated`.  But I really want this one brought to our attention so leaving it log warning.  We can exclude the logger or (better) patch.

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
ahgittin: they are in superclass

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
ahgittin: yes - good spot

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1305,118 +1326,148 @@ public UserCreation(LoginCredentials creds, List<Statement> statements) {
...
+                adminBuilder.adminPassword(passwordToSet);
+                useKey = false;
```
aledsage: This looks wrong, so not sure how it works. Presumably we'll always go into this if branch because `passwordToSet` is never null. Therefore we'll always set `userKey = false`.

core/src/test/java/brooklyn/util/crypto/SecureKeysAndSignerTest.java
ahgittin: maybe it is big enough rat can tell it isn't human readable?

ahgittin: i've added `*.pem` for good measure.  `*_rsa` was already there.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
ahgittin: makes sense.  have changed `get()` to `getPreferredCredential()`, and not used if we've already done an explicit check

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
ahgittin: now using `isEmpty()`

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
ahgittin: yes, good spot.  and similar `isNonBlank` check in codepath above where `user` equals `loginUser` -- fixing

ahgittin: have removed and added quite a bit more logging for this, as it is something we want to be able to drill down on

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
aledsage: sub-classing should be an "is-a" relationship. If not that, then other mechanisms of code-reuse are perhaps more appropriate. This at least deserves a comment in the javadoc of the class.

core/src/main/java/brooklyn/util/crypto/FluentKeySigner.java
aledsage: Sounds sensible, I think. Is there anything beyond bouncy castle currently that we'd want to include in a `BrooklynInitialization`? If it's just bouncy castle, then maybe we live with code as-is for now, with an appropriate comment somewhere (e.g. `LocalManagementContext` or the `Main` class?) about this initialization.

I presume that would be at the level of the `ManagementContext` so wouldn't know about initializing web-apps etc?

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1305,118 +1326,148 @@ public UserCreation(LoginCredentials creds, List<Statement> statements) {
...
+            
+            if (credential.get()==null) {
+                if (!loggedSshKeysHint && !config.containsKey(PRIVATE_KEY_FILE)) {
+                    loggedSshKeysHint = true;
+                    LOG.info("Default SSH keys not found or not usable; will create new keys for each machine. "
+                        + "Create ~/.ssh/id_rsa or "
+                        + "set "+PRIVATE_KEY_FILE.getName()+" / "+PRIVATE_KEY_PASSPHRASE.getName()+" / "+PASSWORD.getName()+" "
+                        + "as appropriate for this location if you wish to be able to log in without Brooklyn.");
                 }
+                KeyPair newKeyPair = SecureKeys.newKeyPair();
```
aledsage: I think that I'd prefer an explicit configuration option to enable creating new keys. The question is what use-case this is covering.

If it is for advanced users, and requires an explicit act to set the config value to blank, then the user could equally well have set a more explicit config key to tell brooklyn to crate the keypair.

If it is for newbies that don't have ssh keys set up yet (and don't intend to), then I think the automatic key-gen is a false economy for them. At some point they'll want to log in, so would have to manually find the private key in the debug log and create an appropriately permissioned file with it. That sounds more onerous than running a standard key-gen command up front.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocationConfig.java
```diff
@@ -71,13 +72,25 @@
     public static final ConfigKey<String> LOGIN_USER_PRIVATE_KEY_FILE = ConfigKeys.newStringConfigKey("loginUser.privateKeyFile",
             "Custom private key for the user who logs in initially", null); 
     public static final ConfigKey<String> EXTRA_PUBLIC_KEY_DATA_TO_AUTH = ConfigKeys.newStringConfigKey("extraSshPublicKeyData",
-            "Additional public key data to add to authorized_keys", null);
+        "Additional public key data to add to authorized_keys", null);
+    @SuppressWarnings("serial")
+    public static final ConfigKey<List<String>> EXTRA_PUBLIC_KEYS_TO_AUTH = ConfigKeys.newConfigKey(new TypeToken<List<String>>() {},
```
aledsage: For consistency with other names, should this be `EXTRA_PUBLIC_KEY_FILES_TO_AUTH`?

core/src/test/java/brooklyn/location/basic/LocationConfigUtilsTest.java
```diff
@@ -21,77 +21,134 @@
...
-        String data = LocationConfigUtils.getPrivateKeyData(config);
+        config.put(LocationConfigKeys.PRIVATE_KEY_FILE, SSH_PRIVATE_KEY_FILE_WITH_TILDE);
+
+        // don't mind if it has a passphrase
+        String data = LocationConfigUtils.getOsCredential(config).doKeyValidation(false).get();
         assertTrue(data != null && data.length() > 0);
     }
     
-    @Test(groups="Integration")
+    @Test(groups="Integration")  // requires ~/.ssh/passphrase-id_rsa
```
aledsage: Instead of adding another external file dependency, can use re-use the `SshToolAbstractIntegrationTest.SSH_KEY_WITH_PASSPHRASE` (i.e. use `~/.ssh/id_rsa_with_passphrase` as the default)?

utils/common/src/main/java/brooklyn/util/crypto/AuthorizedKeysParser.java
```diff
@@ -0,0 +1,132 @@
...
+            }
+            
+        } catch (Exception e) {
+            Exceptions.propagateIfFatal(e);
+            throw new IllegalArgumentException("Error parsing authorized_keys/SSH2 format public key: "+e);
+        }
+    }
+
+    private static int readInt(InputStream stream) throws IOException {
+        return ((stream.read() & 0xFF) << 24) | ((stream.read() & 0xFF) << 16)
```
aledsage: Worth a unit test for this kind of thing, I think (package-private, annotated as `@VisibleForTesting`, and an easy unit test to write). same for the other methods here.

utils/common/src/main/java/brooklyn/util/crypto/AuthorizedKeysParser.java
```diff
@@ -0,0 +1,132 @@
...
+            | ((stream.read() & 0xFF) << 8) | (stream.read() & 0xFF);
+    }
+    
+    private static byte[] readBytesWithLength(InputStream stream) throws IOException {
+        int len = readInt(stream);
+        byte[] result = new byte[len];
```
aledsage: Perhaps guard this - if `len` is big then there's something wrong (given the context of this method). Better to throw a graceful exception rather than cause an OutOfMemoryError.

utils/common/src/main/java/brooklyn/util/crypto/AuthorizedKeysParser.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+    private static void writeBytesWithLength(OutputStream stream, byte[] buf) throws IOException {
+        writeInt(stream, buf.length);
+        stream.write(buf);
+    }
+    
+    private static String readType(InputStream stream) throws IOException { return new String(readBytesWithLength(stream)); }
+    private static BigInteger readBigInt(InputStream stream) throws IOException { return new BigInteger(readBytesWithLength(stream)); }
+
+    public static String encodePublicKey(PublicKey key) {
```
aledsage: Surprised there's not some utility method in one of our dependencies that already does this, but I haven't looked at all.

utils/common/src/main/java/brooklyn/util/crypto/AuthorizedKeysParser.java
aledsage: If we create a new `AuthorizedKeysParserTest.java`, then it should include a reference to `SecureKeysAndSignerTest` where there are tests that use this utility class.

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
aledsage: Ah, I see they are on the super-class (`SecureKeysWithoutBouncyCastle`). Therefore I presume this comment can be ignored (though I haven't gone through to check all methods are there with the same signature).

aledsage:
@ahgittin finished reviewing. Lots of good stuff here.

My biggest comment (because it relates to the user experience) is about an explicit configuration option to tell it to auto-generate a key-pair.

core/src/main/java/brooklyn/util/crypto/SecureKeys.java
aledsage: True that most/all exceptions that come from java 1.0 have a no-arg constructor. It was a great relief when they stopped using those in situations like `ClassCastException`, to actually tell you something about the exception.

For new exception classes that folk are writing, the (vast?) majority of cases it deserves forcing the caller to pass a message, rather than allowing the absence of useful info to the end-user.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
ahgittin: This is for both use cases.  I pretty much agree with your suggestion but that could be a new feature (ie not in this PR).  All I've done here is swap the random per-machine password with a random per-machine key, because (a) it's more secure and (b) an increasing number of VM's don't support password-ssh login.

for the advanced use case, i like an explicit `create_new_key_per_machine`.  in this case i would not write it to the file system, but we make the key available through REST API or GUI.  (by the way, currently the lookup procedure is probably even worse than you expect:  we don't know what machine is being provisioned here, so if there are several machines it's going to trial-and-error to correlate.)

if this config key were not set, then in the new user case we could then create a `brooklyn_id_rsa` and `.pub` in `MGMT_BASE_DIR` (which is used subsequently as the default).  but i would only want to do this if it's just the one ssh key, and i don't want to lose the capability of new keys per machine.

have added comment to this effect.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocationConfig.java
ahgittin: agree, `EXTRA_PUBLIC_KEY_URLS_TO_AUTH` (here we can accept urls, falling back to file format)

core/src/test/java/brooklyn/location/basic/LocationConfigUtilsTest.java
ahgittin: comment was wrong, it uses a key that is added to classpath

utils/common/src/main/java/brooklyn/util/crypto/AuthorizedKeysParser.java
ahgittin: `SecureKeysAndSignerTest` provides enough test coverage i think

utils/common/src/main/java/brooklyn/util/crypto/AuthorizedKeysParser.java
ahgittin: agree, now:

        if (len<10 || len>100000)
            throw new IllegalStateException("Invalid stream header: length "+len);

utils/common/src/main/java/brooklyn/util/crypto/AuthorizedKeysParser.java
ahgittin: i spent a while looking and didn't find one

ahgittin:
Thanks @aledsage .  You happy with the last 3 commits?  I think they address all the main issues.

@richardcloudsoft let me know when you've tried this!

ahgittin:
fixed a typo.  also added some simple CLI stuff; not related, but very small and it would otherwise conflict with this PR.

ahgittin:
merging -- @richardcloudsoft as this should address the worry you had, not compound it!  but please do test and let me know

PR #466, opened 2015-01-23
JS gui improvements - empty catalog and server list

ahgittin:
shows yaml tab (and nice message on catalog tab) if catalog is empty

server list is now after apps list, and there is a way to clear it (new `HighAvailabilityManagerImpl.publishClearNonMaster()` method, REST call, and button on main page)

### DIFF
--- a/api/src/main/java/brooklyn/management/ha/HighAvailabilityManager.java
+++ b/api/src/main/java/brooklyn/management/ha/HighAvailabilityManager.java
@@ -97,7 +97,12 @@ public interface HighAvailabilityManager {
     void setPriority(long priority);
     
     long getPriority();
-    
+
+    /** deletes non-master node records; active nodes (including this) will republish, 
+     * so this provides a simple way to clean out the cache of dead brooklyn nodes */
+    @Beta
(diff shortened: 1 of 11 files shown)


api/src/main/java/brooklyn/management/ha/HighAvailabilityManager.java
```diff
@@ -97,7 +97,12 @@
     void setPriority(long priority);
     
     long getPriority();
-    
+
+    /** deletes non-master node records; active nodes (including this) will republish, 
+     * so this provides a simple way to clean out the cache of dead brooklyn nodes */
+    @Beta
+    void publishClearNonMaster();
```
aledsage: This feels like a strange method name. Why the two verbs of publish and clear?

aledsage:
When the catalog is empty, should I see "The applications catalog is empty" rather than just a white rectangle? I'm still just seeing the empty white rectangle in the add application, when I start with a catalog.xml of `<catalog/>`.

The other changes all work really well for me.

ahgittin:
you're right @aledsage -- fixed that

PR #468, opened 2015-01-23
JcloudsLocation: make more testable

aledsage:
- allow a custom jclouds ComputeService to be injected, so can stub out things like VM creation and deletion


nakomis:
LGTM, tests pass

aledsage:
Thanks @nakomis ; will merge now.

PR #469, opened 2015-01-23
delete port forwarding when VM is deleted

aledsage:
This includes commits from https://github.com/apache/incubator-brooklyn/pull/468 - please review+merge that first.


nakomis:
An additional test which creates two machines, releases one machine then checks if the port mappings for the other machine are still around would be nice. Other than that, LGTM, tests pass

aledsage:
Thanks @nakomis - I've rebased, then added a test for two VMs (ensuring that releasing first does not impact second). Merging.

PR #470, opened 2015-01-26
Use javac compiler where possible

neykov:
Replace groovy-eclipse compiler with javac. All projects' non-test code apart from `groovy-utils` and `brooklyn-test-support` is compiled with javac. **Test code** for the following projects still compiled with groovy-eclipse:
  * core
  * locations-jclouds
  * policy
  * sandbox/database
  * sanbox/nosql
  * software-database
  * software-webapp
  * cli


neykov:
Tests are not passing for me though. I get a stack overflow - `super.configure()` calling into itself.
```
	at brooklyn.location.basic.SshMachineLocation.configure(SshMachineLocation.java:124)
	at brooklyn.location.basic.LocalhostMachineProvisioningLocation$LocalhostMachine.configure(LocalhostMachineProvisioningLocation.java:300)
	at brooklyn.location.basic.LocalhostMachineProvisioningLocation$LocalhostMachine.configure(LocalhostMachineProvisioningLocation.java:241)
	at brooklyn.location.basic.SshMachineLocation.configure(SshMachineLocation.java:124)

```

ahgittin:
i get compile errors in `/camp-base/src/main/java/io/brooklyn/camp/spi/AbstractResource.java` 

from http://docs.oracle.com/javase/tutorial/java/javaOO/nested.html

| Non-static nested classes (inner classes) have access to other members of the enclosing class, even if they are declared private. Static nested classes do not have access to other members of the enclosing class. 

this probably affects a fair few of our builders -- would be good to resolve /cc @aledsage

ahgittin:
with `-Peclipse-compiler` things work, but still, we should fix this!

neykov:
Just to note that this happens with Java 7+

Looks like each compiler version brings it's own set of inconsistencies :).

ahgittin:
ha ha.  i've tried switching those `Builder`s to be non-static, seems to work so far.

ahgittin:
your stack overflow is bizarre though!  what is going on there?

neykov:
Couldn't figure it out even after debugging. As this was supposed to be a no-thrills quick change decided to leave it for later as it was becoming a time sink.

PR #471, opened 2015-01-26
brooklyn.util.net.Urls.mergePaths(String... items) mishandles nulls in input path parts

michaeldye:
Fix for [BROOKLYN-129](https://issues.apache.org/jira/browse/BROOKLYN-129): filter nulls before iterating through path parts.

### DIFF
--- a/utils/common/src/main/java/brooklyn/util/net/Urls.java
+++ b/utils/common/src/main/java/brooklyn/util/net/Urls.java
@@ -25,6 +25,9 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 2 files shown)


sjcorbett:
_Should_ it filter nulls? It strikes me as an error of the caller and something that would be quite surprising in practice. What is the context of the original error in BROOKLYN-129?

aledsage:
@michaeldye @sjcorbett The current behaviour (before this PR) is definitely wrong - of appending `null` to the URL.

Whether or not it should filter nulls automatically is a stylistic discussion for Brooklyn APIs. Unfortunately we have a bit of a mix - e.g. `brooklyn.util.collections.MutableList` creation methods take `null` to mean creating an empty list; other places in our code follows the guava best-practice of treating `null` as a programming error (failing fast, to catch these more easily).

In this situation, I lean towards @sjcorbett 's suggestion - a null in the middle of an array of strings smacks of a programming error, rather than a short-hand convenience for saying "create an empty thing". The caller can always filter out the nulls if it truly is valid to have nulls in that context.

Note that passing in an empty string is valid - that will effectively be ignored.

@michaeldye does that make sense for your use-case, if this were to throw a `NullPointerException` when the vararg array of `items` contains a `null`?

p.s. in case anyone is interested in the history, one reason we accept nulls in things like `MutableMap` etc is because of the handling of config and attributes. The values supplied by the user can include explicit nulls, which leads to NPEs if you try to just use `ImmutableMap.copyOf(...)`.

michaeldye:
@sjcorbett @aledsage Thank you for the discussion. I agree that it really should be up to the caller to provide acceptable arguments to ```Urls.mergePaths(String... items)``` and that this method should merely enforce that contract by throwing an exception given a null arg. Since the null values in this case aren't dereferenced, but rather passed to ```StringBuilder.append()```, does an NPE make the most sense? How about an ```IllegalArgumentException``` and a message including the rest of the path to help someone debug it later?

If you agree, I'll update this PR with the behavior desired for mergePaths and create another PR with updates to the calling code. Thanks!

aledsage:
@michaeldye I agree that: if any of the Strings in `items` is null, it should throw an exception that includes the rest of the path in its message.

I'd prefer a `NullPointerException` rather than an `IllegalArgumentException`. To quote Joshua Bloch's Effective Java item 60: "If a caller passes null in some parameter for which null values are prohibited, convention dictates that NullPointerException be thrown rather than IllegalArgumentException."

You can call the NPE constructor that takes a String, so can still pass in the message.

If you can update the PR accordingly (and ideally add a test case to `brooklyn.util.net.UrlsTest`), then that would be fantastic.

michaeldye:
Sounds good, and thanks for the note on the NPE convention. I've updated the PR.

aledsage:
Looks good; merging.

PR #474, opened 2015-01-27
Crate database entity

sjcorbett:
An entity suitable for single-node deployments of Crate. Subsequent work should support clusters as described by the docs: https://crate.io/docs/en/latest/best_practice/multi_node_setup.html.

### DIFF
--- a/software/database/src/main/java/brooklyn/entity/database/crate/CrateNode.java
+++ b/software/database/src/main/java/brooklyn/entity/database/crate/CrateNode.java
@@ -1,0 +1,40 @@
+package brooklyn.entity.database.crate;
+
+import brooklyn.config.ConfigKey;
+import brooklyn.entity.basic.Attributes;
+import brooklyn.entity.basic.ConfigKeys;
+import brooklyn.entity.basic.SoftwareProcess;
+import brooklyn.entity.java.UsesJava;
+import brooklyn.entity.java.UsesJavaMXBeans;
+import brooklyn.entity.java.UsesJmx;
+import brooklyn.entity.proxying.ImplementedBy;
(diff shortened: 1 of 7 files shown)


sjcorbett:
@adamecjackson Thanks for helping out.

To the reviewer: Adam has submitted an ICLA but I don't think it has been filed yet.

sjcorbett:
ICLA now filed.

software/database/src/main/java/brooklyn/entity/database/crate/CrateNode.java
```diff
@@ -0,0 +1,90 @@
...
+    public static final PortAttributeSensorAndConfigKey CRATE_HTTP_PORT = new PortAttributeSensorAndConfigKey(
+            "crate.httpPort", "The port for HTTP traffic", PortRanges.fromString("4200+"));
+
+    AttributeSensor<String> MANAGEMENT_URL = Sensors.newStringSensor(
+            "crate.managementUri", "The address at which the Crate server listens");
+
+    AttributeSensor<String> SERVER_NAME = Sensors.newStringSensor(
+            "crate.server.name", "The name of the server");
+
+    AttributeSensor<Boolean> SERVER_OK = Sensors.newBooleanSensor(
```
andreaturli: what is the difference between SERVER_OK and SERVER_STATUS? are both needed?

sjcorbett: No idea. They're both reported by Crate so we included both.

andreaturli:
lgtm +1

good stuff, @adamecjackson and @sjcorbett

PR #472, opened 2015-01-26
fixed BROOKLYN-128: ArchiveUtils ignores destFile, prefers optionalDestFile during deploy

michaeldye:
Fix for [BROOKLYN-129](https://issues.apache.org/jira/browse/BROOKLYN-129): issue install commands with guarded destFile instead of optionalDestFile.

PR #475, opened 2015-01-27
Adds ArchiveUtilsTest + fixes ArchiveUtils

aledsage:
- Fixes case where optionalTmpDir not supplied
- Adds various test cases

### DIFF
--- a/core/src/main/java/brooklyn/util/file/ArchiveUtils.java
+++ b/core/src/main/java/brooklyn/util/file/ArchiveUtils.java
@@ -273,7 +273,7 @@ public class ArchiveUtils {
 
             // extract, now using task if available
             MutableList<String> commands = MutableList.copyOf(installCommands(destFile))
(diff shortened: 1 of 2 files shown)


aledsage:
Motivated by the fix in https://github.com/apache/incubator-brooklyn/pull/472 by @michaeldye - tests that scenario, and various others.

@michaeldye If you get a chance to review + comment on this PR, that would be great!

aledsage:
Doh, tests fail when run as unit tests, because the test server isn't setup to be able to ssh. I'll have to mark those tests as `@test(groups="Integration")`, so they are only run on test servers that are set up for ssh.

ahgittin:
LGTM.  @michaeldye ?

michaeldye:
Sorry for the delay. Yup, looks good to me too.

aledsage:
Thanks @michaeldye @ahgittin - merging.

PR #476, opened 2015-01-27
Support scriptDir location config

rdowner:
This modifies `SshMachineLocation` to propagate certain config keys to flags on `SshTool.execCommand()` and `.execScript()` invocations; the first use of this is to propagate the `scriptDir` config. `JcloudsLocation` is also modified to propagate this configuration. The result is that `scriptDir` can be specified in the configuration for a jclouds-based location, and the value of this property will be passed to every invocation of `SshTool.execScript()`.

The use case for this is to allow Brooklyn to be usable on locations where the VM image mounts `/tmp` with `noexec`.

Also adds some integration tests to verify the correct operation of the `scriptDir` flag.

### DIFF
--- a/core/src/main/java/brooklyn/location/basic/SshMachineLocation.java
+++ b/core/src/main/java/brooklyn/location/basic/SshMachineLocation.java
@@ -19,6 +19,8 @@
 package brooklyn.location.basic;
 
 import static brooklyn.util.GroovyJavaMethods.truth;
+
+import com.google.common.annotations.Beta;
 import groovy.lang.Closure;
 
 import java.io.Closeable;
@@ -110,6 +112,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.net.HostAndPort;
 
@@ -215,6 +218,14 @@ public class SshMachineLocation extends AbstractLocation implements MachineLocat
                     return input.getConfigKey().getName();
                 }
             }));
+
+    /**
+     * The set of config keys on this location which become default values for properties when invoking an SSH
+     * operation.
+     */
+    @Beta
+    public static final Set<ConfigKey<?>> SSH_CONFIG_GIVEN_TO_PROPS = ImmutableSet.<ConfigKey<?>>of(
+            SCRIPT_DIR);
 
     private Task<?> cleanupTask;
     /** callers should use {@link #getSshPoolCache()} */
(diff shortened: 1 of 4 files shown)


core/src/test/java/brooklyn/location/basic/SshMachineLocationIntegrationTest.java
```diff
@@ -75,7 +81,36 @@ public void testExtractingConnectablePassphraselessKey() throws Exception {
...
+        // For explanation of (some of) the magic behind this command, see http://stackoverflow.com/a/229606/68898
+        final String command = "if [[ \"$0\" == \"/var/tmp/\"* ]]; then true; else false; fi";
+
+        LocalhostMachineProvisioningLocation lhp = (LocalhostMachineProvisioningLocation) mgmt.getLocationRegistry().resolve("localhost", true, null).orNull();
+        SshMachineLocation sm = lhp.obtain();
+
+        Map<String, Object> props = ImmutableMap.<String, Object>builder()
+                .put(SshTool.PROP_SCRIPT_DIR.getName(), "/var/tmp")
+                .build();
+        sm.execScript(props, "Test script directory execution", ImmutableList.of(command));
```
aledsage: I think you need to assert that the return value is 0.

core/src/test/java/brooklyn/location/basic/SshMachineLocationIntegrationTest.java
```diff
@@ -75,7 +81,36 @@ public void testExtractingConnectablePassphraselessKey() throws Exception {
...
+    @Test(groups = "Integration")
+    public void testLocationScriptDirConfigIsRespected() throws Exception {
+        // For explanation of (some of) the magic behind this command, see http://stackoverflow.com/a/229606/68898
+        final String command = "if [[ \"$0\" == \"/var/tmp/\"* ]]; then true; else false; fi";
+
+        Map<String, Object> locationConfig = ImmutableMap.<String, Object>builder()
+                .put(SshMachineLocation.SCRIPT_DIR.getName(), "/var/tmp")
+                .build();
+
+        LocalhostMachineProvisioningLocation lhp = (LocalhostMachineProvisioningLocation) mgmt.getLocationRegistry().resolve("localhost", true, locationConfig).orNull();
```
aledsage: The default for the `managed` param is true, so you can just miss that out I think.

aledsage:
I had a feeling that brooklyn.properties config such as that below would work, but is that not true?

    brooklyn.location.named.myloc.scriptDir=/var/tmp

richardcloudsoft:
Addressed review comments, and rebased on latest master for good luck.

@aledsage regarding your comment, sadly that property did not work (although it should now). `SshMachineLocation` has always had that config key available to it, but it never wired it up to the `SshTool` invocations in a way that would work.

core/src/main/java/brooklyn/location/basic/SshMachineLocation.java
```diff
@@ -216,6 +217,13 @@ public String apply(HasConfigKey<?> input) {
                 }
             }));
 
+    /**
+     * The set of config keys on this location which become default values for properties when invoking an SSH
+     * operation.
+     */
+    public static final Set<ConfigKey<?>> SSH_CONFIG_GIVEN_TO_PROPS = ImmutableSet.<ConfigKey<?>>of(
```
ahgittin: `private` and/or `@Beta` ?

ahgittin:
LGTM.  Have not tested but I assume you have.

core/src/main/java/brooklyn/location/basic/SshMachineLocation.java
rdowner: Good question. I thought it should be `private` but then I saw that similar fields were `public` so I went with that. `@Beta` sounds like a good idea though.

rdowner:
Thanks @ahgittin. Have added `@Beta` annotation and will now proceed to merge.

PR #477, opened 2015-01-27
Disable SSLv3

neykov:
Also enable `$brooklyn:object` to use TypeCoercions enabled objects (i.e. for objects that don't have setters). This allows one to use:

```
services:
  brooklyn.config:
    webapp.enabledProtocols: ["https"]
    webapp.https.ssl:
      $brooklyn:object:
        objectType: brooklyn.entity.webapp.HttpsSslConfig
        object.fields:
          keystoreUrl: https://artifactory:AP5QT5YLhmBu3ET9Qe8wwkrNs3r@artifactory.mms.ibmcloud.com/artifactory/thirdparty/certs/faisalssaph1_keystore
          keystorePassword: changeit
          keyAlias: amm
```


ahgittin:
looks good to me.  there is no reason anyone should use SSLv3 so disabling it universally makes sense.

will merge.

PR #479, opened 2015-01-27
nginx support for upstream https, disable SSLv3

neykov:
Make the template config closer to the NginxDefaultConfigGenerator, support non-domain config.


ahgittin:
looks good.  i will add comments to the effect that the `SERVER_CONF_TEMPLATE_URL` key is only used if `SERVER_CONF_GENERATOR` is explicitly specified to point to  `NginxTemplateConfigGenerator`.

it would probably be cleaner if the DefaultConfigGenerator looked at `SERVER_CONF_TEMPLATE_URL` key and switched to use it if there is an explicit value set, just in case someone tries to use the key only.  but not needed for this PR.

neykov:
> it would probably be cleaner if the DefaultConfigGenerator looked at SERVER_CONF_TEMPLATE_URL key and switched to use it if there is an explicit value set, just in case someone tries to use the key only. but not needed for this PR.

It was like this in a previous version, but then changed to require explicit `SERVER_CONF_GENERATOR`, so I didn't dare to change it back.

PR #480, opened 2015-01-27
Release ports acquired when checking available

mattunderscorechampion:
Fixes issue [BROOKLYN-130](https://issues.apache.org/jira/browse/BROOKLYN-130). Ensures that the setReuseAddress is called before the port is bound. Removes the attempt to connect to the port before binding it as unnecessary. Corrected off by one error in loop so that port is not incremented beyond MAX_PORT_NUMBER.

### DIFF
--- a/utils/common/src/main/java/brooklyn/util/net/Networking.java
+++ b/utils/common/src/main/java/brooklyn/util/net/Networking.java
@@ -111,7 +111,7 @@ public class Networking {
                 ss.bind(new InetSocketAddress(localAddress, port));
 
                 // Check UDP port
-                ds = new DatagramSocket();
+                ds = new DatagramSocket(null);
                 ds.setSoTimeout(250);
                 ds.setReuseAddress(true);
                 ds.bind(new InetSocketAddress(localAddress, port));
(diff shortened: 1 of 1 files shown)


mattunderscorechampion:
Having reviewed the build failure.

A NullPointerException is thrown when parsing a string to URI.

[URI.create(..)](http://docs.oracle.com/javase/7/docs/api/java/net/URI.html#create(java.lang.String)) throws a NullPointerException if the value passed to it is null.

AbstractControllerImpl.computePortsAndUrls(..) is calling URI.create(..) with the value returned by AbstractControllerImpl.inferUrl(..).

AbstractControllerImpl.inferUrl(..) will return null if the domain is null. Either from the AbstractControllerImpl.getDomain(..), HostAndPort.getHostText(..) or LoadBalancer.HOSTNAME attribute.

It is not clear to me why changing the port assignment would affect the domain name.

Stacktrace fragment:
Caused by: java.lang.NullPointerException: null
	at java.net.URI$Parser.parse(URI.java:3004)
	at java.net.URI.<init>(URI.java:577)
	at java.net.URI.create(URI.java:839)
	at brooklyn.entity.proxy.AbstractControllerImpl.computePortsAndUrls(AbstractControllerImpl.java:314)
	at brooklyn.entity.proxy.AbstractControllerImpl.preStart(AbstractControllerImpl.java:299)

mattunderscorechampion:
Closing and reopening to see if build failure happens again

ahgittin:
LGTM - merging

PR #481, opened 2015-01-28
fix JcloudsMachineNamer

andreaturli:
- remove maxLenght for SoftLayer

### DIFF
--- a/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsMachineNamer.java
+++ b/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsMachineNamer.java
@@ -37,8 +37,6 @@ public class JcloudsMachineNamer extends CloudMachineNamer {
             return 39;
         if ("google-compute-engine".equals( setup.peek(JcloudsLocationConfig.CLOUD_PROVIDER) ))
             return 39;
-        if ("softlayer".equals( setup.peek(JcloudsLocationConfig.CLOUD_PROVIDER) ))
-            return 30;
         // TODO other cloud max length rules
         
         return null;  


aledsage:
Code change looks good, and I know you said you'd tested this - therefore happy to merge now.

aledsage:
For future reference, commit message could be more descriptive (e.g. "JcloudsMachineNamer: increase Softlayer max length").

ahgittin:
@aledsage i think andrea's description "fix JcloudsMachineNamer -- remove maxLenght for SoftLayer" is more accurate than your suggestion, spelling aside :) .  did you not see his expanded message or were you suggesting something else?

i confess i'm never sure how much to put in the one-line summary.  following @richardcloudsoft 's advice i now tend to keep the first line short-ish, so that it renders neatly (though in this case all of andrea's text could have been on one line).  however i do wonder how many people actually see the detail in the rest of the message, particularly when doing code reviews.

[ironic to be discussing this in a PR about max-length!]

andreaturli:
@ahgittin and @aledsage I tend to follow those notes in my git commit 
http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
do you think they are sensible advices?

richardcloudsoft:
@andreaturli that's the same guidelines that I follow, and that was the advice that @ahgittin alluded to! In fact I think it's worth proposing as approved guidelines for the Brooklyn project. Let's continue this discussion on the ML.

aledsage:
I'd missed the additional lines during code review, so am fine with the commit message. An improvement would have been to use more of the 50 chars of the first line to make it clearer in various git / github views.

Switching to mailing list discussion, entitled "[PROPOSAL] Guidelines for commit messages".

PR #482, opened 2015-01-28
Wait for STOP effector to complete before unamanging BrooklynNode

neykov:
Otherwise if there still are running tasks for the entity they will be canceled.

### DIFF
--- a/core/src/main/java/brooklyn/entity/basic/BrooklynTaskTags.java
+++ b/core/src/main/java/brooklyn/entity/basic/BrooklynTaskTags.java
@@ -46,6 +46,7 @@ import brooklyn.util.task.Tasks;
 import brooklyn.util.text.StringEscapes.BashStringEscapes;
 import brooklyn.util.text.Strings;
 
+import com.google.common.annotations.Beta;
 import com.google.common.base.Functions;
 import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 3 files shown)


core/src/main/java/brooklyn/entity/basic/BrooklynTaskTags.java
```diff
@@ -346,6 +347,26 @@ public static boolean isInEffectorTask(Task<?> task, @Nullable Entity entity, @N
         return false;
     }
 
+    @Beta
+    public static Task<?> getEffectorParentTask(Task<?> child, Effector<?> effector) {
```
ahgittin: rename as `getClosestEffectorTask()` ?  it's not necessarily the parent.  `closest` is used in the jquery sense.

might be useful eventually to refactor to use predicates, but probably not yet...

and add a one-line javadoc.

software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
```diff
@@ -81,9 +84,25 @@
         RendererHints.register(WEB_CONSOLE_URI, RendererHints.namedActionWithUrl());
     }
 
-    private class UnmanageThread extends Thread {
+    private static class UnmanageTask implements Runnable {
+        private Task<?> latchTask;
+        private Entity unmanageEntity;
+
+        public UnmanageTask(Task<?> latchTask, Entity unmanageEntity) {
```
ahgittin: mark `latchTask` as `@Nullable`

ahgittin:
good fixes; only minor comments then let's merge

neykov:
@ahgittin Addressed comments.

PR #483, opened 2015-01-28
add support for defining explicit inheritance for config keys

ahgittin:
obeyed for entities getConfig AND getAllConfig;
for locations only obeyed for getConfig.
notes in code explain why.

this is applied to a couple of the QuorumCheck config keys to fix an "on-fire" state in one of the stock examples.

### DIFF
--- a/core/src/main/java/brooklyn/entity/basic/EntityConfigMap.java
+++ b/core/src/main/java/brooklyn/entity/basic/EntityConfigMap.java
@@ -22,7 +22,6 @@ import static brooklyn.util.GroovyJavaMethods.elvis;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import java.util.Collections;
-import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
@@ -38,7 +37,6 @@ import brooklyn.config.ConfigMap;
 import brooklyn.event.basic.StructuredConfigKey;
 import brooklyn.management.ExecutionContext;
 import brooklyn.management.Task;
-import brooklyn.util.collections.MutableList;
 import brooklyn.util.collections.MutableMap;
 import brooklyn.util.config.ConfigBag;
 import brooklyn.util.flags.FlagUtils;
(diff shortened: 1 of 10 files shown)


core/src/main/java/brooklyn/entity/basic/EntityConfigMap.java
```diff
@@ -257,17 +282,35 @@ public void setInheritedConfig(Map<ConfigKey<?>, ? extends Object> vals, ConfigB
...
             } else {
+                // a config bag has discarded the keys, so we must assume default inheritance for things given that way
+                // unless we can infer a key; not a big deal, as we should have the key in inheritedConfig for everything
+                // which originated with a key ... but still, it would be nice to clean up the use of config bag!
                 inheritedConfigBag.putStringKey(name, value);
             }
         }
     }
     
+    private Map<ConfigKey<?>, ?> filterUninheritable(Map<ConfigKey<?>, ?> valsO) {
```
aledsage: This impl looks correct but confusing. I'd have expected something simpler like:

    private Map<ConfigKey<?>, ?> filterUninheritable(Map<ConfigKey<?>, ?> vals) {
        Map<ConfigKey<?>, Object> result = Maps.newLinkedHashMap();
        for (Map.Entry<ConfigKey<?>, ?> entry : vals.entrySet()) {
            if (isInherited(entry.getKey())) {
                result.put(entry.getKey(), entry.getValue());
            }
        }
        return result;
    }

core/src/main/java/brooklyn/location/basic/AbstractLocation.java
```diff
@@ -376,6 +389,10 @@ public boolean hasConfig(ConfigKey<?> key, boolean includeInherited) {
     
     @Override
     public ConfigBag getAllConfigBag() {
+        // TODO see comments in EntityConfigMap and on interface methods.
```
aledsage: I really need to rebase and finish off my PR that refactors config: https://github.com/apache/incubator-brooklyn/pull/320 !

utils/common/src/main/java/brooklyn/config/ConfigInheritance.java
```diff
@@ -0,0 +1,47 @@
...
+
+public abstract class ConfigInheritance {
+
+    public static final ConfigInheritance ALWAYS = new Always();
+    public static final ConfigInheritance NONE = new None();
+    
+    private ConfigInheritance() {}
+    
+    @Beta
+    public abstract boolean isInherited(ConfigKey<?> key, Object from, Object to);
```
aledsage: I wonder if we should use `BrooklynObject` rather than `Object` for `from` and `to`, but no strong feelings.

aledsage:
Looks good; only very minor comments. Good to merge when you've looked through those.

core/src/main/java/brooklyn/entity/basic/EntityConfigMap.java
ahgittin: was some mindless bugfixing, using yours now

core/src/main/java/brooklyn/location/basic/AbstractLocation.java
ahgittin: ooh yes please

utils/common/src/main/java/brooklyn/config/ConfigInheritance.java
ahgittin: config is completely independent of brooklyn.  i like that.

PR #484, opened 2015-01-28
obfuscate passwords and credentials in the gui (literal "shadow passwords")

ahgittin:
applies text shadowing to blur keys that say obvious things like "password" and "credential",
to config table and sensors table. clears up when you click it.
this prevents people looking over your shoulder from seeing things they shouldn't,
but it doesn't block REST access, and if you click on it you can still see it.
(this is a common trick done at AWS & SL, btw.)

a separate feature is to enforce visibility of sensors; this can be done with entitlements on a per-sensor basis
but it might be nice to have an easy entitlements mode where "sensitive" info is not available,
and options on config keys (similar to how i just did it with ConfigInheritance,
in https://github.com/apache/incubator-brooklyn/pull/483) to allow ConfigSensitivity.

an easy way to test is:

    curl -v -X POST -H "Content-Type: application/json" --data \"foo\" http://127.0.0.1:8082/v1/applications/YKH2Dp3E/entities/NN0BJzNA/sensors/my_secret

here's what it looks like:

screen shot 2015-01-28 at 16 05 23

### DIFF
--- a/docs/_extra/deploying-yaml.md
+++ b/docs/_extra/deploying-yaml.md
@@ -27,6 +27,8 @@ you can:
 $ curl -T ./blueprint.yaml -X POST http://localhost:8081/v1/applications
 {% endhighlight %}
 
+You may also need a `-H "Content-Type: application/yaml"` depending on type configuration.
+(Not usually for this, but often for other calls.)
 
 - In the web-console, select the "YAML" tab in the "Add Application" wizard:
 
--- a/usage/jsgui/src/main/webapp/assets/css/base.css
+++ b/usage/jsgui/src/main/webapp/assets/css/base.css
@@ -1455,13 +1455,13 @@ textarea.param-value {
 }
 
 /* For secret things */
-tr.secret-info span.value {
+.secret-info span.value {
     display: none;
 }
-tr.secret-info.secret-revealed span.value {
+.secret-info.secret-revealed span.value {
     display: inherit;
 }
-tr.secret-info.secret-revealed span.secret-indicator {
+.secret-info.secret-revealed span.secret-indicator {
     display: none;
 }
 
--- a/usage/jsgui/src/main/webapp/assets/css/base.css
+++ b/usage/jsgui/src/main/webapp/assets/css/base.css
@@ -1453,3 +1453,20 @@ textarea.param-value {
 #catalog-details-accordion {
     margin-top: 12px;
 }
+
+/* For secret things */
+tr.secret-info span.value {
+    display: none;
+}
+tr.secret-info.secret-revealed span.value {
+    display: inherit;
+}
+tr.secret-info.secret-revealed span.secret-indicator {
+    display: none;
+}
+
+.secret-indicator {
+    /* blur */
+    color: transparent;
+    text-shadow: 0 0 5px rgba(0,0,0,0.5);
+}
--- a/usage/jsgui/src/main/webapp/assets/js/util/brooklyn-utils.js
+++ b/usage/jsgui/src/main/webapp/assets/js/util/brooklyn-utils.js
@@ -162,6 +162,17 @@ define([
             return alternateMessage;
         }
     };
+    
+    secretWords = [ "password", "passwd", "credential", "secret", "private", "access.cert", "access.key" ];
+    
+    Util.isSecret = function (key) {
+        if (!key) return false;
+        key = key.toString().toLowerCase();
+        for (secretWord in secretWords)
+            if (key.indexOf(secretWords[secretWord]) >= 0)
+                return true;
+        return false; 
+    };
 
     return Util;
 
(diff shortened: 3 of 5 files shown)


ahgittin:
@alasdairhodge @bernd-clemenz one of you care to review?

ahgittin:
also note `RendererHints.censoredValue()`

ahgittin:
what is going on with this intermittent failure -- 

    2015-01-28 16:18:45,480 INFO  TESTNG FAILED: "Surefire test" - brooklyn.launcher.BrooklynWebServerTest.verifyHttps() finished in 10 ms
        java.lang.IllegalArgumentException: Unable to access URL brooklyn.webconsole.security.keystore.url: /home/jenkins/jenkins-slave/workspace/incubator-brooklyn-pull-requests%402/usage/launcher/target/test-classes/server.ks

pretty sure it is unrelated to this, but is starting to happen more often.  /cc @aledsage any ideas?

(close/reopen to kick off another test run)

docs/_extra/deploying-yaml.md
```diff
@@ -27,6 +27,8 @@ you can:
 $ curl -T ./blueprint.yaml -X POST http://localhost:8081/v1/applications
 {% endhighlight %}
 
+You may also need a `-H "Content-Type: application/yaml"` depending on type configuration.
+(Not usually for this, but often for other calls.)
```
alasdairhodge: Spurious diff for this PR.

alasdairhodge:
Meh. It's a cute CSS trick, but I don't think it takes us any closer to correctly solving the problem.

IMO, each sensor / CK (regardless of its name) should be optionally annotable as `@Sensitive` (redacted with `*` chars) or `@Hidden` (not even shown). It's up for discussion what should be visible to other Brooklyn entities (esp. those in different apps within the same server session), and what should be presented over the REST interface.

Some might argue that it's unacceptable to be able to access certain values client-side, even if it involves javascript hackery or viewing the HTML/JSON sources.

docs/_extra/deploying-yaml.md
ahgittin: i had to dig that out when testing this.  it's in the unused `_extra` folder of docs which we want to fold in elsewhere.

ahgittin:
@alasdairhodge RTFI (where I = issue).  this addresses the real problem of passwords being blatantly obvious to the person walking past your desk, where you have entitlement and sometimes requirement to see the real information.

controlling people's entitlements to data is a totally separate problem for which the entitlement subsystem can be used currently.  a good addition to this would be the ability to define sets of sensors where we might want to grant RO or RW or NO access (where NO is not an acronym :) ).  however i think a single `@Hidden` annotation is too coarse as it assumes there is only ever one such set.

while the distinction you draw between "you can know it exists but you can't see the value" and "you can't even know it exists" is a valid one, is there really any use case?  i'd drop it, in favour of only the latter.

ahgittin:
if you need to control entitlements see `EntitlementManagerAdapter`.  this could be easily extended to allow fine-grained access to config keys, following the pattern used for sensors.

btw the text which is blurred out is the word `[hidden]`, it is not the actual value.  but the general principle of the brooklyn console is that you either have root access -- in which case all data is accessible -- or you are coming in with a specific entitlement scheme which would among other things block access to the groovy console.

neykov:
Doesn't work for me:

screen shot 2015-01-30 at 16 55 07

Not supported on IE9-.

Also if I take a picture I can reverse-engineer the text  :wink:

alasdairhodge:
Suggest **not** blurring the placeholder text, but definitely style it so it won't be mistaken for the actual value. Perhaps greyed-out and italicised?

alasdairhodge:
(@ahgittin: Thanks for clarifying that it's just the placeholder value that's blurred! I originally assumed it was the actual config value.)

ahgittin:
@neykov can you try again - there was a bug where sometimes it would not apply the `secret-info` class

@neykov what do you mean you can reverse-engineer the text? placeholder text `(hidden)` is all you'd glean from a picture. NB the real data is on the web page, and a click will reveal it, so this is only meant as security from the person looking over your shoulder.

i much prefer the blur to anything grayed out, but let's see what people say

neykov:
Still something weird going on:
  * Creating an app with `my_secret` config, there is some unexplained delay while loading:
screen shot 2015-02-02 at 15 53 15
  * Once loaded the value part of the screen is missing:
screen shot 2015-02-02 at 15 53 19
  * After clicking around the entities and going back, the value is visible:
screen shot 2015-02-02 at 15 53 30
  * After adding a secret sensor (which displays fine), all config values are blurred:
screen shot 2015-02-02 at 15 57 30


As for reverse-engineering, I didn't realized the text is a static "(hidden)", consider it a try for a paranoia joke.

ahgittin:
aha, i've found one thing which might have caused this; could you try again? sorry + thanks

btw if anything funny does happen, can you check the errors in the JS console?

ahgittin:
@neykov could you try this again, see if i've fixed the issue for you?

neykov:
It's working fine now on modern browsers. On IE9- it's just hiding the value.

PR #485, opened 2015-01-29
JcloudsLocation: avoid NPE on user-creation

aledsage:
If jclouds node doesn’t know its operatingSystem, then was
getting NPE. Guard against that, and default to unix.


richardcloudsoft:
Merged. Does exactly what it says on the tin.

PR #486, opened 2015-01-29
refactor RebindManager to have a RebindIteration

ahgittin:
introduce new class and new methods for each phase. (no real changes.)

### DIFF
--- a/core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
+++ b/core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
@@ -1,0 +1,1086 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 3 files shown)


core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
```diff
@@ -0,0 +1,1086 @@
...
+            throw exceptionHandler.onFailed(e);
+            
+        } finally {
+            rebindActive.release();
+            RebindTracker.reset();
+        }
+    }
+    
+    protected void checkEnteringPhase(int targetPhase) {
+        phase++;
```
aledsage: Maybe should use an enum, with ordinal for readability of these calls? No strong feelings.

aledsage:
Looks good; much cleaner structure. I presume the body of the methods are the same as before (with just code moved around). I've really focused on just the structure during this review. Will merge now.

ahgittin:
Great.  Yes, the bodies are unchanged, apart from a few bits wrapping the start and catching exceptions, but I thought hard to confirm that.

I considered using an enum, but we'd need something like http://stackoverflow.com/questions/17664445/java-operator-for-enum -- probably worth doing it, putting a `<T extends Enumeration> T next(T item)` into `Enums`.

PR #487, opened 2015-01-29
Transformer clean-up and catalogItemId

ahgittin:
Transformer code uses simpler XSLT, cleaner java, and supports causing a `catalogItemId` change


neykov:
LGTM. I like how easy it's to introduce new transforms.

PR #488, opened 2015-01-29
set IS_IN_CLUSTER on couchbase primary node, fix problem noticed in #427

ahgittin:


software/nosql/src/main/java/brooklyn/entity/nosql/couchbase/CouchbaseClusterImpl.java
```diff
@@ -229,8 +229,13 @@ protected void doStart() {
...
                 
                 ((CouchbaseNode)getPrimaryNode()).rebalance();
-            } else if (getQuorumSize()>1) {
-                log.warn(this+" is not quorate; will likely fail later, but proceeding for now");
+            } else {
+                if (getQuorumSize()>1) {
+                    log.warn(this+" is not quorate; will likely fail later, but proceeding for now");
+                }
+                if (!serversToAdd.isEmpty()) {
+                    ((EntityInternal) Iterables.getOnlyElement(serversToAdd)).setAttribute(CouchbaseNode.IS_IN_CLUSTER, true);
```
aledsage: How do you know that `serversToAdd` will only have one element? If quorum size was 3, then presumably you could get into this code path with serversToAdd having size 2?

aledsage:
@ahgittin Not sure what the failing build was from - it's too old so has been deleted from apache history. 

Can you address the comment about `getOnlyElement` and thus force the build to be kicked off again? (or if my comment is wrong, then get another build by closing+reopening or whatever).

ahgittin:
@aledsage I think you're right, have fixed

aledsage:
Last test failure looks unrelated:

    io.brooklyn.camp.rest.resource.PlatformRestResourceTest.startServer (from TestSuite)

We've had a few failures for trying to get an available port. Maybe we need to mark those as integration tests :-(

software/nosql/src/main/java/brooklyn/entity/nosql/couchbase/CouchbaseClusterImpl.java
```diff
@@ -229,8 +229,13 @@ protected void doStart() {
                 }
                 
                 ((CouchbaseNode)getPrimaryNode()).rebalance();
-            } else if (getQuorumSize()>1) {
-                log.warn(this+" is not quorate; will likely fail later, but proceeding for now");
+            } else {
+                if (getQuorumSize()>1) {
```
aledsage: Should this really be `if (getQuorumSize() > serversToAdd.size())`? I won't let that block a merge of the PR though.

aledsage:
Looks good; merging.

software/nosql/src/main/java/brooklyn/entity/nosql/couchbase/CouchbaseClusterImpl.java
ahgittin: it's long-winded but i think logic is correct.  we've established there are `upNodes` aka `serversToAdd`, and if we are in this `else` block then also either we're not quorate or we are only adding a single server.  in the latter case we might be quorate hence this check before warning.

PR #489, opened 2015-01-29
Maven javac purity

ahgittin:
builds on #470, fixing issues i saw which appear to be java7 related

the stack overflow @neykov reports at https://github.com/apache/incubator-brooklyn/pull/470#issuecomment-71441391 is truly strange, and worrying.  i do not experience it -- this builds fine for me.

let's see what jenkins has to say!


ahgittin:
well, jenkins has the same problem, which might make it a bit easier to fix :)

ahgittin:
it's being invoked in `InternalLocationFactory`

    ((AbstractLocation)loc).configure(ConfigBag.newInstance().putAll(spec.getFlags()).putAll(spec.getConfig()).getAllConfig());

so it's not a constructor issue.  it makes me wonder:  under what circumstances does a method call go to the line number of the class declaration before going to the method?

ahgittin:
hey hey, this worked.  looks like the dodgy `super.configure(...)` call was getting confused between methods taking `Map<?,?>` and raw `Map` depending on java version.  

something to be watchful of!  /cc @aledsage

will tidy this up and then merge, as it's a trivial extension of #470

PR #490, opened 2015-01-29
Remove non-Apache repos

neykov:
Use `altDeploymentRepository` if alternative repos needed. `brooklyn.deployTo` remains only as a marker to activate source and gpg plugins during build.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,7 +43,7 @@
 
       brooklyn.explicitModules :: only builds explicit modules (instead of default modules)
 
-      brooklyn.deployTo={cloudsoftFilesystem,cloudsoftArtifactory,sonatype,apache} ::
+      brooklyn.deployTo={apache} ::
             :: required when deploying; specify the deployment target
 
       javadoc :: build javadoc (adds a minute or two; enabled automatically for target deploy)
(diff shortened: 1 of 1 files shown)


ahgittin:
on the fence.  we might still want to push to sonatype or cloudsoft repos, and i think it would be a hassle to do so without the extra repos.  OTOH they are not the canonical apache repos so arguably shouldn't be here.

@neykov can you explain the `altDeploymentRepository` suggestion? if there's an easy way to push elsewhere then i'm in favour of this.

@richardcloudsoft @aledsage wdyt?

neykov:
@ahgittin when building you can execute
```
mvn clean deploy -DaltDeploymentRepository=<repo_id>::default::<repo_url>
```
so you don't depend on the repos defined in the project.

rdowner:
@ahgittin The removed fragment of pom.xml could go into `~/.m2/settings.xml` for any person who needs it - it doesn't have to be in pom.xml. (Although it seems @neykov has a neater method with `altDeploymentRepository`)

With my Apache hat on: the only place Apache Brooklyn will ever be deployed to will be the Apache repository, so anything else is unnecessary, and it's probably not good to show favouritism for one company's convenience.

ahgittin:
thanks.  i'll rescue the repo settings and put them in my settings.

let's merge!

PR #491, opened 2015-01-29
None location

sjcorbett:
cc @aledsage

### DIFF
--- a/software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
+++ b/software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
@@ -78,7 +78,6 @@ import brooklyn.util.text.Strings;
 import brooklyn.util.time.Duration;
 
 import com.google.common.annotations.Beta;
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
@@ -603,7 +602,7 @@ public abstract class MachineLifecycleEffectorTasks {
                     if (stoppingProcess == null) {
                         msg.append(". Process stop skipped, pre-stop not finished?");
                     } else {
-                        msg.append(" because process stop has "+
+                        msg.append(" because process stop has ").append(
                                 (stoppingProcess.isDone() ? "finished abnormally" : "not finished"));
                     }
                     log.warn(msg.toString());
--- a/core/src/main/java/brooklyn/config/render/RendererHints.java
+++ b/core/src/main/java/brooklyn/config/render/RendererHints.java
@@ -268,7 +268,10 @@ public class RendererHints {
         return new NamedActionWithUrl<T>("Open", transform);
     }
 
-    /** use with care as this makes REST access difficult */
+    /**
+     * Forces the given sensor or config key's value to be censored. It will be
+     * presented as <code>********</code>.
+     */
     @Beta
     public static <T> DisplayValue<T> censoredValue() {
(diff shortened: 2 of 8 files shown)


sjcorbett:
An alternative to this idea from @aledsage is to add a location like this to your properties:
```
brooklyn.location.named.auto = localhost(name="auto")
brooklyn.location.named.auto.displayName = -- AUTO --
```

software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
```diff
@@ -561,7 +560,7 @@ public void stop(ConfigBag parameters) {
 
         boolean isEntityStopped = entity().getAttribute(SoftwareProcess.SERVICE_STATE_ACTUAL)==Lifecycle.STOPPED;
 
-        DynamicTasks.queue("pre-stop", new Callable<String>() { public String call() {
+        DynamicTasks.queue("pre-stop of " + entity() + ", submitted by " + Tasks.current(), new Callable<String>() { public String call() {
```
ahgittin: entity and submitted are available as metadata on the task already. generally i think it works better in the gui if the task name is short. is there a good reason to include it in the name?

ahgittin:
we've needed support for no-location in the gui. i was half-thinking of a way to allow multiple locations, or none, but for now an expicit "none" option is best.

testing it now.

ahgittin:
works a treat.  good to merge once you've considered the pre-stop task name.

software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
sjcorbett: It was helpful to discover what was being queued from where when a bug in Clocker caused massive task recursion. Hundreds of "pre-stops" in the logs were no help, the message above was. See [this comment](https://github.com/brooklyncentral/clocker/pull/91#issue-55657091) for more.

ahgittin: looking at that comment -- the task name is `pre-stop` but it shows `pre-stop of <entity>, submitted by Task[<ibid>]` (recursing).  where did the extra log info `of ..., submitted by ...` come from in that case?  because if i understand, when you *add* extra data to the task name, and it shows *less* information, i.e. just the name `pre-stop of <entity>, submitted by <other-task>` and not that extra log info.

my gut is saying this isn't the best way to address poor logging, as it would suggest we should apply similar enriched data to most tasks.

ahgittin:
@sjcorbett bump - is there a way we can fix the problem you found with logging in a generic way (ie so it applies to all tasks, not requiring a long-winded rename of individual tasks) ?

software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
sjcorbett: It's just a discrepancy between the code that ran to in the Clocker PR and the code listed above. Since it's contentious I'll just drop it from this pull request.

sjcorbett: But yes, worth noting that the best way for me to solve this was the unscalable approach of a manual addition to the logging.

sjcorbett:
@ahgittin Fire away.

aledsage:
Looks good; merging.

ahgittin:
@sjcorbett something tells me the missing log info is going to come back to haunt us; if it's top of mind maybe you could look at a general way to solve the problem you encountered!

PR #492, opened 2015-01-29
fixes JIRA issue 126

robertgmoss:


neykov:
`registeredTypeName` was deprecated in favor of `symbolicName`. Is the error in rest-client? Better fix it there to follow the changes in rest-server.
If backwards compatibility is needed for existing binaries, then no need to create an additional constructor argument (unless required for json serialization) for `registeredTypeName`, use `symbolicName` to initialize the field.

robertgmoss:
So the problem is that the json sent to the client contains this field, so
in trying to deserialize it expects this. Would it be better to use the
ignoreproperties annotation?

aledsage:
Jenkins build failure is due to errors below - can you fix those please?

    [ERROR] CatalogItemSummary.java[97:1] (whitespace) FileTabCharacter: Line contains a tab character.
    [ERROR] CatalogItemSummary.java[98:1] (whitespace) FileTabCharacter: Line contains a tab character.

robertgmoss:
@aledsage no need, better fix with PR/495 please close this PR

neykov:
@robertgmoss Does this mean that the user is trying to connect to an old instance of brooklyn? If so then keeping `registeredType` in the constructor, but initializing `symbolicName` with it sounds like a better idea, wdyt? The point is that the server will send either one or the other, but they mean the same thing.

ahgittin:
btw @robertgmoss only *you* have the power to close your own issues

this is a side-effect of the way apache use github; the committers can't actually close issues!

PR #493, opened 2015-01-30
fix maven javac purity on JDK7

andreaturli:
- fix for CAMP module: some setters on the builders were breaking the build with javac
- it also fixes javac compiler in intellij

@ahgittin and @neykov can you have a look?

### DIFF
--- a/camp/camp-base/src/main/java/io/brooklyn/camp/spi/AbstractResource.java
+++ b/camp/camp-base/src/main/java/io/brooklyn/camp/spi/AbstractResource.java
@@ -18,27 +18,26 @@
  */
 package io.brooklyn.camp.spi;
 
-import io.brooklyn.camp.commontypes.RepresentationSkew;
-
 import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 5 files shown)


camp/camp-base/src/main/java/io/brooklyn/camp/spi/AbstractResource.java
```diff
@@ -94,66 +93,66 @@ public RepresentationSkew getRepresentationSkew() {
...
-    private void setType(String type) {
+    void setType(String type) {
         this.type = type;
     }
-    private void setRepresentationSkew(RepresentationSkew representationSkew) {
+    void setRepresentationSkew(RepresentationSkew representationSkew) {
         this.representationSkew = representationSkew;
     }
-    public void setCustomAttribute(String key, Object value) {
+    void setCustomAttribute(String key, Object value) {
```
neykov: Did you decrease the visibiliby from `public` on purpose?

andreaturli: not sure it is needed to have `public` here, but happy to revert.

aledsage:
Looks good; merging.

PR #494, opened 2015-01-30
Update "User Guide" documentation

kitforbes:
Small changes to various files to correct spelling, style and clarity.


ahgittin:
this PR merges multiple lines into a single line in many places.  that shouldn't make a difference to markdown but it makes it much harder to work with in most editors, it loses any semantic breaks which the author may have intended to help other developers, and it makes it harder to see changes in github.

the other changes are good but please revert those places where multiple lines are merged together.

PR #495, opened 2015-01-30
Added ignore property for deprecated field

robertgmoss:
This replaces pull request 492 as a fix for JIRA issue 126

### DIFF
--- a/usage/rest-api/src/main/java/brooklyn/rest/domain/CatalogItemSummary.java
+++ b/usage/rest-api/src/main/java/brooklyn/rest/domain/CatalogItemSummary.java
@@ -22,6 +22,7 @@ import java.net.URI;
 import java.util.Map;
 
 import org.apache.commons.lang.builder.EqualsBuilder;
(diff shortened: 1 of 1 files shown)


ahgittin:
btw you can use `#` followed by number for references to other PR's -- eg #492 
and typically BROOKLYN-126 for references to Jira (don't think github picks that up but we will see!)

ahgittin:
nope, github doesn't recognise it, but a good reason to include that jira label is that *jira* will automatically pick it up as a comment

see e.g. https://issues.apache.org/jira/browse/BROOKLYN-126

ahgittin:
who is reading or writing `registeredType` ?  i can't see why it should be there at all.  hesitant about ignoring a property unless we know that there is an acceptable but ignorable reason it is being written.

ahgittin:
confirmed that this was caused by `SNAPSHOT` REST client talking to `M2-incubator` backend.

@robertgmoss can you close this issue?

PR #496, opened 2015-01-30
Fix projects build order

neykov:
The quickstart generated project depends on downstream-parent, but it is not yet built. Change modules build order.


ahgittin:
good fix, merging

PR #497, opened 2015-01-30
More fixes for Java7 pure javac build

ahgittin:
Meant to be a part of #489 but got left off!  I think this is preferred over #493 because it keeps the visibility nice.

@andreaturli can you try this?


andreaturli:
@ahgittin it looks good to me, and I think it is better having the setters totally private.
mvn cli and my IDE are both happy with your changes.

ahgittin:
thx @andreaturli, merging

PR #498, opened 2015-01-30
Java 8 compile fixes

neykov:
Problems are at places where generics are misused. Usually preceded by compiler warnings.


ahgittin:
works fine for me, and looks sensible -- merging

most of these seem to be caused by java 8 not wanting a raw `Thing` passed to a function which wants a `Thing<WithGenerics>` -- is that correct?

neykov:
The problem (with most of the spots) is passing a raw type which causes the return type to be erased. This by itself is not a problem, but if there are any chained methods, they are not resolved correctly.

Looks like related to http://bugs.java.com/bugdatabase/view_bug.do?bug_id=7144506

PR #499, opened 2015-02-01
fix bug where missing script dir causes null script dir explicitly set in ssh tool

ahgittin:
bug introduced in #476

affects all deployments to ssh machines where script dir is not expliclty set, causing NPE


ahgittin:
i'm going to merge this CTR because it is critical.  @aledsage @richardcloudsoft would one of you cast an eye over this?

richardcloudsoft:
Embarrassing :flushed: your fix looks good. I'll shout up if I see a problem in practice, but test coverage is good so I don't anticipate a problem.

PR #500, opened 2015-02-02
Softlayer safe name length

duncangrant:
After extensive testing it would appear that:
- anything up to 55 chars seems to be fine.  
- 56-58 characters you could jclouds exceptions while provisioning.  
- 59 chars up to around 80 ish you get a provisioned machine with no hostname 
- if you set the chars at higher than 100(ish) you get an error saying that the maximum length had been exceeded.

Someone else please test before merging.

### DIFF
--- a/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsMachineNamer.java
+++ b/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsMachineNamer.java
@@ -26,20 +26,22 @@ public class JcloudsMachineNamer extends CloudMachineNamer {
     public JcloudsMachineNamer(ConfigBag setup) {
         super(setup);
     }
-    
+
     /** returns the max length of a VM name for the cloud specified in setup;
      * this value is typically decremented by 9 to make room for jclouds labels */
     public Integer getCustomMaxNameLength() {
         // otherwise, for some known clouds which only allow a short name, use that length
-        if ("vcloud".equals( setup.peek(JcloudsLocationConfig.CLOUD_PROVIDER) )) 
+        if ("vcloud".equals( setup.peek(JcloudsLocationConfig.CLOUD_PROVIDER) ))
             return 24;
-        if ("abiquo".equals( setup.peek(JcloudsLocationConfig.CLOUD_PROVIDER) )) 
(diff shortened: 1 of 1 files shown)


andreaturli:
@duncangrant thanks for your analysis but now I'm confused

CloudLocationConfig.VM_NAME_MAX_LENGTH is 61, so without an explicit JcloudsMachineNamer exception for SoftLayer the max length would be 61 - 9 = 52, which is < 55 so it should be ok, according to your testing. Thoughts?

duncangrant:
I should Have tried running this with a debugger to see what is going on.  I'll let you know.

duncangrant:
@andreaturli 

I'm not sure i understand your question.  The 9 chars are taken from the length either way so when the custom length is set to 55 then 46 is used.

andreaturli:
@duncangrant Does this mean that the hard limit imposed by SoftLayer is 46 not 55, right?

duncangrant:
Yes - I should have been clearer.  The values I was reporting were the values in the JcloudsMachineNAmer - so the fqdn length not the hostname length

andreaturli:
@duncangrant could you please clarify what's the problem with the hostnames in SL with the current master? I've deployed a basic blueprint with jboss and the `hostname` command on the VM returns what I was expecting and `hostname -f` is ok, as well.

duncangrant:
@andreaturli So I think it is something to do with the length of the name of the blueprint.  I'll dig into the code a bit but can you confirm that the following yaml fails for you (try your hostname test):
```yaml
name: JBoss7
location: jclouds:softlayer:ams01
services:
```
whereas the following seems to work:
```yaml
name: jb
location: jclouds:softlayer:ams01
services:
```

andreaturli:
Thanks @duncangrant I think it is now more clear!
Yes, the first YAML fails for me as well. The reason seems to be the following:
using the default maxLength=61 - 9 = 52 allowed characters, but jclouds appends and extra suffix of typically 3 chars (*-xy). That means 55 chars which is exactly the limit you found.

Happy to merge your PR, if you could first rebase and squash into one commit only to have a cleaner git history.

duncangrant:
@andreaturli thanks - Squashed and rebased

PR #501, opened 2015-02-02
fix time-sensitive sensor derivation when rebinding to old sensors

ahgittin:
before this, autoscaler might scale out on rebind, because the
last total reqs might be 100, if data is v old, current value might be 10000;
a recent timestamp is attached to initial value, meaning it computes 9900 new reqs in a 1s window.

this forces an invalid timestamp for enrichers

better fix would be to store timestamp on sensors themselves

### DIFF
--- a/policy/src/main/java/brooklyn/enricher/RollingTimeWindowMeanEnricher.java
+++ b/policy/src/main/java/brooklyn/enricher/RollingTimeWindowMeanEnricher.java
@@ -21,17 +21,18 @@ package brooklyn.enricher;
 import java.util.Iterator;
 import java.util.LinkedList;
 
-import com.google.common.base.Preconditions;
-
-import brooklyn.catalog.Catalog;
+import brooklyn.config.ConfigKey;
 import brooklyn.enricher.basic.AbstractTypeTransformingEnricher;
 import brooklyn.entity.Entity;
+import brooklyn.entity.basic.ConfigKeys;
 import brooklyn.event.AttributeSensor;
 import brooklyn.event.Sensor;
 import brooklyn.event.SensorEvent;
 import brooklyn.util.flags.SetFromFlag;
 import brooklyn.util.javalang.JavaClassNames;
 import brooklyn.util.time.Duration;
+
+import com.google.common.base.Preconditions;
 
 /**
  * Transforms {@link Sensor} data into a rolling average based on a time window.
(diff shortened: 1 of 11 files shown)


ahgittin:
took a while to get all the tests happy with this change but finally there

ahgittin:
@aledsage @grkvlt can you look at this PR also?  without it, odd things happen on rebind with policies.

policy/src/main/java/brooklyn/enricher/RollingTimeWindowMeanEnricher.java
```diff
@@ -103,31 +119,69 @@ public void onEvent(SensorEvent<T> event) {
...
+    public ConfidenceQualifiedNumber getAverage(long fromTimeExact) {
+        return getAverage(fromTimeExact, 0);
+    }
+    
+    public ConfidenceQualifiedNumber getAverage(long fromTime, long graceAllowed) {
         if (timestamps.isEmpty()) {
             return lastAverage = new ConfidenceQualifiedNumber(lastAverage.value, 0.0d);
         }
+        
+        // (previously there was an old comment here, pre-Jul-2014,
```
aledsage: Include a `TODO` in this comment - it's the sort of comment we want to delete if/when we figure out what the old comment meant, or we're happy with it. Hence it justifies a TODO because we want to change it.

policy/src/main/java/brooklyn/enricher/RollingTimeWindowMeanEnricher.java
```diff
@@ -103,31 +119,69 @@ public void onEvent(SensorEvent<T> event) {
...
+                // causing spurious resizes, so allow it to be configured, and
+                // by default require a high value
+
+                // TODO would be nice to include timestamp, etc
+                entity.setAttribute((AttributeSensor<Double>)target, average.value); 
+            }
+        }
     }
     
     public ConfidenceQualifiedNumber getAverage() {
```
aledsage: Nothing seems to call the `getAverage()` method, so the `TIMESTAMP_GRACE_TIME` is never used. Or am I missing something?

aledsage:
@ahgittin I'm not sure I follow the fix.

Is the main fix that `BasicSensorEvent` now defaults to the current timestamp, rather than 0?

I think I follow the change to `Transformer`, but what about `Propagator` where it calls `emitAllAttributes` when first setting all entities? Would the same logic as for Transformer not apply for Propagator as well?

policy/src/main/java/brooklyn/enricher/RollingTimeWindowMeanEnricher.java
ahgittin: In that case I'll take it out

ahgittin:
thanks @aledsage; i've implemented your suggestions, apart from `Propagator` where I've now added a TODO.

you are right in your thinking that we want this for `Propagator`; I simply hadn't noticed it.  it was `Transformer` that was causing the errors i was encountering.  however fixing it for propagator seems a little less clear-cut, and i've not tested it, so all i have done is the todo.

I think to fix, we need to make a timestamp a first-class property of sensors.

okay to merge as changed?

aledsage:
Looks good; merging.

PR #503, opened 2015-02-04
Fixes RabbitMQ installation on Centos 6 and 7

nakomis:


software/messaging/src/main/java/brooklyn/entity/messaging/rabbit/RabbitSshDriver.java
```diff
@@ -24,6 +24,7 @@
 import java.util.List;
 import java.util.Map;
 
+import com.google.api.client.repackaged.com.google.common.base.Strings;
```
sjcorbett: Wrong package. Should be `com.google.common.base.Strings`.

sjcorbett: I'm tempted to set Checkstyle to completely forbid `com.google.api.client.repackaged`.

software/messaging/src/main/java/brooklyn/entity/messaging/rabbit/RabbitSshDriver.java
```diff
@@ -73,7 +81,15 @@ public void install() {
         List<String> urls = resolver.getTargets();
         String saveAs = resolver.getFilename();
 
+        String osMajorVersion = getMachine().getOsDetails().getVersion();
+        osMajorVersion = osMajorVersion.indexOf(".") > 0 ? osMajorVersion.substring(0, osMajorVersion.indexOf('.')) : osMajorVersion;
```
sjcorbett: `osMajorVersion` and `osArchitecture` are both annotated `@Nullable` so there's a chance of `NullPointerExceptions` here.

software/messaging/src/test/java/brooklyn/entity/messaging/rabbit/RabbitEc2LiveTest.java
```diff
@@ -81,7 +81,16 @@ private Channel getAmqpChannel(RabbitBroker rabbit) throws Exception {
         Channel channel = conn.createChannel();
         return channel;
     }
-    
+
+    @Override
+    public void test_CentOS_5_6() throws Exception {
+        // Not supported. The EPEL repository described here at [1] does not contain erlang, and the
```
sjcorbett: Could throw a [`SkipException`](http://testng.org/javadoc/org/testng/SkipException.html) to be explicit to the build.

software/messaging/src/main/java/brooklyn/entity/messaging/rabbit/RabbitSshDriver.java
nakomis: Good idea. I usually spot it when my IDE picks this one, but obviously missed it this time

nakomis:
PR comments addressed

software/messaging/src/main/java/brooklyn/entity/messaging/rabbit/RabbitSshDriver.java
```diff
@@ -72,8 +80,15 @@ public void preInstall() {
     public void install() {
         List<String> urls = resolver.getTargets();
         String saveAs = resolver.getFilename();
+        String osMajorVersion = Strings.nullToEmpty(getMachine().getOsDetails().getVersion());
+        osMajorVersion = osMajorVersion.indexOf(".") > 0 ? osMajorVersion.substring(0, osMajorVersion.indexOf('.')) : osMajorVersion;
+        String epelVersion = Strings.nullToEmpty(CENTOS_VERSION_TO_EPEL_VERSION.get(osMajorVersion));
+        String osArchitecture = Strings.nullToEmpty(getMachine().getOsDetails().getArch());
```
sjcorbett: Is this behaviour desirable? NPEs are guarded against but it could generate URLs like `http://download.fedoraproject.org/pub/epel///epel-release-.noarch.rpm`. Should we pick a default if we can't detect versions and architectures? Or it could even just fail.

nakomis:
Added selection of default for EPEL package download

software/messaging/src/main/java/brooklyn/entity/messaging/rabbit/RabbitSshDriver.java
```diff
@@ -44,6 +45,13 @@
 public class RabbitSshDriver extends AbstractSoftwareProcessSshDriver implements RabbitDriver {
 
     private static final Logger log = LoggerFactory.getLogger(RabbitSshDriver.class);
+
+    // See http://fedoraproject.org/wiki/EPEL/FAQ#howtouse
+    private static final Map<String, String> CENTOS_VERSION_TO_EPEL_VERSION = ImmutableMap.of(
```
aledsage: It would make sense to do this somewhere generic (e.g. `BashCommands`) as EPEL is not specific to Rabbit or even to Erlang. However, I'm ok with merging it here and then it being moved.

I don't like just "waiting for the second usage" before refactoring, because downstream projects may want to use that - those projects should ideally be written without having to change core.

software/messaging/src/test/java/brooklyn/entity/messaging/rabbit/RabbitEc2LiveTest.java
```diff
@@ -81,7 +82,16 @@ private Channel getAmqpChannel(RabbitBroker rabbit) throws Exception {
         Channel channel = conn.createChannel();
         return channel;
     }
-    
+
+    @Override
+    public void test_CentOS_5_6() throws SkipException {
+        // Not supported. The EPEL repository described here at [1] does not contain erlang, and the
```
aledsage: Agree with @sjcorbett `throw new SkipException()` means we don't pretend it worked in the test results.

aledsage:
Thanks @nakomis and @sjcorbett - merging.

PR #504, opened 2015-02-04
Forbid imports from com.google.api.client.repackaged

sjcorbett:
Prompted by #503 and bf38d7f7c27e51938e59d8e6ebc75fa7df889f37. We could also prohibit `com.google.api.client` entirely or `..client.util` additionally.


ahgittin:
LGTM

PR #478, opened 2015-01-27
CLI better default, port/https control, and version reporting

ahgittin:
* adds config key for web-console port
* adds CLI for https
* port defaults to 8443+ if no port specified and using https
* default if no CLI options is to show help and an error
* some https test fixes
* discovers and reports Brooklyn OSGi metadata in log, including git SHA1 (and `brooklyn info`)

PR #506, opened 2015-02-06
Active partial rebind allowing catalog item version change dynamically

ahgittin:
No GUI support (yet) for changing versions, but you can test this in groovy console with

    mgmt.rebindManager.rebindPartialActive(
        brooklyn.entity.rebind.transformer.CompoundTransformer.builder().changeCatalogItemId(
            "item", "1.0", "item", "2.0" /* old and new catalog item+version */).build(),
        "abcd1234" /* entity id */);


ahgittin:
requires #478 

also fixes deserialization of entity specs with catalog items.  see the two `ActivePartial...Test` classes. /cc @neykov @bhairavi25

aledsage:
Test failure is `brooklyn.management.ha.HighAvailabilityManagerInMemoryTest.tearDown`:

    Caused by: java.lang.IllegalStateException: Has different management context: null; expected LocalManagementContext[A6LXdOF0-oBWEBSQn]
    	at brooklyn.management.internal.EntityManagementSupport.onManagementStopping(EntityManagementSupport.java:271)
    	at brooklyn.management.internal.LocalEntityManager$3.apply(LocalEntityManager.java:400)
    	at brooklyn.management.internal.LocalEntityManager$3.apply(LocalEntityManager.java:397)
    	at brooklyn.management.internal.LocalEntityManager.recursively(LocalEntityManager.java:526)
    	at brooklyn.management.internal.LocalEntityManager.recursively(LocalEntityManager.java:531)
    	at brooklyn.management.internal.LocalEntityManager.unmanage(LocalEntityManager.java:397)
    	at brooklyn.management.internal.LocalEntityManager.unmanage(LocalEntityManager.java:356)
    	at brooklyn.management.internal.LocalEntityManager.unmanage(LocalEntityManager.java:352)
    	at brooklyn.entity.basic.AbstractApplication.stop(AbstractApplication.java:222)

api/src/main/java/brooklyn/basic/AbstractBrooklynObjectSpec.java
```diff
@@ -121,4 +121,21 @@ protected final void checkIsImplementation(Class<?> val, Class<? super T> requir
         if (Modifier.isAbstract(val.getModifiers())) throw new IllegalStateException("Implementation "+val+" is abstract, but must be a non-abstract class");
     }
 
+    @Override
+    public boolean equals(Object obj) {
+        if (obj==null) return false;
```
aledsage: I'm slightly hesitant about equals/hashCode on mutable objects. It potentially breaks the contract if two objects that were equal then become not-equal and the hashCode changes (http://stackoverflow.com/a/4718059/1393883).

Guessing this is just used for tests? Perhaps having a utility class for comparisons would be safer?

core/src/main/java/brooklyn/BrooklynVersion.java
```diff
@@ -19,80 +19,215 @@
...
+  
   public static final BrooklynVersion INSTANCE = new BrooklynVersion();
+  
+  private final Properties versionProperties = new Properties();
+  
+  public BrooklynVersion() {
+      // we read the maven pom metadata and osgi metadata and make sure it's sensible
+      // everything is put into a single map for now (good enough, but should be cleaned up)
+      readPropertiesFromMavenResource(BrooklynVersion.class.getClassLoader());
+      readPropertiesFromOsgiResource(BrooklynVersion.class.getClassLoader(), "org.apache.brooklyn.core");
```
aledsage: Use constant `BROOKLYN_CORE_SYMBOLIC_NAME`

core/src/test/java/brooklyn/BrooklynVersionTest.java
```diff
@@ -19,18 +19,63 @@
...
+      Assert.assertTrue(v==null || BrooklynVersion.get().equals(v), v);
+  }
+  
+  @Test
+  public void testGetOsgiSha1() {
+      String sha1 = BrooklynVersion.INSTANCE.getSha1FromOsgiManifest();
+      log.info("sha1: "+sha1);
+      if (Strings.isNonBlank(sha1) || BrooklynVersion.isDevelopmentEnvironment())
+          return;
+      // we might not have a SHA1 if it's a standalone source build; just log warn in that case
```
aledsage: Longer term, would prefer to create a special test group for things to be skipped with a source build. Folk aren't going to notice this warning when running the test suite.

core/src/main/java/brooklyn/BrooklynVersion.java
```diff
@@ -19,80 +19,215 @@
...
+  }
+
+  /** @deprecated since 0.7.0, redundant with {@link #get()} */ @Deprecated
+  public static String getVersionFromStatic() {
+      return VERSION_FROM_STATIC;
   }
 
   public static String get() {
-    return INSTANCE.version;
+    return getVersionFromStatic();
```
aledsage: Would prefer this to call `getVersion()`, rather than the deprecated `getVersionFromStatic()`.

ahgittin:
i've fixed the failed test, hopefully buildhive agrees.  will look through your other comments when you're finished.

also some new comments on a slow location persistence leak (no one unmanages locations created for an entity when that entity is cleaned up, /cc @neykov)

core/src/main/java/brooklyn/catalog/internal/CatalogUtils.java
```diff
@@ -227,4 +228,13 @@ public static String getVersionedId(String id, String version) {
         }
     }
 
+    @Beta
+    public static EntitySpec<?> createEntitySpec(ManagementContext mgmt, CatalogItem<?, ?> catalogItem) {
```
aledsage: This is currently only used for tests - should we move it to `src/test/java` instead?

My concern is that this doesn't necessarily do the full thing that the original `CatalogItem` did - for example, the `CatalogItem.getPlanYaml()` is ignored (i.e. whatever other config was supplied in that yaml will be ignored), and also it assumes the `CatalogItem` is for an `EntitySpec`. That's all fine to assume for tests, but doesn't feel right for core code (even if it is marked `@Beta`).

core/src/main/java/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -130,12 +130,14 @@ public ZoneFailureDetector apply(final String input) {
...
     private static final Function<Collection<Entity>, Entity> defaultRemovalStrategy = new Function<Collection<Entity>, Entity>() {
         @Override public Entity apply(Collection<Entity> contenders) {
-            // choose newest entity that is stoppable
+            // choose newest entity that is stoppable, or if none are stoppable take the newest non-stoppable
             long newestTime = 0;
             Entity newest = null;
 
             for (Entity contender : contenders) {
-                if (contender instanceof Startable && contender.getCreationTime() > newestTime) {
+                boolean newer = contender.getCreationTime() > newestTime;
```
aledsage: Is there a unit test for this, using the `DynamicCluster` with non-startable entities?

core/src/main/java/brooklyn/entity/rebind/ActivePartialRebindIteration.java
```diff
@@ -0,0 +1,158 @@
...
+    protected boolean isRebindingActiveAgain() {
+        return true;
+    }
+    
+    public void setObjectIterator(Iterator<BrooklynObject> objectsToRebind) {
+        this.objectsToRebindInitial = objectsToRebind;
+    }
+    
+    public void applyTransformer(CompoundTransformer transformer) {
+        if (transformer!=null)
```
aledsage: I still have a general preference for us not to ignore nulls like this. It's a serious risk of ignoring problems with surprising behaviour, versus the convenience of the caller not having to check for null. The time I lean towards accepting nulls is if it's in a fluent API.

core/src/main/java/brooklyn/entity/rebind/ActivePartialRebindIteration.java
```diff
@@ -0,0 +1,158 @@
...
+            transformers.add(transformer);
+    }
+    
+    @Override
+    protected void doRun() throws Exception {
+        checkEnteringPhase(1);
+        Preconditions.checkState(readOnlyRebindCount.get()==Integer.MIN_VALUE, "Partial rebind in read-only mode not supported");
```
aledsage: This seems a strange ay to check if the state is read-only mode. I wouldn't have guess it from looking at the variable name.

core/src/main/java/brooklyn/entity/rebind/ActivePartialRebindIteration.java
```diff
@@ -0,0 +1,158 @@
...
+                objectsToRebindFinal.addAll( ((EntityInternal)bo).getEnrichers() );
+                objectsToRebindFinal.addAll( ((EntityInternal)bo).feeds().getFeeds() );
+            }
+        }
+        
+        // get serialization
+        for (BrooklynObject bo: objectsToRebindFinal) {
+            Memento m = ((BrooklynObjectInternal)bo).getRebindSupport().getMemento();
+            BrooklynMementoPersister p = rebindManager.getPersister();
+            String mr = ((BrooklynMementoPersisterToObjectStore)p).getMementoSerializer().toString(m);
```
aledsage: Shame we have to cast this. It makes it feel pointless to leave the old implementations of `BrooklynMementoPersister` deprecated if one will potentially get runtime exceptions using it (better to have deleted the classes to give compile-time errors, depending under what circumstances this code-path will be hit).

The `BrooklynMementoPersisterInMemory` feels like a useful thing to keep in the test packages, but maybe not even worth that if we just accept the performance overhead in our test cases of always using `BrooklynMementoPersisterToObjectStore`.

core/src/main/java/brooklyn/entity/rebind/ActivePartialRebindIteration.java
```diff
@@ -0,0 +1,158 @@
...
+        this.objectsToRebindInitial = objectsToRebind;
+    }
+    
+    public void applyTransformer(CompoundTransformer transformer) {
+        if (transformer!=null)
+            transformers.add(transformer);
+    }
+    
+    @Override
+    protected void doRun() throws Exception {
```
aledsage: Could this be implemented by just overriding `loadManifestFiles()` and `rebuildCatalog()`, rather than `doRun()`?

api/src/main/java/brooklyn/basic/AbstractBrooklynObjectSpec.java
ahgittin: i don't think it breaks the contract of hashcode -- if a spec is modified in such a way that the hashcode changes, then the equality semantics will have changed in the expected way.

as we already have `Object.equals/hashcode` i think this implementation is an improvement, and putting things into a test convenience would just make things harder to find.

making the spec immutable would be the elegant way to solve this; i wouldn't be strongly opposed although i don't see any major benefit at this time for the extra code. (is spec mutability confusion really a problem?)

core/src/test/java/brooklyn/BrooklynVersionTest.java
ahgittin: the warning is useful if you're running the test interactive

yes it would be cleaner to have a group skipped on non-git source builds, or a flag indicating that status, but not really worth the effort now

core/src/main/java/brooklyn/BrooklynVersion.java
ahgittin: now does `return INSTANCE.getVersion();`.  the other option was to return the static constant, but feels safer to have the method fail if `INSTANCE` couldn't be initialized.

core/src/main/java/brooklyn/catalog/internal/CatalogUtils.java
ahgittin: good spot.  i completely agree -- i had missed that config etc would be ignored (d'oh).

have moved to `CatalogTestUtils.createEssentialEntitySpec`.

core/src/main/java/brooklyn/entity/group/DynamicClusterImpl.java
ahgittin: have added

core/src/main/java/brooklyn/entity/rebind/ActivePartialRebindIteration.java
ahgittin: i don't agree but i've changed it :)

core/src/main/java/brooklyn/entity/rebind/ActivePartialRebindIteration.java
ahgittin: have added another check and improved messages

core/src/main/java/brooklyn/entity/rebind/ActivePartialRebindIteration.java
ahgittin: there is a `SingleThreadedInMemoryStoreObjectAccessor` for use with the `BrooklynMementoPersisterToObjectStore`, giving minimal overhead, and a more faithful test pathway.

the deprecated classes aren't used and can be deleted; we just need to cut 070-GA. i don't think there's point in adding support or compile-time checks for the deprecated classes in this pathway.

core/src/main/java/brooklyn/entity/rebind/ActivePartialRebindIteration.java
ahgittin: yes, that's cleaner as it turns out

core/src/main/java/brooklyn/entity/rebind/RebindContextLookupContext.java
```diff
@@ -60,6 +66,9 @@ public RebindContextLookupContext(ManagementContext managementContext, RebindCon
     @Override public Entity lookupEntity(String id) {
         Entity result = rebindContext.getEntity(id);
         if (result == null) {
+            result = managementContext.lookup(id, Entity.class);
```
aledsage: Should this be guarded with `lookInManagementContext`? That field is never used.

aledsage: But neither is the constructor to set `lookInManagementContext` to true. Can we just delete that?

ahgittin: yes - i didn't want to break compatibility, but now i see that i did, and that the only usages are during rebind.  have removed that field and added javadoc that this class looks in both (which helps justify its existence besides the rebind context).

core/src/main/java/brooklyn/management/internal/ManagementTransitionMode.java
```diff
@@ -0,0 +1,121 @@
...
+ * under the License.
+ */
+package brooklyn.management.internal;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Preconditions;
+
+public class ManagementTransitionMode {
```
aledsage: Could do with some javadoc.

core/src/main/java/brooklyn/entity/rebind/RebindManagerImpl.java
```diff
@@ -361,6 +364,38 @@ public void stop() {
         if (persistenceStoreAccess != null) persistenceStoreAccess.stop(true);
     }
     
+        
+    public void rebindPartialActive(CompoundTransformer transformer, Iterator<BrooklynObject> objectsToRebind) {
+        final ClassLoader classLoader = 
+            managementContext.getCatalog().getRootClassLoader();
+        final RebindExceptionHandler exceptionHandler = 
+            RebindExceptionHandlerImpl.builder()
+                .danglingRefFailureMode(danglingRefFailureMode)
```
aledsage: Do we definitely want to use the same exception handler as was used at start-up? What happens if there is an unhandled exception? What state will the entities be left in? I presume that if the exception is early (e.g. when loading/transforming state etc) then it's like a no-op, but if during the entity's manage() then we'll have changed some entities but not others?

Doing the partial-rebind at run-time feels more dangerous than at startup so possibly want different exception handling (e.g. on startup, might be happy to fail-fast, so JVM exits).

core/src/main/java/brooklyn/entity/rebind/RebindManagerImpl.java
```diff
@@ -361,6 +364,38 @@ public void stop() {
...
+        iteration.setObjectIterator(objectsToRebind);
+        iteration.applyTransformer(transformer);
+        iteration.run();
+    }
+    
+    public void rebindPartialActive(CompoundTransformer transformer, String ...objectsToRebindIds) {
+        List<BrooklynObject> objectsToRebind = MutableList.of();
+        for (String objectId: objectsToRebindIds) {
+            BrooklynObject obj = managementContext.lookup(objectId);
+            if (obj instanceof Entity) obj = Entities.deproxy((Entity)obj);
```
aledsage: Do these have to be deproxy'ed here? Callers of `rebindPartialActive(CompoundTransformer, Iterator<BrooklynObject>) ` probably wouldn't realise that they can't just pass proxy objects. Therefore I'd be inclined to have the deproxy'ing happen somehwere in that other `rebindPartialActive` method.

ahgittin: agree

core/src/main/java/brooklyn/entity/rebind/persister/XmlMementoSerializer.java
```diff
@@ -341,5 +353,151 @@ public Object unmarshal(HierarchicalStreamReader reader, UnmarshallingContext co
...
+    
+    protected void acquireXstreamLock() {
+        synchronized (xstreamLockOwner) {
+            while (true) {
+                if (xstreamLockOwner.compareAndSet(null, Thread.currentThread()) || 
+                    Thread.currentThread().equals( xstreamLockOwner.get() )) {
+                    break;
+                }
+                try {
+                    xstreamLockOwner.wait(1000);
```
aledsage: Why is this calling `wait(1000)` when nothing calls `notifyAll()`?

core/src/main/java/brooklyn/entity/rebind/persister/XmlMementoSerializer.java
```diff
@@ -341,5 +353,151 @@ public Object unmarshal(HierarchicalStreamReader reader, UnmarshallingContext co
...
+    AtomicReference<Thread> xstreamLockOwner = new AtomicReference<Thread>();
+    int lockCount;
     
+    /** Must be accompanied by a corresponding {@link #popXstreamCustomClassLoader()} when finished. */
+    @SuppressWarnings("deprecation")
+    protected void pushXstreamCustomClassLoader(BrooklynClassLoadingContext clcNew) {
+        acquireXstreamLock();
+        BrooklynClassLoadingContext oldClc;
+        if (!contexts.isEmpty()) {
+            oldClc = contexts.peek();
```
aledsage: Looks dangerous; no synchronization on `xstreamLockOwner` so is assuming no other thread will have popped in between the check to `isEmpty` and `peek`?

core/src/main/java/brooklyn/entity/rebind/RebindManagerImpl.java
ahgittin: have added a TODO

the update-proxy-then-unmanage-old is mostly atomic so most failures should be pretty safe, but i'm sure there will be edge conditions

when we better understand the failure dimensions (and have visualization) then we can tidy this up

core/src/main/java/brooklyn/entity/rebind/persister/XmlMementoSerializer.java
ahgittin: added `notifyAll` on release

core/src/main/java/brooklyn/entity/rebind/persister/XmlMementoSerializer.java
ahgittin: he has acquired the xstream lock at this point

core/src/main/java/brooklyn/entity/rebind/persister/XmlMementoSerializer.java
```diff
@@ -341,5 +353,151 @@ public Object unmarshal(HierarchicalStreamReader reader, UnmarshallingContext co
...
+    }
+    
+    Stack<BrooklynClassLoadingContext> contexts = new Stack<BrooklynClassLoadingContext>();
+    Stack<ClassLoader> cls = new Stack<ClassLoader>();
+    AtomicReference<Thread> xstreamLockOwner = new AtomicReference<Thread>();
+    int lockCount;
     
+    /** Must be accompanied by a corresponding {@link #popXstreamCustomClassLoader()} when finished. */
+    @SuppressWarnings("deprecation")
+    protected void pushXstreamCustomClassLoader(BrooklynClassLoadingContext clcNew) {
```
aledsage: Gut feel is that this feels complicated, but not sure how to simplify it so happy as-is (but see other related comments).

core/src/main/java/brooklyn/entity/rebind/ActivePartialRebindIteration.java
```diff
@@ -0,0 +1,158 @@
...
+import brooklyn.mementos.BrooklynMementoRawData;
+import brooklyn.mementos.BrooklynMementoRawData.Builder;
+import brooklyn.mementos.Memento;
+import brooklyn.util.collections.MutableList;
+import brooklyn.util.collections.MutableSet;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * Pauses a set of existing entities, writes their state, applies a transformation, then reads them back.
```
aledsage: Could do with more javadoc, to say that this is for use on a running primary management node etc.

core/src/main/java/brooklyn/management/internal/LocalEntityManager.java
```diff
@@ -354,23 +362,23 @@ private void unmanage(final Entity e, ManagementTransitionMode mode, boolean has
...
-            // (called from manage(...)
+            // don't unmanage or even clear its fields, because there might be references to it
             
-            if (mode==ManagementTransitionMode.REBINDING_NO_LONGER_PRIMARY) {
-                // when migrating away, these all need to be called
+            if (mode.wasReadOnly()) {
+                // if coming *from* read only; nothing needed
+            } else {
+                if (!mode.wasPrimary()) {
+                    log.warn("Unexpected mode "+mode+" for unmanage-replace "+e+" (applying anyway)");
```
aledsage: I'm a bit worried about these kind of warnings around the place, in that we might not notice them happening until it's too late (e.g. until a customer reports a problem). Longer term, is there something we can do to more strongly assert what modes are expected/allowed in different circumstances?

(I'm not specifically talking about this particular piece of code; more in general so those maintaining the code know exactly what each code branch is for. On balance, agree it's often better to warn and continue than fail completely in production - but in tests and QA then failing fast is arguably better! Perhaps it should be an `assert`?)

core/src/test/java/brooklyn/entity/rebind/ActivePartialRebindTest.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Test(groups="Integration")
+    public void testRebindCheckingMemoryLeak() throws Exception {
+        TestEntity c1 = origApp.addChild(EntitySpec.create(TestEntity.class));
+        Entities.manage(c1);
+        c1.setConfig(TestEntity.CONF_NAME, Strings.makeRandomId(1000000));
+        
+        gcAndLog("before");
+        long used0 = Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory();
```
aledsage: Minor preference: I'd set assign this to -1 or some such, given it will always be overwritten when `i==5`.

api/src/main/java/brooklyn/basic/AbstractBrooklynObjectSpec.java
aledsage: You're right - it's not braking the official hashCode contract. It would just be wrong to put a spec in a `HashSet` and then modify it.

The spec is used like a builder, so tricky to make it immutable. We could have a `build` method (or `toImmutable` or some such), but I don't think it's a big deal.

Agree moving it to a test utility would be much harder to find (i.e. bad).

I'm happy to leave as-is in this PR.

aledsage:
@ahgittin I've finished reviewing. Lots of great stuff here.

Two general comments: 1) worth e-mailing brooklyn-dev list about this, as it is a significant new feature to be able to do in-life upgrades without restarting the server; 2) worth some more javadoc at the class-level to help future maintainers.

core/src/main/java/brooklyn/management/internal/LocalEntityManager.java
ahgittin: My ideal would be a model where tests fail if they write any WARN messages at all.  (Maybe we support `@AllowLogWarnings` for some cases.)

I'm in the habit of watching out for WARNs when running things.  Everyone should be.

ahgittin:
Thanks @aledsage.  Great comments.  I've added a bunch more javadoc, and I was already in the process of drafting a mail to dev.

I think they're all addressed, and will merge if the latest push passes buildhive, unless you have any further comments.

ahgittin:
spurious failure?  port 65536?  i think there's another PR or comment about fixing that... trying again as it seems unrelated

PR #509, opened 2015-02-09
Adds getConfig method to server REST API to allow client code to read pr...

nakomis:
...operties

### DIFF
--- a/usage/rest-api/src/main/java/brooklyn/rest/api/ServerApi.java
+++ b/usage/rest-api/src/main/java/brooklyn/rest/api/ServerApi.java
@@ -26,6 +26,7 @@ import javax.ws.rs.FormParam;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
@@ -38,6 +39,8 @@ import brooklyn.rest.domain.HighAvailabilitySummary;
 import brooklyn.rest.domain.VersionSummary;
 
 import com.google.common.annotations.Beta;
+import com.wordnik.swagger.core.ApiError;
+import com.wordnik.swagger.core.ApiErrors;
 import com.wordnik.swagger.core.ApiOperation;
 import com.wordnik.swagger.core.ApiParam;
 
@@ -86,6 +89,18 @@ public interface ServerApi {
             responseClass = "String",
             multiValueResponse = false)
     public String getStatus();
+
+    @GET
+    @Path("/config/{configKey}")
+    @ApiOperation(value = "Get the value of the specified config key from brooklyn properties")
+    @ApiErrors(value = {
+            // TODO: This should probably return a 404 if the key is not present, and should return a predictable
+            // value if the value is not set. Behaviour should be consistent with EntityConfigApi.get()
+            @ApiError(code = 204, reason = "Could not find config key")
+    })
+    public String getConfig(
+            @ApiParam(value = "Config key ID", required = true)
+            @PathParam("configKey") String configKey);
 
     @Deprecated /** @deprecated since 0.7.0 use /ha/states */
     @GET
--- a/usage/rest-server/src/main/java/brooklyn/rest/resources/ServerResource.java
+++ b/usage/rest-server/src/main/java/brooklyn/rest/resources/ServerResource.java
@@ -33,6 +33,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
+import brooklyn.config.ConfigKey;
+import brooklyn.entity.basic.ConfigKeys;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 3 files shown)


usage/rest-api/src/main/java/brooklyn/rest/api/ServerApi.java
```diff
@@ -87,6 +88,11 @@ public void shutdown(
             multiValueResponse = false)
     public String getStatus();
 
+    @GET
+    @Path("/getConfig/{configKey}")
```
aledsage: Would be better if this was `/config/{configKey}`, which would more closely match `EntityConfigApi` - that uses `/v1/applications/{application}/entities/{entity}/config/{config}` for retrieving a particular config value of a given entity.

usage/rest-api/src/main/java/brooklyn/rest/api/ServerApi.java
```diff
@@ -87,6 +88,11 @@ public void shutdown(
...
+    @ApiOperation(value = "Get the value of the specified config key from brooklyn properties")
+    public String getConfig(@PathParam("configKey") final String configKey);
```
aledsage: Suggest using something like that below (taken from `EntityConfigApi`):

    public String getConfig(
            @ApiParam(value = "Config key ID", required = true)
            @PathParam("configKey") String configKey)

aledsage: Also say that it will give 404 if not found (again, see `EntityConfigApi.get` for example).

usage/rest-server/src/test/java/brooklyn/rest/resources/ServerResourceTest.java
```diff
@@ -97,5 +99,15 @@ public void testShutdown() throws Exception {
                 assertFalse(getManagementContext().isRunning());
             }});
     }
+
+    @Test
+    void testGetConfig() throws Exception {
+        ((BrooklynProperties)getManagementContext().getConfig()).addFrom(ImmutableMap.of("foo.bar.baz", "quux"));
```
aledsage: Better to do `((ManagementContextInternal)getManagementContext()).getBrooklynProperties()`. It effectively does the same thing, but we know we can always cast to the "internal" version (whereas casting the return value of a getter isn't guaranteed moving forwards).

Also, you can use `.put("foo.bar.baz", "quux")` rather than needing the map.

usage/rest-server/src/test/java/brooklyn/rest/resources/ServerResourceTest.java
```diff
@@ -97,5 +99,15 @@ public void testShutdown() throws Exception {
                 assertFalse(getManagementContext().isRunning());
             }});
     }
+
+    @Test
+    void testGetConfig() throws Exception {
```
aledsage: Worth including a test of behaviour if you ask for a config key that is not there.

usage/rest-server/src/main/java/brooklyn/rest/resources/ServerResource.java
```diff
@@ -238,6 +240,12 @@ public String getStatus() {
         return getHighAvailabilityNodeState().toString();
     }
 
+    @Override
+    public String getConfig(String configKey) {
+        ConfigKey<String> config = ConfigKeys.newStringConfigKey(configKey);
```
aledsage: Worth including an entitlements check, because the return values could contain passwords retrieved from `brooklyn.properties`. For example, something like:

    if (!Entitlements.isEntitled(mgmt().getEntitlementManager(), Entitlements.SEE_ALL_SERVER_INFO, null)) {
        throw WebResourceUtils.unauthorized("User '%s' is not authorized for this operation", Entitlements.getEntitlementContext().user());
    }

nakomis:
PR comments addressed

usage/rest-server/src/test/java/brooklyn/rest/resources/ServerResourceTest.java
```diff
@@ -97,5 +103,38 @@ public void testShutdown() throws Exception {
...
+        boolean keyAlreadyPresent = false;
+        String response = null;
+        if (properties.containsKey(key)) {
+            existingValue = properties.remove(key);
+            keyAlreadyPresent = true;
+        }
+        try {
+            response = client().resource("/v1/server/config/" + key).get(String.class);
+        } catch (UniformInterfaceException e) {
+            assertEquals(e.getResponse().getStatus(), 204);
```
aledsage: Does this test pass?! It's asserting a 204 rather than a 404 (i.e. "No Content').

My personal preference is to put the `Asserts.fail` inside the try block, and not to do a `return` inside the catch block. I find that clearer, but no strong feelings.

aledsage: jenkins shows:

    2015-02-10 10:44:28,237 INFO  TESTNG INVOKING: "Surefire test" - brooklyn.rest.resources.ServerResourceTest.testGetMissingConfigThrowsException()
    2015-02-10 10:44:28,237 INFO  3 * Server in-bound request
    3 > GET /v1/server/config/foo.bar.baz
    3 > 

    2015-02-10 10:44:28,237 INFO  3 * Server out-bound response
    3 < 204
    3 < Content-Type: application/json
    3 < 

We either want to fix comments to say it will be a 204, or fix the impl so it throws a 404 (see stuff like `BrooklynRestResourceUtils.getApplication`, where it does `throw notFound("Application '%s' not found", application)`).

It looks to me like `EntityConfigResource.get(...)` is going to behave the same as your code if the config key isn't known. Maybe we just want to update the annotations in `EntityConfigApi.get(...)` etc to say it gives a 404 if "Could not find application or entity" instead of "Could not find application, entity or config key".

Thoughts?

nakomis:
PR comments addressed

aledsage:
Thanks @nakomis - merging.

PR #511, opened 2015-02-10
Update the build instructions to include the launch command

bostko:


ahgittin:
thanks @bostko and +1, this solves the problem that the script is well buried!  merging

ahgittin:
would you mind signing a contributor agreement if you haven't already?

https://brooklyn.incubator.apache.org/developers/how-to-contribute.html

PR #513, opened 2015-02-13
Strip "Expect: 100-Continue" header in Jclouds object store

neykov:
The 100 Continue response causes the Java 7,8 http client to misbehave,
more so when used in SSL mode.

Sockets closed by GC break the internal bookkeeping
of HttpUrlConnection, leading to invalid handling of the "HTTP/1.1 100 Continue"
response. Coupled with a bug when using SSL sockets reads will block
indefinitely even though a read timeout is explicitly set.


ahgittin:
This was a terrible bug, or rather pair of bugs, incredibly hard to track down, particularly since you couldn't get the softlayer SSL keys to see the https wire traffic.

One doesn't expect Java 8 HTTP code to be buggy!

Tremendous work @neykov .

Merging.

PR #519, opened 2015-02-18
Improve document usability - navigation menu

bostko:
As a new user I first click on the "get started" link and then clicking forward to the next steps and getting back is a little painful because you have to remember all the links you've clicked through the articles.

So I implemented a navigation menu which is automatically generated following the directory structure in the source.
* It is shown for the guides docs.
* It supports weights for every article so they could be sorted easily.

I think it is working and this helps you to no more feeling lost in the docs.

It is not finished yet, there is a lot of things that could be improved, but it is a start.
I am expecting your feedback.


bostko:
I merged with master, try again

neykov:
I like the idea of having the full menu available at all times as an anchor. It took me a long time to realize that clicking on the children of the current menu will drill down, and I am still surprised by how it changes.
It's easier to orient myself in the structure of the site and find specific content with the changes.

Some examples:
 * https://brooklyn.incubator.apache.org/v/latest/ops/locations/more-locations.html and clicking on BYON brings up a different menu with the parent highlighted
 * https://brooklyn.incubator.apache.org/v/latest/misc/release-notes.html and clicking on Developer Guide goes to a completely different hierarchy, maybe a separate section with "See Also" links would be nice.
  * Sometimes clicking on a child link will replace the parent hierarchy, sometimes it will add to it, sometimes it will just keep it the same.

I realize this is a bike-shedding type of discussion but still worth sharing my impressions.

--

There's still polishing needed, but good enough to demonstrate the intent.

Test failure not related.

ahgittin:
I get:

```
Configuration file: /Users/alex/Data/cloudsoft/dev/gits/brooklyn/docs/_config.yml
/Users/alex/Data/cloudsoft/dev/gits/brooklyn/docs/_plugins/hierarchical_menu.rb:3:in `require': cannot load such file -- pry (LoadError)
	from /Users/alex/Data/cloudsoft/dev/gits/brooklyn/docs/_plugins/hierarchical_menu.rb:3:in `<top (required)>'
	from /Users/alex/.rvm/gems/ruby-2.1.2@incubator-brooklyn-site/gems/jekyll-2.5.3/lib/jekyll/plugin_manager.rb:74:in `require'
```

Adding `gem 'pry'` to the `Gemfile` fixed it.  (The other two deps -- `digest/md5` and `open-uri` -- didn't resolve but didn't seem to need such a line.)

bostko:
Just a sec I'll make a commit

bostko:
Pull again

ahgittin:
Now that I can see it, personally I don't really like seeing all the uncles in tree.  This is comparable to the older style (you can see if you browse to the 0.6.0 docs) which people found overwhelming, and I tend to agree.  IMHO it is better to let the user focus on the area they are browsing.

What do other people think?

@bostko maybe paste a few screenshots and let's do a straw poll?

Also, maybe there are other ways to resolve the confusion which you encountered (which I do understand, it just felt like a lesser evil than a too-big-menu!).

bostko:
Here's how it looks like with the navigation menu
screenshot from 2015-02-20 20 10 23
screenshot from 2015-02-20 20 10 28
screenshot from 2015-02-20 20 10 35
screenshot from 2015-02-20 20 10 46

sjcorbett:
Anecdotally, several people at companies I've worked with have expressed surprise at the current behaviour of the menu. I am in favour of changes along the lines of this PR.

aledsage:
I agree with @sjcorbett as well - @bostko's hierarchy will be a more familiar experience (and easier to navigate/use) for users. With the +/- buttons to expand/hide sub-items, I think it should not overwhelm users. We just need to be conscious of the structure to make sure there are not too many top-level pages, and also not too deep a hierarchy.

aledsage:
Looking at the two options again, I see pros and cons for each (I've called the options "full menu" and "current context only").

I wonder if we need to put this on hold until folk have more time to focus on it (or if there are any quick wins that move us towards the best of both options).
We need to be sure to invest our time in what is agreed to be the right future direction.

A big concern with the "full menu" option is that the menu takes up a lot of vertical space once you've navigated to the appropriate page - e.g. expand "User Guide -> YAML Blueprints -> Chef in YAML Blueprints -> About Chef". There are about 30 things shown in the menu.
We don't want to have to scroll to see the menu options.
We don't want it to get harder to use over time as more pages are added.
It's nice to have the menu float (i.e. always visible at the top of page) - that is only possible if scrolling will never be needed, even on low resolution displays. Alternatively we could have separately scrollable sections on the page, but that is out of vogue.
A couple of minor improvements to @bostko's menu would be: why are the lines being wrapped; why is there so much white space between rows.

For the current "current context only" option. I worry it will confuse users when the menu changes - 
it's not clear enough that the breadcrumbs at the top are showing the hierarchy that these items are under.
It's also slightly harder to navigate back if the user goes to the wrong page (because their model of the menu has been entirely changed, with the previous links having disappeared).

For the "current context only" option, perhaps an improvement would be to show the sibling pages indented so it's clear they are child pages of the top-level hierarchy.
There could perhaps be a "..." to show that uncles etc have been hidden.

Another interesting feature of the "current context only" option is that pages can appear in more than one place. For example, "developer guide" also appears under "user guide -> other 0.7.0-M2-incubating resources".
This makes sense for a context-based menu - it's showing the pages that are relevant for where you've navigated to. But it would not make sense when showing someone a full hierarchy.
I'm on the fence as to how useful that is, versus improving our cross-references and our searchability.

I think there are entirely orthogonal issues around searchability and improving the structure so things are more logically grouped in the right places.
Let's not let those considerations affect this particular discussion.

ahgittin:
@bostko close this PR and open a jira issue pointing at this for reference?

bostko:
https://issues.apache.org/jira/browse/BROOKLYN-138

PR #520, opened 2015-02-19
Compiler independent persistence

neykov:
The project recently switched from the eclipse-groovy-compiler to the system provided compiler. It turns out that the compilers generate different indexes for the `this$N` fields in non-static inner classes. This results in failures when rebinding to a state generated from a different compiler.

There is one case where there is not enough information in the XML to figure out the new names reliably - when a dynamic class extends another dynamic class and the classes have different this$N field indexes. This is quite rare (we don't have such persisted classes in Brooklyn, though user code might have), in this case the changes will at least log an error, before the rebind fails, with instructions on how to update the fields.

### DIFF
--- a/core/src/main/java/brooklyn/util/xstream/CompilerIndependentOuterClassFieldMapper.java
+++ b/core/src/main/java/brooklyn/util/xstream/CompilerIndependentOuterClassFieldMapper.java
@@ -21,12 +21,16 @@ package brooklyn.util.xstream;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.thoughtworks.xstream.core.Caching;
 import com.thoughtworks.xstream.mapper.Mapper;
 import com.thoughtworks.xstream.mapper.MapperWrapper;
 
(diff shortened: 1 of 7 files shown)


neykov:
XStream JIRA issue: http://jira.codehaus.org/browse/XSTR-769

ahgittin:
at first blush i really like this.  will test.

@aledsage if you have any high-level thoughts that'd be useful

neykov:
Fixed in XStream, but not in a backwards compatible way - existing state will fail to unmarshal, if compiler is changed, even if using the udpated version. Once state is saved by an updated version, then it's compiler independent.
Changeset: https://fisheye.codehaus.org/changelog/xstream?cs=2358 - renumbers the outer fields by superclass (topmost superclass' field has index 0, extending classes' fields increment the index)

neykov:
Added caching for field names.

core/src/main/java/brooklyn/util/xstream/CompilerIndependentOuterClassFieldMapper.java
```diff
@@ -0,0 +1,167 @@
...
+                    // so we check the target class instead. This should work most of the time, but
+                    // if code is recompiled in such a way that the new instance has fields with
+                    // different names, where only the field of the extending class is renamed and
+                    // the super field is not, then the instance will be deserialized incorrectly -
+                    // the super field will be assigned both times. If the field type is incompatible
+                    // then a casting exception will be thrown, if it's the same then only the warning
+                    // below will indicate of a possible problem down the line - most probably NPE on
+                    // the this$N field.
+                    LOG.warn(msg + "Will use the field " + deepestField + " instead.");
+                    deserializeFieldName = deepestField;
```
aledsage: I'd personally delete the `deepestField` variable (which is only used here), and just assign `deserializeFieldName = Iterables.getOnlyElement(uniqueFieldNames)l`. The name "deepestField" threw me at first.

aledsage:
@neykov looks really good. Only one extremely minor comment. Happy to either merge now or if you want to change it then can merge after that.

Let's hope we can strongly discourage the use of non-static inner classes for stuff that is to be persisted!

core/src/main/java/brooklyn/util/xstream/CompilerIndependentOuterClassFieldMapper.java
neykov: Done. Can't use Iterables.getOnlyElement as there are potentially mutliple items in the collection. deepestField means the bottom-most field in the class hierarchy.

aledsage:
Excellent - thanks @neykov - merging.

PR #524, opened 2015-02-21
Catalog documentation improvements

bostko:
* Include the header in the catalog pages
* Load an entity using one html file

Simplify the build
* Delete the separate configuration file for the catalog
* Move the catalog js and catalog css to the js folder

screenshot from 2015-02-22 00 31 03
screenshot from 2015-02-22 00 30 53


docs/style/js/catalog/common.js
```diff
@@ -25,7 +25,7 @@ var brooklyn = (function ($, _) {
         },
 
         entityCard: _.template(
-            "<a class='plain' data-type='<%= type %>' href='entities/<%= type %>.html'>" +
+            "<a class='plain' data-type='<%= type %>' href='brooklyn-node.html#entities/<%= type %>'>" +
```
neykov: Not an expert in SEO, but wouldn't this hamper findability? Technically it makes more sense, but having a separate page for each entity might be more appropriate google-wise.

docs/website/learnmore/catalog/brooklyn-node.html
```diff
@@ -0,0 +1,90 @@
+---
```
neykov: brooklyn-node refers to the specific `brooklyn.entity.brooklynnode.BrooklynNode` entity. `item.html` as in `Catalog item` would be more appropriate.

bostko: Thank you!
I renamed it.

docs/style/js/catalog/common.js
bostko: Yes you are right about the SEO. Generally it is not good to have Ajax requests in the first place.
Although it is not recognized well in google the catalog classes are listed in google https://www.google.bg/?q=brooklynnode
In order to mimic the current SEO capabilities of the website we have to support https://developers.google.com/webmasters/ajax-crawling/docs/getting-started
_escaped_fragment_ get parameter on the server.
 I will research this more if it is possible to use this get_parameter without implementing it on the server and rather use the capabilities of the modern browsers like pushState. 
The other option is to autogenerate sitemap.xml

docs/_build/build.sh
```diff
@@ -43,7 +43,7 @@ function parse_mode() {
     help
     exit 0 ;;
   website-root)
-    JEKYLL_CONFIG=_config.yml,_build/config-production.yml,_build/config-exclude-guide.yml,_build/config-website-root.yml
+    JEKYLL_CONFIG=_config.yml,_build/config-production.yml,_build/config-exclude-guide.yml
```
ahgittin: `website-root` should include the config for website root, shouldn't it?

docs/_build/build.sh
```diff
@@ -90,7 +90,7 @@ function parse_mode() {
     SUMMARY="user guide files in the root"
     ;;
   test-both)
-    JEKYLL_CONFIG=_config.yml,_build/config-production.yml,_build/config-exclude-root-index.yml,_build/config-website-root.yml,_build/config-guide-latest.yml
+    JEKYLL_CONFIG=_config.yml,_build/config-production.yml,_build/config-exclude-root-index.yml,_build/config-guide-latest.yml
```
ahgittin: again i think this was correct, and it was working when it was written.  can you back out these changes and let's discuss?

docs/_build/config-website-root.yml
```diff
@@ -1,3 +0,0 @@
-path:
-  website: ""
-  guide: "/v/latest"
```
ahgittin: as above, let's keep for now and discuss separately

docs/_config.yml
```diff
@@ -26,7 +26,8 @@ url_root: http://0.0.0.0:4000
 path:
   style: /style
   guide: /guide
-  website: /website
+  #guide: "/v/latest"
```
ahgittin: aha i see why you did what you did -- but it needs to be reverted.

we like to be able to run `jekyll serve` from the root and for the resulting links to be happy.  this means we have to have guide in guide and website it website.  this is why the structural changes are done afterwards production builds.  unfortunate but i don't see a way around it.

docs/_plugins/site_structure.rb
```diff
@@ -116,7 +116,17 @@ def self.find_page_with_path_absolute_or_relative_to(site, path, referrent, stru
       # Pathname API ignores first arg below if second is absolute
 #      puts "converting #{path} wrt #{referrent ? referrent.path : ""}"
       file = Pathname.new(File.dirname(referrent ? referrent.path : "")) + path
-      file += "index.md" if file.to_s.end_with? "/"
```
ahgittin: much better

ahgittin:
a few things to revert on the docs build process but otherwise looks very good

bostko:
@ahgittin In order to simplify the build I removed the `docs/_build/config-website-root.yml` 
and I removed the guide path `/v/latest`

My idea is to use one centralized place for configuration `_config.yml`. This simplifies a lot the development since now you have to move the compiled files from the website folder to the docs root.
Differentiating the jekyll configs leads to а lot of different folders and configs and if some link is not working or if there is some other problem it is harder to debug.

I see the point about versioning the docs and putting them in separate folders /v/latest/guide
but I think it is easier if we do docs versioning as a whole and put everything in /v/version not only the guide. The thing which is still more updateble is the catalog_items.json

ahgittin:
of course it would be much simpler if we could simply version everything but we've had this discussion and the consensus was that we needed something like this.  it simply doesn't make sense to have old verisons of things like contact information, news pages, community info kicking around, hence the separation into one section which is versioned ("`guide`") and one section which is not ("`website`").

please revert those -- they don't belong in a PR about the catalog in any case!

btw previously this was handled by having two separate projects but that was even more complex as we had different JS versions and builds so i genuinely think this is the least of all evils given the requirement not to have old versions of contact info etc!

ahgittin:
looks good now, thanks @bostko .  i've reinstated `docs/index.md` which is useful when running `jekyll serve`, and merged.

a few things it would be nice to clean up:
* search is only on entities and only matches simple class name; should match much more
* hover on an entry shifts the div by a pixel or two; it should highlight only but not shift
* once you've selected an item there is no obvious way to return to the catalog

trivial though, i've merged.  great to have this catalog tidied up!

bostko:
I added this to my todos, thank you

PR #525, opened 2015-02-23
Fix: no-hostname, and cluster.replaceMember

aledsage:


aledsage:
@richardcloudsoft @grkvlt can you take a look at this please?

rdowner:
Hi @aledsage, some comments on the set-hostname commit. I can take a look at the other one later.

aledsage:
@rdowner I've incorporated those comments; can you review again please

core/src/test/java/brooklyn/util/ssh/BashCommandsIntegrationTest.java
```diff
@@ -361,6 +363,99 @@ public void testWaitForPortFreeWhenFreedAfterStart() throws Exception {
...
+        
+        try {
+            List<String> cmd = (includeDomain) ? BashCommands.setHostname(newHostname, newDomain) : BashCommands.setHostname(newHostname);
+            execRequiringZeroAndReturningStdout(loc, cmd).get();
+            
+            assertEquals(getHostname(loc), fqdn);
+            
+            String result = execRequiringZeroAndReturningStdout(loc, "grep -n "+fqdn+" /etc/hosts").get();
+            assertTrue(result.contains("localhost"), "line="+result);
+            log.info("result="+result);
```
rdowner: A good test to do would be:

* Executing `hostname` returns the hostname without the domain name
* Executing `hostname --fqdn` returns the hostname *with* the domain name
* Executing `ping -c1 -n -q ${ hostname }` return success (rc=0)
* Executing `ping -c1 -n -q ${ hostname --fqdn }` return success (rc=0)

The last two are probably the most important - it's these kinds of failures that would also cause an entity to fail to bind.

Unfortunately OSX has slightly different parameters - it's `hostname -s` and `hostname -f` respectively.

rdowner:
Added a couple more comments. With those I think we're ready to merge the hostname changes (no comment yet on the cluster changes).

Note that my knowledge of setting the hostname is based on my experience in the past, but it's possible that modern distributions do things differently in 2015 and perhaps the recommendations have changed. Nevertheless, the old-style way should still work!

aledsage:
@rdowner I've incorporated those comments. Can you please review again, and also review the second commit so we can see about merging this?

rdowner:
@aledsage I'm happy with the hostname changes - this has been an issue for a few entities in the past so it's great to have this fix available in core Brooklyn now. I think it could maybe be reworked in the future - I think maybe my thinking is a bit "old fashioned" and it does seem that OSX does things slightly differently - but what is here will definitely be reliable and fix known problems.

I've also reviewed the cluster changes and they seem sane.

:+1: good to merge. Do you want to squash the hostname commits first?

aledsage:
Squashing commits, and then will merge:

    pick ca0e640 Add BashCommands.setHostname
    fixup f0dfe12 setHostname: incorporate review comments
    fixup 0d57a39 setHostname: incorporate more review comments

PR #526, opened 2015-02-24
LocationTransformer: generate correct location spec strings

alasdairhodge:
Correctly deals with `Maybe` values to avoid spurious warnings like:
````
WARN  Location resolution failed for 'Present[value=localhost]' (will fail shortly):
known resolvers are: [id, named, localhost, byon, single, host, multi, portForwardManager, jclouds, jcloudsByon, pool]
````


aledsage:
@alasdairhodge Looks good; nice and simple. Would be nice to have a unit test in the spirit of writing a unit test for failing functionality to avoid regressions. Will leave that to you for future work as you see fit.

Merging.

aledsage:
One other minor comment: can you stick to approx 50 chars for first line of commit message:
https://github.com/erlang/otp/wiki/Writing-good-commit-messages

PR #528, opened 2015-02-24
Improve persistence transformers

neykov:
* Accept a list of single key-value maps so one file can contain multiple actions with same name
* When renaming a class replace the new name in the class tag, type element, class attribute

### DIFF
--- a/core/src/main/java/brooklyn/entity/rebind/transformer/CompoundTransformer.java
+++ b/core/src/main/java/brooklyn/entity/rebind/transformer/CompoundTransformer.java
@@ -114,10 +114,17 @@ public class CompoundTransformer {
          * Replaces a tag, but while continuing to recurse.
          */
         public Builder xmlRenameTag(String xpathToMatch, String newValue) {
-            return xmlReplaceItem(xpathToMatch, 
+            return xmlReplaceItem(xpathToMatch,
                 "<"+newValue+">"
                     + "<xsl:apply-templates select=\"@*|node()\" />"
                 + "</"+newValue+">"); 
+        }
+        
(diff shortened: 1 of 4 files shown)


aledsage:
Test failure looks unrelated, and has been happening far too often. Perhaps we need to mark everything that requires an open port to be integration tests?!

core/src/main/java/brooklyn/entity/rebind/transformer/CompoundTransformer.java
```diff
@@ -181,6 +188,18 @@ public Builder changeCatalogItemId(String oldSymbolicName, String newSymbolicNam
...
+        /**
+         * Updates all references to a class to a new value
+         * @param oldName the old name of the class
+         * @param newName the new name of the class to be used instead
+         */
+        public Builder renameClass(String oldName, String newName) {
+            return renameClassTag(oldName, newName)
+                .xmlChangeAttribute("//@class[.='" + oldName + "']", newName)
+                .renameType(oldName, newName);
+                //TODO update reference attributes
```
aledsage: Not sure what this TODO means?

aledsage:
Looks good; merging.

core/src/main/java/brooklyn/entity/rebind/transformer/CompoundTransformer.java
neykov: When the same object is used at multiple places it is serialized only once and referenced from all other places through the `reference` attribute along with a relative path to the node. When renaming classes the path changes so the value needs to be updated. Not needed for the advanced-networking transforms so I didn't spend time implementing it.

PR #529, opened 2015-02-25
Redirects web gui to root on logout to prevent login failure after logout

nakomis:
Previously, clicking logout would leave the page at /logout. Trying to log in would then navigate to a URL such as /logout/#v1 etc

### DIFF
--- a/usage/jsgui/src/main/webapp/assets/js/util/brooklyn-utils.js
+++ b/usage/jsgui/src/main/webapp/assets/js/util/brooklyn-utils.js
@@ -174,6 +174,26 @@ define([
         return false; 
     };
 
+    Util.logout = function logout() {
+        $.ajax({
+            type: "POST",
+            dataType: "text",
+            url: "/logout",
+            success: function() {
+                window.location.replace("/");
+            },
+            failure: function() {
+                window.location.replace("/");
+            }
+        });
+    }
+
+    $("#logout-link").on("click", function (e) {
+        e.preventDefault();
+        Util.logout()
+        return false;
+    });
+
     return Util;
 
 });
--- a/usage/jsgui/src/main/webapp/index.html
+++ b/usage/jsgui/src/main/webapp/index.html
@@ -44,7 +44,7 @@ under the License.
 
 <div class="navbar navbar-fixed-top">
     <div class="navbar-inner">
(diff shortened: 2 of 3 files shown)


sjcorbett:
Isn't the problem with this that after clicking logout the user will immediately be logged back in? The browser will just resend their credentials and they will be reauthenticated. Can you include a test case to prove me wrong?

nakomis:
@sjcorbett The user isn't immediately logged back in due to `originalRequest` logic in https://github.com/apache/incubator-brooklyn/blob/master/usage/rest-server/src/main/java/brooklyn/rest/filter/BrooklynPropertiesSecurityFilter.java, the logic of which hasn't been changed. This PR simply changes the behaviour in the browser, which I'm not sure how to write a test for

sjcorbett:
Thanks for pointing me at the at `originalRequest` in `BrooklynPropertiesSecurityFilter`. My initial concern is invalid but there are a few usability issues with the changes here:

* The `onClick` event is not fired when I right click the link and click "Open link in a new tab", so I'm not logged out.
* Every so often - I haven't worked out the cause - clicking Logout navigates me to the dashboard screen but doesn't actually log me out.
* Clicking logout normally makes it impossible to log back in without refreshing the page. Each time I enter my username and password I'm immediately logged out again. The log shows:
```
2015-03-04 10:55:33,823 INFO  REST logging sam out of session 1xf18cj3jx94x59y5dt4kahdp
2015-03-04 10:55:47,359 INFO  REST logging sam out of session 1tb4mq85kwddh199yigwsrwtrw
2015-03-04 10:55:52,173 INFO  REST logging sam out of session 4afpbliikh5i1w29exmvrw534
```
* You can break logout on someone's web console by altering the logout function to either do something different or nothing at all (i.e. open the Javascript console and enter `logout = undefined`). It's broken until they refresh, at least.
* Having the function in the global namespace means there's a chance it will clash with another library that also defines `logout` globally. I'd suggest moving the function to `brooklyn.js` or `brooklyn-util.js`, adding an `id` to the anchor element and binding an `onClick` event there rather than putting it on the element directly.

@nakomis and @aledsage up to you for the first few comments. The last two should be fixed.

nakomis:
PR comments (mostly) addressed

PR #530, opened 2015-02-25
Adds template processing to the MySQL database creation script

nakomis:


software/database/src/main/java/brooklyn/entity/database/mysql/MySqlSshDriver.java
```diff
@@ -186,9 +186,13 @@ protected void copyDatabaseConfigScript() {
...
-        InputStream creationScript = DatastoreMixins.getDatabaseCreationScript(entity);
-        if (creationScript==null) return false;
-        getMachine().copyTo(creationScript, getRunDir() + "/creation-script.sql");
+        String creationScriptContents = DatastoreMixins.getDatabaseCreationScriptAsString(entity);
+        if (creationScriptContents==null) return false;
+
+        creationScriptContents = processTemplateContents(creationScriptContents);
+        Reader creationContents = new StringReader(creationScriptContents);
+
+        getMachine().copyTo(creationContents, getRunDir() + "/creation-script.sql");
```
sjcorbett: You should just use `copyTemplate` instead of the previous four lines.

sjcorbett: Although `copyTemplate` wants the URI of the template so its a bit awkward with `CREATION_SCRIPT_CONTENTS`. I guess you'd have to write it out as a temporary file. I leave it up to you.

aledsage:
Can we get an integration test, that the database creation script is being templated correctly? Or do you think that is a lot more effort than it's worth?

nakomis:
@sjcorbett PR comment addressed
@aledsage Given that the existing integration test simply runs the Vogella, so fitting in a test for the templating could be tricky

software/database/src/main/java/brooklyn/entity/database/mysql/MySqlSshDriver.java
```diff
@@ -186,9 +192,20 @@ protected void copyDatabaseConfigScript() {
...
+        try {
+            templateFile = File.createTempFile("mysql", null);
+            templateFile.deleteOnExit();
+            BufferedWriter writer = new BufferedWriter(new FileWriter(templateFile));
+            writer.write (creationScriptContents);
+        } catch (IOException e) {
+            throw Exceptions.propagate(e);
+        }
+        copyTemplate(templateFile.getAbsoluteFile(), getRunDir() + "/creation-script.sql");
+        templateFile.delete();
```
aledsage: Personally I'd have put the `copyTemplate` inside the try block, and the `templateFile.delete()` in a finally block.
Another slight concern about this approach is that the file permission could mean the file is readable by others for the brief time it exists - if the sql script contained any passwords, that would not be good.

However, I don't feel strongly enough about this to delay the PR. Happy to merge.

aledsage:
Thanks @nakomis - merging.

PR #531, opened 2015-02-25
Basho Riak - Linux Install

bostko:
Install from packagecloud.io

### DIFF
--- a/software/base/src/test/java/brooklyn/entity/AbstractEc2LiveTest.java
+++ b/software/base/src/test/java/brooklyn/entity/AbstractEc2LiveTest.java
@@ -75,10 +75,11 @@ public abstract class AbstractEc2LiveTest extends BrooklynAppLiveTestSupport {
         super.setUp();
     }
 
+    // Image ids for Debian: https://wiki.debian.org/Cloud/AmazonEC2Image/Squeeze
     @Test(groups = {"Live"})
     public void test_Debian_6() throws Exception {
         // release codename "squeeze"
-        runTest(ImmutableMap.of("imageId", "us-east-1/ami-0740476e", "loginUser", "admin", "hardwareId", SMALL_HARDWARE_ID));
+        runTest(ImmutableMap.of("imageId", "us-east-1/ami-5e12dc36", "loginUser", "admin", "hardwareId", SMALL_HARDWARE_ID));
     }
 
     @Test(groups = {"Live"})
--- a/software/base/src/test/java/brooklyn/entity/AbstractEc2LiveTest.java
+++ b/software/base/src/test/java/brooklyn/entity/AbstractEc2LiveTest.java
@@ -108,9 +108,9 @@ public abstract class AbstractEc2LiveTest extends BrooklynAppLiveTestSupport {
     }
 
     @Test(groups = {"Live"})
-    public void test_CentOS_5_6() throws Exception {
-        // Image: {id=us-east-1/ami-49e32320, providerId=ami-49e32320, name=RightImage_CentOS_5.6_x64_v5.7.14, location={scope=REGION, id=us-east-1, description=us-east-1, parent=aws-ec2, iso3166Codes=[US-VA]}, os={family=centos, arch=paravirtual, version=5.6, description=rightscale-us-east/RightImage_CentOS_5.6_x64_v5.7.14.manifest.xml, is64Bit=true}, description=rightscale-us-east/RightImage_CentOS_5.6_x64_v5.7.14.manifest.xml, version=5.7.14, status=AVAILABLE[available], loginUser=root, userMetadata={owner=411009282317, rootDeviceType=instance-store, virtualizationType=paravirtual, hypervisor=xen}}
-        runTest(ImmutableMap.of("imageId", "us-east-1/ami-49e32320", "hardwareId", SMALL_HARDWARE_ID));
+    public void test_CentOS_5() throws Exception {
+        // Image: {id=us-east-1/ami-e4bffe8d, providerId=ami-e4bffe8d, name=RightImage_CentOS_5.9_x64_v12.11.4_EBS, location={scope=REGION, id=us-east-1, description=us-east-1, parent=aws-ec2, iso3166Codes=[US-VA]}, os={family=centos, arch=paravirtual, version=5.0, description=411009282317/RightImage_CentOS_5.9_x64_v12.11.4_EBS, is64Bit=true}, description=RightImage_CentOS_5.9_x64_v12.11.4_EBS, version=12.11.4_EBS, status=AVAILABLE[available], loginUser=root, userMetadata={owner=411009282317, rootDeviceType=ebs, virtualizationType=paravirtual, hypervisor=xen}}
+        runTest(ImmutableMap.of("imageId", "us-east-1/ami-e4bffe8d", "hardwareId", SMALL_HARDWARE_ID));
     }
 
     @Test(groups = {"Live"})
--- a/software/database/src/test/java/brooklyn/entity/database/rubyrep/RubyRepEc2LiveTest.java
+++ b/software/database/src/test/java/brooklyn/entity/database/rubyrep/RubyRepEc2LiveTest.java
@@ -62,7 +62,7 @@ public class RubyRepEc2LiveTest extends AbstractEc2LiveTest {
 
     @Override
     @Test(enabled=false, groups = "Live")
-    public void test_CentOS_5_6() throws Exception { } // Diabling all except Ubuntu 12.0 temporarily
+    public void test_CentOS_5() throws Exception { } // Diabling all except Ubuntu 12.0 temporarily
 
     @Override
     @Test(enabled=false, groups = "Live")
--- a/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
+++ b/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Map;
 
 import brooklyn.util.ssh.BashCommands;
+import brooklyn.util.task.ssh.SshTasks;
 import com.google.api.client.util.Joiner;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -40,6 +41,7 @@ import brooklyn.util.collections.MutableMap;
 import brooklyn.util.net.Urls;
 import brooklyn.util.os.Os;
 import brooklyn.util.task.DynamicTasks;
+import brooklyn.util.text.Strings;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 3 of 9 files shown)


bostko:
Failed tests: 
  RiakNodeEc2LiveTest>AbstractEc2LiveTest.test_CentOS_5_6:112->AbstractEc2LiveTest.runTest:128->doTest:40 » PropagatedRuntime
  RiakNodeEc2LiveTest>AbstractEc2LiveTest.test_Debian_6:81->AbstractEc2LiveTest.runTest:128->doTest:40 » PropagatedRuntime
  RiakNodeEc2LiveTest>AbstractEc2LiveTest.test_Debian_7_2:87->AbstractEc2LiveTest.runTest:128->doTest:40 » PropagatedRuntime

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -156,6 +157,34 @@ public void install() {
...
+    
+    private ImmutableList<String> installRpmBased() {
+        ImmutableList.Builder<String> commands = ImmutableList.<String>builder();
+        commands.add("curl https://packagecloud.io/install/repositories/basho/riak/script.rpm | sudo bash");
+        commands.add("sudo yum install -y riak");
+        return commands.build();
+    }
+
+    private static ImmutableList<String> ifExecutable0Else1(String command, List<String> ifTrue, List<String> otherwise) {
+        ImmutableList.Builder<String> commands = ImmutableList.<String>builder();
```
aledsage: personal preference for a fluent API style when using a builder like this. e.g.

    return ImmutableList.<String>builder()
            .add(String.format("if test -z `which %s`; then", command))
            .addAll(ifTrue)
            .add("else")
            .addAll(otherwise)
            .add("fi")
            .build();

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -156,6 +157,34 @@ public void install() {
...
+    }
+    
+    private ImmutableList<String> installRpmBased() {
+        ImmutableList.Builder<String> commands = ImmutableList.<String>builder();
+        commands.add("curl https://packagecloud.io/install/repositories/basho/riak/script.rpm | sudo bash");
+        commands.add("sudo yum install -y riak");
+        return commands.build();
+    }
+
+    private static ImmutableList<String> ifExecutable0Else1(String command, List<String> ifTrue, List<String> otherwise) {
```
aledsage: Name is a bit confusing, particularly compared to `BashCommands.ifExecutableElse1` and `BashCommands.ifExecutableElse0` - those are semantically like your impl with the `otherwise` being to just return 0 or 1 respectively.

Perhaps `ifExecutableElse` would be better, as there is no 0 or 1 involved?

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -156,6 +157,34 @@ public void install() {
...
+        ImmutableList.Builder<String> commands = ImmutableList.<String>builder();
+        commands.add("curl https://packagecloud.io/install/repositories/basho/riak/script.deb | sudo bash");
+        commands.add("sudo apt-get install --assume-yes riak");
+        return commands.build();
+    }
+    
+    private ImmutableList<String> installRpmBased() {
+        ImmutableList.Builder<String> commands = ImmutableList.<String>builder();
+        commands.add("curl https://packagecloud.io/install/repositories/basho/riak/script.rpm | sudo bash");
+        commands.add("sudo yum install -y riak");
```
aledsage: Suggest you use `.add(BashCommands.sudo("yum install -y riak"))`. If running as the root user, then on some VMs it causes problems to try to execute `sudo`. The `sudo(...)` method gets around that by checking if the user is already root.

Same applies for other uses of sudo.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -95,7 +95,8 @@ public void install() {
         OsDetails osDetails = getMachine().getMachineDetails().getOsDetails();
         List<String> commands = Lists.newLinkedList();
         if (osDetails.isLinux()) {
-            commands.addAll(installLinux(getExpandedInstallDir()));
+//            commands.addAll(installLinux(getExpandedInstallDir()));
```
aledsage: Any reason to keep this line commented out? Can we just delete it?
We like to be strict about no commented out code - it's only allowed if there is a very clear comment that says why the code might be useful.

Actually, it looks like we are no longer calling `installLinux` at all, which means we have lost a lot of the other stuff it did (e.g. installing logrotate libpam0g-dev libssl0.9.8, etc). Why are we no longer doing that?

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
bostko: Thank you for your feedback!
That might fix the problem

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
bostko: We use the scripts provided by package cloud.
It has two bash scripts for debian based and rpm based distros which handle the resolution for choosing a concrete package.
https://packagecloud.io/basho/riak/install

bostko:
Still it has problems with AWS Debian 6 and CentOS_5_6
http://pastebin.com/LP3KR2vj http://pastebin.com/4f8eeeP6
Are these two images working?

bostko:
The image id for debian is updated

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -112,48 +106,65 @@ public void install() {
...
+
+    /**
+     * Returns a command which
+     * executes <code>statement</code> only if <code>command</code> is NOT found in <code>$PATH</code>
+     *
+     * @param command
+     * @param statement
+     * @return command
+     */
+    private static String ifNotExecutable(String command, String statement) {
```
ahgittin: can these come from `BashCommands` ?

software/base/src/test/java/brooklyn/entity/AbstractEc2LiveTest.java
```diff
@@ -75,10 +75,11 @@ public void setUp() throws Exception {
         super.setUp();
     }
 
+    // Image ids for Debian: https://wiki.debian.org/Cloud/AmazonEC2Image/Squeeze
     @Test(groups = {"Live"})
     public void test_Debian_6() throws Exception {
         // release codename "squeeze"
-        runTest(ImmutableMap.of("imageId", "us-east-1/ami-0740476e", "loginUser", "admin", "hardwareId", SMALL_HARDWARE_ID));
+        runTest(ImmutableMap.of("imageId", "us-east-1/ami-5e12dc36", "loginUser", "admin", "hardwareId", SMALL_HARDWARE_ID));
```
ahgittin: is it worth introducing constants for these?

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -112,48 +106,65 @@ public void install() {
                 .execute();
     }
 
-    private List<String> installLinux(String expandedInstallDir) {
```
ahgittin: can we keep the old code -- maybe a new static `newTaskInstallToLinuxFromFiles(...)` method, possibly refactored to a `RiakInstallationUtils` class?

ideally this would build up a task tree as part of the process -- using `Tasks` and `DynamicTasks` -- rather than create an enormous script.  i.e. trying the `packagecloud.io` install first, and if that fails it trying the old-style `fromFiles` link instead.  in the process @bostko you will probably encounter plenty of questions on tasks.  ask on these on IRC, and if necessary put together `tasks.md` page for the documentation.  (this is a bigger piece of work than is necessary, but it's something we've wanted for a while, plus it should be more interesting for you!)

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
bostko: Yes I will do that

software/base/src/test/java/brooklyn/entity/AbstractEc2LiveTest.java
bostko: Are you asking about SMALL_HARDWARE_ID constant or you want constants for the other parameters?

ahgittin: i was thinking the machine ID -- but if it's localised to this one method no need i guess.  really just curious why this one and centos5 changed.  are the old images deprecated?

software/database/src/test/java/brooklyn/entity/database/rubyrep/RubyRepEc2LiveTest.java
```diff
@@ -62,7 +62,7 @@ public void test_CentOS_6_3() throws Exception { } // Diabling all except Ubuntu
 
     @Override
     @Test(enabled=false, groups = "Live")
-    public void test_CentOS_5_6() throws Exception { } // Diabling all except Ubuntu 12.0 temporarily
+    public void test_CentOS_5() throws Exception { } // Diabling all except Ubuntu 12.0 temporarily
```
ahgittin: could this be renabled?

software/base/src/test/java/brooklyn/entity/AbstractEc2LiveTest.java
bostko: 5.6 has problem when using curl with ssl and
It fails to download the ssh installation script.
Also it is over 3 years old and I saw that there is several implementations of AbstractEc2LiveTest which just override the test_CentOS_5_6 method with an empty one because the test is not working so I decided to make a pull request about it.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
bostko: Actually, I tried to do that and it doesn't resolve correctly the BashCommands class, it gets it from brooklyn-jmxmp-agent-shaded*.jar
We had a little discussion with Svet about this but we have to rearrange the dependencies in brooklyn parent which we do not want to do now.

software/database/src/test/java/brooklyn/entity/database/rubyrep/RubyRepEc2LiveTest.java
bostko: I tested it but the tests for CentOS 5 and Ubuntu 12 fail

ahgittin: worth a comment "not supported" - plus detail of error if you have it

ahgittin: ah i see that below, for 5 at least - ignore previous comment :)

bostko:
Besides installing from packagecloud it also support fallback mechanism for installing from the default location specified in the BrooklynNode interface
If an explicit location is specified then it is preferred over the packagecloud.

ahgittin:
this looks good, merging

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
grkvlt: That seems very wrong. As there's nothing special about this entity, does that mean _all_ entities get the wrong BashCommands class? That needs to be fixed asap.

PR #532, opened 2015-02-25
Load catalog.xml items from YAML when version qualifier is used

neykov:
Previously catalog.xml items couldn't be loaded from YAML if version (even the default one - 0.0.0.SNAPSHOT) was used.


sjcorbett:
Looks good. Merging. Thanks @neykov.

PR #534, opened 2015-02-27
fix cobertura build - BROOKLYN-134

ahgittin:
change how we detect dev environment -- just look for a `classes` directory


aledsage:
Looks good; merging.

PR #535, opened 2015-02-27
Misc tidies

ahgittin:
two minor grievances--
* eclipse groovy was giving a funny error due to config type incompatibility
* the dev docs should make it easier to find our jenkins server

### DIFF
--- a/core/src/main/java/brooklyn/entity/basic/AbstractEntity.java
+++ b/core/src/main/java/brooklyn/entity/basic/AbstractEntity.java
@@ -961,8 +961,12 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
     
     // -------- CONFIGURATION --------------
 
-    @Override
-    public ConfigurationSupportInternal config() {
+    @Override 
+    @Beta
+    // the concrete type rather than an interface is returned because Groovy subclasses
+    // complain (incorrectly) if we return ConfigurationSupportInternal
+    // TODO revert to ConfigurationSupportInternal when groovy subclasses work without this (eg new groovy version)
+    public BasicConfigurationSupport config() {
         return config;
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/brooklyn/entity/basic/AbstractEntity.java
```diff
@@ -962,7 +962,9 @@ public void removeAttribute(AttributeSensor<?> attribute) {
     // -------- CONFIGURATION --------------
 
     @Override
-    public ConfigurationSupportInternal config() {
+    // the concrete type rather than an interface is returned because Groovy subclasses
+    // complain (incorrectly) if we return ConfigurationSupportInternal 
+    public BasicConfigurationSupport config() {
```
aledsage: On balance I'm fine with this, but really would prefer if the impl did not leak as that ties us down for future refactoring. Perhaps we should add an `@Beta` to this method, and add javadoc explaining why (saying that it may change to return a `ConfigurationSupportInternal` in a future release, that being the preferred way to use it).

aledsage: Does the same apply to `AbstractEntityAdjunct.config()`, which also returns `ConfigurationSupportInternal`?
And same for `AbstractLocation.config()`?

aledsage:
Looks good. Only one minor comment about potentially adding an `@Beta`. I presume the need for that will go away in a future Groovy version, so at some point we can revert it to return the interface.

@ahgittin happy for you to merge once you've had a look at the comments.

core/src/main/java/brooklyn/entity/basic/AbstractEntity.java
neykov: Isn't it better to fix it in a new subclass, meant just for groovy usage? Or the subclasses are not directly on this class...

ahgittin:
Agree with you @aledsage .  Messy,  Worth an `@Beta` for sure.

PR #538, opened 2015-03-02
Adds addAssociationListener and removeAssociationListener to PortForward...

nakomis:
...Manager

### DIFF
--- a/core/src/main/java/brooklyn/location/access/PortForwardManager.java
+++ b/core/src/main/java/brooklyn/location/access/PortForwardManager.java
@@ -18,14 +18,15 @@
  */
 package brooklyn.location.access;
 
-import java.util.Collection;
-
 import brooklyn.config.ConfigKey;
 import brooklyn.entity.basic.ConfigKeys;
 import brooklyn.location.Location;
-
 import com.google.common.annotations.Beta;
+import com.google.common.base.Objects;
+import com.google.common.base.Predicate;
 import com.google.common.net.HostAndPort;
+
+import java.util.Collection;
 
 /**
  * Acts as a registry for existing port mappings (e.g. the public endpoints for accessing specific
@@ -45,6 +46,57 @@ import com.google.common.net.HostAndPort;
  */
 @Beta
 public interface PortForwardManager extends Location {
+
+    @Beta
+    class AssociationMetadata {
+        private final String publicIpId;
+        private final HostAndPort publicEndpoint;
+        private final Location location;
+        private final int privatePort;
+
+        /**
+         * Users are discouraged from calling this constructor; the signature may change in future releases.
+         * Instead, instances will be created automatically by Brooklyn to be passed to the
+         * {@link AssociationListener#onAssociationCreated(AssociationMetadata)} method.
+         */
+        public AssociationMetadata(String publicIpId, HostAndPort publicEndpoint, Location location, int privatePort) {
+            this.publicIpId = publicIpId;
+            this.publicEndpoint = publicEndpoint;
+            this.location = location;
+            this.privatePort = privatePort;
+        }
+
+        public String getPublicIpId() {
+            return publicIpId;
+        }
+
+        public HostAndPort getPublicEndpoint() {
+            return publicEndpoint;
+        }
+
+        public Location getLocation() {
+            return location;
+        }
+
+        public int getPrivatePort() {
+            return privatePort;
+        }
+
+        public String toString() {
+            return Objects.toStringHelper(this)
+                    .add("publicIpId", publicIpId)
+                    .add("publicEndpoint", publicEndpoint)
+                    .add("location", location)
+                    .add("privatePort", privatePort)
+                    .toString();
+        }
+    }
+
+    @Beta
+    interface AssociationListener {
+        void onAssociationCreated(AssociationMetadata metadata);
+        void onAssociationDeleted(AssociationMetadata metadata);
+    }
 
     /**
      * The intention is that there is one PortForwardManager instance per "scope". If you 
@@ -92,6 +144,16 @@ public interface PortForwardManager extends Location {
      * subsequently be looked up using {@link #lookup(String, int)}.
      */
     public void associate(String publicIpId, HostAndPort publicEndpoint, int privatePort);
+
+    /**
+     * Registers a listener, which will be notified each time a new port mapping is associated. See {@link #associate(String, HostAndPort, int)}
+     * and {@link #associate(String, HostAndPort, Location, int)}.
+     */
+    @Beta
+    public void addAssociationListener(AssociationListener listener, Predicate<? super AssociationMetadata> filter);
+
+    @Beta
+    public void removeAssociationListener(AssociationListener listener);
     
     /**
      * Returns the public ip hostname and public port for use contacting the given endpoint.
@@ -261,4 +323,5 @@ public interface PortForwardManager extends Location {
      */
     @Deprecated
     public PortMapping getPortMappingWithPrivateSide(Location l, int privatePort);
+
 }
--- a/core/src/main/java/brooklyn/location/access/PortForwardManagerImpl.java
+++ b/core/src/main/java/brooklyn/location/access/PortForwardManagerImpl.java
@@ -26,6 +26,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.slf4j.Logger;
@@ -38,8 +39,10 @@ import brooklyn.location.Location;
 import brooklyn.location.basic.AbstractLocation;
 import brooklyn.mementos.LocationMemento;
 import brooklyn.util.collections.MutableMap;
+import brooklyn.util.exceptions.Exceptions;
 
 import com.google.common.base.Objects.ToStringHelper;
+import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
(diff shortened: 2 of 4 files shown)


core/src/main/java/brooklyn/location/access/PortForwardManagerImpl.java
```diff
@@ -38,12 +25,25 @@
...
 import brooklyn.mementos.LocationMemento;
 import brooklyn.util.collections.MutableMap;
-
 import com.google.common.base.Objects.ToStringHelper;
+import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.net.HostAndPort;
+import org.slf4j.Logger;
```
aledsage: Try not to let Intellij reorganise the imports. The vast majority of our code uses the Eclipse defaults for imports, so we should stick to that in classes that use it.

You can set the formatting in Intellij to use the Eclipse defaults, but I'm not sure exactly how. Check with @richardcloudsoft ? If you find out, can you please update https://brooklyn.incubator.apache.org/v/latest/dev/env/ide/index.html

But I won't let that block merging of this PR.

core/src/main/java/brooklyn/location/access/PortForwardManager.java
```diff
@@ -46,6 +47,42 @@
...
 public interface PortForwardManager extends Location {
 
+    @Beta
+    class AssociationMetadata {
+        private final String publicIpId;
+        private final HostAndPort publicEndpoint;
+        private final Location location;
+        private final int privatePort;
+
+        public AssociationMetadata(String publicIpId, HostAndPort publicEndpoint, Location location, int privatePort) {
```
aledsage: Worth a javadoc to say: `Users are discouraged from calling this constructor; the signature may change in future releases. Instead, instances will be created automatically by Brooklyn to be passed to the {@link AssociationListener#onEvent(AssociationMetadata)} method.`

core/src/main/java/brooklyn/location/access/PortForwardManager.java
```diff
@@ -92,6 +129,12 @@
      * subsequently be looked up using {@link #lookup(String, int)}.
      */
     public void associate(String publicIpId, HostAndPort publicEndpoint, int privatePort);
+
+    @Beta
+    public void addAssociationListener(AssociationListener listener, Predicate<? super AssociationMetadata> filter);
```
aledsage: Worth javadoc to say `Registers a listener, which will be notified each time a new port mapping is associated. See {@link #associate(String, HostAndPort, int)} and {@link #associate(String, HostAndPort, Location, int)}.`

core/src/main/java/brooklyn/location/access/PortForwardManager.java
```diff
@@ -46,6 +47,42 @@
...
+        }
+
+        public int getPrivatePort() {
+            return privatePort;
+        }
+    }
+
+    @Beta
+    interface AssociationListener {
+        void onEvent(AssociationMetadata metadata);
```
aledsage: Do you think we should call this on removal of a port mapping? I think we probably should.

The reason is that it impacts the choice of method name, so impacts backwards compatibility. How about changing this to `onAssociationCreated(AssociationMetadata)` and `onAssociationDeleted(AssociationMetadata)`?

For the impl of the latter, see `PortForwardManagerImpl.forgetPortMapping(...)` and `forgetPortMappings(...)` - i.e. everywhere that it removes from the `PortForwardManagerImpl.mappings` map.

core/src/main/java/brooklyn/location/access/PortForwardManagerImpl.java
```diff
@@ -160,11 +162,22 @@ protected int getNextPort() {
...
     @Override
     public void associate(String publicIpId, HostAndPort publicEndpoint, int privatePort) {
         associateImpl(publicIpId, publicEndpoint, null, privatePort);
+        emitAssociationEvent(publicIpId, publicEndpoint, null, privatePort);
+    }
+
+    private void emitAssociationEvent(String publicIpId, HostAndPort publicEndpoint, Location location, int privatePort) {
+        AssociationMetadata metadata = new AssociationMetadata(publicIpId, publicEndpoint, location, privatePort);
+        for (AssociationListener listener : associationListeners.keySet()) {
+            if (associationListeners.get(listener).apply(metadata)) {
```
aledsage: Looks dangerous in a multi-threaded world. You've separated the call of `keySet()` and `get(listener)`, so you've introduced the chance for another thread to remove the listener in between those two calls.

Better to do the following (thus allowing `ConcurrentHashMap` to give its guarantees):

    for (Map.Entry<AssociationListener, Predicate<? super AssociationMetadata> entry : associationListeners.entrySet()) {
        if (entry.getValue().apply(metadata)) {
            entry.getKey().onEvent(metadata);
        }
    }

core/src/main/java/brooklyn/location/access/PortForwardManagerImpl.java
```diff
@@ -78,7 +78,9 @@
     private static final Logger log = LoggerFactory.getLogger(PortForwardManagerImpl.class);
     
     protected final Map<String,PortMapping> mappings = new LinkedHashMap<String,PortMapping>();
-    
+
+    private Map<AssociationListener, Predicate<? super AssociationMetadata>> associationListeners = new ConcurrentHashMap<AssociationListener, Predicate<? super AssociationMetadata>>();
```
aledsage: Declare the field `final`. It makes reasoning about the code simpler for folk reading or modifying it later.

aledsage:
Looks good; a few comments worth addressing. Ping me when it's ready to review again.

core/src/main/java/brooklyn/location/access/PortForwardManager.java
```diff
@@ -46,6 +47,42 @@
 @Beta
 public interface PortForwardManager extends Location {
 
+    @Beta
+    class AssociationMetadata {
```
aledsage: Worth adding `toString()` to this - you rely on a nice toString in a debug message of https://github.com/brooklyncentral/advanced-networking/pull/39

e.g.:

    public String toString() {
        return Objects.toStringHelper()
                .add("publicIpId", publicIpId)
                .add("publicEndpoint", publicEndpoint)
                .add("location", location)
                .add("privatePort", privatePort)
                .build();
    }

core/src/main/java/brooklyn/location/access/PortForwardManagerImpl.java
nakomis: I'm finding inconsistencies in how Eclipse organises the imports. Configuring IntelliJ to have the same imports list as Eclipse unfortunately doesn't work. As @neykov put it: "One of the IDEs (don't remember which) was interleaving packages not specified in the list, while the other was putting them at the start and then listing the remaining packages."

nakomis:
PR comments addressed

core/src/main/java/brooklyn/location/access/PortForwardManagerImpl.java
```diff
@@ -208,6 +223,7 @@ public boolean forgetPortMapping(String publicIpId, int publicPort) {
         PortMapping old;
         synchronized (mutex) {
             old = mappings.remove(makeKey(publicIpId, publicPort));
+            emitAssociationDeletedEvent(associationMetadataFromPortMapping(old));
```
aledsage: Guard this with `if (old != null)` - i.e. was something really removed?

core/src/main/java/brooklyn/location/access/PortForwardManager.java
```diff
@@ -92,6 +144,16 @@
      * subsequently be looked up using {@link #lookup(String, int)}.
      */
     public void associate(String publicIpId, HostAndPort publicEndpoint, int privatePort);
+
+    /**
+     * Registers a listener, which will be notified each time a new port mapping is associated. See {@link #associate(String, HostAndPort, int)}
+     * and {@link #associate(String, HostAndPort, Location, int)}.
+     */
+    @Beta
+    public void addAssociationListener(AssociationListener listener, Predicate<? super AssociationMetadata> filter);
```
aledsage: Could do with a unit test in PortForwardManagerTest, that the `onAssociationCreated` and `onAssociationDeleted` methods get called.

core/src/main/java/brooklyn/location/access/PortForwardManagerImpl.java
```diff
@@ -181,6 +187,15 @@ protected void associateImpl(String publicIpId, HostAndPort publicEndpoint, Loca
         onChanged();
     }
 
+    private void emitAssociationCreatedEvent(String publicIpId, HostAndPort publicEndpoint, Location location, int privatePort) {
+        AssociationMetadata metadata = new AssociationMetadata(publicIpId, publicEndpoint, location, privatePort);
+        for (Map.Entry<AssociationListener, Predicate<? super AssociationMetadata>> entry : associationListeners.entrySet()) {
+            if (entry.getValue().apply(metadata)) {
+                entry.getKey().onAssociationCreated(metadata);
```
aledsage: Should wrap this in try-catch block, to log.warn any exception. An exception in one listener shouldn't cause the other listeners to not be called. However, in the catch block first do `Exceptions.propagateIfFatal(e)`.

Same goes for the call to `onAssociationDeleted(metadata)`.

nakomis:
PR comments addressed again

core/src/main/java/brooklyn/location/access/PortForwardManagerImpl.java
```diff
@@ -181,6 +188,20 @@ protected void associateImpl(String publicIpId, HostAndPort publicEndpoint, Loca
...
 
+    private void emitAssociationCreatedEvent(String publicIpId, HostAndPort publicEndpoint, Location location, int privatePort) {
+        AssociationMetadata metadata = new AssociationMetadata(publicIpId, publicEndpoint, location, privatePort);
+        for (Map.Entry<AssociationListener, Predicate<? super AssociationMetadata>> entry : associationListeners.entrySet()) {
+            if (entry.getValue().apply(metadata)) {
+                try {
+                    entry.getKey().onAssociationCreated(metadata);
+                } catch (Exception e) {
+                    Exceptions.propagateIfFatal(e);
+                    log.warn("Exception thrown when emitting association creation event {}: {}", metadata, e);
```
aledsage: This will just give the toString of the exception, rather than the stacktrace - it would make debugging problems very hard. For example, it might tell you there was an NPE without ever telling you which line that null pointer happened on. Suggest changing it to:

    log.warn("Exception thrown when emitting association creation event "+metadata, e);

Same for log message in `emitAssociationDeletedEvent`.

aledsage:
Thanks - merging.

PR #542, opened 2015-03-06
Mirroring entities should wait for a proper restart of remote

neykov:
... and other restart improvements.

There is a window of time when restarting Brooklyn where it returns 404 for its entities, making mirroring entities unmanage themselves. Fix it by extending the existing filtering logic to return 403 for explicitly marked endpoints while remote is STANDBY or rebinding.
I had to use a delay after master promotion because there is still a small period of time just after the node becomes master but before it starts rebinding where it will return 404 for entities and there are no flags to check for it.

### DIFF
--- a/core/src/main/java/brooklyn/management/ha/HighAvailabilityManagerImpl.java
+++ b/core/src/main/java/brooklyn/management/ha/HighAvailabilityManagerImpl.java
@@ -1077,4 +1077,13 @@ public class HighAvailabilityManagerImpl implements HighAvailabilityManager {
         return result;
     }
     
+    @Override
+    public long getLastStateChange() {
+        if (nodeStateHistory.size() > 0) {
+            return (Long)nodeStateHistory.get(0).get("timestamp");
+        } else {
+            return 0;
+        }
+    }
+    
 }
--- a/software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
+++ b/software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
@@ -47,6 +47,7 @@ import brooklyn.entity.brooklynnode.effector.SetHighAvailabilityModeEffectorBody
 import brooklyn.entity.brooklynnode.effector.SetHighAvailabilityPriorityEffectorBody;
 import brooklyn.entity.effector.EffectorBody;
 import brooklyn.entity.effector.Effectors;
+import brooklyn.entity.software.MachineLifecycleEffectorTasks;
 import brooklyn.entity.trait.Startable;
 import brooklyn.event.feed.ConfigToAttributes;
 import brooklyn.event.feed.http.HttpFeed;
@@ -155,7 +156,18 @@ public class BrooklynNodeImpl extends SoftwareProcessImpl implements BrooklynNod
     @Override
     protected void preStop() {
         super.preStop();
-        shutdownGracefully();
+        if (MachineLifecycleEffectorTasks.canStop(getStopProcessModeParam(), this)) {
+            shutdownGracefully();
+        }
+    }
+
+    private StopMode getStopProcessModeParam() {
+        ConfigBag parameters = BrooklynTaskTags.getCurrentEffectorParameters();
+        if (parameters != null) {
+            return parameters.get(StopSoftwareParameters.STOP_PROCESS_MODE);
+        } else {
+            return StopSoftwareParameters.STOP_PROCESS_MODE.getDefaultValue();
+        }
     }
 
     @Override
--- a/software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
+++ b/software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
@@ -40,6 +40,7 @@ import brooklyn.entity.basic.Entities;
 import brooklyn.entity.basic.Lifecycle;
 import brooklyn.entity.basic.ServiceStateLogic;
 import brooklyn.entity.basic.ServiceStateLogic.ServiceNotUpLogic;
+import brooklyn.entity.basic.SoftwareProcess.StopSoftwareParameters.StopMode;
 import brooklyn.entity.basic.SoftwareProcessImpl;
 import brooklyn.entity.brooklynnode.effector.BrooklynNodeUpgradeEffectorBody;
 import brooklyn.entity.brooklynnode.effector.SetHighAvailabilityModeEffectorBody;
@@ -154,17 +155,30 @@ public class BrooklynNodeImpl extends SoftwareProcessImpl implements BrooklynNod
     @Override
     protected void preStop() {
         super.preStop();
+        shutdownGracefully();
+    }
 
+    @Override
+    protected void preRestart() {
+        super.preRestart();
+        shutdownGracefully();
+        DynamicTasks.queue("post-shutdown", new Runnable() { public void run() {
+            //set by shutdown - clear it so the entity starts cleanly. Does the indicator bring any value at all?
+            ServiceNotUpLogic.clearNotUpIndicator(BrooklynNodeImpl.this, SHUTDOWN.getName());
+        }});
+    }
+
+    private void shutdownGracefully() {
         // Shutdown only if accessible: any of stop_* could have already been called.
         // Don't check serviceUp=true because stop() will already have set serviceUp=false && expectedState=stopping
         if (Boolean.TRUE.equals(getAttribute(BrooklynNode.WEB_CONSOLE_ACCESSIBLE))) {
             queueShutdownTask();
             queueWaitExitTask();
         } else {
-            log.info("Skipping children.isEmpty check and shutdown call, because web-console not up for {}", this);
+            log.info("Skipping graceful shutdown call, because web-console not up for {}", this);
         }
     }
-    
+
     private void queueWaitExitTask() {
         //give time to the process to die gracefully after closing the shutdown call
         DynamicTasks.queue(Tasks.builder().name("wait for graceful stop").body(new Runnable() {
--- a/software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
+++ b/software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
@@ -176,8 +176,9 @@ public class BrooklynNodeImpl extends SoftwareProcessImpl implements BrooklynNod
     @Override
     protected void preRestart() {
         super.preRestart();
+        //restart will kill the process, try to shut down before that
         shutdownGracefully();
-        DynamicTasks.queue("post-shutdown", new Runnable() { public void run() {
+        DynamicTasks.queue("pre-restart", new Runnable() { public void run() {
             //set by shutdown - clear it so the entity starts cleanly. Does the indicator bring any value at all?
             ServiceNotUpLogic.clearNotUpIndicator(BrooklynNodeImpl.this, SHUTDOWN.getName());
         }});
@@ -220,7 +221,7 @@ public class BrooklynNodeImpl extends SoftwareProcessImpl implements BrooklynNod
     @Override
     protected void postStop() {
         super.postStop();
-        if (isStopMachine()) {
+        if (isMachineStopped()) {
             // Don't unmanage in entity's task context as it will self-cancel the task. Wait for the stop effector to complete.
             // If this is not enough (still getting Caused by: java.util.concurrent.CancellationException: null) then
             // we could search for the top most task with entity context == this and wait on it. Even stronger would be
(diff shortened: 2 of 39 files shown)


core/src/main/java/brooklyn/management/ha/HighAvailabilityManagerImpl.java
```diff
@@ -1077,4 +1077,13 @@ public String toString() {
         return result;
     }
     
+    @Override
+    public long getLastStateChange() {
+        if (nodeStateHistory.size() > 0) {
+            return (Long)nodeStateHistory.get(0).get("timestamp");
+        } else {
+            return 0;
```
aledsage: Personal preference for returning `-1` to signify not available (as 0 is an unlikely but valid timestamp)

software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
```diff
@@ -153,17 +158,41 @@ protected void preStopConfirmCustom() {
...
         super.preStop();
+        if (MachineLifecycleEffectorTasks.canStop(getStopProcessModeParam(), this)) {
+            shutdownGracefully();
+        }
+    }
+
+    private StopMode getStopProcessModeParam() {
+        ConfigBag parameters = BrooklynTaskTags.getCurrentEffectorParameters();
+        if (parameters != null) {
+            return parameters.get(StopSoftwareParameters.STOP_PROCESS_MODE);
```
aledsage: also need to consider the deprecated `STOP_MACHINE` parameter option?

software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
```diff
@@ -153,17 +158,41 @@ protected void preStopConfirmCustom() {
...
+        } else {
+            return StopSoftwareParameters.STOP_PROCESS_MODE.getDefaultValue();
+        }
+    }
 
+    @Override
+    protected void preRestart() {
+        super.preRestart();
+        shutdownGracefully();
+        DynamicTasks.queue("post-shutdown", new Runnable() { public void run() {
```
aledsage: Should be `"pre-restart`" presumably?

software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
```diff
@@ -153,17 +158,41 @@ protected void preStopConfirmCustom() {
...
+            return parameters.get(StopSoftwareParameters.STOP_PROCESS_MODE);
+        } else {
+            return StopSoftwareParameters.STOP_PROCESS_MODE.getDefaultValue();
+        }
+    }
 
+    @Override
+    protected void preRestart() {
+        super.preRestart();
+        shutdownGracefully();
```
aledsage: Deserves a comment to say why doing shutdown here

software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
```diff
@@ -202,6 +229,15 @@ protected void postStop() {
         }
     }
 
+    private boolean isStopMachine() {
```
aledsage: I was thrown by this method name when I looked at the impl.

Should this method be called `isStoppingMachine()` or `isMachineStopped()` (I prefer the former)?

software/base/src/test/java/brooklyn/entity/software/MachineLifecycleEffectorTasksTest.java
```diff
@@ -23,11 +23,17 @@
...
 
+import brooklyn.entity.basic.BasicEntityImpl;
+import brooklyn.entity.basic.Lifecycle;
+import brooklyn.entity.basic.SoftwareProcess;
 import brooklyn.entity.basic.SoftwareProcess.StopSoftwareParameters.StopMode;
 
 public class MachineLifecycleEffectorTasksTest {
     public static boolean canStop(StopMode stopMode, boolean isEntityStopped) {
-        return MachineLifecycleEffectorTasks.canStop(stopMode, isEntityStopped);
+        BasicEntityImpl entity = new BasicEntityImpl();
```
aledsage: ok for now, but at some point calling the entity constructor directly may give us very surprising subsequent behaviour (e.g. because not managed, and no management context injected).

usage/jsgui/src/test/java/brooklyn/rest/jsgui/BrooklynJavascriptGuiLauncherTest.java
```diff
@@ -60,8 +61,15 @@ public void testJavascriptWithRest() throws Exception {
         checkUrlContains("/v1/catalog/entities", "Tomcat");
     }
 
-    protected void checkUrlContains(String path, String text) {
-        HttpTestUtils.assertContentContainsText(rootUrl()+path, text);
+    protected void checkUrlContains(final String path, final String text) {
+        //Server may return 403 until it loads completely, wait a bit
+        //until it stabilizes.
+        Asserts.succeedsEventually(new Runnable() {
```
aledsage: Can use `HttpTestUtils.assertContentEventuallyContainsText(rootUrl()+path, text)`.

I'd really like consistency for our assertXxxEventuallyYyy versus assertXxxYyyEventually. Personally, I much prefer the latter because we often also have an assertXxxYyy method - having the Eventually as a suffix keeps the methods together more. But that's an aside...

usage/rest-server/src/main/java/brooklyn/rest/filter/HaHotStateRequired.java
```diff
@@ -0,0 +1,28 @@
...
+package brooklyn.rest.filter;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.TYPE})
+public @interface HaHotStateRequired {}
```
aledsage: worth javadoc saying what annotation means, and where it should be used

usage/rest-server/src/main/java/brooklyn/rest/resources/ApplicationResource.java
```diff
@@ -82,6 +83,7 @@
 import com.google.common.collect.FluentIterable;
 import com.google.common.collect.Iterables;
 
+@HaHotStateRequired
```
aledsage: Is it right to have these annotations on the Resource rather than the API interface? No strong feelings.

usage/rest-server/src/test/java/brooklyn/rest/BrooklynRestApiLauncher.java
```diff
@@ -150,6 +152,11 @@ public BrooklynRestApiLauncher withoutJsgui() {
         this.deployJsgui = false;
         return this;
     }
+    
+    public BrooklynRestApiLauncher disableHighAvailability(boolean value) {
```
aledsage: Personal preference for fields/methods to be written in the positive rather than the negative.
We could call the method/field `enableHighAvailability(boolean)`, defaulting to true.

Other such examples include:

* `JcloudsLocation.USE_JCLOUDS_SSH_INIT` rather than `DO_NOT_USE_JCLOUDS_SSH_INIT`.
* `org.testng.annotations.Test.enabled` (rather than disabled)

usage/rest-server/src/test/java/brooklyn/rest/HaHotCheckTest.java
```diff
@@ -0,0 +1,142 @@
...
+import com.sun.jersey.api.client.ClientResponse;
+import com.sun.jersey.api.client.WebResource.Builder;
+import com.sun.jersey.api.core.ResourceConfig;
+
+public class HaHotCheckTest extends BrooklynRestResourceTest {
+
+    private ManagementContextMock mgmtMock;
+
+    //Treat before/after class methods as before/after method
+    //otherwise they should be static.
```
aledsage: I don't follow. The same instance is used for all the test methods, so `@BeforeClass` should not be static.

usage/rest-server/src/test/java/brooklyn/rest/HaHotCheckTest.java
```diff
@@ -0,0 +1,142 @@
...
+    @BeforeClass(alwaysRun=true)
+    public void setUpClass() {}
```
aledsage: It's a bit confusing overriding the setUp and changing it from `@BeforeClass` in super to `@BeforeMethod`.

usage/rest-server/src/test/java/brooklyn/rest/testing/mocks/HighAvailabilityManagerMock.java
```diff
@@ -0,0 +1,121 @@
...
+
+import java.util.Map;
+
+import brooklyn.management.ha.HighAvailabilityManager;
+import brooklyn.management.ha.HighAvailabilityMode;
+import brooklyn.management.ha.ManagementNodeState;
+import brooklyn.management.ha.ManagementPlaneSyncRecord;
+import brooklyn.management.ha.ManagementPlaneSyncRecordPersister;
+
+public class HighAvailabilityManagerMock implements HighAvailabilityManager {
```
aledsage: Strictly speaking, this is a stub rather than a mock (http://martinfowler.com/articles/mocksArentStubs.html).
Personal preference for the class name to have `Stub` suffix instead.

aledsage:
Looks really good! A few minor comments to be addressed, but very close to being merged.

FYI I did a tiny bit of this in https://github.com/apache/incubator-brooklyn/pull/547 (see commits https://github.com/aledsage/incubator-brooklyn/commit/fe5ecca123525b4d6c8cd9f5c0049e3d5b6b12b6 "BrooklynNode: only unmanage if VM terminating", https://github.com/aledsage/incubator-brooklyn/commit/f736dfc84f79d3aa6275747ad3bdf6e4d9181a0d "Add SoftwareProcess.preRestart (for BrooklynNode)", and https://github.com/aledsage/incubator-brooklyn/commit/d38b652519cf4b46632c1b8dd06c52498eb3cdd3 "BrooklynCluster: add memberSpec + test").

I prefer your impl of `preRestartCustom` (that you included the method on the superclass as well, and that you also added `postRestart` for symmetry.

The test I added is useful though. I can look at that again once this is merged and I rebase my PR, or you can pull in my test if you fancy.

ahgittin:
agree, changes here look really good, and would be nice to pull in @aledsage 's addl test. /cc @neykov

software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeImpl.java
neykov: This concerns the process only, stop machine mode not relevant.

usage/rest-server/src/main/java/brooklyn/rest/resources/ApplicationResource.java
neykov: Both ways can be used, but having class-wide annotation is better in this case as it is too easy to miss the annotation in newly added methods. If we need to be more selective, then method-scope is implemented as well.

usage/rest-server/src/test/java/brooklyn/rest/BrooklynRestApiLauncher.java
neykov: I agree with you, I also prefer positive naming. In this case though it's more specific. When the flag is `disableHighAvailability=true` the code calls
```
 mgmt.getHighAvailabilityManager().disabled();
```


Having it named `enableHighAvailability=true` I would expect that it calls start on the HA. By adding the flag I wanted to be able to say "don't disable HA".

So wdyt, I don't mind changing it, but `enableHighAvailability` could lead to unmet expectations?

neykov:
Thanks for the great review @aledsage. Addressed comments and added your test case.

Merge conflicts need to be resolved due to https://github.com/apache/incubator-brooklyn/pull/553, needs some extra attention.

usage/rest-server/src/test/java/brooklyn/rest/BrooklynRestApiLauncher.java
aledsage: Makes sense; agreed.

aledsage:
@neykov once you've resolved merge conflicts, this will be ready to merge.

neykov:
Rebased, merge conflicts resolved.

neykov:
Rebased, merge conflicts resolved.

aledsage:
Thanks; merging now.

ahgittin:
just came across the rest filter -- very nice piece of work.  it breaks some things in the gui when connecting to a standby server, and i wondered why.  not a big problem, and probably wiser to fix that in the JS side, keeping the filter changes here, but wanted to note that.

also it might be nice to merge `HaHotCheckResourceFilter` with the `HaMasterCheckFilter` but again minor.  i've put TODO's in the code in PR coming shortly.

PR #543, opened 2015-03-06
Props given as arguments to SshMachineLocation.connectSsh trump others

sjcorbett:
So, for example, the user that the SSH connection is made as can be overridden by providing `PROP_USER` in the flags. Previously it was overwritten by the inclusion of the values in `config().getBag()`.

Happy to be informed that there's already a way to switch the user that the SSH connection is made as.


aledsage:
Looks sensible; merging
(would be nice to have a test as well).

PR #545, opened 2015-03-06
Encapsulate all CAMP calls in a single place. Remove code repetition.

andreaturli:
@neykov as it is based on your #436 could you please make sure I'm not missing anything here?

### DIFF
--- a/core/src/main/java/brooklyn/management/internal/EntityManagementUtils.java
+++ b/core/src/main/java/brooklyn/management/internal/EntityManagementUtils.java
@@ -124,12 +124,12 @@ public class EntityManagementUtils {
         return (EntitySpec<? extends Application>) createSpec(mgmt, yaml, loader);
     }
 
-    private static <T, SpecT> SpecT createEntitySpec(String symbolicName, CampPlatform camp,
+    private static EntitySpec<? extends Application> createEntitySpec(String symbolicName, CampPlatform camp,
                                                                       DeploymentPlan plan, BrooklynClassLoadingContext loader) {
         AssemblyTemplate at = registerDeploymentPlan(camp, plan, loader);
         AssemblyTemplateInstantiator instantiator = newInstantiator(at);
         if (instantiator instanceof AssemblyTemplateSpecInstantiator) {
-            return (SpecT) ((AssemblyTemplateSpecInstantiator)instantiator).createNestedSpec(at, camp, loader, MutableSet.of(symbolicName));
+            return (EntitySpec<? extends Application>) ((AssemblyTemplateSpecInstantiator)instantiator).createNestedSpec(at, camp, loader, MutableSet.of(symbolicName));
         } else {
             throw new IllegalStateException("Unable to instantiate YAML; incompatible instantiator " + instantiator + " for " + at);
         }
--- a/core/src/main/java/brooklyn/management/internal/EntityManagementUtils.java
+++ b/core/src/main/java/brooklyn/management/internal/EntityManagementUtils.java
@@ -18,14 +18,14 @@
  */
 package brooklyn.management.internal;
 
-import io.brooklyn.camp.CampPlatform;
-import io.brooklyn.camp.spi.Assembly;
-import io.brooklyn.camp.spi.AssemblyTemplate;
-import io.brooklyn.camp.spi.instantiate.AssemblyTemplateInstantiator;
+import static com.google.common.base.Preconditions.checkNotNull;
 
+import java.io.Reader;
 import java.io.StringReader;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
-import java.util.Set;
+import java.util.Map;
 import java.util.concurrent.Callable;
 
 import javax.annotation.Nullable;
@@ -33,7 +33,19 @@ import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.annotations.Beta;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+
+import brooklyn.basic.AbstractBrooklynObjectSpec;
 import brooklyn.camp.brooklyn.api.AssemblyTemplateSpecInstantiator;
+import brooklyn.catalog.CatalogItem;
+import brooklyn.catalog.internal.BasicBrooklynCatalog.BrooklynLoaderTracker;
+import brooklyn.catalog.internal.CatalogItemBuilder;
+import brooklyn.catalog.internal.CatalogItemDtoAbstract;
+import brooklyn.catalog.internal.CatalogUtils;
 import brooklyn.config.BrooklynServerConfig;
 import brooklyn.config.ConfigKey;
 import brooklyn.entity.Application;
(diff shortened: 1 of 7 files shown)


core/src/main/java/brooklyn/management/internal/EntityManagementUtils.java
```diff
@@ -82,41 +100,90 @@
...
+
     @SuppressWarnings("unchecked")
-    public static <T extends Application> T createUnstarted(ManagementContext mgmt, AssemblyTemplate at) {
-        CampPlatform camp = getCampPlatform(mgmt).get();
-        AssemblyTemplateInstantiator instantiator;
+    public static EntitySpec<? extends Application> createEntitySpec(ManagementContext mgmt, Reader yaml, BrooklynClassLoadingContext loader) {
+        return (EntitySpec<? extends Application>) createSpec(mgmt, yaml, loader);
+    }
+
+    private static <T, SpecT> SpecT createEntitySpec(String symbolicName, CampPlatform camp,
```
neykov: No need for the generics, can return `EntitySpec<? extends Application>`

usage/camp/src/main/java/io/brooklyn/camp/brooklyn/YamlLauncherAbstract.java
```diff
@@ -79,8 +80,8 @@ public Application launchAppYaml(String url) {
 
     public Application launchAppYaml(String url, boolean waitForTasksToComplete) {
         try {
-            Reader input = Streams.reader(new ResourceUtils(this).getResourceFromUrl(url));
-            Application app = launchAppYaml(input, waitForTasksToComplete);
```
neykov: This is the wrong method to update, replace the one that was called here. For a global check if the patch was successful check for references like `registerDeploymentPlan`, `instantiate`, they should we restricted to the `EntityManagementUtils` class only (outside of camp projects).

neykov:
I am getting the following error when trying to start an app from yaml:
`com.sun.proxy.$Proxy78 cannot be cast to brooklyn.entity.Application`
with yaml
```
location: broo
services:
- type: brooklyn.entity.webapp.tomcat.TomcatServer
```

Something wrong with my build?

andreaturli:
thanks @neykov for your comments. I've addressed them but I think there are still problems 

Failed tests:
  ApplicationResourceTest.testDeployApplicationYaml:236 expected [simple-app-yaml] but found [Application (p6lRzMdf)]
  ApplicationResourceTest.testReferenceCatalogEntity:249 response is Client response status: 500 expected [true] but found [false]

not sure what is causing them to fail. can you @neykov help on that?

andreaturli:
@neykov looks like the problem is with 'AssemblyTemplateSpecInstantiator.createNestedSpec' used by 'EntityManagementUtils.createEntitySpec' 
I know you spent some time on it, so maybe it is easier for you to understand the issue

usage/camp/src/main/java/io/brooklyn/camp/brooklyn/YamlLauncherAbstract.java
```diff
@@ -130,4 +126,13 @@ public void destroyAll() {
         }
     }
 
+   private Application createFromUrl(Reader reader) {
```
neykov: The argument is no longer a URL, rename method. Also the TODO below is not needed in this case.

core/src/main/java/brooklyn/management/internal/EntityManagementUtils.java
```diff
@@ -82,41 +100,90 @@
...
+    public static EntitySpec<? extends Application> createEntitySpec(ManagementContext mgmt, Reader yaml, BrooklynClassLoadingContext loader) {
+        return (EntitySpec<? extends Application>) createSpec(mgmt, yaml, loader);
+    }
+
+    private static EntitySpec<? extends Application> createEntitySpec(String symbolicName, CampPlatform camp,
+                                                                      DeploymentPlan plan, BrooklynClassLoadingContext loader) {
+        AssemblyTemplate at = registerDeploymentPlan(camp, plan, loader);
+        AssemblyTemplateInstantiator instantiator = newInstantiator(at);
+        if (instantiator instanceof AssemblyTemplateSpecInstantiator) {
+            return (EntitySpec<? extends Application>) ((AssemblyTemplateSpecInstantiator)instantiator).createNestedSpec(at, camp, loader, MutableSet.of(symbolicName));
```
neykov: Make sure that `createNestedSpec` is called only from `BasicBrooklynCatalog` code (as the `createEntitySpec` wrapping it). Atm it's used for creating apps, thus the error.
Also rename the methods coming from `BasicBrooklynCatalog` to indicate that they are used for catalog items.

core/src/main/java/brooklyn/management/internal/EntityManagementUtils.java
```diff
@@ -82,41 +100,90 @@
...
+    }
+
+    public static AbstractBrooklynObjectSpec<?, ?> createSpec(ManagementContext mgmt, String yaml, BrooklynClassLoadingContext loader) {
+        return createSpec(mgmt, new StringReader(yaml), loader);
+    }
+
+    public static AbstractBrooklynObjectSpec<?, ?> createSpec(ManagementContext mgmt, Reader yaml, BrooklynClassLoadingContext loader) {
+        Preconditions.checkNotNull(mgmt, "mgmt");
+        Preconditions.checkNotNull(mgmt, "yaml");
+        Preconditions.checkNotNull(mgmt, "loader");
```
neykov: Copy-paste discrepancy

core/src/main/java/brooklyn/management/internal/EntityManagementUtils.java
```diff
@@ -82,41 +100,90 @@
...
+        DeploymentPlan plan = parseDeploymentPlan(camp, yaml);
+        return createSpec(null, mgmt, plan, loader);
+    }
+
+    private static AbstractBrooklynObjectSpec<?, ?> createSpec(String symbolicName, ManagementContext mgmt,
+                                                               DeploymentPlan plan, BrooklynClassLoadingContext loader) {
+        if (isPolicyPlan(plan)) {
+            return createPolicySpec(plan, loader);
+        } else {
+            return createEntitySpec(symbolicName, getCampPlatform(mgmt), plan, loader);
```
neykov: Here is where things go sideways. The method `createEntitySpec` called here is coming from `BasicBrooklynCatalog`, calling into `createNestedSpec`. You should instead use https://github.com/neykov/incubator-brooklyn/commit/55305ea9bf05598a8ac2c811691008818edbf321#diff-b713fc5de648db4fede03fa71cfd1c6fR315 which calls into `AssemblyTemplateSpecInstantiator.createSpec`.

andreaturli: Notice that currently BasicBrooklynCatalog looks like https://github.com/apache/incubator-brooklyn/blob/1b0037b609b4ad11a803522e696d6bfcac00f57f/core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java#L299-L325
so I think here it is doing the same

ahgittin:
great idea here @andreaturli .  i'm not sure `EntityManagementUtils` is the natural home but then i'm not sure what is.

it will probably need redoing however given changes since this PR and also in #585.  tedious, but i don't think there are any difficult changes.

andreaturli:
Closing as #760 is what we want to merge

PR #546, opened 2015-03-10
Add recursive sanitation and move to own class

robertgmoss:
Deprecates the existing method in Entities and updates all references.  Partial fix for BROOKLYN-10

### DIFF
--- a/core/src/main/java/brooklyn/entity/basic/Entities.java
+++ b/core/src/main/java/brooklyn/entity/basic/Entities.java
@@ -270,17 +270,20 @@ public class Entities {
                 (v instanceof CharSequence&& ((CharSequence)v).length() == 0);
     }
 
+    /**
+     * @deprecated since 0.7
+     */
+    @Deprecated
     public static Map<String,Object> sanitize(ConfigBag input) {
-        return sanitize(input.getAllConfig());
+        return Sanitizer.sanitize(input.getAllConfig());
     }
 
+    /**
+     * @deprecated since 0.7
+     */
+    @Deprecated
     public static <K> Map<K,Object> sanitize(Map<K,?> input) {
-        Map<K,Object> result = Maps.newLinkedHashMap();
-        for (Map.Entry<K,?> e: input.entrySet()) {
-            if (isSecret(""+e.getKey())) result.put(e.getKey(), "xxxxxxxx");
-            else result.put(e.getKey(), e.getValue());
-        }
-        return result;
+        return Sanitizer.sanitize(input);
     }
 
     public static void dumpInfo(Iterable<? extends Entity> entities) {
--- a/core/src/main/java/brooklyn/entity/basic/Entities.java
+++ b/core/src/main/java/brooklyn/entity/basic/Entities.java
@@ -127,7 +127,10 @@ public class Entities {
     /**
      * Names that, if they appear anywhere in an attribute/config/field indicates that it
      * may be private, so should not be logged etc.
+     * 
+     * @deprecated since 0.7; instead use {@link Sanitizer#SECRET_NAMES}
      */
+    @Deprecated
     public static final List<String> SECRET_NAMES = ImmutableList.of(
             "password",
             "passwd",
@@ -250,12 +253,12 @@ public class Entities {
         return invokeEffector(callingEntity, entitiesToCall, effector, Collections.<String,Object>emptyMap());
     }
 
+    /**
+     * @deprecated since 0.7; instead use {@link Sanitizer#IS_SECRET_PREDICATE.apply(Object)}
+     */
+    @Deprecated
     public static boolean isSecret(String name) {
-        String lowerName = name.toLowerCase();
-        for (String secretName : SECRET_NAMES) {
-            if (lowerName.contains(secretName)) return true;
-        }
-        return false;
+        return Sanitizer.IS_SECRET_PREDICATE.apply(name);
     }
 
     public static boolean isTrivial(Object v) {
--- a/core/src/main/java/brooklyn/entity/basic/Entities.java
+++ b/core/src/main/java/brooklyn/entity/basic/Entities.java
@@ -271,15 +271,15 @@ public class Entities {
     }
 
     /**
-     * @deprecated since 0.7
+     * @deprecated since 0.7; instead use {@link Sanitizer#sanitize(ConfigBag)}
      */
     @Deprecated
     public static Map<String,Object> sanitize(ConfigBag input) {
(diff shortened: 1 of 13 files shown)


core/src/main/java/brooklyn/entity/basic/Entities.java
```diff
@@ -270,17 +270,20 @@ public static boolean isTrivial(Object v) {
                 (v instanceof CharSequence&& ((CharSequence)v).length() == 0);
     }
 
+    /**
+     * @deprecated since 0.7
+     */
+    @Deprecated
     public static Map<String,Object> sanitize(ConfigBag input) {
-        return sanitize(input.getAllConfig());
+        return Sanitizer.sanitize(input.getAllConfig());
```
aledsage: I'd call `Sanitize(input)`

core/src/main/java/brooklyn/entity/basic/Entities.java
```diff
@@ -270,17 +270,20 @@ public static boolean isTrivial(Object v) {
                 (v instanceof CharSequence&& ((CharSequence)v).length() == 0);
     }
 
+    /**
+     * @deprecated since 0.7
```
aledsage: Can you include in javadoc which method to call (e.g. `@deprecated since 0.7; instead use {@link Sanitizer#sanitize(ConfigBag)}`

core/src/main/java/brooklyn/entity/basic/Sanitizer.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public static <K> Map<K, Object> sanitize(Map<K, ?> input) {
+        return sanitize(input, Sets.newHashSet());
+    }
+
+    public static <K> Map<K, Object> sanitize(Map<K, ?> input, Set<Object> visited) {
+        Map<K, Object> result = Maps.newLinkedHashMap();
+        for (Map.Entry<K, ?> e : input.entrySet()) {
+            if (Entities.isSecret("" + e.getKey()))
```
aledsage: Maybe move `Entities.isSecret` to `Sanitizer` as well? And `Entities.SECRET_NAMES` (deprecated in old place of course).

core/src/main/java/brooklyn/entity/basic/Sanitizer.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public static Map<String, Object> sanitize(ConfigBag input) {
+        return sanitize(input.getAllConfig());
+    }
+
+    public static <K> Map<K, Object> sanitize(Map<K, ?> input) {
+        return sanitize(input, Sets.newHashSet());
+    }
+
+    public static <K> Map<K, Object> sanitize(Map<K, ?> input, Set<Object> visited) {
```
aledsage: Personally I'd have this as a package-private method - the second arg is an implementation detail. But no strong feelings.

core/src/main/java/brooklyn/entity/basic/Sanitizer.java
```diff
@@ -0,0 +1,59 @@
...
+        Map<K, Object> result = Maps.newLinkedHashMap();
+        for (Map.Entry<K, ?> e : input.entrySet()) {
+            if (Entities.isSecret("" + e.getKey()))
+                result.put(e.getKey(), "xxxxxxxx");
+            else if (e.getValue() instanceof Map) {
+                if (visited.contains(e.getValue())) {
+                    continue;
```
aledsage: I think we still need a `result.put(...)` for if already visited?

core/src/main/java/brooklyn/entity/basic/Sanitizer.java
```diff
@@ -0,0 +1,59 @@
...
+                }
+                visited.add(e.getValue());
+                result.put(e.getKey(), sanitize((Map<?, ?>) e.getValue(), visited));
+            } else {
+                result.put(e.getKey(), e.getValue());
```
aledsage: Could also have a `sanitize(List<?> input, Set<Object> visited)`, and same for `Set`? Feels like simple to add, and will possibly catch some more places we should sanitize (either now or in the future).

aledsage:
@robertgmoss looks good - only very minor comments.

core/src/main/java/brooklyn/entity/basic/Sanitizer.java
ahgittin: +1

core/src/main/java/brooklyn/entity/basic/Sanitizer.java
ahgittin: Actually I'd build out a non-static implementation whereby the secret detection can be controlled.

    public static Sanitizer(Predicate<Object> sanitizingNeededCheck) { new Sanitizer(sanitizingNeededCheck); }
    protected Sanitizer(Predicate<Object> sanitizingNeededCheck) { ... }
    public <K> Map<K,Object> apply(Map<K,?> input) { return apply(input, Sets.newHashSet()); }
    private <K> Map<K,Object> apply(Map<K,?> input, Set<Object> visitied) {
      // old visited code
    }

then moving `SECRET_NAMES` in becomes a default that `newInstance()` can use, but it can now be customised

ahgittin: (the static `sanitize(x)` methods can stay as conveniences calling `newInstance().apply(x)`)

robertgmoss:
@aledsage I've addressed your comments.

core/src/main/java/brooklyn/entity/basic/Sanitizer.java
```diff
@@ -0,0 +1,152 @@
...
+    public static final List<String> SECRET_NAMES = ImmutableList.of(
+            "password", 
+            "passwd", 
+            "credential", 
+            "secret", 
+            "private",
+            "access.cert", 
+            "access.key");
+    
+    public static final Predicate<Object> IS_SECRET_PREDICATE = new Predicate<Object>() {
```
ahgittin: note there are `StringPredicates` and `StringFunctions` including `toLowerCase` and `containsLiteral` which you can combine, but lists are especially tedious so i wouldn't try a pure functional build-up ... however if this logic is needed elsewhere might be worth moving the body there as a static method `Predicate<Object> containsAnyLiteralIgnoringCase(Iterable<String>)`

not needed for this PR, just wanted to point you at those classes

ahgittin:
LGTM

aledsage:
Merging.

PR #548, opened 2015-03-10
Docs: improve rebind and SoftwareProcess docs

aledsage:
We could also do with moving the persistence guidelines / best practices into somewhere more prominent (e.g. a top-level page)?

This PR hopefully gives the content required.

### DIFF
--- a/docs/guide/java/entities.md
+++ b/docs/guide/java/entities.md
@@ -32,8 +32,6 @@ What to Extend -- Implementation Classes
 A common lifecycle pattern is that the ``start`` effector (see more on effectors below) is invoked, 
 often delegating either to a driver (for software processes) or children entities (for clusters etc).
 
-See ``JBoss7Server`` and ``MySqlNode`` for exemplars.
-
 
 Configuration
 -------------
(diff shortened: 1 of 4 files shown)


docs/guide/java/entities.md
```diff
@@ -123,4 +121,90 @@ TODO more drivers such as jmx, etc are planned
 Testing
 -------
 
-* Run in a mock ``SimulatedLocation``, defining new metaclass methods to be able to start there and assert the correct behaviour when that is invoked
+* Unit tests can make use of `SimulatedLocation` and `TestEntity`, and can extend `BrooklynAppUnitTestSupport`.
+* Integration tests and use a `LocalhostMachineProvisioningLocation`, and can also extend `BrooklynAppUnitTestSupport`.
+
+
+<a name="SoftwareProcess-lifecycle"></a>
```
ahgittin: links are now generated automatically by the doc tools, so these `<a ...` links can be removed i think.  you can even have menu-updates-as-you-scroll behaviour; see e.g. learn more -> features.

ahgittin:
good info, merging.

i think persistence fits under operations, but should probably be linked from the java section, or maybe persistence should be split between java and ops, depending what is being covered. but hardly important exactly where, as long as it is there and not too hard to find.  would not move to a higher level -- everything is important at that higher level!

PR #550, opened 2015-03-13
Fix AbstractEntity.config()

aledsage:
- When return type of config() was changed to BasicConfigurationSupport,
  it broke any entity impls that tried to call config().set(...) etc
- Solution is to change BasicConfigurationSupport to be protected
  rather than private, and mark as @Beta because we’ll change it
  back to private at some point.


ahgittin:
lgtm - shame we have to jump through this groovy hoop

PR #552, opened 2015-03-14
Riak Linux - fix

bostko:
- Move the utility methods to the appropriate
  BashCommands class

### DIFF
--- a/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
+++ b/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
@@ -190,14 +190,14 @@ public class RiakNodeSshDriver extends AbstractSoftwareProcessSshDriver implemen
     private ImmutableList<String> installDebianBased() {
         return ImmutableList.<String>builder()
                 .add("curl https://packagecloud.io/install/repositories/basho/riak/script.deb | " + BashCommands.sudo("bash"))
-                .add(BashCommands.sudo("apt-get install --assume-yes riak"))
+                .add(BashCommands.sudo("apt-get install --assume-yes riak=" + getEntity().getFullVersion() + "-1"))
                 .build();
     }
 
     private ImmutableList<String> installRpmBased() {
         return ImmutableList.<String>builder()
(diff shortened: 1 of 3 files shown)


software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -190,45 +190,14 @@ public void install() {
     private ImmutableList<String> installDebianBased() {
         return ImmutableList.<String>builder()
                 .add("curl https://packagecloud.io/install/repositories/basho/riak/script.deb | " + BashCommands.sudo("bash"))
-                .add(BashCommands.sudo("apt-get install --assume-yes riak"))
+                .add(BashCommands.sudo("apt-get install --assume-yes riak=" + getEntity().getFullVersion() + "-1"))
```
ahgittin: why the `-1` ?

ahgittin: and should it really be `riak=` not `riak-` ?

ahgittin:
looks good.  i've some concerns whether the new `BashCommands` do the right thing in all situations (e.g. if you give a list of commands, but then it is run as a single CLI rather than a script) but that applies to others as well, it can be cleaned up as needed.  (in general i'd like to move away from them in any case, to pure yaml.)

minor comment about `=` and `-1` in the filename to be installed.  confirm that's correct and/or add comment and then happy to merge.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
bostko: This is how packagecloud works https://packagecloud.io/basho/riak/packages/ubuntu/precise/riak_2.0.5-1_amd64.deb
https://packagecloud.io/basho/riak/packages/el/5/riak-2.0.5-1.x86_64.rpm

bostko:
I tested it and it is ok.
-1 is part of the version and it is included when you specify the version in packagecloud. However it seems that it is not exactly part of the version since the package name could end with 2.0.5-OSX http://docs.basho.com/riak/latest/downloads/ .
 So to ease my work I wrote it that way.

PR #553, opened 2015-03-16
Destroys machine on expunge, even if stopped

nakomis:
Fixes an issue where stopping an entity (with
stopMachineMode of NEVER) and then attempting
to expunge the top level application would
unmanage the entity, but not destroy the VM


aledsage:
Looks good; merging.

aledsage:
Next time though, can you use a shorter first line for commit message. See discussion that we had previously on dev@brooklyn.

PR #554, opened 2015-03-17
fixes for BrooklynNode and java appservers to run in clocker, using subnet hostname

ahgittin:
confirmed working w jboss and brooklyn node in clocker.  i want to test with tomcat, and explore port-forward manager bug, then ready for merging.

### DIFF
--- a/core/src/main/java/brooklyn/entity/rebind/dto/MementosGenerators.java
+++ b/core/src/main/java/brooklyn/entity/rebind/dto/MementosGenerators.java
@@ -37,7 +37,9 @@ import brooklyn.entity.Feed;
 import brooklyn.entity.Group;
 import brooklyn.entity.basic.EntityDynamicType;
 import brooklyn.entity.basic.EntityInternal;
+import brooklyn.entity.rebind.AbstractBrooklynObjectRebindSupport;
 import brooklyn.entity.rebind.TreeUtils;
+import brooklyn.entity.rebind.persister.BrooklynPersistenceUtils;
 import brooklyn.event.AttributeSensor;
 import brooklyn.event.AttributeSensor.SensorPersistenceMode;
 import brooklyn.event.feed.AbstractFeed;
@@ -61,6 +63,7 @@ import brooklyn.util.collections.MutableMap;
 import brooklyn.util.config.ConfigBag;
 import brooklyn.util.flags.FlagUtils;
 
+import com.google.common.annotations.Beta;
 import com.google.common.base.Function;
 import com.google.common.base.Predicates;
 
@@ -68,10 +71,23 @@ public class MementosGenerators {
 
     private MementosGenerators() {}
     
-    /**
-     * Inspects a brooklyn object to create a corresponding memento.
-     */
+    /** @deprecated since 0.7.0 use {@link #newBasicMemento(BrooklynObject)} */
     public static Memento newMemento(BrooklynObject instance) {
+        return newBasicMemento(instance);
+    }
+    
+    /**
+     * Inspects a brooklyn object to create a basic corresponding memento.
+     * <p>
+     * The memento is "basic" in the sense that it does not tie in to any entity-specific customization;
+     * the corresponding memento may subsequently be customized by the caller.
+     * <p>
+     * This method is intended for use by {@link AbstractBrooklynObjectRebindSupport#getMemento()}
+     * and callers wanting a memento for an object should use that, or the
+     * {@link BrooklynPersistenceUtils#newObjectMemento(BrooklynObject)} convenience.
+     */
+    @Beta
+    public static Memento newBasicMemento(BrooklynObject instance) {
         if (instance instanceof Entity) {
             return newEntityMemento((Entity)instance);
         } else if (instance instanceof Location) {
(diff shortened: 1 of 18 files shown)


core/src/main/java/brooklyn/entity/rebind/dto/MementosGenerators.java
```diff
@@ -68,10 +71,23 @@
 
     private MementosGenerators() {}
     
+    /** @deprecated since 0.7.0 use {@link #newBasicMemento(BrooklynObject)} */
```
aledsage: Why rename from `newMemento(BrooklynObject)` to `newBasicMemento(BrooklynObject)`? Do we have anything but "basic", or are we expecting that in the near future? Personally I prefer `newMemento`, as "basic" implies there is some other kind of memento out there.

core/src/test/java/brooklyn/location/access/PortForwardManagerRebindTest.java
```diff
@@ -86,6 +96,41 @@ public void testAssociationPreservedOnRebind() throws Exception {
...
+        PortForwardManager origPortForwardManager = origEntity.getConfig(MyEntity.PORT_FORWARD_MANAGER);
+
+        origPortForwardManager.associate(publicIpId, HostAndPort.fromParts(publicAddress, 40080), origSimulatedMachine, 80);
+
+        String label = origManagementContext.getManagementNodeId()+"-"+Time.makeDateSimpleStampString();
+        PersistenceObjectStore targetStore = BrooklynPersistenceUtils.newPersistenceObjectStore(origManagementContext, null, 
+            "tmp/web-persistence-"+label+"-"+Identifiers.makeRandomId(4));
+        File dir = ((FileBasedObjectStore)targetStore).getBaseDir();
+        // only register the parent dir because that will prevent leaks for the random ID
+        Os.deleteOnExitEmptyParentsUpTo(dir.getParentFile(), dir.getParentFile());
```
aledsage: Personal preference for tearDown or try-finally, rather than "deleteOnExit". Stuff gets left behind on dev machines when tests are terminated prematurely. But no strong feelings.

core/src/test/java/brooklyn/location/access/PortForwardManagerRebindTest.java
```diff
@@ -86,6 +96,41 @@ public void testAssociationPreservedOnRebind() throws Exception {
...
+        BrooklynPersistenceUtils.writeMemento(origManagementContext, targetStore, MementoCopyMode.LOCAL);            
+
+        RebindTestUtils.waitForPersisted(origApp);
+        log.info("Using manual export dir "+dir+" for rebind instead of "+mementoDir);
```
aledsage: Why use the manual export dir? If we need to do that here, why do we not need to do it in all our other tests?

ahgittin: I'm testing rest-like manual export distinct from normal persist-and-rebind.

core/src/main/java/brooklyn/entity/rebind/dto/MementosGenerators.java
ahgittin: This isn't the real memento as per new java doc.  The old name confused me
and caused the manual export failures.

software/webapp/src/main/java/brooklyn/entity/webapp/JavaWebAppSshDriver.java
```diff
@@ -89,11 +89,11 @@ protected String inferRootUrl() {
         if (isProtocolEnabled("https")) {
             Integer port = getHttpsPort();
             checkNotNull(port, "HTTPS_PORT sensors not set; is an acceptable port available?");
-            return String.format("https://%s:%s/", getHostname(), port);
+            return String.format("https://%s:%s/", getSubnetHostname(), port);
```
aledsage: Am I right that `getSubnetHostname()` will return the same as `getHostname()` if not running in Clocker or a `SubnetTier`?

What about for AWS - will the subnetHostname still be the public hostname, rather than the private IP of the VM?

Assuming so, this looks sensible.

aledsage:
@ahgittin looks good - only a couple of minor comments/questions.

core/src/main/java/brooklyn/entity/rebind/dto/MementosGenerators.java
```diff
@@ -127,13 +143,16 @@ public static BrooklynMemento newBrooklynMemento(ManagementContext managementCon
     
     /**
      * Inspects an entity to create a corresponding memento.
+     * <p>
+     * @deprecated since 0.7.0, see {@link #newBasicMemento(BrooklynObject)}
      */
+    @Deprecated
```
neykov: Do we really need to keep all of these? The code is so specific that I am pretty sure no one know it exists. Also removing it won't mess with serialized state. I'd say delete it right now.

ahgittin: I've been burned by that assumption before.  I'm almost certain you're right but I think not worth it now.  We will have fun as soon as 070 is out getting rid of all these!!!

software/webapp/src/main/java/brooklyn/entity/webapp/JavaWebAppSshDriver.java
ahgittin: Testing EC2 now to be sure, but from the code I think so.

ahgittin: Discovered a problem in EC2 with the naming strategy introduced in 89c126340dcf350a3d887782a5ea79a672fe2f17 -- have fixed it with an additional commit here, and detailed comments.

Now confirmed that subnet hostname in AWS does the right thing.

Merging with the add'l commit but @aledsage please look at b68363e8636a3e0c3cea28a2d39aa448e8d50650 in case you see a better way.  (Bear in mind jclouds wants a `Function<...Image...>` and `Image` doesn't make the cloud provider available!)

PR #555, opened 2015-03-17
BROOKLYN-136: persist dynamic locations

aledsage:
See javadoc on `BasicLocationRegistry` for a description of how this all hangs together.


ahgittin:
failure is `io.brooklyn.camp.brooklyn.EmptySoftwareProcessYamlTest.testWithAppAndEntityLocations`

```
java.util.concurrent.ExecutionException: brooklyn.util.exceptions.PropagatedRuntimeException: Error invoking start at BasicApplicationImpl{id=gl0S5Wg7}: Error invoking start at BasicApplicationImpl{id=gl0S5Wg7}: Error invoking start at EmptySoftwareProcessImpl{id=AVEPF1IO}: Cannot establish ssh connection to jenkins @ SshMachineLocation[SshMachineLocation:tUEe:127.0.0.1/127.0.0.1:22@tUEehcA5] (Exhausted available authentication methods).
```

am guessing this should be an integration test although why it's appearing now i don't know

utils/common/src/main/java/brooklyn/util/yaml/Yamls.java
```diff
@@ -66,6 +66,47 @@
...
+     * In the given path:
+     * <ul>
+     *   <li>A vanilla string is assumed to be a key into a map.
+     *   <li>A string in the form like "[0]" is assumed to be an index into a list
+     * </ul>
+     * 
+     * Returns {@code null} if that path does not exist. 
+     */
+    @SuppressWarnings("unchecked")
+    public static Object getAtPreParsed(Object current, List<String> path) {
```
ahgittin: ugly method name.  if it has already been parsed is there any difference between this and `Jsonya.of(current).at(path).get()` ?

(at least mark `@Beta`)

ahgittin:
very useful, and looks sensible.  would be good to have a bigger review but as this is new functionality i'm not opposed to merging it and reviewing as we go along.  good tests.

the other thing which occurs to me is that maybe we could get rid of some of the redundancy between definitions, specs, etc, by switching to catalog items.  not sure about this but something to think about.

nakomis:
A couple of minor comments, but other than that, LGTM. EmptySoftwareProcessYamlTest.testWithAppAndEntityLocations passes on my machine

aledsage:
Thanks @ahgittin @nakomis 

For the failing `EmptySoftwareProcessYamlTest.testWithAppAndEntityLocations`, other tests were previously failing non-deterministically for me because the `start()` effector was kicked off asynchronously. For this test, I presume nothing was checking if the effector had actually executed successfully. Now that the code blocks for `start()` to complete, we get this failure. Have marked those as integration.

Have incorporated review comments, and rebased. Will wait for jenkins, and then merge.

PR #558, opened 2015-03-18
Fix UsageResourceTest.testListApplicationUsages

aledsage:
Context: this test is checking that the REST api for application-usage will report usage in the specified time range. If we ask for usage between `Wed Mar 18 11:45:25 GMT 2015` and `Wed Mar 18 11:45:25 GMT 2015` then we expect to be told about all apps that were in use during that single second (and not to be told about the usage of those apps outside of that specific second).

- Was failing occasionally if the requested start/end date rounded down to a different value than the postStart date.
- I suspect 1 in 1000 chance, because depended if afterPostStart was in a different second from postStart - it was adding 1ms to it. I suspect the original intent was to add 1 second.

### DIFF
--- a/usage/rest-server/src/test/java/brooklyn/rest/resources/UsageResourceTest.java
+++ b/usage/rest-server/src/test/java/brooklyn/rest/resources/UsageResourceTest.java
@@ -91,7 +91,12 @@ public class UsageResourceTest extends BrooklynRestResourceTest {
         Date preStart = new Date();
         String appId = createApp(simpleSpec);
         Date postStart = new Date();
-
+        
+        // We will retrieve usage from one millisecond after start; this guarantees to not be  
+        // told about both STARTING+RUNNING, which could otherwise happen if they are in the 
+        // same milliscond.
+        Date afterPostStart = new Date(postStart.getTime()+1);
+        
         // Check that app's usage is returned
         ClientResponse response = client().resource("/v1/usage/applications").get(ClientResponse.class);
         assertEquals(response.getStatus(), Response.Status.OK.getStatusCode());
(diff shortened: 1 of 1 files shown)


neykov:
I was the last one who took a stab at fixing this test so can provide some context.
The 1ms delay is on purpose due to the check in `UsageResource`
```
            if (eventStartDate.compareTo(endDate) > 0 || eventEndDate.compareTo(startDate) < 0) {
                continue;
            }
```
We have to make sure that we are requesting times at least 1ms *after* `eventStartDate` (thus the +1ms). The `waitForFuture(postStart+1ms)` is because `eventEndDate=new Date()`, to make sure that the requested time is not in the future. The comparison is done on the full values, they are cropped at serialization time only, so no need to add 1s.

Your other half of the fix is correct though, the test should assert against `afterPostStart` due to
```
            if (eventStartDate.compareTo(startDate) < 0) {
                eventStartDate = startDate;
            }
            if (eventEndDate.compareTo(endDate) > 0) {
                eventEndDate = endDate;
            }
```

I can see another potential problem. It could (quite unlikely) happen that `afterPostStart` is equal to `eventEndDate = new Date()` Due to the `eventEndDate.compareTo(startDate) < 0` the test will fail. To make sure this doesn't happen I'd suggest changing the original test to
```
long afterPostStart = postStart.getTime()+1;
waitForFuture(afterPostStart+1);
```
so that we are completely sure that `afterPostStart` is at least 1ms after the entity start time and at least 1ms before `new Date()`.

aledsage:
Thanks @neykov - I've incorporated your suggestions.

Not sure I entirely follow. We're using `.compareTo(...) > 0` and `.compareTo(...) < 0`, so they should cope fine with it being equal. It should always return us event details about `startDate` to `endDate` (i.e. `afterPostStart` to `afterPostStart`, because that is what we're now passing in as the time range).

I've fine with the test taking an extra 1ms to be cautious though.

usage/rest-server/src/test/java/brooklyn/rest/resources/UsageResourceTest.java
```diff
@@ -426,6 +432,10 @@ public void release(SshMachineLocation machine) {
         }
     }
 
+    private void waitForFuture(Date futureTime) throws InterruptedException {
```
neykov: Looks like this is not used any more?

neykov:
You are right, the comparison is inclusive. LGTM with or without the extra ms.

PR #559, opened 2015-03-19
group rescan should be triggered by rebind, but not be in critical rebind path

ahgittin:
(since it might unmanage things, esp in the DynamicMultiGroup subclass,
which causes deadlock, as it can't unmanage until management start transition completes)

### DIFF
--- a/core/src/main/java/brooklyn/entity/basic/DynamicGroupImpl.java
+++ b/core/src/main/java/brooklyn/entity/basic/DynamicGroupImpl.java
@@ -32,10 +32,12 @@ import brooklyn.entity.Entity;
 import brooklyn.event.Sensor;
 import brooklyn.event.SensorEvent;
 import brooklyn.event.SensorEventListener;
+import brooklyn.management.Task;
 import brooklyn.management.internal.CollectionChangeListener;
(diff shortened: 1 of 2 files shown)


core/src/main/java/brooklyn/entity/basic/DynamicGroupImpl.java
```diff
@@ -163,13 +165,20 @@ public void onManagementBecomingMaster() {
...
-            Exceptions.propagateIfFatal(e);
-        }
+        Task<Object> rescan = Tasks.builder().name("rescan entities").body(
+            new Runnable() {
+                @Override
+                public void run() {
+                    try {
+                        rescanEntities();
+                    } catch (Exception e) {
+                        log.warn("Error rescanning entities when rebinding; may be a group set against an unknown entity: "+e);
```
aledsage: This is not just called on rebind - it's called when the entity is very created+managed as well.
Perhaps a better message would just be `"Error rescanning for members of "+DynamicGroupImpl.this+"; may be a ..."`

aledsage:
Looks good; only one very minor comment.

PR #560, opened 2015-03-19
Fix NPE in removeMember

grkvlt:
Check that the `DELEGATE_ENTITY` attribute exists and is set before comparing it with the entity to be removed.


ahgittin:
LGTM

PR #562, opened 2015-03-20
Riak Linux - add flag for turning off sysctl optimizations

bostko:


grkvlt:
@bostko I have incorporated this idea into #563 including a suggestion from @ahgittin to mark the generated script as `inessential()` so I will close this. can you have a look at the changes I made in the other PR, please?

PR #564, opened 2015-03-23
Update to RiakNode to fix issue with Clocker and cluster resize

grkvlt:


andreaturli:
lgtm, it solves problem in Clocker

it still needs some debugging for the cluster management, though, but not in this PR

grkvlt:
@andreaturli Thanks, will merge and then we can look at the cluster management issues seperately

grkvlt:
Somehow I managed to get the commits for this PR included twice. I don't think it did any harm, though. I expect it was caused by me creating the PR from `master` rather than a branch, which was a mistake.

PR #566, opened 2015-03-23
Relax synchronisation in EntityManager

grkvlt:
Relax synchronization in `LocalEntitymanager` to prevent a deadlock on rebind.


aledsage:
Looks good to me; merging.

PR #568, opened 2015-03-24
Wait for join to finish before committing cluster

grkvlt:
Blocks on the `joinCluster` effector task before starting the `commitCluster` effector.


grkvlt:
Tested with Clocker, resize now commits changes to the Riak cluster :frog:

PR #569, opened 2015-03-24
Fix the CAMP planId so that it is not inherited

grkvlt:
Because the CAMP planId uses the basic `ConfigKey` settings it is inherited, so attempts to resolve entities by ID will return all children. Creates the plan ID as non-inherited configuration so it will only be found on the actual entity.

### DIFF
--- a/utils/common/src/main/java/brooklyn/config/ConfigInheritance.java
+++ b/utils/common/src/main/java/brooklyn/config/ConfigInheritance.java
@@ -22,6 +22,7 @@ import java.io.Serializable;
 
 import com.google.common.annotations.Beta;
 
+@SuppressWarnings("serial")
 public abstract class ConfigInheritance implements Serializable {
 
     public static final ConfigInheritance ALWAYS = new Always();
--- a/utils/common/src/main/java/brooklyn/config/ConfigInheritance.java
+++ b/utils/common/src/main/java/brooklyn/config/ConfigInheritance.java
@@ -18,9 +18,11 @@
  */
 package brooklyn.config;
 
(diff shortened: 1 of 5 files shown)


neykov:
You can disable `ConfigKey` inheritance, see https://github.com/apache/incubator-brooklyn/blob/master/core/src/main/java/brooklyn/entity/basic/ServiceStateLogic.java#L378 for example.

grkvlt:
@neykov that's interesting... I'll maybe change the code, then

grkvlt:
OK, the last failure is just `java.lang.RuntimeException: unable to find a free port at or above 8080` so I'm happy to merge now :frog:

utils/common/src/main/java/brooklyn/config/ConfigInheritance.java
```diff
@@ -18,9 +18,11 @@
  */
 package brooklyn.config;
 
+import java.io.Serializable;
+
 import com.google.common.annotations.Beta;
 
-public abstract class ConfigInheritance {
+public abstract class ConfigInheritance implements Serializable {
```
neykov: Add `serialVersionUID` to avoid warnings?

neykov:
+1, I don't see potential problems with rebinding, good to merge.

utils/common/src/main/java/brooklyn/config/ConfigInheritance.java
grkvlt: I'll just ignore the warnings, since it's not needed for actual serialisation here.

PR #571, opened 2015-03-25
new policies: SshConnectionFailure, ConditionalSuspendPolicy

neykov:
* SshConnectionFailure emits CONNECTION_FAILURE if it can't make ssh connection to the machine of the entity
  * ConditionalSuspendPolicy suspends a target policy if it receives a sensor event (CONNECTION_FAILURE by default)

### DIFF
--- a/policy/src/main/java/brooklyn/policy/ha/AbstractFailureDetector.java
+++ b/policy/src/main/java/brooklyn/policy/ha/AbstractFailureDetector.java
@@ -1,0 +1,359 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package brooklyn.policy.ha;
+
+import static brooklyn.util.time.Time.makeTimeStringRounded;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import brooklyn.config.ConfigKey;
+import brooklyn.entity.basic.BrooklynTaskTags;
+import brooklyn.entity.basic.ConfigKeys;
+import brooklyn.entity.basic.EntityInternal;
+import brooklyn.entity.basic.EntityLocal;
+import brooklyn.event.Sensor;
+import brooklyn.management.Task;
+import brooklyn.policy.basic.AbstractPolicy;
+import brooklyn.policy.ha.HASensors.FailureDescriptor;
+import brooklyn.util.collections.MutableMap;
+import brooklyn.util.exceptions.Exceptions;
+import brooklyn.util.flags.SetFromFlag;
+import brooklyn.util.task.BasicTask;
+import brooklyn.util.task.ScheduledTask;
+import brooklyn.util.time.Duration;
+import brooklyn.util.time.Time;
+
+import com.google.common.reflect.TypeToken;
+
+public abstract class AbstractFailureDetector extends AbstractPolicy {
+
+    // TODO Remove duplication from ServiceFailureDetector, particularly for the stabilisation delays.
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractFailureDetector.class);
+
+    private static final long MIN_PERIOD_BETWEEN_EXECS_MILLIS = 100;
+
+    public static final ConfigKey<Duration> POLL_PERIOD = ConfigKeys.newDurationConfigKey(
+            "failureDetector.pollPeriod", "", Duration.ONE_SECOND);
+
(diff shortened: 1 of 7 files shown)


grkvlt:
There look really useful @neykov

neykov:
Added option to `ServiceFailureDetector` to republish failed state periodically.

policy/src/main/java/brooklyn/policy/ha/AbstractFailureDetector.java
```diff
@@ -0,0 +1,359 @@
...
+import brooklyn.util.time.Duration;
+import brooklyn.util.time.Time;
+
+import com.google.common.reflect.TypeToken;
+
+public abstract class AbstractFailureDetector extends AbstractPolicy {
+
+    // TODO Remove duplication from ServiceFailureDetector, particularly for the stabilisation delays.
+
+    private static final Logger LOG = LoggerFactory.getLogger(ConnectionFailureDetector.class);
```
aledsage: Wrong class.

policy/src/main/java/brooklyn/policy/ha/SshMachineFailureDetector.java
```diff
@@ -0,0 +1,84 @@
...
+            SshMachineLocation sshMachine = sshMachineOption.get();
+            try {
+                Duration timeout = config().get(CONNECT_TIMEOUT);
+                Map<String, ?> flags = ImmutableMap.of(
+                        SshTool.PROP_CONNECT_TIMEOUT.getName(), timeout.toMilliseconds(),
+                        SshTool.PROP_SESSION_TIMEOUT.getName(), timeout.toMilliseconds(),
+                        SshTool.PROP_SSH_TRIES.getName(), 1);
+                int exitCode = sshMachine.execCommands(flags, SshMachineFailureDetector.class.getName(), ImmutableList.of("exit"));
+                return new BasicCalculatedStatus(exitCode == 0, sshMachine.toString());
+            } catch (Exception e) {
```
aledsage: Risks swallowing `InterruptedException`. Worth doing `Exceptions.propagateIfFatal(e)`.

Do we want to log it (at debug) when it's the first time we've seen the ssh exception? No strong feelings. Certainly don't want to log it every time (unless at trace).

policy/src/test/java/brooklyn/policy/ha/ServiceFailureDetectorTest.java
```diff
@@ -312,6 +316,54 @@ public void testReportsFailureWhenAlreadyOnFireOnRegisteringPolicy() throws Exce
...
+        
+        // Now recover
+        ServiceStateLogic.ServiceProblemsLogic.clearProblemsIndicator(e1, "test");
+        EntityTestUtils.assertAttributeEqualsEventually(e1, TestEntity.SERVICE_STATE_ACTUAL, Lifecycle.RUNNING);
+        assertHasEventEventually(HASensors.ENTITY_RECOVERED, Predicates.<Object>equalTo(e1), null);
+        
+        TimeUnit.MILLISECONDS.sleep(500);
+
+        //can't assert number of republish events due to jitter, warn below if it deviates from expectation
+        assertTrue(events.size() > 2, "events="+events);
```
aledsage: I'm hesitant about time-sensitive tests like this. We're probably ok because it's integration...

For those that used to run in cloudbees, we'd occassionally see failures that suggested the thread had been starved for several seconds. I resorted to 30 second timeouts to be on the safe side, so we wouldn't get false negatives in functional tests.

policy/src/test/java/brooklyn/policy/ha/ServiceFailureDetectorTest.java
```diff
@@ -312,6 +316,54 @@ public void testReportsFailureWhenAlreadyOnFireOnRegisteringPolicy() throws Exce
...
+
+        SensorEvent<FailureDescriptor> prevEvent = null;
+        for (SensorEvent<FailureDescriptor> event : events) {
+            if (prevEvent != null) {
+                long repeatOffset = event.getTimestamp() - prevEvent.getTimestamp();
+                long deviation = Math.abs(repeatOffset - republishPeriod.toMilliseconds());
+                if (deviation > republishPeriod.toMilliseconds()/10 &&
```
aledsage: Same here: very time sensitive. We might well get a 10ms starvation of the required thread when running on a cloud test server.

For this kind of time test (to avoid false negatives), I resorted to number that were more like republishPeriod being at least 1 second (rather than 100ms), and accepting anything up to 500ms difference.

If we start running our integration tests on apache hardware, we might find that will sometimes fail as well!

policy/src/test/java/brooklyn/policy/ha/ServiceFailureDetectorTest.java
```diff
@@ -312,6 +316,54 @@ public void testReportsFailureWhenAlreadyOnFireOnRegisteringPolicy() throws Exce
...
+                        //warn only if recovered is too far away from the last failure
+                        (!event.getSensor().equals(HASensors.ENTITY_RECOVERED) ||
+                        repeatOffset > republishPeriod.toMilliseconds())) {
+                    log.error("The time between failure republish (" + repeatOffset + "ms) deviates too much from the expected " + republishPeriod + ". prevEvent=" + prevEvent + ", event=" + event);
```
aledsage: I see it's just a log rather than a failure. Is that because of the risk of false negatives? Would be good if we can have a test that will genuinely fail if the republish isn't working. For example, we could assert that we eventually get 2 entity_failed events before we set the entity has recovered.

aledsage:
Looks good (and agree looks very useful). Only some minor comments from me.

policy/src/test/java/brooklyn/policy/ha/ServiceFailureDetectorTest.java
neykov: Republish is already checked for above ` assertTrue(events.size() > 2, "events="+events);` - 1 event for initial FAIL, 1 event for RECOVERED, any additional events for republish. The updated code now waits for 10 events before continuing.

The check here is to make sure we are really emitting republish FAILs at the requested frequency, but this is something that can't be asserted as it will vary even on idle systems.

neykov:
Addressed comments. Republish events are checked but the frequency between them can't be asserted, that's why I left it at `log.error`.

aledsage:
Looks good; merging.

PR #572, opened 2015-03-25
UsageListener: improve API

aledsage:
- Deprecates old interface
- Adds new UsageListener interface that is more flexible and more backwards compatible.
- Marks new interface as @Beta
- Adds tests for new UsageListener, and refactors old tests for old-style UsageListener.
- Adds Entities.proxy(Entity)


nakomis:
Looks good, tests pass. Have also done some testing / changes against this branch and it seems to work :-)

aledsage:
Thanks @nakomis - merging.

PR #575, opened 2015-03-26
Fix issue with Riak nodes not leaving cluster

grkvlt:
This includes #574 and also some other minor Riak fixes


grkvlt:
Still needs more work, but the commits here are useful, so merging as-is.

PR #576, opened 2015-03-28
Smoother js gui yaml flow

ahgittin:
@ggrunin had the great idea that in the JS gui add-application wizard, if a *yaml* catalog item is selected and it includes a `location` or `locations` block, proceed to show the yaml in an editable way, rather than the (unhelpful) location-selection-and-configuration page.  we expect this will make it easier for people to get started with yaml and for carefully chosen catalog examples, encourage editting these yaml files.

This PR causes the wizard to be smart about what is displayed after an item is chosen.  if it's a Java template or a YAML template with no `location` or `locations` block, it acts as before.  however if it's a YAML template with a `location` it shows the edit YAML tab. (also note that comments are preserved in this tab so the catalog item yaml can include comments to help people edit it.)

note this introduces a yaml parser which seems pretty slick -- https://github.com/nodeca/js-yaml -- so we could now very easily do yaml validation and other fun things client-side.

@ggrunin can you sign a CLA as this includes your commit making the primary change behind this PR?


ahgittin:
one final change, clicking "preview" now shows the *editable* yaml view, removing the old separate preview step view

usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -398,9 +426,10 @@ define([
         addTemplateLozenge: function(that, item) {
             var $tempel = _.template(CreateStepTemplateEntryHtml, {
                 id: item.id,
-                name: item.name,
+                name: item.name || item.id,
                 description: item.description,
-                iconUrl: item.iconUrl
+                planYaml:  item.planYaml,
+                iconUrl: item.iconUr
```
sjcorbett: `iconUr`? Typo or intentional?

usage/jsgui/src/main/webapp/assets/js/libs/js-yaml.js
```diff
@@ -0,0 +1,3666 @@
+/* js-yaml 3.2.7 https://github.com/nodeca/js-yaml */!function(e){if("object"==typeof exports&&"undefined"!=typeof module)module.exports=e();else if("function"==typeof define&&define.amd)define([],e);else{var f;"undefined"!=typeof window?f=window:"undefined"!=typeof global?f=global:"undefined"!=typeof self&&(f=self),f.jsyaml=e()}}(function(){var define,module,exports;return (function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
```
sjcorbett: Need another entry in `LICENSE`?

usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
sjcorbett: Definitely a typo.

sjcorbett:
In the logs I see this error message: "Uncaught ReferenceError: veryFirstViewLoad is not defined". I don't think it's been triggered by the changes here but worth noting.

sjcorbett:
Jumping to the YAML is great.

sjcorbett:
I would like the "Back" button to be maintained when clicking Preview. Its disappearance is surprising.

sjcorbett:
Doesn't work when the "None" location is selected. Error: Uncaught JS-YAML: unacceptable kind of an object to dump [object Undefined]. Last step in unobfuscated jsgui code ~ `renderCurrentStep`.

sjcorbett:
The alternating tab presence/absence is also a bit jarring. Open add app modal: Catalog/YAML/Other tab bar visible. Choose catalog entry and click next. Tab bar disappears. Click preview. Tab bar reappears with YAML selected. Would prefer it if the tabs were just visible the whole time. Comparatively minor!

sjcorbett:
Looks really good. A couple of comments to address then happy to see this merged.

ahgittin:
@sjcorbett thanks - great review.

the `veryFirstViewLoad` message i've fixed in a separate PR.  it was introduced elsewhere.

agree it would be nicer to have one flow.  it's a few hours' work though -- could this be merged without it?  i have put TODO's about it.

the others i will fix.

usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
ahgittin: yes, done

usage/jsgui/src/main/webapp/assets/js/libs/js-yaml.js
ahgittin: yes, done

ahgittin:
fixed everything but the tab/buttons where i agree with you but don't have the couple hours to tidy. thanks for great review @sjcorbett .  if you have thoughts on why `set("location",[])` isn't yaml parsing i'd be interested (See comment in last commit)

ahgittin:
@sjcorbett can this be merged?  it's especially nice with https://github.com/apache/incubator-brooklyn/pull/585 !

sjcorbett:
@ahgittin I'll double check and merge.

PR #578, opened 2015-03-30
Adds name of currently logged in user to LocationEvent and ApplicationEv...

nakomis:
...ent

### DIFF
--- a/core/src/main/java/brooklyn/management/usage/ApplicationUsage.java
+++ b/core/src/main/java/brooklyn/management/usage/ApplicationUsage.java
@@ -38,14 +38,20 @@ public class ApplicationUsage {
     public static class ApplicationEvent {
         private final Date date;
         private final Lifecycle state;
+        private final String user;
 
-        public ApplicationEvent(Lifecycle state) {
-            this(new Date(), state);
+        public ApplicationEvent(Lifecycle state, String user) {
+            this(new Date(), state, user);
         }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/brooklyn/management/usage/LocationUsage.java
```diff
@@ -73,25 +75,29 @@ public String getEntityType() {
...
         public int hashCode() {
-            return Objects.hashCode(date, state, entityId, entityType, applicationId);
+            return Objects.hashCode(date, state, entityId, entityType, applicationId, user);
         }
         
         @Override
         public String toString() {
             return Objects.toStringHelper(this).add("date", date).add("state", state).add("entityId", entityId)
-                    .add("appId", applicationId).toString();
+                    .add("appId", applicationId).add("entitlementContext", user).toString();
```
aledsage: maybe `add("user", user)` instead?

core/src/main/java/brooklyn/management/usage/ApplicationUsage.java
```diff
@@ -38,14 +38,24 @@
     public static class ApplicationEvent {
         private final Date date;
         private final Lifecycle state;
+        private final String user;
 
         public ApplicationEvent(Lifecycle state) {
```
aledsage: Do we still need this constructor, or should it be deleted?

aledsage:
Looks good; only a couple of minor comments.

PR #579, opened 2015-03-30
web app deploy - use temporary file for upload

neykov:
**WARN:** The fix changes the semantics of the deploy method. Previously it always succeeded regardless of the result of the copy operation. Now it will fail if either the copy or the rename operations fail.

Uploading the file directly to the webapps folder results in autodeploy triggering before the upload is complete. A problem arises when autodeploy triggers in the same second (before) the upload completes - it will fail due to incomplete file, but will not retry because the timestamp of the file won't change. Change deploy to upload to a temporary file and rename it to a .war only when done. This has a few benefits:
  * The application won't be undeployed at the start of the upload and be down for the duration of the upload
  * Once the server observers the updated file we can be sure that it will read its full content
  * If the upload fails (timeout, connection reset, full disk...) the old app won't be undeployed.

### DIFF
--- a/software/webapp/src/main/java/brooklyn/entity/webapp/JavaWebAppSshDriver.java
+++ b/software/webapp/src/main/java/brooklyn/entity/webapp/JavaWebAppSshDriver.java
@@ -27,6 +27,9 @@ import java.util.Set;
 import brooklyn.entity.basic.Attributes;
 import brooklyn.entity.java.JavaSoftwareProcessSshDriver;
 import brooklyn.location.basic.SshMachineLocation;
+import brooklyn.util.task.DynamicTasks;
+import brooklyn.util.task.Tasks;
+import brooklyn.util.task.ssh.SshTasks;
 
 import com.google.common.collect.ImmutableList;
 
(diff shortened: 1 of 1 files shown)


aledsage:
Good catch! Changes look good; merging.

software/webapp/src/main/java/brooklyn/entity/webapp/JavaWebAppSshDriver.java
```diff
@@ -146,15 +149,39 @@ public void deploy(File f, String targetName) {
...
-        // create a backup
-        getMachine().execCommands("backing up old war", ImmutableList.of(String.format("mv -f %s %s.bak > /dev/null 2>&1", dest, dest))); //back up old file/directory
-        int result = copyResource(url, dest);
-        log.debug("{} deployed {} to {}:{}: result {}", new Object[]{entity, url, getHostname(), dest, result});
-        if (result!=0) log.warn("Problem deploying {} to {}:{} for {}: result {}", new Object[]{url, getHostname(), dest, entity, result}); 
+    public String deploy(final String url, final String targetName) {
+        final String canonicalTargetName = getFilenameContextMapper().convertDeploymentTargetNameToFilename(targetName);
+        final String dest = getDeployDir() + "/" + canonicalTargetName;
+        //write to a .tmp so autodeploy is not triggered during upload
+        final String tmpDest = dest + ".tmp";
```
ahgittin: make this `dest+"."+Strings.makeRandomId(8)+".tmp"` ?

that should prevent collisions if the same thing is deployed simultaneously

PR #580, opened 2015-03-30
bump jclouds version to 1.9.0

andreaturli:


nakomis:
Required addition of loginUser brooklyn.properties setting for GCE, but other than that, looks good. Probably a lot more testing required though!

grkvlt:
Tested extensively by @grkvlt @andreaturli and others with Clocker on SoftLayer and AWS, including excercising the jclouds-labs Docker driver. The GCE issue can be worked around by setting properties appropriately. I'm happy to merge this now :frog:

grkvlt:
Missed one occurrence in downstream project POM, fixed in d08289911dcdd739820e3f4fcf530ba12a544d30

andreaturli:
thanks @grkvlt for fixing that!
is there any better way to set the property in the downstream project so that we don't missed it again next time?

PR #581, opened 2015-04-01
Riak - Create Bucket type effector

bostko:
This is the initial pull request for supporting riak bucket types in brooklyn.
http://docs.basho.com/riak/latest/dev/advanced/bucket-types/

Bucket types are more used for dynamic configurations, for example changing easily properties of a running cluster.
Riak supports adding, updating and deleting bucket types.

I added an effector which creates a bucket type.
I still can not figure out how can we add some monitoring /sensors/ to those bucket types?

### DIFF
--- a/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
+++ b/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
@@ -143,7 +143,6 @@ public interface RiakNode extends SoftwareProcess {
     MethodEffector<Void> JOIN_RIAK_CLUSTER = new MethodEffector<Void>(RiakNode.class, "joinCluster");
     MethodEffector<Void> LEAVE_RIAK_CLUSTER = new MethodEffector<Void>(RiakNode.class, "leaveCluster");
     MethodEffector<Void> REMOVE_FROM_CLUSTER = new MethodEffector<Void>(RiakNode.class, "removeNode");
-    MethodEffector<Void> COMMIT_RIAK_CLUSTER = new MethodEffector<Void>(RiakNode.class, "commitCluster");
 
     AttributeSensor<Integer> RIAK_NODE_GET_FSM_TIME_MEAN = Sensors.newIntegerSensor("riak.node_get_fsm_time_mean", "Time between reception of client read request and subsequent response to client");
     AttributeSensor<Integer> RIAK_NODE_PUT_FSM_TIME_MEAN = Sensors.newIntegerSensor("riak.node_put_fsm_time_mean", "Time between reception of client write request and subsequent response to client");
@@ -182,6 +181,9 @@ public interface RiakNode extends SoftwareProcess {
 
     String getOsMajorVersion();
 
+    // TODO add commitCluster() effector and add effectors joinCluster, leaveCluster, removeNode, recoverFailedNode which do not execute commitCluster()
+    // the commit where the commitCluster effector was available is adbf2dc1cb5df98b1e52d3ab35fa6bb4983b722f
+
     @Effector(description = "Join the Riak cluster on the given node")
     void joinCluster(@EffectorParam(name = "nodeName") String nodeName);
 
@@ -193,9 +195,6 @@ public interface RiakNode extends SoftwareProcess {
 
     @Effector(description = "Recover and join the Riak cluster on the given node")
     void recoverFailedNode(@EffectorParam(name = "nodeName") String nodeName);
-
-    @Effector(description = "Commit changes made to the Riak cluster")
-    void commitCluster();
 
     @Effector(description = "Create or modify a bucket type before activation")
     void bucketTypeCreate(@EffectorParam(name = "bucketTypeName") String bucketTypeName,
--- a/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
+++ b/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
@@ -197,7 +197,20 @@ public interface RiakNode extends SoftwareProcess {
     @Effector(description = "Commit changes made to the Riak cluster")
     void commitCluster();
 
-    @Effector(description = "Apply bucket type")
-    void applyBucketType(@EffectorParam(name = "bucketTypeName") String bucketTypeName,
-                         @EffectorParam(name = "bucketTypeProperties") String bucketTypeProperties);
+    @Effector(description = "Create bucket type")
+    void bucketTypeCreate(@EffectorParam(name = "bucketTypeName") String bucketTypeName,
+                          @EffectorParam(name = "bucketTypeProperties") String bucketTypeProperties);
+
+    @Effector(description = "List bucket types")
+    List<String> bucketTypeList();
+
+    @Effector(description = "Status bucket type")
+    List<String> bucketTypeStatus(@EffectorParam(name = "bucketTypeName") String bucketTypeName);
+
+    @Effector(description = "Update bucket type")
+    void bucketTypeUpdate(@EffectorParam(name = "bucketTypeName") String bucketTypeName,
+                          @EffectorParam(name = "bucketTypeProperties") String bucketTypeProperties);
+
+    @Effector(description = "Activate bucket type")
+    void bucketTypeActivate(@EffectorParam(name = "bucketTypeName") String bucketTypeName);
 }
--- a/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
+++ b/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
@@ -197,20 +197,20 @@ public interface RiakNode extends SoftwareProcess {
     @Effector(description = "Commit changes made to the Riak cluster")
     void commitCluster();
 
-    @Effector(description = "Create bucket type")
+    @Effector(description = "Create or modify a bucket type before activation")
     void bucketTypeCreate(@EffectorParam(name = "bucketTypeName") String bucketTypeName,
                           @EffectorParam(name = "bucketTypeProperties") String bucketTypeProperties);
 
-    @Effector(description = "List bucket types")
+    @Effector(description = "List all currently available bucket types and their activation status")
     List<String> bucketTypeList();
 
-    @Effector(description = "Status bucket type")
+    @Effector(description = "Display the status and properties of a specific bucket type")
     List<String> bucketTypeStatus(@EffectorParam(name = "bucketTypeName") String bucketTypeName);
 
-    @Effector(description = "Update bucket type")
+    @Effector(description = "Update a bucket type after activation")
     void bucketTypeUpdate(@EffectorParam(name = "bucketTypeName") String bucketTypeName,
                           @EffectorParam(name = "bucketTypeProperties") String bucketTypeProperties);
 
-    @Effector(description = "Activate bucket type")
+    @Effector(description = "Activate a bucket type")
     void bucketTypeActivate(@EffectorParam(name = "bucketTypeName") String bucketTypeName);
 }
--- a/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
+++ b/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
@@ -183,18 +183,21 @@ public interface RiakNode extends SoftwareProcess {
     String getOsMajorVersion();
 
     @Effector(description = "Join the Riak cluster on the given node")
-    public void joinCluster(@EffectorParam(name = "nodeName") String nodeName);
+    void joinCluster(@EffectorParam(name = "nodeName") String nodeName);
 
     @Effector(description = "Leave the Riak cluster")
-    public void leaveCluster();
+    void leaveCluster();
 
     @Effector(description = "Remove the given node from the Riak cluster")
-    public void removeNode(@EffectorParam(name = "nodeName") String nodeName);
+    void removeNode(@EffectorParam(name = "nodeName") String nodeName);
 
     @Effector(description = "Recover and join the Riak cluster on the given node")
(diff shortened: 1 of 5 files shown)


software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
```diff
@@ -183,18 +183,21 @@
...
     @Effector(description = "Recover and join the Riak cluster on the given node")
-    public void recoverFailedNode(@EffectorParam(name = "nodeName") String nodeName);
+    void recoverFailedNode(@EffectorParam(name = "nodeName") String nodeName);
 
     @Effector(description = "Commit changes made to the Riak cluster")
-    public void commitCluster();
+    void commitCluster();
 
+    @Effector(description = "Apply bucket type")
+    void applyBucketType(@EffectorParam(name = "bucketTypeName") String bucketTypeName,
```
aledsage: Is it feasible to have an integration and/or live test that uses this (including subsequently using the bucket-type that is created + activated)?

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -474,6 +475,18 @@ public void commitCluster() {
     }
 
     @Override
+    public void applyBucketType(String bucketTypeName, String bucketTypeProperties) {
```
aledsage: perhaps "createBucketType" would be a better method name, given that we all Riak's `bucket-type create ...`? (even though it means create or modify - still arguably better to be consistent with Riak's commands and documentation). Same goes for effector method on RiakNode as well (but the effector description can say "Create or modify a bucket type before activation" [1]).

Do we also need to activate the bucket type: "Like all bucket types, this type needs to be activated to be usable within the cluster." [1]. That is, call `riak-admin bucket-type activate ...`?

[1] http://docs.basho.com/riak/latest/dev/advanced/bucket-types/

aledsage:
@bostko not sure what has happened to this PR - github thinks it includes the two commits that have been merged to master in other pull requests! Out of interest, how did you `git pull` or `git rebase` your branch?

bostko:
Thank you for your comment.
It is still not ready for merging I just wanted some feedback

bostko:
I included also a fix for the cluster join command
http://docs.basho.com/riak/latest/ops/building/basic-cluster-setup/#Add-a-Second-Node-to-Your-Cluster

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -395,12 +390,12 @@ public void joinCluster(String nodeName) {
             if (!hasJoinedCluster()) {
                 ScriptHelper joinClusterScript = newScript("joinCluster")
                         .body.append(sudo(format("%s cluster join %s", getRiakAdminCmd(), nodeName)))
+                        .body.append(sudo(format("%s cluster plan", getRiakAdminCmd())))
+                        .body.append(sudo(format("%s cluster commit", getRiakAdminCmd())))
```
neykov: Looking at the driver API looks like .commitCluster() is meant to be called as a separate command. Why call commit here and not for the other methods?

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -474,6 +463,79 @@ public void commitCluster() {
...
+    }
+
+    @Override
+    public void bucketTypeUpdate(String bucketTypeName, String bucketTypeProperties) {
+        ScriptHelper bucketTypeStatusScript = newScript("bucket-type_update")
+                .body.append(sudo(format("%s bucket-type update %s %s",
+                        getRiakAdminCmd(),
+                        bucketTypeName,
+                        escapeLiteralForDoubleQuotedBash(bucketTypeProperties))))
+                .noExtraOutput()
```
neykov: no need for `noExtraOutput` if you are not parsing stdout

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -542,4 +602,10 @@ public String getOsMajorVersion() {
         String osVersion = osDetails.getVersion();
         return osVersion.contains(".") ? osVersion.substring(0, osVersion.indexOf(".")) : osVersion;
     }
+
+    private void addRiakOnPath(ScriptHelper scriptHelper) {
+        Map<String, String> newPathVariable = ImmutableMap.of("PATH", sbinPath);
+//        log.warn("riak command not found on PATH. Altering future commands' environment variables from {} to {}", getShellEnvironment(), newPathVariable);
```
neykov: Why comment it out? If really not needed better remove it altogether.

neykov:
Brooklyn side changes looking good, can't comment on the riak commands.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
bostko: For distributions that require this, It is annoying to show it all the time in the log

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
bostko: You are right

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
bostko: Very good note.
The point here is that this clusterCommit() method is made as effector.
If we go to the use case of that command it is as following: executing one of the following effectors: leaveCluster(), joinCluster(), removeNode() and maybe recoverFailedNode()
So let's say we want to support executing these commands and then invoking commitCluster() at the end.

However when setting up a new cluster blueprint it will lead that at the end of blueprint initialization the user should manually go to the UI and click on commitCluster()
The way we could workaround this is by introducing new version of the methods leaveCluster(), joinCluster(), removeNode() like leaveClusterEffector(), joinClusterEffector(), removeNodeEffector() which is without invoking commitCluster() and leave the current methods as they are.
That way deploying and operations queue would work but then it won't be very well supported in the UI, it will be easy for the user to miss to invoke commitCluster().

ahgittin: agree, it feels like our default behaviour should be to commit.

we could (in time) add an optional parameter "commitChange" which defaults to true, that way if someone wanted to they could batch/preview changes.

i'd say add a TODO with this (and reference the commit which includes the commitCluster effector, in case we want to revive that we can copy-paste from here.)

ahgittin:
btw @bostko to answer your question in original post:  if we want sensors/effectors for individual bucket types what i would do is add entities for those bucket types (and same for buckets, and for databases in DBMSs etc).

not needed at this stage.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
```diff
@@ -183,18 +183,31 @@
...
+    void removeNode(@EffectorParam(name = "nodeName") String nodeName);
 
     @Effector(description = "Recover and join the Riak cluster on the given node")
-    public void recoverFailedNode(@EffectorParam(name = "nodeName") String nodeName);
+    void recoverFailedNode(@EffectorParam(name = "nodeName") String nodeName);
 
-    @Effector(description = "Commit changes made to the Riak cluster")
-    public void commitCluster();
+    @Effector(description = "Create bucket type")
+    void bucketTypeCreate(@EffectorParam(name = "bucketTypeName") String bucketTypeName,
```
ahgittin: verb first is the usual pattern for methods isn't it?   so `createBucketType`

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
```diff
@@ -183,18 +183,31 @@
...
-    public void removeNode(@EffectorParam(name = "nodeName") String nodeName);
+    void removeNode(@EffectorParam(name = "nodeName") String nodeName);
 
     @Effector(description = "Recover and join the Riak cluster on the given node")
-    public void recoverFailedNode(@EffectorParam(name = "nodeName") String nodeName);
+    void recoverFailedNode(@EffectorParam(name = "nodeName") String nodeName);
 
-    @Effector(description = "Commit changes made to the Riak cluster")
-    public void commitCluster();
+    @Effector(description = "Create bucket type")
```
ahgittin: i hate descriptions and javadoc which simply restate the name.  say `Creates a new bucket type with the given name and properties, with properties in <?what kind of ?> format`.  (or should properties be a map?)

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
bostko: I am following the riak commands flow.
This way it will also stay sorted in the UI, so bucket stuff will be next to each other

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
```diff
@@ -183,18 +183,31 @@
...
+    void bucketTypeCreate(@EffectorParam(name = "bucketTypeName") String bucketTypeName,
+                          @EffectorParam(name = "bucketTypeProperties") String bucketTypeProperties);
 
+    @Effector(description = "List bucket types")
+    List<String> bucketTypeList();
```
ahgittin: as above, call this `listBucketTypes` ?  (note the plural.)  and expand the description, at least a little.  `List the names all bucket types in the system.`

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
```diff
@@ -183,18 +183,31 @@
...
+
+    @Effector(description = "Status bucket type")
+    List<String> bucketTypeStatus(@EffectorParam(name = "bucketTypeName") String bucketTypeName);
```
ahgittin: don't understand this -- is it `lookupBucketTypeStatus` ?  what is the meaning of the return `List<String>` ?

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
```diff
@@ -183,18 +183,31 @@
...
+
+    @Effector(description = "Update bucket type")
+    void bucketTypeUpdate(@EffectorParam(name = "bucketTypeName") String bucketTypeName,
```
ahgittin: verb first again

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
```diff
@@ -183,18 +183,31 @@
...
+                          @EffectorParam(name = "bucketTypeProperties") String bucketTypeProperties);
+
+    @Effector(description = "Activate bucket type")
+    void bucketTypeActivate(@EffectorParam(name = "bucketTypeName") String bucketTypeName);
```
ahgittin: and again

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -445,32 +438,81 @@ public void removeNode(String nodeName) {
...
     }
 
     @Override
-    public void commitCluster() {
-        if (hasJoinedCluster()) {
-            ScriptHelper commitClusterScript = newScript("commitCluster")
-                    .body.append(sudo(format("%s cluster plan", getRiakAdminCmd())))
-                    .body.append(sudo(format("%s cluster commit", getRiakAdminCmd())))
-                    .failOnNonZeroResultCode();
+    public void bucketTypeCreate(String bucketTypeName, String bucketTypeProperties) {
```
ahgittin: again, change method names here to match effector rename

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
ahgittin: as discussed -- you are following the riak CLI syntax, with nice side effect that effectors are nicely sorted.  makes a lot of sense.  disregard these comments.  :)

ahgittin:
looks good.  add the TODO and expand descriptions, then i think good to merge.

can test whether we can abandon `sudo` in a subsequent PR.

bostko:
I fixed the bucket type effectors' description and I added a TODO in https://github.com/bostko/incubator-brooklyn/commit/343ed278035e4fe58d9273add9c4f16076d09a24

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
bostko: Actually I found why it is implemented that way.
I think the initial idea was to support cluster queue operations. It is correct not to commit just after the cluster join command. Other than that it was executed for one of the members.
Instead of that commit was executed after the joinings are done and then in the RiakCluster.start at the end it was executed clusterCommit()
The bug here is that the commit is not executed properly because every time joining was executed against random member of the ring instead of using one, so commiting should be done against the nodes or the node which triggered cluster join. This is the actual bug. And the other bug is may be that leaveCluster had plan and commit command which is unnecessary.

My point is that it could support queue cluster operations but the task requires bigger effort.

ahgittin:
LGTM.  Needs a lot of testing, but going ahead and merging.

PR #583, opened 2015-04-02
made entity summary Serializable

andreschembri:
made brooklyn.rest.domain.EntitySummary Serializable

### DIFF
--- a/usage/rest-api/src/main/java/brooklyn/rest/domain/EntitySummary.java
+++ b/usage/rest-api/src/main/java/brooklyn/rest/domain/EntitySummary.java
@@ -24,10 +24,11 @@ import org.codehaus.jackson.annotate.JsonProperty;
 import org.codehaus.jackson.map.annotate.JsonSerialize;
 import org.codehaus.jackson.map.annotate.JsonSerialize.Inclusion;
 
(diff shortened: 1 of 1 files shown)


ahgittin:
@andreschembri - all the domain classes are serializable and you rightly point out they aren't marked as such.  but why single out this particular class to be serializable?

andreschembri:
@ahgittin 
This particular class was effecting my rest api calls, the other classes were not producing any side effects directly to me so I didn't really look at them, however I agree that these should be consistent hence they should be all serialzable, shall I work on this?

ahgittin:
@andreschembri how are you using the REST API that this is a problem?  i think we shouldn't need java `Serializable` as we use jackson serialization instead.

aledsage:
@ahgittin it was calling the REST API in a mule flow, using `brooklyn.rest.client.BrooklynApi` to generate the required http calls, and to deserialize the results into Java pojos. The call returned `EntitySummary`. The mule flow then tried to serialise the returned Java object (I think for use in a subsequent step of the flow), which failed.

I agree that all of these classes should be `Serializable`. I've no objection with making this one Serializable in this PR, and then we can address the others in a subsequent PR.

PR #585, opened 2015-04-07
Add multiple items to catalog with a single YAML

ahgittin:
Useful as we move towards a bigger catalog.  The multi-item format is described at:

    https://github.com/ahgittin/incubator-brooklyn/blob/catalog-multi-item/docs/guide/ops/catalog/index.md

The catalog gui has been updated accordingly.  @sjcorbett @neykov this is probably in your wheelhouse!


ahgittin:
@neykov fixed.  but i'm guessing you're not finished reviewing!  :)

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -491,43 +508,127 @@ public void load() {
...
-        final String description = (String) catalog.getMaybe("description").orNull();
-        final String iconUrl = (String) catalog.getMaybe("iconUrl").orNull();
-        final String iconUrlUnderscore = (String) catalog.getMaybe("icon_url").orNull();
-        final String deprecated = (String) catalog.getMaybe("deprecated").orNull();
+        // now look at the actual item, first correcting the sourceYaml and interpreting the catalog metadata
+        String itemYaml = Yamls.getTextOfYamlAtPath(sourceYaml, "item").getMatchedYamlTextOrWarn();
+        if (itemYaml!=null) sourceYaml = itemYaml;
+        else sourceYaml = new Yaml().dump(item);
+        
+        CatalogItemType itemType = TypeCoercions.coerce(getFirstAs(catalogMetadata, Object.class, "item.type", "itemType", "item_type").orNull(), CatalogItemType.class);
```
neykov: Can't see the reason why introduce an explicit type?

ahgittin: you mean because we can infer?  inference seems dangerous to me, better to require for anything other than an entity i think.  (also application v template is now slightly different)

ahgittin:
i've discouraged using the "top-level syntax", and explained why it is important:  it is needed for CAMP compliance, where the *same* blueprint can be posted either to the catalog endpoint or the applications endpoint

ahgittin:
intermittent test failure :( /cc @aledsage 

`ServerPoolTest.testDeployReleaseDeploy` gives

    java.lang.AssertionError: entity=ServerPoolImpl{id=yfGmsgI3}; attribute=Sensor: pool.available (java.lang.Integer) expected [3] but found [2]

(as the last commit was docs only it is clearly not the culprit! -- but closing and reopening to try to get a good report)

ahgittin:
@neykov good to go with the last set of mods?

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -538,19 +639,17 @@ public void load() {
             } else {
                 catalogSymbolicName = id;
             }
-        } else if (Strings.isNonBlank(name)) {
-            catalogSymbolicName = name;
-        } else if (Strings.isNonBlank(plan.getName())) {
+        } else if (plan!=null && Strings.isNonBlank(plan.getName())) {
             catalogSymbolicName = plan.getName();
-        } else if (plan.getServices().size()==1) {
+        } else if (plan!=null && plan.getServices().size()==1) {
```
neykov: Same check needed for all plan.getXXX calls following. What's the use case for registering invalid yaml?

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -664,10 +858,28 @@ private DeploymentPlan makePlanFromYaml(String yaml) {
...
+        return Iterables.getOnlyElement(addItems(yaml, forceUpdate));
+    }
+    
+    @Override
+    public List<? extends CatalogItem<?,?>> addItems(String yaml, boolean forceUpdate) {
         log.debug("Adding manual catalog item to "+mgmt+": "+yaml);
         checkNotNull(yaml, "yaml");
-        CatalogItemDtoAbstract<?,?> itemDto = getAbstractCatalogItem(yaml);
+        List<CatalogItemDtoAbstract<?, ?>> result = addAbstractCatalogItems(yaml, forceUpdate);
+        // previously we did this here, but now we do it on each item, in case #2 refers to #1
```
neykov: I see where you are coming from, but when combined with the "don't overwrite existing items" default behaviour users will have hard time testing and developing catalog items blueprints if addition is not atomic.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -491,43 +508,127 @@ public void load() {
...
-        @SuppressWarnings("rawtypes")
-        Maybe<Map> possibleCatalog = plan.getCustomAttribute("brooklyn.catalog", Map.class, true);
-        MutableMap<String, Object> catalog = MutableMap.of();
-        if (possibleCatalog.isPresent()) {
-            @SuppressWarnings("unchecked")
-            Map<String, Object> catalog2 = (Map<String, Object>) possibleCatalog.get();
-            catalog.putAll(catalog2);
+        List<CatalogItemDtoAbstract<?, ?>> result = MutableList.of();
+        
+        addAbstractCatalogItems(Yamls.getTextOfYamlAtPath(yaml, "brooklyn.catalog").getMatchedYamlTextOrWarn(),
```
neykov: Above you are looking at both `brooklyn.catalog` and `catalog`, but here it's the first one only. Why support both at all, better stick to one of them. I know that source availability is optional, but still looks wrong.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -491,43 +508,127 @@ public void load() {
...
-        if (possibleCatalog.isPresent()) {
-            @SuppressWarnings("unchecked")
-            Map<String, Object> catalog2 = (Map<String, Object>) possibleCatalog.get();
-            catalog.putAll(catalog2);
+        List<CatalogItemDtoAbstract<?, ?>> result = MutableList.of();
+        
+        addAbstractCatalogItems(Yamls.getTextOfYamlAtPath(yaml, "brooklyn.catalog").getMatchedYamlTextOrWarn(), 
+            catalogMetadata, result, null, whenAddingAsDtoShouldWeForce);
+        
+        itemDef.remove("brooklyn.catalog");
```
neykov: Same here, better remove `catalog` support.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
ahgittin: agree

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
ahgittin: yes.  if i've supplied a policy then `plan` wouldn't be set.  (or if i've deliberately supplied a broken template which i expect people to fix, but that's not something we need to support.)

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -491,43 +508,127 @@ public void load() {
...
-        final String version = Strings.toString(catalog.getMaybe("version").orNull());
-        final String symbolicName = (String) catalog.getMaybe("symbolicName").orNull();
-        final String name = (String) catalog.getMaybe("name").orNull();
-        final String displayName = (String) catalog.getMaybe("displayName").orNull();
-        final String description = (String) catalog.getMaybe("description").orNull();
-        final String iconUrl = (String) catalog.getMaybe("iconUrl").orNull();
-        final String iconUrlUnderscore = (String) catalog.getMaybe("icon_url").orNull();
-        final String deprecated = (String) catalog.getMaybe("deprecated").orNull();
+        // now look at the actual item, first correcting the sourceYaml and interpreting the catalog metadata
+        String itemYaml = Yamls.getTextOfYamlAtPath(sourceYaml, "item").getMatchedYamlTextOrWarn();
```
neykov: When viewing the YAML of a catalog item in the UI all the catalog meta info is lost. Just the CAMP blueprint remains. `id` and `version` are displayed by other means, but `libraries` is lost. Better include the original YAML, or Yamls.dump the merged blueprint.

neykov: Nicest would be to show the whole YAML and highlight the item CAMP plan, but that would require additional work on the UI.

neykov: You are keeping track of three views on the same markup: 
  * the string (sub-)yaml
  * the parsed map
  * the camp plan

Why not deal with just the parsed YAML and pass it directly to pdp parser? This will greatly simplify the code. The only place where `Yamls.getTextOfYamlAtPath` is really needed is for the markers to highlight in UI (if we go that way at all).

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
neykov: The old code used `makePlanFromYaml` for policies, why throw now?

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
neykov: (from discussion)

agreed:  in most cases it can be inferred

also noted:  for item_type entity it would be better to expect a service specification { type: ... } rather than a full camp plan

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
ahgittin: in the case of templates we want to display the source yaml

the camp plan is not valid for cases where we expect item in `{ type: xxx }` syntax

however i think we *need* the camp plan as that is the best way to resolve catalog items.  if we want to support `{ type: other-catalog-item }` syntax -- which i think we do -- then i think it gets messier:

* we want to use the camp plan parser to create the spec (since we need the underlying java)
* so we need to create a camp plan if `{type: xxx}` syntax is used
* but we first need to look up the type of `xxx` by trying a class-load + catalog-lookup in order to know how to generate the plan

it's doable but tedious.  any thoughts on an easier way?

ahgittin: re the UI i agree we should show catalog metadata in the catalog item view.  however i think it shouldn't be in the "plan" -- we should have a dedicated space in the gui and rest api calls for metadata (if not already there?).  i think that doesn't need to be part of this PR.

ahgittin:
great comments @neykov - i'm working through addressing them.  most were easy, now working on the atomic add and the type inference.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
ahgittin: back to resolving catalog items -- the full camp plan is the easiest way to create a spec -- but as you pointed out we only use the spec to determine the catalog item type, which now we know, so we don't need the spec and i think everything will become quite a bit simpler

ahgittin:
@neykov i think it's all addressed now.  all the commits are different now as i've rebased -- merge conflicts around rest api for listing locations etc.  only the last commit is "new", and you'll see in the comment there how i've handled your points.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -529,145 +641,315 @@ public void load() {
...
+            this.loader = loader;
+            this.itemsDefinedSoFar = itemsDefinedSoFar;
         }
-    }
 
-    private boolean isApplicationSpec(EntitySpec<?> spec) {
-        return !Boolean.TRUE.equals(spec.getConfig().get(EntityManagementUtils.WRAPPER_APP_MARKER));
+        public PlanInterpreterGuessingType reconstruct() {
+            attemptType(null, CatalogItemType.ENTITY);
+            attemptType(null, CatalogItemType.TEMPLATE);
```
neykov: This will never succeed, either the ENTITY above will resolve or this will fail as well, for the same reason.

neykov:
Finished review, useful functionality moving us closer to getting rid of XML.
Only major concern is CI resolution will fail if id:version reference used. As soon as addressed the PR can be merged. Worth adding a test as well.
The changes are a bit on the dirty side, missing some corner cases with the validation (when child catalog item references are involved), but on the other hand the changes are self contained which is good - we can fix this additionally if it proves to be a problem in real world use.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
ahgittin: the `TEMPLATE check will succeed is if `catalogItemType` has been pre-set as `TEMPLATE`; i'll change the code to make this explicit

ahgittin:
@neykov Thanks!  Great stuff.  I think I've applied all the fixes.

There is one issue I'm trying to track down, pretty sure it wasn't introduced here but if it's easy to fix it here I'd like to.  As per discussion on IRC.  Will update this when that's resolved.

PR #587, opened 2015-04-08
Parse service types in YAML using pluggable resolver impls

grkvlt:
Adds a `ServiceLoader` plugin capability for parsing the `serviceType: xxx:yyy` part of the YAML blueprint. Includes `brooklyn:`, `java:`, `catalog:` and `chef:` implementations. This will be useful for a `docker:` service type prefix in Clocker.

This is based on my comment on #436 and @neykov's suggestions there.

### DIFF
--- a/usage/camp/src/main/java/io/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
+++ b/usage/camp/src/main/java/io/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
@@ -19,7 +19,8 @@
 package io.brooklyn.camp.brooklyn.spi.creation;
 
 import io.brooklyn.camp.brooklyn.BrooklynCampConstants;
-import io.brooklyn.camp.brooklyn.spi.dsl.methods.BrooklynDslCommon;
+import io.brooklyn.camp.brooklyn.spi.creation.service.BrooklynServiceTypeResolver;
+import io.brooklyn.camp.brooklyn.spi.creation.service.ServiceTypeResolver;
 import io.brooklyn.camp.spi.AbstractResource;
 import io.brooklyn.camp.spi.ApplicationComponentTemplate;
 import io.brooklyn.camp.spi.AssemblyTemplate;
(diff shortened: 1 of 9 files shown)


grkvlt:
@neykov @ahgittin I have tested this with a new `ServiceTtypeResolver` for Clocker in https://github.com/brooklyncentral/clocker/pull/100 which works nicely, using types like `docker:mysql:5.7` or docker:jboss/wildfly:8.1.0.Final` etc.

usage/camp/src/main/java/io/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
```diff
@@ -77,59 +74,80 @@
...
-            if (type!=null) {
-                if (type.startsWith("brooklyn:") || type.startsWith("java:")) return BrooklynComponentTemplateResolver.class;
-                if (type.equalsIgnoreCase("chef") || type.startsWith("chef:")) return ChefComponentTemplateResolver.class;
-                // TODO other BrooklynComponentTemplateResolver subclasses detected here 
-                // (perhaps use regexes mapping to subclass name, defined in mgmt?)
+            return findService(context, type);
+        }
+
+        protected static ServiceTypeResolver findService(BrooklynClassLoadingContext context, String type) {
+            String prefix = Splitter.on(":").splitToList(type).get(0);
```
ahgittin: first do `if (type.indexOf(':')==-1) return null;` ?  in case type matches a prefix but there is no `:`

ahgittin:
This looks really good to me.  One minor comment is all.  @neykov if you have the chance to look this over too that would be useful.

usage/camp/src/main/java/io/brooklyn/camp/brooklyn/spi/creation/service/ServiceTypeResolver.java
```diff
@@ -0,0 +1,73 @@
...
+ * @see ChefServiceTypeResolver
+ */
+public interface ServiceTypeResolver {
+
+    String DEFAULT_TYPE_PREFIX = "brooklyn";
+
+    /**
+     * The service type prefix the resolver is responsible for.
+     */
+    String getTypePrefix();
```
aledsage: I wonder about also having an `accepts(String serviceType)` or an `accepts(BrooklynComponentTemplateResolver resolver, String serviceType)` - i.e. similar to `LocationResolver` - but no strong feelings.

PR #588, opened 2015-04-09
added representation of registeredType to CatalogItemSummary

andreschembri:
added String registeredType;  since rest client was failing because it could not represent the JSON returned correctly .


andreschembri:
it seems that there are calls that use this class that do not yet provide the registeredType property, hence my modification was breaking the code. I am going to change this and reopen a different pull request.

ahgittin:
@andreschembri does #585 address this sufficiently?  it's not the main focus of that PR but it adds the annotation `@JsonIgnoreProperties(ignoreUnknown = true)` which should stop errors

PR #589, opened 2015-04-09
make ConfigBag thread-safe

ahgittin:
`ConfigBag` has not been thread-safe, but we have been accessing it from multiple threads.  This can cause `ConcurrentModificationException` to be thrown in many places.

Such a higher-level shared class should be thread-safe, and this makes it so by synchronizing on the `ConfigBag` class appropriately.  As we do not have iterators here it was pretty straightforward.  (You get a copy of the map to iterate -- that has always been the case.)

### DIFF
--- a/core/src/main/java/brooklyn/util/config/ConfigBag.java
+++ b/core/src/main/java/brooklyn/util/config/ConfigBag.java
@@ -20,7 +20,7 @@ package brooklyn.util.config;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import java.util.Collections;
+import java.util.ConcurrentModificationException;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
@@ -37,6 +37,7 @@ import brooklyn.util.flags.TypeCoercions;
 import brooklyn.util.guava.Maybe;
 import brooklyn.util.javalang.JavaClassNames;
 
+import com.google.common.annotations.Beta;
 import com.google.common.base.Objects;
 import com.google.common.collect.Sets;
 
@@ -47,7 +48,9 @@ import com.google.common.collect.Sets;
  * It is recommended to use {@link ConfigKey} instances to access,
  * although in some cases (such as setting fields from flags, or copying a map)
  * it may be necessary to mark things as used, or put, when only a string key is available.
- * 
+ * <p>
+ * This bag is order-preserving and thread-safe except where otherwise indicated.
+ * <p>
  * @author alex
  */
 public class ConfigBag {
@@ -71,6 +74,9 @@ public class ConfigBag {
 
     /**
      * Creates an instance that is backed by a "live map" (e.g. storage in a datagrid).
+     * The order-preserving nature of this class is only guaranteed if the
+     * provided storage has those properties. External modifications to the store can cause
+     * {@link ConcurrentModificationException} to be thrown, here or elsewhere. 
      */
     public static ConfigBag newLiveInstance(Map<String,Object> storage) {
         return new ConfigBag(checkNotNull(storage, "storage map must be specified"));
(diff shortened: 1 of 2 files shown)


neykov:
Looks good to merge.

core/src/main/java/brooklyn/util/config/ConfigBag.java
```diff
@@ -96,14 +102,42 @@ public static ConfigBag newInstanceCopying(final ConfigBag configBag) {
...
+    @Beta
+    public static ConfigBag newInstanceExtending(final ConfigBag parentBag) {
+        return new ConfigBagExtendingParent(parentBag).copy(parentBag);
+    }
+
+    /** @see #newInstanceExtending(ConfigBag) */
+    public static class ConfigBagExtendingParent extends ConfigBag {
+        ConfigBag parentBag;
+        private ConfigBagExtendingParent(ConfigBag parentBag) {
+            this.parentBag = parentBag;
```
aledsage: Was a bit surprised this didn't do `copy(parentBag)`. Should we leave that up to the caller, or do it inside `ConfigBagExtendingParent`'s constructor? I lean towards the latter because otherwise it hasn't really extended it. But no strong feelings.

core/src/main/java/brooklyn/util/config/ConfigBag.java
```diff
@@ -132,12 +166,12 @@ public String getDescription() {
     
     /** current values for all entries 
      * @return non-modifiable map of strings to object */
-    public Map<String,Object> getAllConfig() {
-        return Collections.unmodifiableMap(config);
+    public synchronized Map<String,Object> getAllConfig() {
```
aledsage: Preference to not synchronize on `this` - if other code synchronizes on the object then leads to a risk of deadlock. It leaks the implementation detail. Better to declare a `private final` field, and synchronise on that.

aledsage: Ah, I see that `getAllConfigMutable` means that we may need to synchronize on this. I really don't like our uses of `getAllConfigMutable`. Having public methods like that is an anti-pattern in my opinion.

Longer term, we should deprecate that and also `getUnusedConfigMutable`. We could have a `getMutex()` method for what to synchronize on, with that marked `@Beta`?

core/src/test/java/brooklyn/util/config/ConfigBagTest.java
```diff
@@ -0,0 +1,191 @@
...
+        bag.put(K3, "v3");
+        runConcurrentTest(bag, 20, Duration.seconds(5));
+    }
+    
+    private void runConcurrentTest(final ConfigBag bag, int numThreads, Duration time) throws InterruptedException {
+        List<Thread> threads = MutableList.of();
+        final Map<Thread,Exception> exceptions = new ConcurrentHashMap<Thread,Exception>();
+        final AtomicInteger successes = new AtomicInteger();
+        for (int i=0; i<numThreads; i++) {
+            Thread t = new Thread() {
```
aledsage: Tests should use `ExecutorService executor = Executors.newCachedThreadPool()`, and in a finally block (or in tearDown) should do `executor.shutdownNow`. There are very few situations that folk should do `new Thread` these days, in my opinion.

aledsage:
Finished reviewing; one big thing (whether to synchronize on `this`) and the rest minor.

core/src/main/java/brooklyn/util/config/ConfigBag.java
ahgittin: let's make the class private

ahgittin: and do in constructor

core/src/main/java/brooklyn/util/config/ConfigBag.java
ahgittin: +1 to phasing out the mutable accesses

as for `this`, it's a common and well understood pattern, and the sync dangers are no different to elsewhere.  there may be times someone wants to batch items so it is handy to expose the object for synching.  in time perhaps we'll replace with a RW lock, which would be the elegant thing, but since most `ConcurrentMap` impls don't do that it seems like overkill for now.

have added a comment in javadoc about synching and possibility of that changing

PR #590, opened 2015-04-09
Catalogitemsummaryfix

andreschembri:
Not sure if this is the correct way to resolve this issue:

- There are some code that is using this POJO that are providing the registeredType and others that do not. 
- If the registeredType is present and the constructor does not take it , the mapping would fail, if registeredType is not present and the constructor expects the registeredType the number of arguments would not match.
- I ended up adding a new constructor that takes the registeredType, and leaving the one that does not in place

### DIFF
--- a/usage/rest-api/src/main/java/brooklyn/rest/domain/CatalogItemSummary.java
+++ b/usage/rest-api/src/main/java/brooklyn/rest/domain/CatalogItemSummary.java
@@ -16,6 +16,8 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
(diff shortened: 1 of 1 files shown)


ahgittin:
as noted in #588 this might be fixed a better way in #585 (not caring about `registeredType` which is mostly an internal detail anyway, which hopefully we can get rid of in time...)

if not, can you include a test case showing the failure which is fixed here so we can better understand?

andreschembri:
Yep this makes sense to me ! hence I am closing this pull request.

Thanks,
Andre

PR #592, opened 2015-04-12
allow open ports to be specified in provisioning properties

robertgmoss:


software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
```diff
@@ -429,8 +432,21 @@ protected final void callStartHooks() {}
...
                 if (p != null && !p.isEmpty()) ports.add(p.iterator().next());
+            }   
+        }
+        
+        Map<String, Object> provisioningProperties = getConfig(PROVISIONING_PROPERTIES);
+        if(provisioningProperties.containsKey("required.ports")){
+            Object requiredPorts = provisioningProperties.get("required.ports");
+            if (requiredPorts instanceof Integer){
+                ports.add((Integer)requiredPorts);
+            }else if (requiredPorts instanceof List) {
```
aledsage: Better to accept `Iterable` rather than just list?

software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
```diff
@@ -429,8 +432,21 @@ protected final void callStartHooks() {}
             if (PortRange.class.isAssignableFrom(k.getType())) {
                 PortRange p = (PortRange)getConfig(k);
                 if (p != null && !p.isEmpty()) ports.add(p.iterator().next());
+            }   
+        }
+        
+        Map<String, Object> provisioningProperties = getConfig(PROVISIONING_PROPERTIES);
+        if(provisioningProperties.containsKey("required.ports")){
```
aledsage: The other keys in `PROVISIONING_PROPERTIES` correspond to `ConfigKey`s, such as those in `CloudLocationConfig`. It would be good to include `required.ports` as a `ConfigKey` there, rather than as a hard-coded string. If we do that, then it also needs to be respected by `JcloudsLocation` etc.

Also, there is a `CloudLocationConfig.INBOUND_PORTS` config key. If you set that in `PROVISIONING_PROPERTIES` then it (unfortunately but understandably) overrides all other values from the `getRequiredOpenPorts`. We could augment this with a `ADDITIONAL_INBOUND_PORTS` config key perhaps?

aledsage:
Thanks @robertgmoss - this will be really useful.

@duncangrant was also talking about adding this. Robert, can you sync up with Duncan to get his input as well please?

ahgittin:
on IRC we discussed:

> maybe we should expand getRequiredOpenPorts() to look for any config which has a segment port and an integer value, e.g. artifactory.http.port ?  alternatively an explicit required.ports config key but that seems clunky.

looks like you've implemented the latter.  could we instead do the former, e.g. in `getRequiredOpenPorts()` also loop through the `getConfig()` and `getSensors()` finding any where the key's name matches regex `.*\.port` and checking if their *raw* value is an integer or a string matching `[0-9]+` and if so then add it?  this is a bit of convention magic but it's such a common convention, and avoids needing to duplicate the ports to open if they are defined in config keys.

software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
grkvlt: Won't a cooma-separated `String` or single `Integer` get coerced to a `List<Integer>` using a `ConfigKey` or at least you can call the appropriate `TypeCoercions#coerce(...)` method?

robertgmoss:
@ahgittin this was the approach I initially took, but I seem to be missing something.  

None of the keys being iterated by `getRequiredOpenPorts()` currently,  i.e `getEntityType().getConfigKeys()`, is the one I have set under `brooklyn.config` in the YAML (perhaps this is an earlier snapshot before they have been set, I don't know). 

If you mean I should iterate `getConfig()` without the prior call to `getEntityType()` the only visible methods by that name are `getConfig(ConfigKey)` and `getConfig(HasConfigKey)`, which are obviously not iterable.

ahgittin:
@robertgmoss use `config().getBag().getAllConfig()` and iterate through the keys there.  if you use `getEntityType()` you are looking at the type declaration (in java) which excludes the dynamic keys set at runtime.

robertgmoss:
thanks @ahgittin, I knew I was missing something.

@aledsage, is the `ADDITIONAL_INBOUND_PORTS` config key useful in addition to the convention approach that Alex is suggesting? i.e should I keep what I have in my last commit?

ahgittin:
`ADDITIONAL_INBOUND_PORTS` might be useful - wdyt?  maybe keep it and mark it `@Beta` ?

software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
```diff
@@ -414,6 +417,14 @@ protected final void callStartHooks() {}
         result.putAll(getConfig(PROVISIONING_PROPERTIES));
         if (result.get(CloudLocationConfig.INBOUND_PORTS) == null) {
             Collection<Integer> ports = getRequiredOpenPorts();
+            Object requiredPorts = result.get(CloudLocationConfig.ADDITIONAL_INBOUND_PORTS);
+            if (requiredPorts instanceof Integer){
+                ports.add((Integer)requiredPorts);
+            }else if (requiredPorts instanceof Iterable) {
```
grkvlt: Formatting (should be spaces between language keywords and braces/parentheses)

robertgmoss:
@grkvlt formatting fixed.

ahgittin:
LGTM

PR #594, opened 2015-04-13
REST api domain objects Serializable

aledsage:
- Makes everything in brooklyn.rest.domain.* serialisable
- Useful for users of brooklyn.rest.client.BrooklynApi, where
  they use these Java objects. Serialisable is not necessary,
  but for subsequent usage (e.g. in a mule flow) then it is
  useful if they are.

See related PR #583 which has already been merged.

### DIFF
--- a/usage/rest-api/src/main/java/brooklyn/rest/domain/EntitySpec.java
+++ b/usage/rest-api/src/main/java/brooklyn/rest/domain/EntitySpec.java
@@ -19,14 +19,19 @@
 package brooklyn.rest.domain;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+
 import com.google.common.collect.ImmutableMap;
+
 import org.codehaus.jackson.annotate.JsonProperty;
 
+import java.io.Serializable;
 import java.util.Collections;
 import java.util.Map;
 
-public class EntitySpec implements HasName {
+public class EntitySpec implements HasName, Serializable {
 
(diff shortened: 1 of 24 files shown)


usage/rest-api/src/main/java/brooklyn/rest/domain/EntitySpec.java
```diff
@@ -19,14 +19,19 @@
...
 import com.google.common.collect.ImmutableMap;
+
 import org.codehaus.jackson.annotate.JsonProperty;
 
+import java.io.Serializable;
 import java.util.Collections;
 import java.util.Map;
 
-public class EntitySpec implements HasName {
+public class EntitySpec implements HasName, Serializable {
```
neykov: Does this restrict the possible values for spec configuration to serializable instances only?

neykov: I see they are already strings, so not an issue.

ahgittin:
several files are missing `Serializable` -- e.g. `CatalogEntitySummary` ?

aledsage:
@ahgittin `CatalogEntitySummary` extends `CatalogItemSummary` so is `Serializable`. Do you think we want to include `implements Serializable` on the sub-class as well, in case someone refactors it later and doesn't look at the super-class? I'm fine with as-is.

ahgittin:
Ah of course. Merge away.

grkvlt:
Merging :frog:

PR #595, opened 2015-04-13
Make YAML more powerful

ahgittin:
allowing a `firstMemberSpec` for cluster, make enrichers easier to set up in yaml (with good example), and better support for referring to sensors within maps (esp when configuring enrichers)

### DIFF
--- a/core/src/main/java/brooklyn/config/internal/AbstractConfigMapImpl.java
+++ b/core/src/main/java/brooklyn/config/internal/AbstractConfigMapImpl.java
@@ -1,0 +1,110 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package brooklyn.config.internal;
+
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.concurrent.Future;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import brooklyn.config.ConfigKey;
+import brooklyn.config.ConfigKey.HasConfigKey;
+import brooklyn.config.ConfigMap;
+import brooklyn.entity.basic.ConfigMapViewWithStringKeys;
+import brooklyn.event.basic.StructuredConfigKey;
+import brooklyn.util.flags.TypeCoercions;
+import brooklyn.util.task.DeferredSupplier;
+
+public abstract class AbstractConfigMapImpl implements ConfigMap {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractConfigMapImpl.class);
+    
+    protected final ConfigMapViewWithStringKeys mapViewWithStringKeys = new ConfigMapViewWithStringKeys(this);
+    
+    /**
+     * Map of configuration information that is defined at start-up time for the entity. These
+     * configuration parameters are shared and made accessible to the "children" of this
+     * entity.
+     */
+    protected Map<ConfigKey<?>,Object> ownConfig = Collections.synchronizedMap(new LinkedHashMap<ConfigKey<?>, Object>());
+
+    public <T> T getConfig(ConfigKey<T> key) {
+        return getConfig(key, null);
+    }
+    
+    public <T> T getConfig(HasConfigKey<T> key) {
+        return getConfig(key.getConfigKey(), null);
+    }
+    
+    public <T> T getConfig(HasConfigKey<T> key, T defaultValue) {
+        return getConfig(key.getConfigKey(), defaultValue);
+    }
+    
+    @Override @Deprecated
+    public Object getRawConfig(ConfigKey<?> key) {
+        return getConfigRaw(key, true).orNull();
+    }
+    
+    protected Object setConfigPrep1(ConfigKey<?> key, Object v) {
+        Object val;
+        if ((v instanceof Future) || (v instanceof DeferredSupplier)) {
+            // no coercion for these (coerce on exit)
+            val = v;
+        } else if (key instanceof StructuredConfigKey) {
+            // no coercion for these structures (they decide what to do)
+            val = v;
+        } else if ((v instanceof Map || v instanceof Iterable) && key.getType().isInstance(v)) {
+            // don't do coercion on put for these, if the key type is compatible, 
+            // because that will force resolution deeply
+            val = v;
+        } else {
+            try {
+                // try to coerce on input, to detect errors sooner
+                val = TypeCoercions.coerce(v, key.getTypeToken());
+            } catch (Exception e) {
+                throw new IllegalArgumentException("Cannot coerce or set "+v+" to "+key, e);
+                // if can't coerce, we could just log, and *throw* the error when we retrieve the config
+                // but for now, fail fast (above)
+//                Exceptions.propagateIfFatal(e);
+//                LOG.warn("Cannot coerce or set "+v+" to "+key+" (ignoring): "+e, e);
+//                val = v;
+            }
+        }
+        return val;
+    }
+
+    
+    @Override
+    public Map<String,Object> asMapWithStringKeys() {
+        return mapViewWithStringKeys;
+    }
+
+    @Override
+    public int size() {
+        return ownConfig.size();
+    }
+
+    @Override
+    public boolean isEmpty() {
+        return ownConfig.isEmpty();
+    }
+}
--- a/core/src/main/java/brooklyn/config/internal/AbstractConfigMapImpl.java
+++ b/core/src/main/java/brooklyn/config/internal/AbstractConfigMapImpl.java
@@ -64,7 +64,7 @@ public abstract class AbstractConfigMapImpl implements ConfigMap {
         return getConfigRaw(key, true).orNull();
     }
     
-    protected Object setConfigPrep1(ConfigKey<?> key, Object v) {
+    protected Object coerceConfigVal(ConfigKey<?> key, Object v) {
         Object val;
         if ((v instanceof Future) || (v instanceof DeferredSupplier)) {
             // no coercion for these (coerce on exit)
@@ -82,8 +82,9 @@ public abstract class AbstractConfigMapImpl implements ConfigMap {
                 val = TypeCoercions.coerce(v, key.getTypeToken());
             } catch (Exception e) {
                 throw new IllegalArgumentException("Cannot coerce or set "+v+" to "+key, e);
-                // if can't coerce, we could just log, and *throw* the error when we retrieve the config
-                // but for now, fail fast (above)
+                // if can't coerce, we could just log as below and *throw* the error when we retrieve the config
+                // but for now, fail fast (above), because we haven't encountered strong use cases
+                // where we want to do coercion on retrieval, except for the exceptions above
 //                Exceptions.propagateIfFatal(e);
 //                LOG.warn("Cannot coerce or set "+v+" to "+key+" (ignoring): "+e, e);
 //                val = v;
--- a/core/src/main/java/brooklyn/enricher/basic/Joiner.java
+++ b/core/src/main/java/brooklyn/enricher/basic/Joiner.java
@@ -1,0 +1,128 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package brooklyn.enricher.basic;
+
+import java.util.Map;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import brooklyn.config.ConfigKey;
+import brooklyn.entity.Entity;
+import brooklyn.entity.basic.ConfigKeys;
+import brooklyn.entity.basic.EntityLocal;
+import brooklyn.event.AttributeSensor;
+import brooklyn.event.Sensor;
+import brooklyn.event.SensorEvent;
+import brooklyn.event.SensorEventListener;
+import brooklyn.event.basic.BasicSensorEvent;
+import brooklyn.util.collections.MutableList;
+import brooklyn.util.flags.SetFromFlag;
+import brooklyn.util.text.StringEscapes;
+import brooklyn.util.text.Strings;
+
+import com.google.common.reflect.TypeToken;
+
+//@Catalog(name="Transformer", description="Transforms attributes of an entity; see Enrichers.builder().transforming(...)")
+@SuppressWarnings("serial")
+public class Joiner<T> extends AbstractEnricher implements SensorEventListener<T> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(Joiner.class);
+
+    public static ConfigKey<Entity> PRODUCER = ConfigKeys.newConfigKey(Entity.class, "enricher.producer");
+    public static ConfigKey<Sensor<?>> SOURCE_SENSOR = ConfigKeys.newConfigKey(new TypeToken<Sensor<?>>() {}, "enricher.sourceSensor");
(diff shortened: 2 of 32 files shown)


docs/guide/yaml/yaml-reference.md
```diff
@@ -142,11 +145,17 @@ However you can combine locations using `multi`:
...
-    until the given `sensor` from the component `ID` has a "truthy" (i.e. non-trivial, non-empty, non-zero) value
-  * `.config("key")` will insert the value set against the given key at this entity (or nearest ancestor);
-    can be used to supply config at the root which is used in multiple places in the plan
+* `$brooklyn:attributeWhenReady("sensor")` will store a future which will be blocked when it is accessed,
+  until the given `sensor` from the component `ID` has a "truthy" (i.e. non-trivial, non-empty, non-zero) value
+* `$brooklyn:config("key")` will insert the value set against the given key at this entity (or nearest ancestor);
+  can be used to supply config at the root which is used in multiple places in the plan
+* `$brooklyn:sensor("sensor.name")` returns the given sensor on the current entity if found, or an untyped (Object) sensor;
+  `$brooklyn:sensor("io.brooklyn.ContainingEntityClass", "sensor.name")` returns the strongly typed sensor defined in the given class
+* `$brooklyn:component("ID")` refers to a Brooklyn component with the given ID; you can then access the following subfields,
```
aledsage: Ah, I see from this (and the code) that if you omit `component("ID")` then it refers to this entity.

I'd prefer our examples here to include component, such as `$brooklyn:component("ID").config("key")`. I believe that is the most common usage, so that is what we should show people (e.g. they'll want to copy-paste). We can then say that if you omit `component(...)` then it defaults to referring to this entity. We can also move the definition of `$brooklyn:component(...)` to be first.

docs/guide/yaml/yaml-reference.md
```diff
@@ -157,12 +166,14 @@ concise DSL defined here:
 * `$brooklyn:formatString("pattern e.g. %s %s", "field 1", "field 2")` returns a future which creates the formatted string
   with the given parameters, where parameters may be strings *or* other tasks such as `attributeWhenReady`
 * `$brooklyn:literal("string")` returns the given string as a literal (suppressing any `$brooklyn:` expansion)
-* `$brooklyn:sensor("io.brooklyn.ContainingEntityClass", "sensor.name")` returns the strongly typed sensor defined in the given class
+* `$brooklyn:object(Map)` creates an object, using keys `type` to define the java type,
+  and either `object.fields` or `brooklyn.config` to supply bean/constructor/flags to create an instance
 * `$brooklyn:entitySpec(Map)` returns a new `ServiceSpecification` as defined by the given `Map`,
```
aledsage: Longer term, this could do with a cross-reference to a section that describes an `EntitySpec` (for those who don't look at the Java api), and we should avoid java generics syntax like `ConfigKey<EntitySpec>`.

docs/guide/yaml/yaml-reference.md
```diff
@@ -157,12 +166,14 @@ concise DSL defined here:
 * `$brooklyn:formatString("pattern e.g. %s %s", "field 1", "field 2")` returns a future which creates the formatted string
   with the given parameters, where parameters may be strings *or* other tasks such as `attributeWhenReady`
 * `$brooklyn:literal("string")` returns the given string as a literal (suppressing any `$brooklyn:` expansion)
-* `$brooklyn:sensor("io.brooklyn.ContainingEntityClass", "sensor.name")` returns the strongly typed sensor defined in the given class
+* `$brooklyn:object(Map)` creates an object, using keys `type` to define the java type,
```
aledsage: Longer term, we should include an example here. Folk not familiar with yaml will likely struggle with what `object(Map)` means.

core/src/main/java/brooklyn/config/internal/AbstractConfigMapImpl.java
```diff
@@ -0,0 +1,110 @@
...
+            // don't do coercion on put for these, if the key type is compatible, 
+            // because that will force resolution deeply
+            val = v;
+        } else {
+            try {
+                // try to coerce on input, to detect errors sooner
+                val = TypeCoercions.coerce(v, key.getTypeToken());
+            } catch (Exception e) {
+                throw new IllegalArgumentException("Cannot coerce or set "+v+" to "+key, e);
+                // if can't coerce, we could just log, and *throw* the error when we retrieve the config
```
aledsage: Include a `TODO` in the comment for things that are to be changed in the future. Can you include in the comment why for now we are failing fast, so when would someone want to change this.

core/src/main/java/brooklyn/config/internal/AbstractConfigMapImpl.java
```diff
@@ -0,0 +1,110 @@
...
+    public <T> T getConfig(HasConfigKey<T> key, T defaultValue) {
+        return getConfig(key.getConfigKey(), defaultValue);
+    }
+    
+    @Override @Deprecated
+    public Object getRawConfig(ConfigKey<?> key) {
+        return getConfigRaw(key, true).orNull();
+    }
+    
+    protected Object setConfigPrep1(ConfigKey<?> key, Object v) {
```
aledsage: Is there a better name for this, such as `coerceConfigVal(...)` perhaps?

core/src/main/java/brooklyn/event/basic/MapConfigKey.java
```diff
@@ -125,6 +130,20 @@ protected Object applyEntryValueToMap(Entry value, Map target) {
...
+            // supplier or other unexpected value
+            if (k instanceof Supplier) {
+                // TODO not thread-safe
+                Object mapAtRoot = target.get(this);
+                if (mapAtRoot==null) {
+                    mapAtRoot = new LinkedHashMap();
+                    target.put(this, mapAtRoot);
+                }
+                if (mapAtRoot instanceof Map) {
+                    synchronized (mapAtRoot) {
```
aledsage: Not convinced we should just synchronize on this map, without knowing anything about it. That feels dangerous and ad hoc. We can presumably rely on the map's put method to be synchronized if the map really needs to be synchronized. What do we expect this map (returned by `target.get(this)`) to be?

ahgittin: often the map will be what we `put` just above -- ie an LHMap, and sometimes it will come from another source, e.g. yaml parse (also often an LHMap).

in most cases synching on the map is the standard way to synchronize.  it's not foolproof of course.  but feels likely to cause a problem than a CME on an LHMap.

will add a comment to that effect.

docs/guide/yaml/yaml-reference.md
ahgittin: agree with these, esp for `attributeWhenReady` -- have fixed text and added explicit example for `component...attributeWhenReady`.  (`config` in contrast in my experience is mostly used for local refs.)

docs/guide/yaml/yaml-reference.md
ahgittin: agree -- have inserted TODO for both of these

core/src/main/java/brooklyn/config/internal/AbstractConfigMapImpl.java
ahgittin: +1

core/src/main/java/brooklyn/config/internal/AbstractConfigMapImpl.java
ahgittin: have elaborated on comment.  i don't think it's a TODO, but rather a reminder if someone encounters this exception and thinks we've made the wrong call here.

ahgittin:
great comments @aledsage - think i've addressed them all

core/src/main/java/brooklyn/enricher/basic/Joiner.java
```diff
@@ -0,0 +1,128 @@
...
+    @SetFromFlag("separator")
+    public static ConfigKey<String> SEPARATOR = ConfigKeys.newStringConfigKey("enricher.joiner.separator", "Separator string to insert between each argument", ",");
+    @SetFromFlag("quote")
+    public static ConfigKey<Boolean> QUOTE = ConfigKeys.newBooleanConfigKey("enricher.joiner.quote", "Whether to bash-escape each parameter and wrap in double-quotes, defaulting to true", true);
+    @SetFromFlag("minimum")
+    public static ConfigKey<Integer> MINIMUM = ConfigKeys.newIntegerConfigKey("enricher.joiner.minimum", "Minimum number of elements to join; if fewer than this, sets null; default 0 (no minimum)");
+    @SetFromFlag("maximum")
+    public static ConfigKey<Integer> MAXIMUM = ConfigKeys.newIntegerConfigKey("enricher.joiner.maximum", "Maximum number of elements to join; default null means all elements always taken");
+    
+//    protected Function<? super SensorEvent<T>, ? extends U> transformation;
```
aledsage: Delete commented out code, or include a TODO comment to say why it's useful.

core/src/main/java/brooklyn/enricher/basic/Transformer.java
```diff
@@ -52,7 +57,7 @@
 
     public static ConfigKey<Sensor<?>> TARGET_SENSOR = ConfigKeys.newConfigKey(new TypeToken<Sensor<?>>() {}, "enricher.targetSensor");
     
-    protected Function<? super SensorEvent<T>, ? extends U> transformation;
+//    protected Function<? super SensorEvent<T>, ? extends U> transformation;
```
aledsage: Delete commented out code, or include a TODO comment to say why it's useful.

core/src/main/java/brooklyn/enricher/basic/Transformer.java
```diff
@@ -102,13 +95,65 @@ public void setEntity(EntityLocal entity) {
         }
     }
 
+    /** returns a function for transformation, for immediate use only (not for caching, as it may change) */
+    @SuppressWarnings("unchecked")
+    protected Function<SensorEvent<T>, U> getTransformation() {
```
aledsage: Why might the result of `getTransformation()` change from one call to the other? Is it not based on the config values? If so and if we're catering for config values changing, then we could override `doReconfigureConfig(...)`.

core/src/test/java/brooklyn/enricher/EnrichersTest.java
```diff
@@ -431,4 +435,45 @@ protected void doUpdatingMapChecks(AttributeSensor mapSensor) {
...
+        EntityTestUtils.assertAttributeEqualsEventually(entity, TestEntity.NAME, "");
+        
+        // null causes null
+        entity.setAttribute(LIST_SENSOR, null);
+        EntityTestUtils.assertAttributeEqualsEventually(entity, TestEntity.NAME, null);
+    }
+
+    
+    @Test
+    public void testJoinerUnquoted() {
```
aledsage: Need more tests (e.g. that minimum/maximum are respected - when list length is increasing and when it's decreasing, etc).

docs/guide/yaml/example_yaml/test-app-with-enrichers-slightly-simpler.yaml
```diff
@@ -0,0 +1,57 @@
+#
+# example showing how enrichers can be set
```
aledsage: I'd prefer a file name that is more descriptive (particularly for when someone is deciding what yaml to look at). It's unclear from the name whether this is simpler use-cases of enrichers or simple yaml format for achieving the same thing with enrichers.

Assuming the latter, if there are old files with the old way, I'd prefer us to rename those yaml files to "legacy" or some such. If there aren't old files, then why call this "slightly-simpler"?

docs/guide/yaml/yaml-reference.md
aledsage: Going through the rest of the PR, I see several examples of `brooklyn:sensor(...)` from enrichers/policies. I hadn't used those much before, so now see that accessing "this" entity is more common than I thought.

However, still happy for it to show the `component("ID")` first in the docs.

docs/guide/yaml/example_yaml/test-app-with-enrichers-slightly-simpler.yaml
```diff
@@ -0,0 +1,57 @@
...
+    # pick one uri as the main one to use
+    brooklyn.config:
+      enricher.sourceSensor: $brooklyn:sensor("urls.list")
+      enricher.targetSensor: $brooklyn:sensor("brooklyn.entity.basic.Attributes", "main.uri")
+      quote: false
+      maximum: 1
+brooklyn.enrichers:
+- type: brooklyn.enricher.basic.Propagator
+  # if nothing specified for `propagating` or `sensorMapping` then 
+  # Propagator will do all but the usual lifecycle defaults, handy at the root!
```
aledsage: Slightly hesitant about this being the default? I can see why people would want that, but it's probably a bit surprising. 

It's also a bit surprising that there's not an explicit config key corresponding to this default (so that one can be explicit about behaviour, rather than relying on implicit defaults).

My gut feel was that you'd need (or at least be able to specify) a config key to indicate intent - e.g. "propagateAllButGenericHealth". But obviously I dislike that name; not sure what a better name would be.

docs/guide/yaml/example_yaml/test-app-with-enrichers-slightly-simpler.yaml
```diff
@@ -0,0 +1,57 @@
...
+      uniqueTag: urls.list.comma_separated.max_2
+  - type: brooklyn.enricher.basic.Joiner
+    # pick one uri as the main one to use
+    brooklyn.config:
+      enricher.sourceSensor: $brooklyn:sensor("urls.list")
+      enricher.targetSensor: $brooklyn:sensor("brooklyn.entity.basic.Attributes", "main.uri")
+      quote: false
+      maximum: 1
+brooklyn.enrichers:
+- type: brooklyn.enricher.basic.Propagator
```
aledsage: In the same way as we have short-hand for "Autoscaler" not needing to be fully qualified, we should probably have short-hand for these ones. The argument being that these are the things explicitly on `Enrichers.builder()`.

docs/guide/yaml/yaml-reference.md
```diff
@@ -37,7 +37,10 @@ the entity being defined, with these being the most common:
 
 * `brooklyn.policies`: a list of policies, each as a map described with their `type` and their `brooklyn.config` as keys
 
-* `brooklyn.enrichers`: a list of enrichers, each as a map described with their `type` and their `brooklyn.config` as keys
+* `brooklyn.enrichers`: a list of enrichers, each as a map described with their `type` and their `brooklyn.config` as keys;
+  see the keys declared on individual enrichers; 
+  also see [this enricher example](example_yaml/test-app-with-enrichers-slightly-simpler.yaml) for a detailed and commented illustration
```
aledsage: Would definitely like a different name - the "slightly simpler" should just be "normal" in the eyes of the user, so is redundant and confusing in the name.

docs/guide/yaml/example_yaml/test-app-with-enrichers-slightly-simpler.yaml
```diff
@@ -0,0 +1,57 @@
...
+    $brooklyn:entitySpec:
+      type: brooklyn.test.entity.TestEntity
+      brooklyn.enrichers:
+      - type: brooklyn.enricher.basic.Transformer
+        # transform "ip" (which we expect a feed, not shown here, to set) to a URL;
+        # you can curl an address string to the sensors/ip endpoint an entity to trigger these enrichers 
+        brooklyn.config:
+          enricher.sourceSensor: $brooklyn:sensor("ip")
+          enricher.targetSensor: $brooklyn:sensor("url")
+          enricher.targetValue: $brooklyn:formatString("http://%s/", $brooklyn:attributeWhenReady("ip"))
```
aledsage: Great examples, and great that we can now support all of these. I like the yaml here.

Rambling thoughts follow, not pertinent to this PR!

I wonder about things like `$brooklyn:sensor("brooklyn.entity.basic.Attributes", "main.uri")`. As long as people hardly ever write it, then I'm ok. Otherwise, need to think of the trade-off for letting them just create a new untyped sensor with the same name, versus it finding the statically defined sensor. The untyped obviously wouldn't have the description and type metadata, but would be simpler in the yaml.

core/src/main/java/brooklyn/event/basic/MapConfigKey.java
```diff
@@ -125,6 +131,25 @@ protected Object applyEntryValueToMap(Entry value, Map target) {
...
+                    mapAtRoot = new LinkedHashMap();
+                    target.put(this, mapAtRoot);
+                }
+                // TODO above is not thread-safe, and below is assuming synching on map 
+                // is the best way to prevent CME's, which is often but not always true
+                if (mapAtRoot instanceof Map) {
+                    if (mapAtRoot instanceof ConcurrentMap) {
+                        return ((Map)mapAtRoot).put(k, value.getValue());
+                    } else {
+                        synchronized (mapAtRoot) {
```
aledsage: Or the supplier of the map might have done `Collections.synchronizedMap(...)`. I'm ok with leaving this synchornized in here, but think that we should never be relying on it - if synchronization is important, then the map should have been created with that in mind (e.g. `Collections.synchronizedMap(...)` means that you never need to synchronize explicitly except for multi-call actions such as iterating).

aledsage:
Finished reviewing; looks really good. Happy for you to merge once you've looked through the rest of my comments.

core/src/main/java/brooklyn/enricher/basic/Transformer.java
ahgittin: it wouldn't change, but this design makes it easier to support changing config in future. if it's an efficiency hole we can switch to populate this at start.  will add a comment to this effect.

docs/guide/yaml/example_yaml/test-app-with-enrichers-slightly-simpler.yaml
ahgittin: a refactoring of these files would be nice.  let's wait until we have a massively simpler syntax however.

docs/guide/yaml/example_yaml/test-app-with-enrichers-slightly-simpler.yaml
ahgittin: it's what i would expect as the default (the only reasonable option).  however if we think of a better name i'm okay with marking this beta.

docs/guide/yaml/example_yaml/test-app-with-enrichers-slightly-simpler.yaml
ahgittin: agree.  not in this PR though, part of massively simpler enricher yaml.  :)

could we use a default catalog for this (and need to support enrichers there) ?

PR #596, opened 2015-04-13
fix DynamicFabricImpl so it accepts locations in yaml instead of as parameter to start

ahgittin:


aledsage:
Looks good; merging.

PR #600, opened 2015-04-15
allow dsl commands to get an entity by type

robertgmoss:
e.g $brooklyn:descendant("brooklyn.entity.proxy.nginx.NginxController")
will return the first descendent entity that has that type

### DIFF
--- a/core/src/main/java/brooklyn/entity/basic/EntityPredicates.java
+++ b/core/src/main/java/brooklyn/entity/basic/EntityPredicates.java
@@ -157,20 +157,20 @@ public class EntityPredicates {
         }
     }
     
-    public static Predicate<Entity> classNameEquals(final String name) {
-        return classNameSatisfies(Predicates.equalTo(name));
+    public static Predicate<Entity> typeNameEquals(final String name) {
+        return typeNameSatisfies(Predicates.equalTo(name));
     }
 
-    public static Predicate<Entity> classNameSatisfies(
+    public static Predicate<Entity> typeNameSatisfies(
             final Predicate<? super String> condition) {
-        return new ClassNameSatisfies(condition);
+        return new TypeNameSatisfies(condition);
     }
 
-    protected static class ClassNameSatisfies implements
+    protected static class TypeNameSatisfies implements
             SerializablePredicate<Entity> {
         protected final Predicate<? super String> condition;
 
-        protected ClassNameSatisfies(Predicate<? super String> condition) {
+        protected TypeNameSatisfies(Predicate<? super String> condition) {
             this.condition = condition;
         }
 
@@ -183,7 +183,7 @@ public class EntityPredicates {
 
         @Override
         public String toString() {
-            return "classNameSatisfies(" + condition + ")";
+            return "typeNameSatisfies(" + condition + ")";
         }
     }
 
--- a/core/src/main/java/brooklyn/entity/basic/EntityPredicates.java
+++ b/core/src/main/java/brooklyn/entity/basic/EntityPredicates.java
@@ -156,6 +156,36 @@ public class EntityPredicates {
             return "applicationIdSatisfies("+condition+")";
         }
     }
+    
+    public static Predicate<Entity> classNameEquals(final String name) {
+        return classNameSatisfies(Predicates.equalTo(name));
+    }
+
+    public static Predicate<Entity> classNameSatisfies(
+            final Predicate<? super String> condition) {
+        return new ClassNameSatisfies(condition);
+    }
+
+    protected static class ClassNameSatisfies implements
+            SerializablePredicate<Entity> {
+        protected final Predicate<? super String> condition;
+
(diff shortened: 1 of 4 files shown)


core/src/main/java/brooklyn/entity/basic/EntityPredicates.java
```diff
@@ -156,6 +156,36 @@ public String toString() {
             return "applicationIdSatisfies("+condition+")";
         }
     }
+    
+    public static Predicate<Entity> classNameEquals(final String name) {
```
aledsage: Perhaps should be `typeNameEquals` to better match `input.getEntityType().getName()`, which is subsequently called. This (normally?) returns the entity's interface name (rather than the implementation class name), which is another reason `typeNameEquals` would be clearer.

usage/camp/src/main/java/io/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -122,14 +124,15 @@ public Entity call() throws Exception {
                 default:
                     throw new IllegalStateException("Unexpected scope "+scope);
             }
-            
-            Optional<Entity> result = Iterables.tryFind(entitiesToSearch, EntityPredicates.configEqualTo(BrooklynCampConstants.PLAN_ID, componentId));
+            Predicate<Entity> configEqualTo = EntityPredicates.configEqualTo(BrooklynCampConstants.PLAN_ID, componentId);
+            Predicate<Entity> classNameEquals = EntityPredicates.classNameEquals(componentId);
+            Optional<Entity> result = Iterables.tryFind(entitiesToSearch, Predicates.or(configEqualTo, classNameEquals));
```
aledsage: If the string matches both an id and a class name, then would be good to guarantee it will first match on id rather than it being dependent on the order that it walks through the entities. I'd therefore go for the more verbose:

    Optional<Entity> result = Iterables.tryFind(entitiesToSearch, configEqualTo);
    if (!result.isPresent()) result = Iterables.tryFind(entitiesToSearch, classNameEquals);

aledsage:
@robertgmoss looks good; only a couple of minor comments with code.

I have a slight hesitation for overloading the meaning of `$brooklyn:descendant("abc")` so that "abc" is interpreted as meaning multiple possible things. However, I see that it is certainly useful and is the simplest way for the YAML author so agree with you.

robertgmoss:
Oops have commited some WIP.  If merging, don't include 56c816e.

usage/camp/src/main/java/io/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -122,14 +124,16 @@ public Entity call() throws Exception {
...
             }
+            Predicate<Entity> configEqualTo = EntityPredicates.configEqualTo(BrooklynCampConstants.PLAN_ID, componentId);
+            Optional<Entity> result = Iterables.tryFind(entitiesToSearch, configEqualTo);
+            if (result.isPresent()) return result.get();
             
-            Optional<Entity> result = Iterables.tryFind(entitiesToSearch, EntityPredicates.configEqualTo(BrooklynCampConstants.PLAN_ID, componentId));
-            
-            if (result.isPresent())
-                return result.get();
+            Predicate<Entity> typeNameEquals = EntityPredicates.typeNameEquals(componentId);
```
aledsage: Would be good to have a unit test for this. For example, see `EntitiesYamlTest.testComponent`. We should also have a test for the id lookup approach with `$brooklyn:descendant("myid")`.

ahgittin:
neat idea -- but some concerns

share concern about overloading `descendant(type)` vs `id`, esp as type is likely to have many more matches.

i'm also concerned that `type` is too specific here, it two ways.  we don't check against java supertypes (it's just a string match), and if we are using catalog items which extend other catalog items and java types we don't match those.

i wonder if there is a better way to achieve what this is trying to achieve.  what is the use case?

(also we should update docs, yaml-reference.md at least, when introducing such features)

robertgmoss:
The original use case was:

    - type: brooklyn.entity.webapp.ControlledDynamicWebAppCluster
      brooklyn.config:
        subnet.publiclyForwardedPorts:
        - $brooklyn:descendant("brooklyn.entity.proxy.nginx.NginxController").attributeWhenReady("proxy.http.port")

although this turned out not to be the solution to that particular problem.  But it seems useful in general.

ahgittin:
agree there are times that it's useful but i worry it is an anti-pattern, esp given the ambiguity around types.

we could solve this situation instead by having CDWAC assign an id to the controller, e.g. `controller`, and then using `descendant("controller")`.  (might need a little bit of work but i think currently `id` from camp yaml becomes `camp.id` and we will match that at runtime so it shouldn't be hard.)

i think that's cleaner.  @robertgmoss @aledsage wdyt?

ahgittin:
@robertgmoss close this?

PR #601, opened 2015-04-15
Adds support for ${attribute['attribute.name']} sensor access in freemarker templates

nakomis:


core/src/main/java/brooklyn/util/text/TemplateProcessor.java
```diff
@@ -225,6 +229,40 @@ public String toString() {
...
+        }
+
+        @Override
+        public TemplateModel get(String key) throws TemplateModelException {
+            Object result;
+            try {
+                result = Entities.submit(entity, DependentConfiguration.attributeWhenReady(entity,
+                        Sensors.builder(Object.class, key).persistence(AttributeSensor.SensorPersistenceMode.NONE).build())).get();
+            } catch (Exception e) {
+                throw Exceptions.propagate(e);
```
grkvlt: Do `throw new TemplateModelException("Error obtaining sensor data for " + key, e)` here instead

nakomis:
@grkvlt Any other comments, or should I go ahead and make the change?

nakomis:
PR comments addressed

grkvlt:
@nakomis the other issues (sorry, I forgot to mention them before) is that the `attributeWhenReady` call may block for ever, so perhaps we should add a timeout to the `get()`? Also I'd like to see a test validating what happens when the attribute is _not_ present at all, and perhaps also what happens when the attribute is present but `null`.

ahgittin:
i think `${attribute[...]}` should return the current attribute value, whether or not set.  it would be surprising if that blocked.  we could have an analogous `${attributeWhenReady[...]}` which does wait.

no timeout necessary -- tasks should be interrupted at a higher level.

agree with a few more tests for null and blank cases.

this is a great addition btw!

ahgittin:
actually thinking about it we can always set a config key to attribute or attribute when ready and then use `${config[...]}`.  so this addition isn't necessary -- though i'm happy to have it.

ahgittin:
@nakomis close this?

neykov:
I like the change - will alleviate the need for the tons of getXXX accessors on the drivers created only because the attributes can't be accessed in templates. All of them just do `entity.getAttribute(...)` so I think having the default block is sensible. Adding a non-blocking alternative could be done as well, but I haven't seen any of those accessors check for present value.

PR #602, opened 2015-04-16
Insert LICENSE/NOTICE/DISCLAIMER for binary build and Maven artifacts, plus documentation

rdowner:
Adds documentation of how to write LICENSE and NOTICE to the developer's guide.

Adds LICENSE, NOTICE and DISCLAIMER tailored for the binary distribution.

Removes unnecessary shaded jars.

Fixes LICENSE, NOTICE and DISCLAIMER for three Maven modules, and for the remaining shaded jar, which don't use the unmodified boilerplate ASL2 files


rdowner:
Failure is: `Port already in use: 40127`
So I suspect this is a false alarm, but testing locally to confirm.

rdowner:
Unable to reproduce test failure - assuming it's a test environment transient error.

usage/dist/src/main/license/LICENSE
```diff
@@ -0,0 +1,1691 @@
...
+-----------------------------------------------------------------------------
+
+com.fasterxml.jackson.core-jackson-annotations-2.4.0.jar
+com.fasterxml.jackson.core-jackson-core-2.4.2.jar
+com.fasterxml.jackson.core-jackson-databind-2.4.2.jar
+com.fasterxml.jackson.jaxrs-jackson-jaxrs-base-2.4.2.jar
+com.fasterxml.jackson.jaxrs-jackson-jaxrs-json-provider-2.4.2.jar
+com.fasterxml.jackson.module-jackson-module-jaxb-annotations-2.4.2.jar
+
+Licensed under the Apache License v2.0
```
aledsage: Why not list these under the previously "apache license" section? Are these different copyright? Do we need to say so?

usage/dist/src/main/license/LICENSE
```diff
@@ -0,0 +1,1691 @@
...
+
+Licensed under the Apache License, Version 2.0
+
+-----------------------------------------------------------------------------
+
+xml-apis-xml-apis-1.0.b2.jar
+
+Composed of materials with the following licenses:
+
+----------
```
aledsage: (Pedant) Different number of dashes - good to be consistent with above?

aledsage: Ah, I see - it is the licenses of xml-apis. Will delete comment.

usage/dist/src/main/license/LICENSE
```diff
@@ -0,0 +1,1691 @@
...
+  Stefan Haustein
+  Aleksander Slominski
+
+-----------------------------------------------------------------------------
+
+xpp3-xpp3_min-1.1.4c.jar
+
+Indiana University Extreme! Lab Software License
+
+Version 1.1.1
```
aledsage: Why does this say 1.1.1, when the jar says 1.1.4c?

usage/dist/src/main/license/LICENSE
rdowner: They are grouped by the organisation that created them. Some organisations include a Copyright statement (like jcommander in this example) so it is convenient to group by organisation. Not all do (like fasterxml in this example).

No strong opinion on this one; I can group by license instead of organisation if that is preferred.

usage/dist/src/main/license/LICENSE
rdowner: I'm guessing that 1.1.1 refers to the version of the license, which is almost always versioned differently to the software (think GPLv2 and GPLv3)

aledsage:
@rdowner Excellent, great work! Looks really good to me. Just a couple of really minor questions.

If the mentors don't get a chance to review soon, then I suggest we merge and then they can review a 0.7.0-M3 release candidate.

usage/dist/src/main/license/LICENSE
ahgittin: This grouping is sensible to me.  If we have a lot of singletons without special copyright then maybe change.

ahgittin:
Great work.  I think @rdowner you've addressed all @aledsage 's comments.  Merging.

PR #603, opened 2015-04-17
Switch to Tomcat for ControlledDynamicWebAppCluster

aledsage:
- Was previously a cluster of AS7 nodes; now Tomcat
- The reasons are:
  - Tomcat is a popular Apache project; Brooklyn is also in Apache
  - AS7 doesn’t support Java 8


grkvlt:
Great, will merge. Thanks @aledsage :frog:

PR #605, opened 2015-04-18
simplify yaml for averaging + summing enricher, plus test

ahgittin:


ggrunin:
tested #605  - works well.

core/src/main/java/brooklyn/enricher/Enrichers.java
```diff
@@ -216,13 +219,25 @@ public B fromHardcodedProducers(Iterable<? extends Entity> val) {
...
+        @SuppressWarnings({ "unchecked", "rawtypes" })
         public B computing(Function<? super Collection<S>, ? extends T> val) {
-            this.computing = checkNotNull(val);
+            this.computingSupplier = (Supplier)Suppliers.ofInstance(checkNotNull(val));
             return self();
         }
-        @SuppressWarnings({ "unchecked", "rawtypes" })
         public B computingSum() {
-            // relies of TypeCoercion of result from Number to T, and type erasure for us to get away with it!
+            this.computingSupplier = new Supplier<Function<? super Collection<S>, ? extends T>>() {
```
neykov: Better to use named classes to avoid rebind issues.

ahgittin: The `Supplier` is local to the builder, resolved before the spec is built, so I think it's fine.  The `Function` itself is a static class.

neykov: Oh I see it now, shouldn't trust github's enclosing scope method signature.

neykov:
Looks good to merge.

PR #606, opened 2015-04-18
Store additional config in the riak.conf

bostko:
- additional config moved from the RiakEnterpriseNode
- bump up default Riak version

### DIFF
--- a/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
+++ b/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
@@ -56,7 +56,6 @@ import brooklyn.util.text.Strings;
 
 import com.google.common.base.Joiner;
 import com.google.common.base.Optional;
-import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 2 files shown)


software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -251,6 +251,15 @@ public void customize() {
             String saveAsRiakConf = Urls.mergePaths(getRunDir(), "riak.conf");
             DynamicTasks.queueIfPossible(SshEffectorTasks.put(saveAsRiakConf).contents(riakConfTemplate));
             commands.add(sudo("mv " + saveAsRiakConf + " " + getRiakEtcDir()));
+
+            if(!osDetails.isMac()) {
```
neykov: Why leave out additional config for Mac?

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -251,6 +251,15 @@ public void customize() {
...
+                if(!com.google.common.base.Strings.isNullOrEmpty(entity.getConfig(RiakNode.RIAK_CONFIG_FILE))) {
```
neykov: Isn't it better to leave the default to be empty string and check for non-null only here? Any reason empty string should be treated specially?

bostko: There is no point in doing this task if the string is empty

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -251,6 +251,15 @@ public void customize() {
...
+                    String configTemplate = processTemplateContents(entity.getConfig(RiakNode.RIAK_CONFIG_FILE));
+                    String saveAsConfig = Urls.mergePaths(getRunDir(), "riak_extra.conf");
+                    DynamicTasks.queueIfPossible(SshEffectorTasks.put(saveAsConfig).contents(configTemplate));
+                    commands.add(sudo("cat " + saveAsConfig + " >> " + getRiakEtcDir() + "/riak.conf"));
```
neykov: `customize` step is called on each start, which will lead to multiple copies of the config appended.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
bostko: I see that there is RIAK_CONF_TEMPLATE_URL_MAC and RIAK_CONF_TEMPLATE_URL_LINUX .
I am not familiar with differences between Riak conf on Mac and Linux but I guess your are right and we can use RIAK_CONFIG_FILE for both.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -251,6 +251,15 @@ public void customize() {
             String saveAsRiakConf = Urls.mergePaths(getRunDir(), "riak.conf");
             DynamicTasks.queueIfPossible(SshEffectorTasks.put(saveAsRiakConf).contents(riakConfTemplate));
             commands.add(sudo("mv " + saveAsRiakConf + " " + getRiakEtcDir()));
+
+            if(!osDetails.isMac()) {
+                if(!com.google.common.base.Strings.isNullOrEmpty(entity.getConfig(RiakNode.RIAK_CONFIG_FILE))) {
+                    String configTemplate = processTemplateContents(entity.getConfig(RiakNode.RIAK_CONFIG_FILE));
+                    String saveAsConfig = Urls.mergePaths(getRunDir(), "riak_extra.conf");
+                    DynamicTasks.queueIfPossible(SshEffectorTasks.put(saveAsConfig).contents(configTemplate));
```
neykov: Driver methods are executed in effector context so it's safe to use DynamicTasks.queue (somebody correct me if I am wrong?). At the very least you should follow `queueIfPossible` with `orSubmitAsync` to be sure the task is executed.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
neykov: Oh I see the config is uploaded above, so should be safe to cat it here.

neykov: But note that if `isVersion1()` then the config file is called `app.config`. I think it's better to prepare the config once and then push it to the machine, will save you from needing to execute bash commands to append.

neykov:
Never mind the previous two comments, I was looking at a single commit only. Looks good now. Squash the commits so the history is cleaner, then can be merged.

ahgittin:
LGTM

PR #607, opened 2015-04-21
Non-blocking jetty connector for https

neykov:
Use non-blocking HTTPS connector, HTTP case was already using non-blocking.
HTTPS connector wasn't setting the bind address - something to have in mind as could lead to unexpected behaviour changes in existing installations.


ahgittin:
good cleanup, merging

PR #609, opened 2015-04-22
Coerce aggregated sensor values

neykov:
Also ignore SERVICE_UP failures from non-RUNNING children. Avoids setting the parent on fire while spinning up a child. /cc @ahgittin as he was the initial contributor of the SERVICE_UP quorum logic.

### DIFF
--- a/core/src/test/java/brooklyn/entity/basic/ServiceStateLogicTest.java
+++ b/core/src/test/java/brooklyn/entity/basic/ServiceStateLogicTest.java
@@ -24,9 +24,7 @@ import org.testng.annotations.Test;
 
 import brooklyn.entity.BrooklynAppUnitTestSupport;
 import brooklyn.entity.Entity;
-import brooklyn.entity.Group;
 import brooklyn.entity.basic.EntitySubscriptionTest.RecordingSensorEventListener;
-import brooklyn.entity.basic.QuorumCheck.QuorumChecks;
 import brooklyn.entity.basic.ServiceStateLogic.ComputeServiceIndicatorsFromChildrenAndMembers;
 import brooklyn.entity.basic.ServiceStateLogic.ServiceNotUpLogic;
 import brooklyn.entity.basic.ServiceStateLogic.ServiceProblemsLogic;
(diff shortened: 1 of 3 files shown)


core/src/test/java/brooklyn/entity/basic/ServiceStateLogicTest.java
```diff
@@ -240,7 +254,33 @@ public void testManuallySettingIndicatorsOnApplications() throws Exception {
...
+    @Test
+    public void testQuorumWithStringStates() {
+        final DynamicCluster cluster = app.createAndManageChild(EntitySpec.create(DynamicCluster.class)
+                .configure(DynamicCluster.MEMBER_SPEC, EntitySpec.create(TestEntityWithoutEnrichers.class))
+                .configure(DynamicCluster.INITIAL_SIZE, 1));
+
+        RecordingSensorEventListener r = new RecordingSensorEventListener();
+        app.subscribe(cluster, Attributes.SERVICE_STATE_ACTUAL, r);
+
+        cluster.start(ImmutableList.of(new SimulatedLocation()));
```
aledsage: There is an `app.newSimulatedLocation()`, which uses the `LocationSpec` and `LocationManager` to instantiate the location.

core/src/test/java/brooklyn/entity/basic/ServiceStateLogicTest.java
```diff
@@ -240,7 +254,33 @@ public void testManuallySettingIndicatorsOnApplications() throws Exception {
...
     }
-        
+
+    @Test
+    public void testQuorumWithStringStates() {
+        final DynamicCluster cluster = app.createAndManageChild(EntitySpec.create(DynamicCluster.class)
+                .configure(DynamicCluster.MEMBER_SPEC, EntitySpec.create(TestEntityWithoutEnrichers.class))
+                .configure(DynamicCluster.INITIAL_SIZE, 1));
+
+        RecordingSensorEventListener r = new RecordingSensorEventListener();
```
aledsage: Is this listener ever used (beyond registering it)?

aledsage:
Looks really good.

Am a tiny bit concerned about us ignoring entity's serviceUp when serviceState is null (as that is a change in behaviour for a situation where the expected behaviour is unclear. However, on balance I'm fine with this.

Merging.

PR #610, opened 2015-04-22
temporary local copy of fix for HVM/paravirtual issue on AWS

ahgittin:
fixes https://issues.apache.org/jira/browse/BROOKLYN-140

as noted, when new jclouds lands we should revert this

@duncangrant can you let us know if this fixes your problem?

### DIFF
--- a/locations/jclouds/src/main/java/org/jclouds/ec2/compute/domain/EC2HardwareBuilder.java
+++ b/locations/jclouds/src/main/java/org/jclouds/ec2/compute/domain/EC2HardwareBuilder.java
@@ -1,0 +1,860 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.ec2.compute.domain;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Predicates.not;
+import static org.jclouds.compute.domain.Volume.Type.LOCAL;
+import static org.jclouds.compute.predicates.ImagePredicates.any;
+import static org.jclouds.compute.predicates.ImagePredicates.idIn;
+
+import java.net.URI;
+import java.util.List;
+import java.util.Map;
+
+import org.jclouds.compute.domain.Hardware;
+import org.jclouds.compute.domain.HardwareBuilder;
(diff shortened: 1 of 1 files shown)


ahgittin:
i've confirmed with this PR i can create an instance with `hardwareId: r3.large` which is one of the types which does only hvm.

note that this may introduce some changed default images, with more changed default images landing with the next jclouds (when new generation instance types are preferred). /cc @aledsage

aledsage:
Build fails with:

    [INFO] There are 1 checkstyle errors.
    [ERROR] EC2HardwareBuilder.java[686:1] (whitespace) FileTabCharacter: Line contains a tab character.

aledsage:
This fix is non-deterministic. There is no guarantee of classpath order, for whether the jclouds class or the replacement class will be first on the classpath. It would need to be created as a patch, and added to the `lib/patch/` dir to guarantee classpath ordering.

ahgittin:
isn't it going to be loaded either in alphabetical order (release) or maven order (ide) ?

so i think it *is* deterministic.  slightly naughty but pragmatic, for a temporary fix.  (haven't we used this trick before for overriding jclouds items (gce in particular)?  the only times problems i'm aware of problems is when the overriding jar filename comes first alphabetically.)

ahgittin:
btw - fixed checkstyle (tab from jclouds project).  also have confirmation from @duncangrant that this fixes his problem case.

neykov:
Good point by @aledsage , agree with him. We already had an issue with jars ordered differently - one of the jars was shadowing the others, but only if it was before them in the classpath.

aledsage:
For jar ordering, it is deterministic on a given platform+jvm but different platforms seem to order it differently. There was a fix done like this one before that worked fine locally but then failed when Brooklyn was run on (if I recall correctly) ubuntu with OpenJDK.

This relates to the distro, with the classpath set to include `lib/brooklyn/*.jar`. Ordering in IDEs probably behave completely differently again, as might maven command line builds.

ahgittin:
@aledsage should we add a `lib/deps/` dir and order the `dist` classpath so that `lib/brooklyn/*.jar` is first?  that seems the easiest way to solve issues with the dist.  i'd rather not introduce a patch projects.  or as most environments seem to work we could just leave it as is and accept that the aws `hvm` error might remain an issue in some cases until we have a new jclouds cut.

neykov:
Any way we can make the patch work transparently for downstream projects as well? Can't see how it could be approached.

Having `lib/deps` would be fine for Brooklyn, but downstream projects will face the same problem again, because `brooklyn-location-jclouds` will be mixed with the other dependencies.
I like the idea of a separate patches project, with each downstream making sure it puts it in its patches folder. This won't be the last time there is need to override a dependency, so why not go all the way and make it robust. It's a one time change so next time patching will be transparent.

ahgittin:
Downstream patches could be an elegant solution but I can't help feeling we're trying to do too much.  Maintaining a patch dir downstream which grabs patches from upstream feels like a lot of work which is best avoided.  Also it won't solve any errors which occur if the downstream project is being run from the IDE.

That said I'm not convinced these errors are severe because both maven order and alpha order give us the desired result and those are the majority strategies.

The right way to support a downstream project, would be to take a tagged jclouds cut with this patch.  I'd rather invest in that if required than passing patches, I think.  (But even that I don't think is urgent.)

However happy to be overruled if @aledsage feels strongly.

ahgittin:
@aledsage bump. could we merge this before someone else hits hvm issues and if a full-blown patches project is needed we do that (or resolve it with one of the other techniques above).

ahgittin:
merging this; if we want to do a better patching strategy we can shortly

PR #612, opened 2015-04-22
Nodejs riak sample app

bostko:
- Refactoring the NodeJsWebAppSshDriver .
Install packages from package.json as non-root

### DIFF
--- a/examples/simple-web-cluster/src/main/resources/brooklyn/demo/nodejs-riak-todo.yaml
+++ b/examples/simple-web-cluster/src/main/resources/brooklyn/demo/nodejs-riak-todo.yaml
@@ -1,0 +1,46 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: "Node.JS Todo Application"
+origin: "https://github.com/amirrajan/nodejs-todo/"
+location:
+  jclouds:aws-ec2:us-west-1:
+    imageId: us-west-1/ami-c33cdd87
+services:
+- type: brooklyn.entity.nosql.riak.RiakCluster
+  initialSize: 2
+  id: mycluster
+  brooklyn.config:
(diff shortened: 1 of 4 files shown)


software/webapp/src/main/java/brooklyn/entity/webapp/nodejs/NodeJsWebAppSshDriver.java
```diff
@@ -103,6 +103,7 @@ public void install() {
                         BashCommands.installPackage("software-properties-common python-software-properties python g++ make"),
                         BashCommands.sudo("add-apt-repository ppa:chris-lea/node.js"))))
                 .add(BashCommands.installPackage(MutableMap.of("yum", "git nodejs npm", "apt", "git-core nodejs"), null))
+                .add("npm install n") // Create ~/.npm as the current user instead of root
```
neykov: I thought that global install was not recommended, should use normal when possible? Not a NodeJS expert though so might as well be wrong.

examples/simple-web-cluster/src/main/resources/brooklyn/demo/nodejs-riak-todo.yaml
```diff
@@ -0,0 +1,46 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: "Node.JS Todo Application"
+origin: "https://github.com/amirrajan/nodejs-todo/"
```
neykov: I think this is where the blueprint lives, not where the app is forked from. Not important though.

software/webapp/src/main/java/brooklyn/entity/webapp/nodejs/NodeJsWebAppSshDriver.java
neykov: If it's just to create the folder, then create it explicitly, instead of installing the package locally and globally twice.

neykov:
Looks good, minor comments only.

software/webapp/src/main/java/brooklyn/entity/webapp/nodejs/NodeJsWebAppSshDriver.java
```diff
@@ -128,12 +124,19 @@ public void customize() {
...
         } else if (Strings.isNonBlank(archiveUrl)) {
             ArchiveUtils.deploy(archiveUrl, getMachine(), getRunDir());
         } else {
             throw new IllegalStateException("At least one of Git or archive URL must be set for " + getEntity());
         }
 
+        commands.add(BashCommands.ifFileExistsElse1("package.json", "npm install"));
+        List<String> packages = getEntity().getConfig(NodeJsWebAppService.NODE_PACKAGE_LIST);
+        if (packages != null && packages.size() > 0) {
+            commands.add(BashCommands.sudo("npm install -g " + Joiner.on(' ').join(packages)));
```
bostko: What would you say to remove sudo here?

sjcorbett:
@bostko Thanks, will merge.

PR #615, opened 2015-04-28
Add Riak Solr support

bostko:


bostko:
This code is working and all of the entities are in green although I didn't test with search queries against riak.
I will be grateful if you give me some feedback.

I tried to set Solr as a child 
```java
EntitySpec.create(RiakNode.class)
        .configure(RiakNode.RIAK_CONF_ADDITIONAL_CONTENT, "search = on")
        .child(getConfig(
                SOLR_MEMBER_SPEC,
                EntitySpec.create(SolrServer.class))
                .configure(ConfigKeys.newIntegerSensorAndConfigKey("solr.http.port", "Solr HTTP communications port", 8983), 8983))
        .configure(SoftwareProcess.CHILDREN_STARTABLE_MODE, SoftwareProcess.ChildStartableMode.FOREGROUND));
```
but this breaks SolrServerImpl.getSolrPort() and it returns null. If the key is not an integer then either in Solr or in the RiakNode the port is incremented since it cannot set a port twice on one machine.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakCluster.java
```diff
@@ -54,6 +55,16 @@
...
 
+    @SetFromFlag("search")
+    ConfigKey<Boolean> SEARCH = ConfigKeys.newBooleanConfigKey(
+            "riak.search", "Deploy Solr and configure Riak to use it", false);
+
+    @SetFromFlag("memberSpec")
+    ConfigKey<EntitySpec<?>> SOLR_MEMBER_SPEC = ConfigKeys.newConfigKey(
+            new TypeToken<EntitySpec<?>>() {
+            },
+            "riak.solr.memberspec", "Solr entity spec for creating new cluster members", null);
```
neykov: No need to set null as the default value (this is the value used if no explicit one is specified).

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakClusterImpl.java
```diff
@@ -79,7 +82,7 @@ public void start(Collection<? extends Location> locations) {
 
         //FIXME: add a quorum to tolerate failed nodes before setting on fire.
         Optional<Entity> anyNode = Iterables.tryFind(getMembers(), Predicates.and(
-                Predicates.instanceOf(RiakNode.class),
+                Predicates.or(Predicates.instanceOf(RiakNode.class), Predicates.instanceOf(SameServerEntity.class)),
```
neykov: I don't see the point of this check, probably better to just remove it.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakClusterImpl.java
```diff
@@ -79,7 +82,7 @@ public void start(Collection<? extends Location> locations) {
...
                 EntityPredicates.attributeEqualTo(RiakNode.RIAK_NODE_HAS_JOINED_CLUSTER, true),
```
neykov: The entity assumes that the members are RiakNodes, thus relaying on specific behaviour from them. For example `RIAK_NODE_HAS_JOINED_CLUSTER` will be set by the node, it's not managed by the cluster. Another example is calling the effector `REMOVE_FROM_CLUSTER` on the members.

neykov:
Tested the current branch - I get NPE in Solr:
```
java.lang.NullPointerException: null value in entry: solrPort=null
	at com.google.common.collect.CollectPreconditions.checkEntryNotNull(CollectPreconditions.java:33)
	at com.google.common.collect.ImmutableMap.entryOf(ImmutableMap.java:135)
	at com.google.common.collect.ImmutableMap$Builder.put(ImmutableMap.java:206)
	at brooklyn.entity.nosql.solr.SolrServerSshDriver.getPortMap(SolrServerSshDriver.java:100)
...
```

Riak also fails (tested on Ubuntu box) because apt-get tries to install both entities at the same time.

Also the above considerations about `SameServerEntity`

I think the way to go is:
  * Make Solr child of Riak
  * Change the `SEARCH_SOLR_PORT` to a plain ConfigKey<Integer>, it shouldn't be reserving ports on the machine, it's only used for config.
  * The above could be a custom `AttributeSensorAndConfigKey<PortRange,Integer>, propagating the value from Solr to Riak, but not sure if worth the effort.

neykov:
Failing with NPE because it is still trying to reserve the port - this is done regardless of the type of the config value as the key is coming from `getEntityType().getSensors()` which are inferred from the class fields (returning key `PortAttributeSensorAndConfigKey` which tries to obtain a free port).

From the log:
```
SolrServerImpl{id=xMgHHmbI}: port 8983 not available for solr.http.port
```

I don't see how we can work around two  `PortAttributeSensorAndConfigKey` fighting for the same port, can't be overridden with config only. As I wrote above changing `SEARCH_SOLR_PORT` to a `ConfigKey<Integer>` will solve the problem.

---

When Solr is created as a child of Riak, might be better to set `SoftwareProcess.CHILDREN_STARTABLE_MODE=FOREGROUND_LATE` so the two installs don't fight over apt-get. Looks like Riak is happy to start before there's a Solr port to talk to, but this might prove to be a problem as well.

bostko:
I included non Solr specific fixes also.
@neykov Can you look at this. I think it is acceptable solution.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakClusterImpl.java
```diff
@@ -91,8 +94,16 @@ public void start(Collection<? extends Location> locations) {
     }
 
     protected EntitySpec<?> getMemberSpec() {
-        return getConfig(MEMBER_SPEC, EntitySpec.create(RiakNode.class));
-
+        if(getConfig(SEARCH)) {
+            return getConfig(
+                    MEMBER_SPEC,
+                    EntitySpec.create(SameServerEntity.class)
```
neykov: `SameServerEntity` breaks the assumptions made by the entity. See previous comments.

bostko:
I didn't pay attention that Riak installs Solr on its own http://docs.basho.com/riak/latest/ops/advanced/configs/search/#JVM-Installation
So we need just to install java and Basho's yokozuna is taking care for the Solr process.
I left the Solr ports as an integer config since it is used just for the riak.conf file, these ports do not have to be exposed.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -72,6 +72,11 @@ public RiakNodeSshDriver(final RiakNodeImpl entity, final SshMachineLocation mac
     }
 
     @Override
+    protected String getLogFileLocation() {
+        return Urls.mergePaths("/var/log/riak", "solr.log");
```
neykov: No point in using `mergePaths` in this case, just return a constant string.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -558,6 +563,13 @@ public void recoverFailedNode(String nodeName) {
         }
     }
 
+    @Override
+    public void setup() {
+        if(entity.getConfig(RiakNode.SEARCH_ENABLED)) {
```
neykov: Probably worth a comment while guard `super.setup()`

neykov:
Looking great now, minor comments only.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
```diff
@@ -118,8 +119,12 @@
     PortAttributeSensorAndConfigKey EPMD_LISTENER_PORT = new PortAttributeSensorAndConfigKey("epmdListenerPort", "Erlang Port Mapper Daemon Listener Port", "4369");
     PortAttributeSensorAndConfigKey ERLANG_PORT_RANGE_START = new PortAttributeSensorAndConfigKey("erlangPortRangeStart", "Erlang Port Range Start", "6000+");
     PortAttributeSensorAndConfigKey ERLANG_PORT_RANGE_END = new PortAttributeSensorAndConfigKey("erlangPortRangeEnd", "Erlang Port Range End", "7999+");
-    PortAttributeSensorAndConfigKey SEARCH_SOLR_PORT = new PortAttributeSensorAndConfigKey("search.solr.port", "Solr port", "8093+");
-    PortAttributeSensorAndConfigKey SEARCH_SOLR_JMX_PORT = new PortAttributeSensorAndConfigKey("search.solr.jmx_port", "Solr port", "8985+");
+
+    @SetFromFlag("searchEnabled")
+    ConfigKey<Boolean> SEARCH_ENABLED = ConfigKeys.newBooleanConfigKey("riak.search", "Deploy Solr and configure Riak to use it", false);
+
+    ConfigKey<Integer> SEARCH_SOLR_PORT = ConfigKeys.newIntegerConfigKey("search.solr.port", "Solr port", 8983);
```
aledsage: Why not use `PortAttributeSensorAndConfigKey`, out of interest? A consequence of using `ConfigKey<Integer>` is that we won't auto-detect that it is a port so won't know to open the port in AWS security groups and/or iptables etc.

Is that deliberate?

aledsage:
Looks good. Would be nice to have test(s) that show this works. Or are they already covered by the existing Riak tests?

@neykov you've reviewed more thoroughly. I believe your most recent comments were addressed. Are you happy for this to be merged now?

neykov:
Yes, comments seem to be addressed.
The point of switching to `ConfigKey` was exactly to avoid opening the port, as the Solr instance will be used only by the riak instance on the same host. 
Would be nice if we took the additional steps to make sure the port is not reachable regardless of the provider used, but it's out of the scope for the PR.

Good to merge.

bostko:
I tested with [these queries](http://docs.basho.com/riak/latest/dev/using/search) and it works correctly

sjcorbett:
@bostko thanks. Will merge.

neykov:
@bostko Could you close this.

PR #616, opened 2015-04-28
EntityPredicates.hasInterfaceMatching

sjcorbett:
And uses as optional query parameter in ApplicationApi.list.

### DIFF
--- a/core/src/main/java/brooklyn/entity/basic/EntityPredicates.java
+++ b/core/src/main/java/brooklyn/entity/basic/EntityPredicates.java
@@ -19,6 +19,7 @@
 package brooklyn.entity.basic;
 
 import java.util.Collection;
+import java.util.regex.Pattern;
 
 import javax.annotation.Nullable;
 
@@ -30,6 +31,7 @@ import brooklyn.event.AttributeSensor;
 import brooklyn.location.Location;
 import brooklyn.util.collections.CollectionFunctionals;
 import brooklyn.util.guava.SerializablePredicate;
+import brooklyn.util.javalang.Reflections;
 import brooklyn.util.text.StringPredicates;
 
 import com.google.common.base.Objects;
@@ -271,9 +273,38 @@ public class EntityPredicates {
     // ---------------------------
 
     /**
+     * @param typeRegex a regular expression
+     * @return true if any of the interfaces implemented by the entity (including those derived) match typeRegex.
+     */
+    public static Predicate<Entity> hasInterfaceMatching(String typeRegex) {
+        return new ImplementsInterface(typeRegex);
+    }
+
+    protected static class ImplementsInterface implements SerializablePredicate<Entity> {
+        protected final Pattern pattern;
+
+        public ImplementsInterface(String typeRegex) {
+            this.pattern = Pattern.compile(typeRegex);
+        }
+
+        @Override
+        public boolean apply(@Nullable Entity input) {
+            if (input == null) return false;
+            for (Class<?> cls : Reflections.getAllInterfaces(input.getClass())) {
+                if (pattern.matcher(cls.getName()).matches()) {
+                    return true;
+                }
+            }
+            return false;
+        }
+    }
+
+    // ---------------------------
+
+    /**
      * Returns a predicate that determines if a given entity is a direct child of this {@code parent}.
      */
-    public static <T> Predicate<Entity> isChildOf(final Entity parent) {
+    public static Predicate<Entity> isChildOf(final Entity parent) {
         return new IsChildOf(parent);
     }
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/brooklyn/entity/basic/EntityPredicates.java
```diff
@@ -271,9 +274,38 @@ public boolean apply(@Nullable Entity input) {
...
+        protected final Pattern pattern;
+
+        public ImplementsInterface(String typeRegex) {
+            this.pattern = Pattern.compile(typeRegex);
+        }
+
+        @Override
+        public boolean apply(@Nullable Entity input) {
+            if (input == null) return false;
+            for (Class<?> cls : ClassUtils.getAllInterfaces(input.getClass())) {
```
aledsage: We also have `brooklyn.util.javalang.Reflections.getAllInterfaces(Class)`, which does the same as in `ClassUtils`.

If we're going to keep the dependency on apache-commons lang3, then we should probably delete that brooklyn code! (Or more specifically deprecate it, and have its impl delegate to the lang3 version).

But that's not for this PR.

usage/rest-server/src/main/java/brooklyn/rest/resources/ApplicationResource.java
```diff
@@ -216,9 +222,18 @@ public JsonNode fetch(String entityIds) {
...
+
+    @Override
+    public List<ApplicationSummary> list(String typeRegex) {
+        if (Strings.isBlank(typeRegex)) {
+            typeRegex = ".*";
+        }
         return FluentIterable
                 .from(mgmt().getApplications())
                 .filter(EntitlementPredicates.isEntitled(mgmt().getEntitlementManager(), Entitlements.SEE_ENTITY))
+                .filter(EntityPredicates.hasInterfaceMatching(typeRegex))
```
aledsage: For the record, another option would be to use `EntityType.getName()`, but I prefer yours - good to be able to find everything of type for example `DynamicCluster` including sub-types.

usage/rest-server/src/main/java/brooklyn/rest/resources/ApplicationResource.java
```diff
@@ -79,9 +83,11 @@
 import brooklyn.util.collections.MutableMap;
 import brooklyn.util.exceptions.Exceptions;
 
+import com.google.common.base.Predicate;
 import com.google.common.base.Throwables;
 import com.google.common.collect.FluentIterable;
 import com.google.common.collect.Iterables;
+import com.wordnik.swagger.core.ApiParam;
```
aledsage: unused import (and same for `Predicate`)?

aledsage:
@sjcorbett  looks really good. Do you want to get rid of the unused import in `ApplicationResource` and then merge?

core/src/main/java/brooklyn/entity/basic/EntityPredicates.java
sjcorbett: commons-lang3 is coming from camp-base rather than core directly. I'll swap it for Reflections.

usage/rest-api/src/main/java/brooklyn/rest/api/ApplicationApi.java
```diff
@@ -76,6 +76,12 @@ public JsonNode fetch(
             value = "Fetch list of applications, as ApplicationSummary objects",
             responseClass = "brooklyn.rest.domain.ApplicationSummary"
     )
+    public List<ApplicationSummary> list(
+            @ApiParam(value = "Regular expression to filter by", required = false)
```
neykov: It's not clear from the description what the regex is applied on.

usage/rest-api/src/main/java/brooklyn/rest/api/ApplicationApi.java
```diff
@@ -76,6 +76,12 @@ public JsonNode fetch(
...
+            @DefaultValue(".*")
+            @QueryParam("typeRegex") String typeRegex);
+
+    /** As {@link #list(String)}, filtering for <code>.*</code>. */
     public List<ApplicationSummary> list();
```
neykov: I guess you are keeping the method for backwards compatibility with rest-client? But in this case it should have it's own set of annotations, so that client calls know what to do.

neykov:
Looks like I'm a bit late to the party :), nvm just minor comments.

usage/rest-api/src/main/java/brooklyn/rest/api/ApplicationApi.java
sjcorbett: Good point.

PR #614, opened 2015-04-27
Startup info

ahgittin:
Start webserver first, and report server up-ness and health, with jsgui now checking these in a cleaner way, discouraging access until everything is started.

Prevents issue where GUI does not load for several seconds while things are starting up.  Also prepares us to allow people to explore and correct errors in the GUI.

Also tidies up how we blocked `HaHotStateRequired` access, to know whether a rebind is still in progress.  (It is probable that server start sequence will still need to wait on initial rebind, else risk saying "up" too early -- and same for loading of the catalog; that will be tidied up shortly.)

PR #617, opened 2015-04-30
CLI commands for manipulating catalog, and cleaner catalog loading

ahgittin:
approx as discussed on mailing list, except annotations now also supported (easier to maintain compatibility, in tests)

for review only at this point; i want to do more testing


ahgittin:
BTW it's just the last 3 commits, the others are from #614

ahgittin:
spurious failure - load balancing soak test

neykov:
@ahgittin Could you rebase to remove changes from #614

ahgittin:
@neykov thanks for the comments so far - updated and rebased

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,377 @@
...
+import brooklyn.util.text.Strings;
+
+import com.google.common.annotations.Beta;
+import com.google.common.base.Function;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+
+@Beta
+public class CatalogInitialization implements ManagementContextInjectable {
```
robertgmoss: I'm concerned about the Thread safety of this class.  There is a mutex, but it is not clear what state it is guarding; all reads and writes of the state it is guarding should be guarded by this lock to ensure visibility - that is not happening, this much is clear.  Additionally, the state should be made private to prevent read/write access without the lock.

ahgittin: the mutex governs the population of the catalog, to prevent two threads from running this at the same time.  looks to me that this is functioning correctly.  pure defensive programming could lock this down further and make everything private but i think that would be overkill -- the fields and methods are already protected or package private and the class marked `@Beta` which is to say use caution when extending.  that's the right level of defensiveness for something which is evolving IMHO.

robertgmoss: It's the boolean variables that are concerning me.  Sometimes they are updated/read under the lock, sometimes they are not.

api/src/main/java/brooklyn/mementos/BrooklynMementoPersister.java
```diff
@@ -97,10 +97,13 @@
...
     void checkpoint(BrooklynMemento memento, PersistenceExceptionHandler exceptionHandler);
-    
-    void checkpoint(BrooklynMementoRawData newMemento, PersistenceExceptionHandler exceptionHandler);
 
+    /** applies a full checkpoint (write) of all state */  
+    void checkpoint(BrooklynMementoRawData newMemento, PersistenceExceptionHandler exceptionHandler);
+    /** applies a partial write of state delta */  
     void delta(Delta delta, PersistenceExceptionHandler exceptionHandler);
+    /** inserts an additional delta to be written on the next delta request */
+    void queueDelta(Delta delta);
```
aledsage: We already have the `PeriodicDeltaChangeListener`, which wraps calls to `persister.delta`. It handles queues, and does smart stuff like if you do a modify then a remove within the same time period, it will discard the modify and just do the remove.

Should we try to reuse `PeriodicDeltaChangeListener` somehow, rather than having `queueDelta` here?

aledsage: The semantics of `queueDelta` (to only be applied on the next delta request) is clear and good enough for current usage, but feels strange to have on an api. If means your queue won't be executed until either you call delta (in which case why did you queue it) or someone else calls delta (in which case you don't know when that will happen).

Worth at least having `@Beta` on this.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -186,12 +192,7 @@ public CatalogDo getCatalog() {
             return null;
         }
 
-        String versionedId = CatalogUtils.getVersionedId(symbolicName, fixedVersionId);
-        CatalogItemDo<?, ?> item = null;
-        //TODO should remove "manual additions" bucket; just have one map a la osgi
-        if (manualAdditionsCatalog!=null) item = manualAdditionsCatalog.getIdCache().get(versionedId);
-        if (item == null) item = catalog.getIdCache().get(versionedId);
-        return item;
+        return catalog.getIdCache().get( CatalogUtils.getVersionedId(symbolicName, fixedVersionId) );
```
aledsage: I don't follow the use of `manualAdditionsCatalog`. Can we safely ignore it in this method?

I see manualAdditionsCatalog is used in the deprecated `addItem(CatalogItem<?,?> item)`, but `catalog` is not.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -750,7 +780,36 @@ private String setFromItemIfUnset(String oldValue, Map<?,?> item, String fieldAt
         return oldValue;
     }
 
-    
+    /** scans the given libraries for annotated items, or if null scans the local classpath */
```
aledsage: I'd prefer it if `null` does not mean "local scan". In (too) many other places we treat null the same as an empty list. Treating it as meaning "local" will be surprising.

If non-null, does it scan the local classpath as well? I'm finding it hard to tell from the code what triggers it to scan the local classpath.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -750,7 +780,36 @@ private String setFromItemIfUnset(String oldValue, Map<?,?> item, String fieldAt
         return oldValue;
     }
 
-    
+    /** scans the given libraries for annotated items, or if null scans the local classpath */ 
+    private Collection<CatalogItemDtoAbstract<?, ?>> scanAnnotations(ManagementContext mgmt, Collection<CatalogBundle> libraries) {
+//        CatalogDto dto = CatalogDto.newDefaultLocalScanningDto(CatalogClasspathDo.CatalogScanningModes.ANNOTATIONS);
```
aledsage: Don't comment out code without a comment to say when and why it would be useful to uncomment it (or some such).

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -986,7 +1045,12 @@ private DeploymentPlan makePlanFromYaml(String yaml) {
     }
     
     private CatalogItem<?,?> addItemDto(CatalogItemDtoAbstract<?, ?> itemDto, boolean forceUpdate) {
-        checkItemNotExists(itemDto, forceUpdate);
+        CatalogItem<?, ?> existingDto = checkItemIsDuplicateOrDisallowed(itemDto, true, forceUpdate);
```
aledsage: Name `checkItemIsDuplicateOrDisallowed` is confusing. The `check...` prefix elsewhere follows the guava convention: it throws an exception if the check fails (e.g. `checkNotNull` throws if the condition "not null" fails). Reading this suggests we would throw if "is duplicate or disallowed" fails. But it's the opposite: we're checking that either it's not a duplicate or is allowed.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -986,7 +1045,12 @@ private DeploymentPlan makePlanFromYaml(String yaml) {
...
+        if (existingDto!=null) {
+            // it's a duplicate, and not forced, just return it
+            log.trace("Using existing duplicate for catalog item {}", itemDto.getId());
```
aledsage: log.debug instead of trace: we're treating the caller's request as a no-op, so should log that we've done nothing on the caller's request. Or does this get called a lot by some automated mechanism?

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -1008,8 +1072,19 @@ private DeploymentPlan makePlanFromYaml(String yaml) {
         return itemDto;
     }
 
-    private void checkItemNotExists(CatalogItem<?,?> itemDto, boolean forceUpdate) {
-        if (!forceUpdate && getCatalogItemDo(itemDto.getSymbolicName(), itemDto.getVersion()) != null) {
+    /** returns item DTO if item is an allowed duplicate, null if it should be added, or false if the item is an allowed duplicate,
```
aledsage: Comment seems wrong - we don't return false.

core/src/main/java/brooklyn/catalog/internal/CatalogDo.java
```diff
@@ -254,8 +270,10 @@ CatalogDo addCatalog(CatalogDto child) {
     public synchronized void addToClasspath(String ...urls) {
         if (dto.classpath == null)
             dto.classpath = new CatalogClasspathDto();
-        for (String url: urls)
-            dto.classpath.addEntry(url);
+        for (String url: urls) {
+            if (url!=null)
```
aledsage: Let's hope we never hit this with `url == null`. I'd expect (at least some times) that will be caused by a programming error or missing resource, which we should tell the caller. This code will now ignore (without any indication to the user).

This relates to a more general discussion of how we should handle nulls. I think we should avoid nulls where feasible and accept them where required (e.g. getAttribute() and getConfig() can always return null). But I don't feel strongly enough to say more at this time!

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,377 @@
...
+
+    /*
+
+    A1) if not persisting, go to B1
+    A2) if --catalog-reset, delete persisted catalog items
+    A3) read persisted catalog items (possibly deleted in A2), go to C1
+    A4) go to B1
```
aledsage: To make this clearer, should say that A3 only goes to C1 if there was persisted state. I presume that is just non-blank state, rather than it having to have catalog items.

api/src/main/java/brooklyn/mementos/BrooklynMementoPersister.java
ahgittin: +1 `@Beta`.  I couldn't see a way to add this batch-ahead-of-time functionality to the  `PDCL`.  Agree it's now suboptimal having two batches.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
ahgittin: i'll stick the `isPopulating` boolean in the mutex.  i think it was fine before but it will be clearer with it inside.  i'll also add comments to `mutex`.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
ahgittin: we can radically simplify everything -- perhaps even just rewrite it -- when we abandon the XML catalog

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
ahgittin: for non-null is does *not* scan the local classpath.  i've refactored to make 3 methods -- easier to read for sure, but at a cost of more private methods in the class.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
ahgittin: it could get called a lot in hot backup.  and we verified that what we would do would be no-op before we did nothing so not a useful log message i think.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
ahgittin: yeah, semantics got updated, tidied as part of method rename

core/src/main/java/brooklyn/catalog/internal/CatalogDo.java
ahgittin: the only non-test use is when we are registering declared osgi bundles.  a pre-installed bundle might have a null url.  we could filter this before calling but it is more work.

agree it's worth warning about nulls in situations where programming errors could crop up, but it seems unlikely here and the investment not worth it.

ahgittin:
@aledsage @robertgmoss these comments all addressed

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,377 @@
...
+        this.additionsUri = additionUri;
+        this.force = force;
+    }
+    
+    public CatalogInitialization() {
+        this(null, false, null, false);
+    }
+
+    public void injectManagementContext(ManagementContext managementContext) {
+        if (this.managementContext!=null && managementContext!=null && !this.managementContext.equals(managementContext))
```
aledsage: I'd prefer us to NPE if the injected managementContext is null. That would presumably be either a programming error or very lazy test code.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,377 @@
...
+    List<Function<CatalogInitialization, Void>> callbacks = MutableList.of();
+    boolean hasRunBestEffort = false, hasRunOfficial = false, isPopulating = false;
+    
+    ManagementContext managementContext;
+    boolean isStartingUp = false;
+    boolean failOnStartupErrors = false;
+    
+    Object mutex = new Object();
+    
+    public CatalogInitialization(String initialUri, boolean reset, String additionUri, boolean force) {
```
aledsage: Worth javadoc to say what `initialUri` and `additionUri` are - not clear without digging through the code.
Rename to `additionsUri` to match the field.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,377 @@
...
+
+    C1) if --catalog-add, read and add those items
+
+    D1) if persisting, read the rest of the persisted items (entities etc)
+
+     */
+
+    private static final Logger log = LoggerFactory.getLogger(CatalogInitialization.class);
+    
+    String initialUri;
```
aledsage: Why package private and not final?

It leaves the reader wondering if it is supposed to support the field being modified at any point by other code in the same package - which makes it much harder to reason about the code if anyone is going to modify it later or investigate any bugs that are reported.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,377 @@
...
+        return reset;
+    }
+
+    public boolean hasRunOfficial() { return hasRunOfficial; }
+    public boolean hasRunIncludingBestEffort() { return hasRunOfficial || hasRunBestEffort; }
+
+    /** makes or updates the mgmt catalog, based on the settings in this class */
+    public void populateCatalog(boolean needsInitial, Collection<CatalogItem<?, ?>> optionalItemsForResettingCatalog) {
+        try {
+            isPopulating = true;
```
aledsage: To @robertgmoss's point, why deliberately do the `isPopulating` and the finally block outside of the mutex?

For example, the other method to be synchronized (`populateBestEffort `) grabs the mutex and the does:
    `if (hasRunOfficial || hasRunBestEffort || isPopulating) return;`
Therefore `populateBestEffort ` return immediately even though this method hasn't actually populated the catalog yet.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,377 @@
...
+
+                populateCatalog(catalog, needsInitial, true, optionalItemsForResettingCatalog);
+            }
+        } finally {
+            hasRunOfficial = true;
+            isPopulating = false;
+        }
+    }
+
+    private void populateCatalog(BasicBrooklynCatalog catalog, boolean needsInitial, boolean runCallbacks, Collection<CatalogItem<?, ?>> optionalItemsForResettingCatalog) {
```
aledsage: I'd prefer to not bother passing in arguments that are always true. For example, `runCallbacks` is true in every caller, so we can miss it out.

I'd also go for a name like `populateCatalogImpl` to indicate that it's the internal implementation (but no strong feelings on that one).

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,377 @@
...
+            return;
+        }
+        
+        String catalogUrl = managementContext.getConfig().getConfig(BrooklynServerConfig.BROOKLYN_CATALOG_URL);
+        if (Strings.isNonBlank(catalogUrl)) {
+            populateInitialFromUri(catalog, catalogUrl, PopulateMode.AUTODETECT);
+            return;
+        }
+        
+        catalogUrl = Urls.mergePaths(BrooklynServerConfig.getMgmtBaseDir( managementContext.getConfig() ), "catalog.bom");
```
aledsage: If getMgmtBaseDir returns a dir, then clearer to use `Os.mergePaths()`. If it returns a URL, then can't do `new File(catalogUrl).exists()` below, I presume.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,377 @@
...
+                result = catalog.addItems(contents);
+            } catch (Exception e) {
+                Exceptions.propagateIfFatal(e);
+                if (problem==null) problem = e;
+            }
+        }
+        
+        if (result==null && contents!=null && (mode==PopulateMode.XML || mode==PopulateMode.AUTODETECT)) {
+            // then try XML
+            problem = populateInitialFromUriXml(catalog, catalogUrl, problem, contents);
```
aledsage: Minor preference: I think the code would be clearer if the try-catch was in this method (similar to the YAML code above), rather than passing the `problem` in as a parameter in the method call here.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,377 @@
...
+                log.debug("Added to catalog from CLI: "+items);
+            else
+                log.debug("Added to catalog from CLI: count "+Iterables.size(items));
+            
+            hasRunAdditions = true;
+        }
+    }
+
+    protected void populateViaCallbacks(BasicBrooklynCatalog catalog) {
+        for (Function<CatalogInitialization, Void> callback: callbacks)
```
aledsage: Would these be better called "actions" rather than "callbacks"? When I first saw the callbacks being passed in, I wondered if they were listeners of some description (being called on population).

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,377 @@
...
+                }
+                populateCatalog(catalog, true, true, null);
+            } finally {
+                hasRunBestEffort = true;
+                isPopulating = false;
+            }
+        }
+    }
+
+    public void setStartingUp(boolean isStartingUp) {
```
aledsage: The lifecycle of this class is quite hard to follow. I presume that the caller is supposed to call `setStartingUp(true)` when first initialising, and `setStartingUp(false)` when done (but to not call that on rebind?). However, I don't feel confident about understanding that context by just looking at this code.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,377 @@
...
+        this.isStartingUp = isStartingUp;
+    }
+
+    public void setFailOnStartupErrors(boolean startupFailOnCatalogErrors) {
```
aledsage: Personal preference to have the setters at the top - they are presumably being used like a builder'y type of thing. They are conveniences, rather than having a huge number of args in the constructor. Having all of those together at the top make it easier to see how the behaviour of the class is set.

core/src/main/java/brooklyn/entity/rebind/PeriodicDeltaChangeListener.java
```diff
@@ -247,68 +221,73 @@ void stop() {
...
-            if (System.currentTimeMillis() > maxEndtime) {
-                throw new TimeoutException("Timeout waiting for pending complete of rebind-periodic-delta, after "+Time.makeTimeStringRounded(timeout));
+        CountdownTimer timer = timeout.isPositive() ? CountdownTimer.newInstanceStarted(timeout) : CountdownTimer.newInstancePaused(Duration.PRACTICALLY_FOREVER);
+        // wait for mutex, so we aren't tricked by an in-progress who has already recycled the collector
+        if (persistingMutex.tryAcquire(timer.getDurationRemaining().toMilliseconds(), TimeUnit.MILLISECONDS)) {
+            try {
+                // now no one else is writing
+                if (!deltaCollector.isEmpty()) {
+                    // but there is data that needs to be written
+                    persistNowSafely(true);
```
aledsage: I can see this is simpler than the previously `writeCount`. However, it means that test code (calling `waitForPendingComplete`) will change the behaviour of the persistence. It no longer just waits, but actually does the persist. Therefore it means a class of bug will no longer be detected where the persisting was not happening for whatever reason.

That's probably not an issue (because such an errors are unlikely, thought they could arise from some deadlock scenarios for example).

No particularly strong feelings.

core/src/main/java/brooklyn/management/internal/AbstractManagementContext.java
```diff
@@ -443,6 +441,22 @@ public void setManagementNodeUri(URI uri) {
         return uri;
     }
     
+    @Override
+    public CatalogInitialization getCatalogInitialization() {
+        if (catalogInitialization!=null) return catalogInitialization;
```
aledsage: Double-checked locking is broken in Java, unless you declare the variable volatile (http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html).

This code is not performance sensitive, so just delete this line to keep the code simpler.

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
```diff
@@ -341,38 +340,67 @@ protected void rebuildCatalog() {
...
             }
         }
+
+        // See notes in CatalogInitialization
         
-        // Register catalogue items with the management context. Loads the bundles in the OSGi framework.
-        CatalogLoadMode catalogLoadMode = managementContext.getConfig().getConfig(BrooklynServerConfig.CATALOG_LOAD_MODE);
+        Collection<CatalogItem<?, ?>> catalogItems = rebindContext.getCatalogItems();
+        CatalogInitialization catInit = ((ManagementContextInternal)managementContext).getCatalogInitialization();
+        catInit.injectManagementContext(managementContext);
```
aledsage: Why would we have to inject the management context here, when we've obtained the instance by calling  `managementContext.getCatalogInitialization()`? That should return an instance which already has the management context.

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
```diff
@@ -341,38 +340,67 @@ protected void rebuildCatalog() {
...
+                    if (shouldLogRebinding()) {
+                        LOG.info(message);
+                    }
+                }
+                logRebindingDebug(message);
+
+                itemsForResettingCatalog = MutableList.<CatalogItem<?,?>>of();
+                
+                PersisterDeltaImpl delta = new PersisterDeltaImpl();
+                for (String catalogItemId: mementoRawData.getCatalogItems().keySet()) {
```
aledsage: just do `delta.removedCatalogItemIds.addAll(mementoRawData.getCatalogItems().keySet())`

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
```diff
@@ -341,38 +340,67 @@ protected void rebuildCatalog() {
...
+            } else {
+                logRebindingDebug("RebindManager will initialize catalog locally because catalog persistence is disabled");
+                needsInitialCatalog = true;
+            }
         }
+
+        // TODO in read-only mode, perhaps do this less frequently than entities etc ?
+        // both in RW and in RO mode, the first run reads the initialization data;
+        // maybe not desired for RO as it defers problems, although if it's standalone it is desired
+        catInit.populateCatalog(needsInitialCatalog, itemsForResettingCatalog);
```
aledsage: Personal preference: I like `Optional<Collection<CatalogItem<?,?>>>` rather than using null to indicate absence. That makes the behaviour of the method more explicit.

core/src/main/java/brooklyn/entity/rebind/PeriodicDeltaChangeListener.java
```diff
@@ -348,13 +327,40 @@ private void addReferencedObjects(DeltaCollector deltaCollector) {
...
+            persistNowInternal(alreadyHasMutex);
+            metrics.noteSuccess(Duration.of(timer));
+            return true;
+        } catch (RuntimeInterruptedException e) {
+            LOG.debug("Interrupted persisting change-delta (rethrowing)", e);
+            metrics.noteFailure(Duration.of(timer));
+            metrics.noteError(e.toString());
+            Thread.currentThread().interrupt();
+            return false;
+        } catch (Exception e) {
```
aledsage: I'd probably merge the `catch Exception` and the `catch Throwable`, with:

        } catch (Throwable t) {
            // Don't rethrow: the behaviour of executionManager is different from a scheduledExecutorService,
            // if we throw an exception, then our task will never get executed again
            LOG.error("Problem persisting change-delta", t);
            metrics.noteFailure(Duration.of(timer));
            metrics.noteError(t.toString());
            Exceptions.propagateIfFatal(t);
            return false;

Downside is that log message is always error, and doesn't include whether or not it is "(rethrowing)".

core/src/main/java/brooklyn/entity/rebind/persister/AbstractBrooklynMementoPersister.java
```diff
@@ -110,6 +115,11 @@ public void delta(Delta delta, PersistenceExceptionHandler exceptionHanlder) {
         memento.updateEnricherMementos(delta.enrichers());
         memento.updateCatalogItemMementos(delta.catalogItems());
     }
+
+    @Override
+    public void queueDelta(Delta delta) {
+        log.warn("Legacy persister ignoring queued delta: "+delta);
```
aledsage: Perhaps `throw new IllegalStateException("Not supported; use "+BrooklynMementoPersisterToObjectStore.class)`.
With this code, we'll log and ignore the requested delta, so the delta will never be applied. Better to fail than to do that, I think (unless the delta really is empty!)

core/src/main/java/brooklyn/entity/rebind/persister/BrooklynMementoPersisterToObjectStore.java
```diff
@@ -560,9 +563,18 @@ public void checkpoint(BrooklynMemento newMemento, PersistenceExceptionHandler e
     public void delta(Delta delta, PersistenceExceptionHandler exceptionHandler) {
         checkWritesAllowed();
 
+        while (!queuedDeltas.isEmpty()) {
+            Delta extraDelta = queuedDeltas.remove(0);
+            doDelta(extraDelta, exceptionHandler, false);
```
aledsage: why does this pass `false` for `previouslyQueued`?

usage/launcher/src/main/java/brooklyn/launcher/BrooklynLauncher.java
```diff
@@ -682,6 +676,8 @@ private void initManagementContext() {
             brooklynProperties.addFromMap(brooklynAdditionalProperties);
         }
         
+        ((ManagementContextInternal)managementContext).setCatalogInitialization(catalogInitialization);
```
aledsage: guard call, to only do if `catalogInitialization` is non-null.

aledsage: Overall I'm ok with this, but I do have a gut feeling that I don't like us injecting the `catalogInitialization`.

It feels like we want to inject config to control the catalog initialization. That could be used to configure the actual catalogInitialization object, which would be purely internal.

Injecting a full-blown `CatalogInitialization` object is obviously more flexible, but the contract of what that object must do is not very well defined for it to be on an interface. I'd expect we'd only want to inject a different impl in tests, or in really extreme cases. If we really want to support that kind of thing, we should think seriously about a dependency injection framework.

core/src/main/java/brooklyn/management/internal/AbstractManagementContext.java
```diff
@@ -443,6 +441,22 @@ public void setManagementNodeUri(URI uri) {
...
+            if (catalogInitialization!=null) return catalogInitialization;
+            CatalogInitialization ci = new CatalogInitialization();
+            setCatalogInitialization(ci);
+            return ci;
+        }
+    }
+    
+    @Override
+    public synchronized void setCatalogInitialization(CatalogInitialization catalogInitialization) {
+        if (catalogInitialization!=null) catalogInitialization.injectManagementContext(this);
```
aledsage: I'd much prefer us to fail if the user passes in null, rather than failing on an NPE in some unrelated code/thread later.

I'd do something like:

    this.catalogInitialization  = checkNotNull(catalogInitialization, "catalogInitialization");
    catalogInitialization.injectManagementContext(this);

core/src/test/java/brooklyn/catalog/internal/CatalogScanTest.java
```diff
@@ -161,13 +162,13 @@ public void testMoreTypesThanAnnotationsForApps() {
...
-        loadAnnotationsOnlyCatalog();
         loadTheDefaultCatalog();
+        int numInDefault = Iterables.size(defaultCatalog.getCatalogItems(Predicates.alwaysTrue()));
         
+        loadAnnotationsOnlyCatalog();
         int numFromAnnots = Iterables.size(annotsCatalog.getCatalogItems(Predicates.alwaysTrue()));
-        int numInDefault = Iterables.size(defaultCatalog.getCatalogItems(Predicates.alwaysTrue()));
         
-        Assert.assertEquals(numFromAnnots, numInDefault);
+        Assert.assertEquals(numInDefault, numFromAnnots);
```
aledsage: Could also assert that `> 0`; otherwise could be a bug and a coincidence that both are zero?

docs/guide/start/blueprints.md
```diff
@@ -45,17 +45,14 @@ application. Your application will be shown as "Starting" on the web console's f
 
 Instead of pasting the YAML blueprint each time,
 this blueprint can be [added to the catalog](../ops/catalog/).
-Or, even easier, you can download a sample [catalog.xml](catalog.xml).
-Install this to your `~/.brooklyn/` folder and relaunch Brooklyn
-(navigating to the "Help" tab in order to shutdown Brooklyn *and* the application you launched in the previous step).
-
-Now when the web console is re-opened, the catalog contains our blueprints.
-Select the "Demo Web Cluster with DB" and click "Next".
+With this YAML blueprint added, including the location, the Add Application dialog will offer
```
aledsage: Is this the only mention in the docs that "including the location" will treat the catalog item as a template? Deserves a section of its own in `docs/guide/ops/catalog/index.md`, including how to add the catalog item as a "tile" (i.e. as a clickable icon that will not show the yaml, but will instead allow it to be deployed directly - a customer was recently stressing the importance of that when they demo Brooklyn to their own customers).

usage/cli/src/main/java/brooklyn/cli/ItemLister.java
```diff
@@ -66,6 +68,9 @@
 public class ItemLister {
```
aledsage: I didn't know about this before! Needs documented somewhere, perhaps in `docs/guide/ops/launch.md` where other CLI options are documented such as  `brooklyn cloud-compute`.

aledsage: Ah, yes I did know about it - it's for us generating documentation (config/attributes/effectors) about all our entities, policies, etc.

Anyway, the comment about docs doesn't need addressed in this PR.

docs/guide/ops/catalog/index.md
```diff
@@ -284,6 +291,19 @@ When referencing a blueprint, if a version number is not specified
 the latest non-snapshot version will be loaded when an entity is instantiated.
 
 
+### CLI Options
+
+The `brooklyn` CLI includes several commands for working with the catalog.
+
+* `--catalogAdd <file.bom>` will add the catalog items in the `bom` file
```
aledsage: The term `bom` seems never to be defined.

usage/cli/src/main/java/brooklyn/cli/Main.java
```diff
@@ -266,7 +290,7 @@ private String readLine(InputStream in) throws IOException {
         // looks like: {@linktourl https://gist.github.com/47066f72d6f6f79b953e}
         @Beta
         @Option(name = { "-sk", "--stopOnKeyPress" },
-                description = "After startup, shutdown on user text entry (default false)")
+                description = "Shutdown immediately on user text entry after startup (useful for debugging and demos)")
```
aledsage: Definite improvement; we should also remove mention of Whirr in comment above this (can be in a separate PR)

usage/cli/src/main/java/brooklyn/cli/Main.java
```diff
@@ -536,9 +570,10 @@ protected BrooklynLauncher createLauncher() {
             BrooklynLauncher launcher;
             launcher = BrooklynLauncher.newInstance();
             launcher.localBrooklynPropertiesFile(localBrooklynProperties)
-                    .ignorePersistenceErrors(ignorePersistenceErrors)
-                    .ignoreWebErrors(ignoreWebErrors)
-                    .ignoreAppErrors(ignoreAppErrors)
+                    .ignorePersistenceErrors(!startupFailOnPersistenceErrors)
```
aledsage: Should we (in separate PR?) align the method names in launcher with the cli names? e.g. `failOnPersistenceErrors(boolean)`, and deprecate the `ignorePersistenceErrors(boolean)`?

usage/cli/src/main/java/brooklyn/cli/lister/ItemDescriptors.java
```diff
@@ -58,8 +65,17 @@
         List<Map<String, Object>> itemDescriptors = Lists.newArrayList();
         
         for (Class<? extends BrooklynObject> type : types) {
-            Map<String, Object> itemDescriptor = toItemDescriptor(type, headingsOnly);
-            itemDescriptors.add(itemDescriptor);
+            try {
+                Map<String, Object> itemDescriptor = toItemDescriptor(type, headingsOnly);
+                itemDescriptors.add(itemDescriptor);
+            } catch (Throwable throwable) {
+                if (throwable instanceof InterruptedException)
```
aledsage: Can just do `Exceptions.propagateIfFatal(throwable)`.

usage/dist/src/main/dist/conf/brooklyn/default.catalog.bom
```diff
@@ -0,0 +1,41 @@
...
+
+  # load everything in the classpath with a @Catalog annotation
+  - scanJavaAnnotations: true
+
+  - id: server
+    description: |
+      Provision a server, with customizable provisioning.properties and credentials installed, 
+      but no other special software process or scripts executed.
+    item:
+      type: brooklyn.entity.basic.EmptySoftwareProcess
```
aledsage: I'd prefer if we have consistent structure, rather than demonstrating different possible structures within the same file (that will just confuse users, as to whether they mean the same thing).

Suggest we include `services: `, as is done in `server-template`.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
ahgittin: agree, moved

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
ahgittin: i think this is the only one i didn't fully agree with -- could go either way, but i think of them as callbacks because typically they are passed in by custom startup classes to have a callback to populate catalog in their space.  (yes, they are actions, but that seems less clear to me.)

ahgittin:
ah i caught up with your comments while on the plane just now but i see you've also been busy @aledsage .  will look in to your latest batch of comments.

usage/dist/src/main/dist/conf/brooklyn/default.catalog.bom
ahgittin: templates *must* have a services block -- everything else generally should *not* -- as per current catalog description

usage/rest-server/src/main/java/brooklyn/rest/resources/ServerResource.java
```diff
@@ -126,45 +126,62 @@ public void shutdown(final boolean stopAppsFirst, final boolean forceShutdownOnE
...
+                    complete();
+                
+                    if (!hasAppErrorsOrTimeout.get() || forceShutdownOnError) {
+                        //give the http request a chance to complete gracefully
+                        Time.sleep(delayForHttpReturn);
+                        System.exit(0);
+                    }
+                    
+                    // There are app errors, don't exit the process, allowing any exception to continue throwing
+                    log.warn("Abandoning shutdown because there were errors and shutdown was not forced.");
```
aledsage: personal preference to put this kind of thing in an else block (even though the above block will have done `System.exit`, so can only reach here if we didn't go into the if).

utils/common/src/main/java/brooklyn/util/javalang/AggregateClassLoader.java
```diff
@@ -69,8 +70,25 @@ public void add(int index, ClassLoader classLoader) {
         if (classLoader != null) classLoaders.add(index, classLoader);
     }
     
-    /** Returns the _live_ (and modifiable) list of classloaders 
-     * @return */ 
+    /** Resets the classloader shown here to be the given set */
+    public void reset(Collection<? extends ClassLoader> newClassLoaders) {
+        synchronized (classLoaders) {
```
aledsage: What is the point of the synchronized? Is it just to guard against multiple concurrent calls to `reset()`? None of the other modifications or accesses to `classLoaders` is synchronized.

aledsage: Clever adding all to end, and then removing `count` from start. However, it will break if there is a concurrent call to `add(int index, ClassLoader classLoader)`.

Personally I'd prefer a simpler synchronization model, that does not try to be clever.

utils/common/src/main/java/brooklyn/util/javalang/Threads.java
```diff
@@ -44,7 +46,15 @@ public static boolean removeShutdownHook(Thread hook) {
             return Runtime.getRuntime().removeShutdownHook(hook);
         } catch (IllegalStateException e) {
             // probably shutdown in progress
-            log.debug("cannot remove shutdown hook "+hook+": "+e);
+            String text = Exceptions.collapseText(e);
+            if (text.contains("Shutdown in progress")) {
+                if (log.isTraceEnabled()) {
+                    log.trace("Could not remove shutdown hook "+hook+": "+text);
```
aledsage: Add trace and debug level, we should just include the entire stacktrace rather than truncating. Otherwise we risk losing important information for debugging problems that customers send via log files.

aledsage:
Have finished going through this huge pull request! Lots of excellent stuff here; really like what it does.

Vast majority of my comments are quite low level.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
aledsage: Ok, no strong feelings. A line or two of javadoc would have removed the uncertainty in my mind.

ahgittin:
Thanks @aledsage -- superb review.  I've got most of them done, I should get the rest in tonight.

Sorry for the large volume of foolishness.  :)  It went through a few rewrites along the way.

utils/common/src/main/java/brooklyn/util/javalang/Threads.java
ahgittin: Normally I'd agree but this is a known exception which is noise in the logs, only arising from one place.  And it appears at the very end of the log and so is very visible, giving the wrong impression if it looks like a real problem.

utils/common/src/main/java/brooklyn/util/javalang/AggregateClassLoader.java
ahgittin: have synched on all accesses, including access to `iterator()`

usage/cli/src/main/java/brooklyn/cli/lister/ItemDescriptors.java
ahgittin: you can get various `*Class*Error` instances here which we don't want to treat as fatal.
probably we should revisit `propagateIfFatal` not to crash and burn on arbitrary `Error` instances.

usage/cli/src/main/java/brooklyn/cli/Main.java
ahgittin: +1 -- let's wait and make sure these names (and logic) is good though

docs/guide/ops/catalog/index.md
ahgittin: i expect we'll be adding a lot more, explaining this, and converting `*.yaml` to `*.bom` in near future

docs/guide/start/blueprints.md
ahgittin: i expect we'll tidy up the gui and this discussion of the gui in future

core/src/main/java/brooklyn/entity/rebind/persister/BrooklynMementoPersisterToObjectStore.java
ahgittin: good spot

core/src/main/java/brooklyn/entity/rebind/persister/AbstractBrooklynMementoPersister.java
ahgittin: this will be deleted as soon as `0.7.0` is cut.  i'm sure we could delete it now, i'm just being super cautious.  no need to make perfect log messages here!

core/src/main/java/brooklyn/entity/rebind/PeriodicDeltaChangeListener.java
ahgittin: handling log messages correctly, with the right message, is important to prevent spurious errors on shutdown

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
ahgittin: meh i dislike `Optional` except when null is ambiguous or you want to chain or maintain an error on access (and in that last case `Maybe` is better).

ahgittin:
A few I disagreed on, I've commented there, the rest should all be addresed in the last PR.  (I've squashed what were three PR's.)

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -597,7 +612,22 @@ private void collectCatalogItems(String sourceYaml, Map<?,?> itemMetadata, List<
...
+            // don't scan
+        } else {
+            // scan for annotations: if libraries here, scan them; if inherited libraries error; else scan classpath
+            if (!libraryBundlesNew.isEmpty()) {
+                result.addAll(scanAnnotationsFromBundles(mgmt, libraryBundlesNew));
+            } else if (libraryBundles.isEmpty()) {
+                result.addAll(scanAnnotationsFromLocal(mgmt));
+            } else {
+                throw new IllegalStateException("Cannot scan catalog node no local bundles, and with inherited bundles we will not scan the classpath");
+            }
```
neykov: Might be better to have two scan keywords - one for local bundles, the other for brooklyn classpath. This will avoid the context dependency, make intent explicit, easier for users to understand.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -750,7 +780,37 @@ private String setFromItemIfUnset(String oldValue, Map<?,?> item, String fieldAt
...
+    private Collection<CatalogItemDtoAbstract<?, ?>> scanAnnotationsFromBundles(ManagementContext mgmt, Collection<CatalogBundle> libraries) {
+        String[] urls = null;
+        CatalogDto dto = CatalogDto.newNamedInstance("Bundles Scanned Catalog", "All annotated Brooklyn entities detected in the classpath", "scanning-bundles-classpath-"+libraries.hashCode());
+        urls = new String[libraries.size()];
+        int i=0;
+        for (CatalogBundle b: libraries)
+            urls[i++] = b.getUrl();
+            
+        CatalogDo subCatalog = new CatalogDo(dto);
+        subCatalog.addToClasspath(urls);
```
neykov: * Will it work if annotated classes depend on a bundle not explicitly listed (no OSGi dependency mechanism in play here)?
* Can't figure out how the bundles in `urls` can see the brooklyn classpath. The classloader created for scanning is limited to only this list.
* Even if the brooklyn classpath is visible how are conflicting versions handled?
* The libraries list can contain name:version pairs only, leaving it to OSGi to figure out the bundle. `getUrl` will return `null` in this case.

neykov: I see that below you are handling the `null` case, but then if all libraries are name-value pairs, the code path will go in the "scanning catalog root classpath" case. Better move the null check here.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
ahgittin: personally i'd rather keep the yaml tight.  and do away with local classpath autoscan altogether.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
ahgittin: i think the what-to-scan logic is correct.  we do

     Collection<CatalogBundle> libraryBundlesNew = CatalogItemDtoAbstract.parseLibraries(librariesNew);

which accepts both `name:version` pairs and URL's; that collection won't be empty if either is supplied, and so it then won't `scanAnnotationsFromLocal`.

the brooklyn classpath will always be available because it is the bootstrap class loader, won't it?  it just won't be scanned when any bundles are defined.

i've added a log warning when only `name:version` is supplied, and comments in several places for the other issues (all of which i agree), including the docs for the `scanJavaAnnotations [experimental]`.  i think this is the right balance until we know whether we want to support scanning of OSGi or instead have some other tool to extract a `bom` (or even a TOSCA definition) which we expect included in the bundle.

ahgittin:
unrelated test failure:

    2015-05-20 00:01:48,202 INFO  TESTNG FAILED: "Surefire test" - brooklyn.rest.jsgui.BrooklynJavascriptGuiLauncherTest.testJavascriptWithoutRest() finished in 9455 ms

@aledsage @neykov anything else?

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
neykov: Made a quick test - tried scanning `brooklyn-test-osgi-entities.jar` (after annotating `SimpleEntity`). Fails with

```
2015-05-20 17:46:39,482 ERROR Error loading catalog item 'brooklyn.catalog.internal.CatalogItemDo[CatalogEntityItemDto[brooklyn.osgi.tests.SimpleEntity:0.0.0.SNAPSHOT/Tomcat Server]]': java.lang.IllegalStateException: Error creating class brooklyn.entity.proxying.EntitySpec brooklyn.osgi.tests.SimpleEntity: java.lang.IllegalStateException: Unable to load brooklyn.osgi.tests.SimpleEntity from []: Invalid class: brooklyn.osgi.tests.SimpleEntity: java.lang.ClassNotFoundException: brooklyn.osgi.tests.SimpleEntity
```
with the yaml

```
brooklyn.catalog:
- libraries:
  - file:///.../core/src/test/resources/brooklyn/osgi/brooklyn-test-osgi-entities.jar
  scanJavaAnnotations: true
```

Worth adding it to the PR. But no classpath:// protocol supported by the catalog CL, as used in other tests.

If there's no explicit parent class loader (as in our case) only the bootstrap class loader will be consulted which doesn't have access to the app classes.

neykov: > that collection won't be empty if either is supplied, and so it then won't scanAnnotationsFromLocal

It will go through `scanAnnotationsFromBundles`, but in `CatalogClasspathDo.load` the list will be filtered out of the non-url bundles, making it go through [this code branch](https://github.com/apache/incubator-brooklyn/blob/09a64755e2c16c25f076af6a32f93ee0ff8c2599/core/src/main/java/brooklyn/catalog/internal/CatalogClasspathDo.java#L137)

I see you've already changed the code, just wanted to clear up what I meant.

ahgittin:
btw failure again is unrelated -- either there is something up with ports on this server or our logic (maybe around networks?) is not what we expect

```
	at io.brooklyn.camp.CampServer.start(CampServer.java:104)
	at io.brooklyn.camp.test.fixture.AbstractRestResourceTest.startServer(AbstractRestResourceTest.java:51)
```

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
```diff
@@ -341,38 +340,64 @@ protected void rebuildCatalog() {
...
+                }
+                logRebindingDebug(message);
+
+                itemsForResettingCatalog = MutableList.<CatalogItem<?,?>>of();
+                
+                PersisterDeltaImpl delta = new PersisterDeltaImpl();
+                delta.removedCatalogItemIds.addAll(mementoRawData.getCatalogItems().keySet());
+                getPersister().queueDelta(delta);
+                
+                mementoRawData.clearCatalogItems();
```
neykov: This looks like too much manual housekeeping / wrong level of abstraction.
Better let the persister know how to clear the catalog instead of introducing two lower-level APIs (queueDelta & clearCatalogItems). Moreover `clearCatalogItems` breaks the immutable nature of `mementoRawData`.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,380 @@
...
+        
+        String catalogUrl = managementContext.getConfig().getConfig(BrooklynServerConfig.BROOKLYN_CATALOG_URL);
+        if (Strings.isNonBlank(catalogUrl)) {
+            populateInitialFromUri(catalog, catalogUrl, PopulateMode.AUTODETECT);
+            return;
+        }
+        
+        catalogUrl = Os.mergePaths(BrooklynServerConfig.getMgmtBaseDir( managementContext.getConfig() ), "catalog.bom");
+        if (new File(catalogUrl).exists()) {
+            populateInitialFromUri(catalog, "file:"+catalogUrl, PopulateMode.YAML);
```
neykov: better use `new File(catalogUrl).toURI()`, same below.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    private enum PopulateMode { YAML, XML, AUTODETECT }
+    
+    protected void populateInitial(BasicBrooklynCatalog catalog) {
+        if (disallowLocal) {
+            if (!hasRunOfficial()) {
+                log.debug("CLI initial catalog not being read with disallow-local mode set.");
+            }
+            return;
```
neykov: If the user has set `initialUri`, but at the same time has `CATALOG_LOAD_MODE=LOAD_PERSISTED_STATE` configured, I'd expect the new style to take precedence. Or even fail early and refuse to start with a descriptive message.
Letting the old style win with only a debug message is confusing.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,380 @@
...
+
+        populateViaCallbacks(catalog);
+    }
+
+    private enum PopulateMode { YAML, XML, AUTODETECT }
+    
+    protected void populateInitial(BasicBrooklynCatalog catalog) {
+        if (disallowLocal) {
+            if (!hasRunOfficial()) {
+                log.debug("CLI initial catalog not being read with disallow-local mode set.");
```
neykov: disallow-local has no meaning to users, better use "brooklyn.catalog.mode=LOAD_PERSISTED_STATE"

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    boolean hasRunAdditions = false;
+    protected void populateAdditions(BasicBrooklynCatalog catalog) {
+        if (Strings.isNonBlank(additionsUri)) {
+            if (disallowLocal) {
+                if (!hasRunAdditions) {
+                    log.warn("CLI additions supplied but not supported in disallow-local mode; ignoring.");
+                }
+                return;
```
neykov: Same as above - if the user explicitly uses `--catalogAdd` on the command line, better to take precedence over deprecated configuration or fail with a descriptive message.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    private Object setFromCLMMutex = new Object();
+    private boolean setFromCatalogLoadMode = false;
+
+    /** @deprecated since introduced in 0.7.0, only for legacy compatibility with 
+     * {@link CatalogLoadMode} {@link BrooklynServerConfig#CATALOG_LOAD_MODE},
+     * allowing control of catalog loading from a brooklyn property */
+    @Deprecated
+    public void applyCatalogLoadMode() {
```
neykov: Can this be private to the class and called at `injectManagementContext` time instead of requiring the "user" to call it?

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
```diff
@@ -0,0 +1,380 @@
...
+        this.reset = reset;
+        this.additionsUri = additionUri;
+        this.force = force;
+    }
+    
+    public CatalogInitialization() {
+        this(null, false, null, false);
+    }
+
+    public void injectManagementContext(ManagementContext managementContext) {
```
neykov: Missing `@Override`

core/src/main/java/brooklyn/catalog/internal/CatalogItemDtoAbstract.java
```diff
@@ -162,6 +163,37 @@ public String getPlanYaml() {
...
+        if (!Objects.equal(javaType, other.javaType)) return false;
+        if (!Objects.equal(nullIfEmpty(libraries), nullIfEmpty(other.libraries))) return false;
+        if (!Objects.equal(getCatalogItemId(), other.getCatalogItemId())) return false;
+        if (!Objects.equal(version, other.version)) return false;
+        if (!Objects.equal(deprecated, other.deprecated)) return false;
+        if (!Objects.equal(description, other.description)) return false;
+        if (!Objects.equal(displayName, other.displayName)) return false;
+        if (!Objects.equal(iconUrl, other.iconUrl)) return false;
+        if (!Objects.equal(tags, other.tags)) return false;
+        // 'type' not checked, because deprecated, we might want to allow removal in future
```
neykov: `symbolicName` is set to `type.getName()`, so shouldn't matter.

core/src/main/java/brooklyn/entity/rebind/PeriodicDeltaChangeListener.java
```diff
@@ -164,14 +163,12 @@ public void remove(BrooklynObject instance) {
...
-    
-    private final AtomicLong writeCount = new AtomicLong();
-    
+        
     private DeltaCollector deltaCollector = new DeltaCollector();
 
     private volatile boolean running = false;
 
-    private volatile boolean stopped = false;
+    private volatile boolean stopping = false, stopCompleted = false;
```
neykov: `stopping`, `stopCompleted` and `running` could be merged into a single enum to manage state easier, can be done in a future PR.

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
```diff
@@ -341,38 +340,64 @@ protected void rebuildCatalog() {
...
-            if (shouldResetCatalog) {
-                // Reset catalog with previously persisted state
-                logRebindingDebug("RebindManager resetting management context catalog to previously persisted state");
-                managementContext.getCatalog().reset(rebindContext.getCatalogItems());
-            } else if (shouldLoadDefaultCatalog) {
-                // Load catalogue as normal
-                // TODO in read-only mode, should do this less frequently than entities etc
-                logRebindingDebug("RebindManager loading default catalog");
-                ((BasicBrooklynCatalog) managementContext.getCatalog()).resetCatalogToContentsAtConfiguredUrl();
+            if (!catInit.hasRunOfficial() && catInit.isInitialResetRequested()) {
```
neykov: How does catalog initialization work with the different HA states? It doesn't matter because we get a brand new ManagementContext with each rebind, which will do the full catalog init procedure, is that right?
If so should we use the rebind logger logic during catalog initialization as well to prevent excessive logging?

What happens if we go master -> standby -> master, either due to persistence store issues or force manually through rest? Will the catalog be re-initialized on each promotion?

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
```diff
@@ -341,38 +340,64 @@ protected void rebuildCatalog() {
...
-                    LOG.info(message + " There {} {} item{} persisted.", new Object[]{
-                            catalogLoadMode, catalogItems.size() == 1 ? "was" : "were", catalogItems.size(), Strings.s(catalogItems)});
-                } else if (LOG.isDebugEnabled()) {
-                    logRebindingDebug(message, catalogLoadMode);
+                if (!isEmpty) {
+                    logRebindingDebug("RebindManager clearing local catalog and loading from persisted state");
+                    itemsForResettingCatalog = rebindContext.getCatalogItems();
+                    needsInitialCatalog = false;
+                } else {
+                    if (catInit.hasRunOfficial()) {
```
neykov: If we have already initialized the

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
neykov: Reset rebindContext catalog items as well?

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
```diff
@@ -341,38 +340,64 @@ protected void rebuildCatalog() {
...
+                    message += "Catalog was empty anyway.";
+                } else {
+                    message += "Deleting "+catalogItems.size()+" persisted catalog item"+Strings.s(catalogItems)+": "+catalogItems;
+                    if (shouldLogRebinding()) {
+                        LOG.info(message);
+                    }
+                }
+                logRebindingDebug(message);
+
+                itemsForResettingCatalog = MutableList.<CatalogItem<?,?>>of();
```
neykov: If there are rebind errors [above](https://github.com/ahgittin/incubator-brooklyn/blob/d6ae0511c35ef4070c8a1637bad1ef3b113bda81/core/src/main/java/brooklyn/entity/rebind/RebindIteration.java#L308-L342) we would get a failed startup even though the catalog will be reset?

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
neykov: After discussion - clarify "official" as being first master rebind OR (in non-rebind mode) the load triggered by CLI
Then, feels like the cases should be
```
if (catalog persistence enabled) {
  if (!hasBeenMaster) {
    if (reset || isEmpty) {
       don't use persistence catalog items
       init initial catalog
    }
    init add'tl catalog
  }
} else {//catalog persistence disabled
  if (didn't initialize the catalog) {
       init initial catalog
       init add'tl catalog
  }
}
```

core/src/main/java/brooklyn/entity/rebind/PeriodicDeltaChangeListener.java
```diff
@@ -348,13 +336,40 @@ private void addReferencedObjects(DeltaCollector deltaCollector) {
     }
     
     @VisibleForTesting
-    public void persistNow() {
-        if (!isActive()) {
+    public boolean persistNowSafely(boolean alreadyHasMutex) {
```
neykov: Since the mutex is private it makes sense to expose to the outside only with alreadyHasMutex=false (i.e. a separate public metod which passes false), but not important.

neykov:
Is `riak.png`renamed on purpose?

ahgittin:
yes `riak.png` is deliberately moved, from `src/main/java` to `src/main/resources`.  maven sometimes ignores it in the former location so the icon doesn't show up when we view it.

usage/cli/src/main/java/brooklyn/cli/lister/ClassFinder.java
```diff
@@ -126,6 +129,8 @@ public boolean apply(Class<?> input) {
                         log.info("Cannot read "+file+"; not a file or directory");
                     }
                 }
+            } else {
+                result.add(new URL("file://"+tidiedFile));
```
neykov: Use tidiedFile.toUri.toUrl

utils/common/src/main/java/brooklyn/util/javalang/AggregateClassLoader.java
```diff
@@ -57,24 +61,65 @@ public static AggregateClassLoader newInstanceWithNoLoaders() {
...
+     * @deprecated since 0.7.0 */
+    @Deprecated
     public List<ClassLoader> getList() {
         return classLoaders;
     }
 
+    public Iterator<ClassLoader> iterator() {
+        synchronized (classLoaders) {
+            // provides iterator of snapshot
+            return classLoaders.iterator();
```
neykov: Should it be unmodifiable iterator? Why give access to the internals at all?

utils/common/src/main/java/brooklyn/util/javalang/Threads.java
```diff
@@ -44,7 +46,15 @@ public static boolean removeShutdownHook(Thread hook) {
             return Runtime.getRuntime().removeShutdownHook(hook);
         } catch (IllegalStateException e) {
             // probably shutdown in progress
-            log.debug("cannot remove shutdown hook "+hook+": "+e);
+            String text = Exceptions.collapseText(e);
+            if (text.contains("Shutdown in progress")) {
```
neykov: This is a good candidate for a typed exception. Always wondered why the general dislike to typed exceptions in Brooklyn?

neykov:
Finished review. It's a great set of changes, making the yaml catalog a usable and useful feature. Really happy to see Brooklyn get to this point.

Two points stand out that need addressing:
  * `scanJavaAnnotations` doesn't work with bundle items
  * catalog initialization and HA states interplay

neykov:
Also would be nice if `--app` is treated analogously to `--catalogInit` with regard to rebind state, can be addressed in a separate PR.

usage/cli/src/main/java/brooklyn/cli/Main.java
```diff
@@ -756,9 +781,10 @@ public ToStringHelper string() {
                     .add("bindAddress", bindAddress)
                     .add("noConsole", noConsole)
                     .add("noConsoleSecurity", noConsoleSecurity)
-                    .add("ignorePersistenceErrors", ignorePersistenceErrors)
-                    .add("ignoreWebErrors", ignoreWebErrors)
-                    .add("ignoreAppErrors", ignoreAppErrors)
+                    .add("startupFailOnPersistenceErrors", startupFailOnPersistenceErrors)
+                    .add("startupFailsOnCatalogErrors", startupFailOnCatalogErrors)
+                    .add("startupContinueOnWebErrors", startupContinueOnWebErrors)
+                    .add("startupFailOnManagedAppsErrors", startupFailOnManagedAppsErrors)
```
neykov: Add new catalog commands.

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
ahgittin: yes to `rebindContext`, good spot. as for the level of abstraction it feels right as is to me, with the persister being relatively dumb apart from persisting things, all notions of what to be persisted is controlled from here. it is too bad to lose immutability on `mementoRawData` but the alternative i think would be either to replace it (which would be bad since the rawData might have been passed elsewhere) or analyse the catalog earlier (introducing new phases). i think it makes sense that the rawData can be mutable when needed.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
ahgittin: i see `initialUri` as often being embedded in a script, so that deliberately it only has effect on a new catalog-initializing run. so it is normal for options to be combined to mask it, hence to warning or failure. the log will say where catalog is being loaded from, and so it shouldn't be too hard to work out that the initial uri is not being applied, and the description is clear why it wouldn't be (it's not the initial run).  also `CATALOG_LOAD_MODE` is now deprecated; it had a short shelf life so unlikely this code path will get executed.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
ahgittin: have amended

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
ahgittin: here we get a warning.  i think that's fine for a deprecated option.  have updated message.

core/src/main/java/brooklyn/catalog/internal/CatalogInitialization.java
ahgittin: probably it could, but just in case the properties are set later i wanted to reflect the original logic. this method should be removed soon (as soon as 0.7.0 is out) so not too concerned.

core/src/main/java/brooklyn/entity/rebind/PeriodicDeltaChangeListener.java
ahgittin: +1 that's cleaner, have applied

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
ahgittin: will change the ambiguous `official` and check whether it has been in master state yet

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
ahgittin: that's the right behavior isn't it? it won't persist the catalog clearance, so it's okay, and the instructions were invalid but were followed. what would you have it do instead?

core/src/main/java/brooklyn/entity/rebind/PeriodicDeltaChangeListener.java
ahgittin: agree - have done

utils/common/src/main/java/brooklyn/util/javalang/AggregateClassLoader.java
ahgittin: semantics of `CopyOnWriteList.iterator()` is to take a snapshot, with unmodifiable iterator; have expanded comment

utils/common/src/main/java/brooklyn/util/javalang/Threads.java
ahgittin: i'd prefer a typed exception here, but the exception is thrown by java, not us, if you try to add a shutdown hook while a shutdown is in progress; this is just trying to clean up the log messages which are shown at the tail of the log on a shutdown

nothing against typed exceptions where useful but in general paths which might want typed exception handling i prefer the error-listener / callback pattern, avoiding throwing altogether -- for instance `Maybe.absentThrowing` and `ReferenceWithError`.

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
neykov: My first thought was that if we are resetting the catalog then there's no point in downloading all the bundles and dealing with missing bundles/catalog items, but these cases will result in soft errors (i.e. be logged only), won't abort the rebind.
Only schema changes will abort the rebind, which for catalog items is not so easy to do.

So there's no point in loading the old catalog if resetting, but won't hurt if we do so.

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
ahgittin: i've set it up so we distinguish between `unofficial` initialization, `official final` initialization, and `official transient` initialization.  that last is done if we run rebind but aren't master, and it allows subsequent runs also to apply initialization.  if we are master (or are a non-persisting launch sequence) then we set `official final` and subsequent re-initialization does *not* apply of the parameters.

added quite a bit more doc around it as well.  unfortunate it is so complex but i'm not sure of a way around it, unless we distinguished things you could do on first run from on subsequent run.

(the edge cases are very edge, as most CLI uses should be when there is no brooklyn running; but i think we do now handle them nicely.)

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
ahgittin: i see -- we could get failures while loading previously persisted catalog items, even if we are resetting.  and yes this is an argument that the right thing to do, if we are going to reset the catalog anyway, is to clear things out earlier or skip the catalog item load.  let me look at this.

ahgittin: agree with your synopsis. have added comments:

                // we will have unnecessarily tried to load the catalog item manifests earlier in this iteration,
                // and problems there could fail a rebind even when we are resetting;
                // it might be cleaner to check earlier whether a reset is happening and not load those items at all,
                // but that would be a significant new code path (to remove a directory in the persistent store, essentially),
                // and as it stands we don't do much with those manifests (e.g. we won't register them or fail on missing types)
                // so we think it's only really corrupted XML or CatalogItem schema changes which would cause such problems.
                // in extremis someone might need to wipe their store but for most purposes i don't think there will be any issue
                // with loading the catalog item manifests before wiping all those files.

ahgittin:
@neykov Thanks for excellent comments.  The last commit addresses them, I think.  Code is clearer now, even if still not as simple as I'd like.  I have done a fair bit of testing with HA modes and catalog initialization and no surprises :) .

neykov:
I don't see changes related to `scanJavaAnnotations` and bundles. Is this working now?

ahgittin:
@neykov `scanJavaAnnotations` addressed in https://github.com/ahgittin/incubator-brooklyn/commit/9cc7293eb45efd045edd81793db31fa8b6f64110 (mainly with comments and docs however as not sure how far we want to go supporting load-time scan for this, as opposed a dev/compile-time manifest -- it is still an experimental feature)

neykov:
Saw the commit and the change are good, but my reading of the code and subsequent testing shows that `scanJavaAnnotations` doesn't work for bundles. Could you add a test to prove me wrong :)

ahgittin:
Thanks for keeping me straight @neykov .

I'll add your test, and try to fix access to the brooklyn classpath:

```
brooklyn.catalog:
- libraries:
  - file:///.../core/src/test/resources/brooklyn/osgi/brooklyn-test-osgi-entities.jar
  scanJavaAnnotations: true
```

ahgittin:
@neykov test added. the issue wasn't in the scanner - it could see brooklyn types - but in transferring the metadata including libraries into to the scanner-generated dto. hackish in places but happy path now works, while we figure out how much to continue with this experimental feature. good catch @neykov!

core/src/test/java/brooklyn/catalog/internal/CatalogScanTest.java
```diff
@@ -161,13 +162,14 @@ public void testMoreTypesThanAnnotationsForApps() {
     
     @Test
     public void testAnnotationIsDefault() {
```
neykov: This test started failing for me after the merge of #634 
Seems to be caused by me having a default catalog in ~/.brooklyn which replaces the `default.catalog.bom` expected here.

neykov:
Finished review of new commits. Couldn't understand why you didn't prefer the `URLStreamHandler` approach for fetching the classpath:// URLs. Seems cleaner.
Mostly minor comments, can be merged after `testAnnotationIsDefault` can handle user catalogs.

ahgittin:
@neykov I think these are all fixed.  The test case now does not depend on `~/.brooklyn/`.  Can you confirm it works for you?

The `URLConnection` class is an ugly beast, that's why I backed out of the `URLStreamHandler` approach.

neykov:
Tests passing now, let's merge this beast :).

PR #619, opened 2015-04-30
add support to control java version installed

andreaturli:
- add more control to the java version installed for the entities that UsesJava

### DIFF
--- a/software/base/src/main/java/brooklyn/entity/java/JavaSoftwareProcessSshDriver.java
+++ b/software/base/src/main/java/brooklyn/entity/java/JavaSoftwareProcessSshDriver.java
@@ -30,6 +30,17 @@ import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Function;
+import com.google.common.base.Joiner;
+import com.google.common.base.Optional;
+import com.google.common.base.Splitter;
+import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.google.gson.internal.Primitives;
+
 import brooklyn.entity.basic.AbstractSoftwareProcessSshDriver;
 import brooklyn.entity.basic.Attributes;
 import brooklyn.entity.basic.Entities;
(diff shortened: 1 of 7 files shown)


software/base/src/main/java/brooklyn/entity/java/JavaSoftwareProcessSshDriver.java
```diff
@@ -272,69 +273,24 @@ public String getJmxContext() {
...
-            log.debug("Java version {} already installed at {}@{}", new Object[]{version.get(), getEntity(), getLocation()});
-            return true;
-        } else {
-            // Let's hope not!
-            if (version.isPresent()) {
-                log.debug("Found old Java version {} on {}@{}. Going to install latest Java version.",
-                        new Object[]{version.get(), getEntity(), getLocation()});
+    protected boolean checkForAndInstallJava(String requiredVersion) {
+        List<String> requiredVersionParts = Splitter.on(".").splitToList(requiredVersion);
+        int requiredJavaMinor = Integer.valueOf(requiredVersionParts.get(1));
```
neykov: Worth checking the length in case invalid format was supplied with more information exception message.

neykov:
Looks good and very useful!
Wondering if we should support the single number style versioning as well, i.e. 6,7,8?

andreaturli:
thanks @neykov I was thinking about the same but java -version returns something like `"1.7.0_71"` so looked to me more consistent

software/base/src/main/java/brooklyn/entity/java/JavaSoftwareProcessSshDriver.java
```diff
@@ -360,7 +316,7 @@ protected int tryJavaInstall(String version, String command) {
      * Checks for the version of Java installed on the entity's location over SSH.
      * @return An Optional containing the version portion of `java -version`, or absent if no Java found.
      */
-    protected Optional<String> getCurrentJavaVersion() {
+    protected Optional<String> getInstalledJavaVersion() {
```
sjcorbett: `getCurrentJavaVersion` should be deprecated rather than removed.

software/base/src/main/java/brooklyn/entity/java/JavaSoftwareProcessSshDriver.java
sjcorbett: +1. I think it should also log a warning in case someone gives a version like 1.7.XX - it will be a surprise when the XX portion is ignored.

ahgittin: java versioning is bizarre, eh?  whoever heard of java `1.7.1` so i think it's okay to ignore the `XX`.  as @neykov says supporting `7` as input might be nice but also very minor.

software/base/src/main/java/brooklyn/entity/java/JavaSoftwareProcessSshDriver.java
ahgittin: @andreaturli strongly agree, this class is often subclassed so let's ensure we maintain api compatibility even on protected methods.  apart from this good to merge in my view.

software/base/src/main/java/brooklyn/entity/java/JavaSoftwareProcessSshDriver.java
andreaturli: better to support `1.7`, `7` or both?

ahgittin: 1.7 is necessary and sufficient.  7 is a maybe-nice-to-have i'd say.

--A

sjcorbett:
@andreaturli Could you resolve the merge conflicts?

andreaturli:
thanks @sjcorbett, conflicts merged, squashed and rebased

sjcorbett:
@andreaturli Thanks. Will merge.

PR #620, opened 2015-04-30
Don't retry jclouds requests on exception if unsafe request method

neykov:
The request could have already been processed on the server when the exception occurs, retry only for request methods without side effects.

Any ideas how this could be tested?

Make sense to use the change for all providers, though the problem has been observed on Softlayer specifically: `Read timed out connecting to POST https://api.softlayer.com/rest/v3/SoftLayer_Virtual_Guest`. Because of the retries a machine has been provisioned 6 times.

@andreaturli do you think this should be filed as a bug report/patch in jclouds?


andreaturli:
@neykov I think it makes more sense to apply this fix at jclouds level, as Brooklyn should expect jclouds to take care of re-tries and other low-level stuff.

Great work by the way, mind open a PR to jclouds or happy to it myself?

neykov:
Thanks @andreaturli. I'll PR it against jclouds, let's keep this open until it is resolved there.

neykov:
Corresponding jclouds PR:
https://github.com/jclouds/jclouds/pull/744

neykov:
jclouds PR https://github.com/jclouds/jclouds/pull/744 merged, so closing this. Changes will be available in next 1.9 release.

PR #621, opened 2015-04-30
Wildfly 8 Support (Partial)

mikezaccardo:
Wildfly 8 is now partially supported.  `brooklyn.entity.webapp.jboss.WildflyServer` can now be used in place of `brooklyn.entity.webapp.jboss.JBoss7Server` in the example blueprints.

However, Wildfly 8's Undertow subsystem does not yet provide several required statistics that are used by certain policies -- for example, an auto-scaling policy configured to use the requests-per-second metric would not yet work.

The JBoss / Wildfly live tests have been refactored and Wildfly 8 live tests have been added (integration tests not yet added).  Note that they will not yet pass due to the missing metrics.

See: https://issues.apache.org/jira/browse/BROOKLYN-142


software/webapp/src/main/java/brooklyn/entity/webapp/jboss/Wildfly8Server.java
```diff
@@ -0,0 +1,101 @@
...
+    @SetFromFlag("managementHttpsPort")
+    PortAttributeSensorAndConfigKey MANAGEMENT_HTTPS_PORT =
+            new PortAttributeSensorAndConfigKey("webapp.jboss.managementHttpsPort", "Management port", "9443+");
+
+    @SetFromFlag("managementNativePort")
+    PortAttributeSensorAndConfigKey MANAGEMENT_NATIVE_PORT =
+            new PortAttributeSensorAndConfigKey("webapp.jboss.managementNativePort", "Management native port", "10999+");
+
+    /**
+     * Port increments are the standard way to run multiple instances of AS7 on the same machine.
```
aledsage: Wildlfy, rather than AS7.

software/webapp/src/main/java/brooklyn/entity/webapp/jboss/Wildfly8ServerImpl.java
```diff
@@ -0,0 +1,196 @@
...
+
+    public static final Logger log = LoggerFactory.getLogger(Wildfly8ServerImpl.class);
+
+    private volatile HttpFeed httpFeed;
+
+    public Wildfly8ServerImpl(){
+        super();
+    }
+
+    public Wildfly8ServerImpl(@SuppressWarnings("rawtypes") Map flags){
```
aledsage: Don't need these constructors (just the no-arg constructor, so can miss them out entirely).

We only have them in other (older) entities for backwards compatibility reasons, where folk called the constructors directly. Now everything should go through the EntitySpec mechanism for creating an entity.

software/webapp/src/main/java/brooklyn/entity/webapp/jboss/Wildfly8ServerImpl.java
```diff
@@ -0,0 +1,196 @@
...
+    
+    public int getManagementNativePort() {
+        return getAttribute(MANAGEMENT_NATIVE_PORT);
+    }
+    
+    public int getPortOffset() {
+        return getConfig(PORT_INCREMENT);
+    }
+    
+    public boolean isWelcomeRootEnabled() {
```
aledsage: What is this for? Can't see it used anywhere (including not from freemarker).

software/webapp/src/main/java/brooklyn/entity/webapp/jboss/Wildfly8ServerImpl.java
```diff
@@ -0,0 +1,196 @@
...
+    public String getManagementBindAddress() {
+        return getConfig(BIND_ADDRESS);
+    }
+    
+    public String getUnsecureBindAddress() {
+        return getConfig(BIND_ADDRESS);
+    }
+    
+    // If empty-string, disables Management security (!) by excluding the security-realm attribute
+    public String getHttpManagementInterfaceSecurityRealm() {
```
aledsage: Appears not to be used. I see you took it from the JBoss7 impl. Not sure what uses that one either!

software/webapp/src/main/java/brooklyn/entity/webapp/jboss/Wildfly8SshDriver.java
```diff
@@ -0,0 +1,249 @@
...
+
+        newScript(INSTALLING)
+                // don't set vars yet -- it resolves dependencies (e.g. DB) which we don't want until we start
+                .environmentVariablesReset()
+                .body.append(commands)
+                .execute();
+    }
+
+    /**
+     * AS7 config notes and TODOs:
```
aledsage: Are all of these applicable to Wildfly as well?

software/webapp/src/main/java/brooklyn/entity/webapp/jboss/Wildfly8SshDriver.java
```diff
@@ -0,0 +1,249 @@
...
+    @Override
+    public void kill() {
+        newScript(MutableMap.of(USE_PID_FILE, true), KILLING).execute();
+    }
+
+    @Override
+    protected List<String> getCustomJavaConfigOptions() {
+        return MutableList.<String>builder()
+                .addAll(super.getCustomJavaConfigOptions())
+                .add("-Xms200m")
```
aledsage: I'm fine with this for now because it's the same as JBoss7Server...

However, would be good if this was easier to control via config. If someone wanted a different Xmx then they'd currently (probably?) have to write code to override this, rather than being able to do it in YAML.

For example, maybe we should look at using `UsesJava.JAVA_OPTS` (here, in Tomcat, and in a whole bunch of other entities).

Let's not do that in this PR though!

software/webapp/src/test/java/brooklyn/entity/webapp/WebAppLiveIntegrationTest.groovy
```diff
@@ -81,7 +83,8 @@ public class WebAppLiveIntegrationTest {
         TomcatServer tomcat = new TomcatServerImpl(parent:application, httpPort:DEFAULT_HTTP_PORT, jmxPort:DEFAULT_JMX_PORT)
         JBoss6Server jboss6 = new JBoss6ServerImpl(parent:application, portIncrement:PORT_INCREMENT, jmxPort:DEFAULT_JMX_PORT)
         JBoss7Server jboss7 = new JBoss7ServerImpl(parent:application, httpPort:DEFAULT_HTTP_PORT, jmxPort:DEFAULT_JMX_PORT)
-        return [ [ tomcat ], [ jboss6 ], [ jboss7 ] ]
+		Wildfly8Server wildfly8 = new Wildfly8ServerImpl(parent:application, httpPort:DEFAULT_HTTP_PORT, jmxPort:DEFAULT_JMX_PORT)
```
aledsage: Ah, I see we still do have code that calls entity constructors directly (and that wants to do it for Wildfly8ServerImpl).

We should convert this class to Java, and see about using EntitySpec. But don't want to add yet more to this PR so let's address this another time.

software/webapp/src/test/java/brooklyn/entity/webapp/jboss/JBoss6ServerAwsEc2LiveTest.java
```diff
@@ -18,55 +18,32 @@
...
+    @Override
+    public void test_CentOS_6_3() throws Exception {
+        super.test_CentOS_6_3();
+    }
+
     @Test(enabled=false)
     public void testDummy() {} // Convince testng IDE integration that this really does have test methods  
+
+	@Override
+	protected Class<? extends JavaWebAppSoftwareProcess> getServerType() {
```
aledsage: strange indentation

software/webapp/src/test/java/brooklyn/entity/webapp/jboss/Wildfly8ServerAwsEc2LiveTest.java
```diff
@@ -0,0 +1,49 @@
...
+public class Wildfly8ServerAwsEc2LiveTest extends JBossServerAwsEc2LiveTest {
+
+    @Override
+    protected void doTest(Location loc) throws Exception {
+    	super.doTest(loc);
+    }
+    
+    @Test(groups = {"Live", "Live-sanity"})
+    @Override
+    public void test_CentOS_6_3() throws Exception {
```
aledsage: Will this fail, because `JBossServerAwsEc2LiveTest` asserts things like `assertNotNull(server.getAttribute(WebAppServiceMetrics.REQUEST_COUNT)`. But in Wildfly8ServerImpl it says "TODO Re-enable these metrics once they are supported by Wildfly", and does not set these metrics.

If so, worth ensuring the tests are marked in the "WIP" group. And having a comment to explain why they'll fail.

software/webapp/src/test/java/brooklyn/entity/webapp/jboss/Wildfly8ServerAwsEc2LiveTest.java
```diff
@@ -0,0 +1,49 @@
...
+        super.test_CentOS_6_3();
+    }
+
+    @Test(enabled=false)
+    public void testDummy() {} // Convince testng IDE integration that this really does have test methods  
+
+	@Override
+	protected Class<? extends JavaWebAppSoftwareProcess> getServerType() {
```
aledsage: strange indent

software/webapp/src/test/java/brooklyn/entity/webapp/jboss/JBoss7ServerGoogleComputeLiveTest.java
```diff
@@ -0,0 +1,49 @@
...
+    @Override
+    public void test_DefaultImage() throws Exception {
+        super.test_DefaultImage();
+    }
+
+    @Test(enabled=false)
+    public void testDummy() {} // Convince testng IDE integration that this really does have test methods
+    
+	@Override
+	protected Class<? extends JavaWebAppSoftwareProcess> getServerType() {
```
aledsage: strange indent

software/webapp/src/test/java/brooklyn/entity/webapp/jboss/JBossPasswordHashingTest.java
```diff
@@ -18,42 +18,44 @@
...
  * Expected values in tests were generated by AS7's add-user.sh script and copied here.
  */
-public class JBoss7PasswordHashingTest {
+public class JBossPasswordHashingTest {
 
     @Test
     public void testPasswordForManagementRealm() {
         assertEquals(
-                JBoss7SshDriver.hashPassword("username", "password", "ManagementRealm"),
+        		JavaWebAppSshDriver.hashPassword("username", "password", "ManagementRealm"),
```
aledsage: Strange indent - tabs instead of spaces perhaps?

software/webapp/src/test/java/brooklyn/entity/webapp/jboss/Wildfly8ServerDockerLiveTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.testng.annotations.Test;
+
+import brooklyn.entity.webapp.JavaWebAppSoftwareProcess;
+import brooklyn.location.Location;
+
+/**
+ * A simple test of installing+running Wildfly 8 on Docker, using various OS distros and versions.
+ */
+public class Wildfly8ServerDockerLiveTest extends JBossServerDockerLiveTest {
```
aledsage: As with the Wildfly AWS test, will this fail because the attributes aren't being set? In which case we should include a comment explaining why.

Could also parameterise the `super.doTest()` to say whether to assert those attributes are non-null.

software/webapp/src/test/java/brooklyn/entity/webapp/jboss/Wildfly8ServerGoogleComputeLiveTest.java
```diff
@@ -0,0 +1,49 @@
...
+    @Override
+    public void test_DefaultImage() throws Exception {
+        super.test_DefaultImage();
+    }
+
+    @Test(enabled=false)
+    public void testDummy() {} // Convince testng IDE integration that this really does have test methods
+    
+	@Override
+	protected Class<? extends JavaWebAppSoftwareProcess> getServerType() {
```
aledsage: strange indent

aledsage:
Thanks @mikezaccardo - looking good. A few minor comments, some of which should be addressed and many of which should be deferred.

You said in the description that the live tests "will not yet pass due to the missing metrics". Given we'll not get those metrics out of Wildfly in the short term, is there something we should do to get the live tests passing? For example, parameterise the `doTest()` method to say whether to assert the attributes have been set?

Would be good to have some tests that pass, so we can regression test this after it's been merged.

software/webapp/src/main/java/brooklyn/entity/webapp/jboss/Wildfly8SshDriver.java
mikezaccardo: Created ticket: https://issues.apache.org/jira/browse/BROOKLYN-152

software/webapp/src/main/java/brooklyn/entity/webapp/jboss/Wildfly8SshDriver.java
mikezaccardo: Will investigate further!

mikezaccardo:
As per discussion with @aledsage, this is not going to be merged just yet.  I will work adding tests that will pass given the incomplete state of Wildfly's metrics, and cleanup whatever else is deemed necessary.

software/webapp/src/test/java/brooklyn/entity/webapp/jboss/Wildfly8ServerDockerLiveTest.java
mikezaccardo: Moved all to WIP with explanation.

software/webapp/src/test/java/brooklyn/entity/webapp/jboss/JBossPasswordHashingTest.java
mikezaccardo: Yes since writing this code I changed my Eclipse settings so this won't happen in the future.

software/webapp/src/test/java/brooklyn/entity/webapp/jboss/Wildfly8ServerAwsEc2LiveTest.java
mikezaccardo: Done

software/webapp/src/test/java/brooklyn/entity/webapp/WebAppLiveIntegrationTest.groovy
mikezaccardo: Added TODO

PR #622, opened 2015-05-01
Revert the eager loading of the catalog items

bostko:
Currently loading the documentation of a catalog item is little slow. I think for the catalog items particularly, there is no reason for lazy loading the resource.
Loading via ajax also doesn't allow the browser to cache the js resource and it is reloaded each time when switching back and forth between index.html and catalog-item.html

The other change I made is to include the catalog items generation in the documentation build


docs/_build/build.sh
```diff
@@ -179,9 +179,20 @@ function test_site() {
 }
 
 function make_jekyll() {
+  if [ -f ../usage/dist/target/brooklyn-dist/bin/brooklyn ]; then
+    ITEMS_JS=style/js/catalog/items.js
```
sjcorbett: Does `items.js` need to be checked in if it is generated by this function?

bostko: Yes it needs to be putted in the project

sjcorbett:
@bostko Thanks. Will merge.

PR #624, opened 2015-05-04
Don't fail on startup in case of an isRunning exception

neykov:
During entity startup treat isRunning exceptions as temporary failures and wait until timeout, don't propagate them on the spot. Could be caused by a temporary problem, for example connectivity issues.

### DIFF
--- a/software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
+++ b/software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
@@ -494,17 +494,26 @@ public abstract class SoftwareProcessImpl extends AbstractEntity implements Soft
         CountdownTimer timer = startTimeout.countdownTimer();
         boolean isRunningResult = false;
         long delay = 100;
+        Exception firstFailure = null;
         while (!isRunningResult && !timer.isExpired()) {
             Time.sleep(delay);
             try {
                 isRunningResult = driver.isRunning();
+                if (log.isDebugEnabled()) log.debug("checked {}, 'is running' returned: {}", this, isRunningResult);
             } catch (Exception  e) {
-                ServiceStateLogic.setExpectedState(this, Lifecycle.ON_FIRE);
-                // provide extra context info, as we're seeing this happen in strange circumstances
-                if (driver==null) throw new IllegalStateException(this+" concurrent start and shutdown detected");
-                throw new IllegalStateException("Error detecting whether "+this+" is running: "+e, e);
+                Exceptions.propagateIfFatal(e);
+
+                isRunningResult = false;
+                if (driver != null) {
+                    log.error("checked " + this + ", 'is running' threw an exception", e);
+                } else {
+                    // provide extra context info, as we're seeing this happen in strange circumstances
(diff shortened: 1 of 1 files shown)


software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
```diff
@@ -494,17 +494,24 @@ public void waitForEntityStart() {
...
             try {
                 isRunningResult = driver.isRunning();
+                if (log.isDebugEnabled()) log.debug("checked {}, 'is running' returned: {}", this, isRunningResult);
             } catch (Exception  e) {
-                ServiceStateLogic.setExpectedState(this, Lifecycle.ON_FIRE);
-                // provide extra context info, as we're seeing this happen in strange circumstances
-                if (driver==null) throw new IllegalStateException(this+" concurrent start and shutdown detected");
-                throw new IllegalStateException("Error detecting whether "+this+" is running: "+e, e);
+                Exceptions.propagateIfFatal(e);
+                firstFailure = e;
```
sjcorbett: `firstFailure` would be better named `lastFailure`.

software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
```diff
@@ -494,17 +494,24 @@ public void waitForEntityStart() {
...
         boolean isRunningResult = false;
         long delay = 100;
+        Exception firstFailure = null;
         while (!isRunningResult && !timer.isExpired()) {
             Time.sleep(delay);
             try {
                 isRunningResult = driver.isRunning();
+                if (log.isDebugEnabled()) log.debug("checked {}, 'is running' returned: {}", this, isRunningResult);
             } catch (Exception  e) {
-                ServiceStateLogic.setExpectedState(this, Lifecycle.ON_FIRE);
```
sjcorbett: What are the semantics of expected state being `ON_FIRE` vs `RUNNING`? Previously we set the former and threw, with this it's the latter.

software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
neykov: Good spot, prefer to catch the first one, so will add a guarding check.

software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
neykov: I'd like to treat exceptions from `isRunning` analogously to returning false. This means setting the expected state to `RUNNING` and the actual state (coming from the enrichers) to 'ON-FIRE' (because of the failing check).

neykov: Essentially this (on purpose) changes the behaviour from treating `isRunning` failures as fatal to treating them as temporary.

neykov:
Fixed `firstFailure` initialization, logging subsequent failures at `debug` level.

sjcorbett:
@neykov Thanks. Will merge.

PR #625, opened 2015-05-04
Support URLs in brooklyn.webconsole.security.keystore.url property

neykov:
Even though the property is named url it supported local files only.


sjcorbett:
Looks good.

PR #627, opened 2015-05-05
White list strong TLS ciphers

neykov:
Default config uses weak/compromised ciphers. Followed recommendations from https://wiki.mozilla.org/Security/Server_Side_TLS.

Upgrade of Jetty was needed for the support of white-listing (previous version supported black lists only).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -144,7 +144,7 @@
         <sshj.version>0.8.1</sshj.version>
         <felix.framework.version>4.4.0</felix.framework.version>
         <reflections.version>0.9.9-RC1</reflections.version>
-        <jetty.version>8.1.4.v20120524</jetty.version>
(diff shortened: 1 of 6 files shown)


neykov:
IBM java supports only SSL_ prefix for ciphers, repeated the list with SSL_ prefix as well.

The first part of the cipher name specifies the protocol, but in practice the cipher can be used with either protocol. Java uses the TLS cipher suite naming convention, but IBM java uses only SSL_ prefix - https://www-01.ibm.com/support/knowledgecenter/SSYKE2_7.0.0/com.ibm.java.security.component.70.doc/security-component/jsse2Docs/ciphersuites.html

ahgittin:
nice improvements.  merging.

PR #628, opened 2015-05-05
respect disabling root flag and test for it

ahgittin:


ggrunin:
+1 tested the fix - it works

PR #629, opened 2015-05-05
Add initial support for PaaS locations

mbarrientos:
An initial support for PaaS locations is provided by adding a new inference rule to [ReflectiveEntityDriverFactory](https://github.com/mbarrientos/incubator-brooklyn/blob/f32fa3cff2174dae18758533e51bddb482eef2c2/core/src/main/java/brooklyn/entity/drivers/ReflectiveEntityDriverFactory.java). The new location [PaasLocation](https://github.com/apache/incubator-brooklyn/blob/f32fa3cff2174dae18758533e51bddb482eef2c2/core/src/main/java/brooklyn/location/paas/PaasLocation.java) is not based on SSH access and is not supposed to be provisioned first (even though it might be the case for some PaaS provider). 

Note that this is part of an ongoing work for integrating PaaS providers, like CloudFoundry, as a Brooklyn Location, which process can be tracked [here](https://github.com/SeaCloudsEU/brooklyn-location-cloudfoundry). That work will be extending the one presented in this PR.

### DIFF
--- a/core/src/test/java/brooklyn/entity/drivers/ReflectiveEntityDriverFactoryTest.java
+++ b/core/src/test/java/brooklyn/entity/drivers/ReflectiveEntityDriverFactoryTest.java
@@ -52,11 +52,6 @@ public class ReflectiveEntityDriverFactoryTest {
     public void tearDown() {
         // nothing to tear down; no management context created
     }
-
-    protected void assertDriverIs(Class<?> clazz) {
-        MyDriver driver = factory.build(entity, sshLocation);
-        assertTrue(driver.getClass().equals(clazz), "driver="+driver+"; should be "+clazz);
-    }
     
     protected void assertDriverIs(Class<?> clazz, Location location) {
         MyDriver driver = factory.build(entity, location);
@@ -65,7 +60,7 @@ public class ReflectiveEntityDriverFactoryTest {
     
     @Test
     public void testInstantiatesSshDriver() throws Exception {
-        assertDriverIs(MySshDriver.class);
+        assertDriverIs(MySshDriver.class, sshLocation);
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


core/src/test/java/brooklyn/entity/drivers/ReflectiveEntityDriverFactoryTest.java
```diff
@@ -54,12 +58,22 @@ protected void assertDriverIs(Class<?> clazz) {
         assertTrue(driver.getClass().equals(clazz), "driver="+driver+"; should be "+clazz);
     }
     
+    protected void assertDriverIs(Class<?> clazz, Location location) {
```
robertgmoss: why not eliminate duplicate code and use this method for the existing test, too?

core/src/test/java/brooklyn/location/basic/PaasLocationTest.java
```diff
@@ -0,0 +1,37 @@
...
+import brooklyn.location.paas.PaasLocation;
+import brooklyn.management.ManagementContext;
+import brooklyn.test.location.TestPaasLocation;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class PaasLocationTest {
+
+    private PaasLocation location;
+    private ManagementContext mgmt;
```
robertgmoss: This ManagementContext field is not used

core/src/test/java/brooklyn/entity/drivers/ReflectiveEntityDriverFactoryTest.java
mbarrientos: @robertgmoss I agree. I didn't wanted to include many changes on the rest of tests but that's cleaner.

andreaturli:
@robertgmoss do you think it is good to merge now?

andreaturli:
thanks @robertgmoss 

Thanks @mbarrientos! Looking forward for the other PRs

PR #631, opened 2015-05-06
Fix version conflicts in brooklyn-downstream-parent

rdowner:
Previously caused an issue if an external project inherits from brooklyn-downstream-parent and depends on brooklyn-rest-client, as these had conflicting jersey.version properties.

First commit reorganises version properties in root pom.xml to make it easier to compare with brooklyn-downstream-parent pom.xml, but does not *change* any versions.

Second commit updates brooklyn-downstream-parent pom.xml with the same versions as the root pom.xml.


sjcorbett:
Looks good to me.

PR #632, opened 2015-05-07
switch docs to kramdown and replace triple backticks with triple tilde

ahgittin:
tilde is more portable, but apart from that kramdown is more widely used,
and it permits mixing html which we do on the brooklyn landing page
(with redcarpet that didn't render right)

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -18,7 +18,7 @@
 #
 
 encoding: utf-8
-markdown: redcarpet
+markdown: kramdown
 
 # where this will publish
 url_root: http://0.0.0.0:4000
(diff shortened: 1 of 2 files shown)


sjcorbett:
@ahgittin Looks fine to merge. Out of curiosity, what is the difference between backticks and tildes?

ahgittin:
@sjcorbett no difference where both are supported but the triple backtick is a githubbism and not so widely supported while triple tilde is supported everywhere afaik

(personally i prefer the triple backtick, but better is having a processor which allows us to write docs which parse nicely when viewed in the source tree at github.com and which display nicely when rendered.)

PR #633, opened 2015-05-08
Add Tomcat 8 Support

mikezaccardo:


sjcorbett:
@mikezaccardo Thanks for this. What are the differences in the entities between Tomcat 7 and 8? From a scan, there's not much - the new server.xml modifies the AccessLogValve prefix and removes this line:
```xml
  <Listener className="org.apache.catalina.core.JasperListener" />
```
The different web.xml file changes the file's schemaLocation and version.

I don't think this warrants a separate package for v8.

I wonder if a simpler approach could maintain the package structure and reuse the existing classes: 
* Move server.xml and web.xml to ```.tomcat``` and rename them to something like ~ tomcat8-server.xml and tomcat8-web.xml.
* Add a `Tomcat8Server` interface that extends `TomcatServer` and redefines `SERVER_XML_RESOURCE`, `WEB_XML_RESOURCE` and the appropriate download/version keys. e.g.:
```java
    @SetFromFlag("server.xml")
    ConfigKey<String> SERVER_XML_RESOURCE = ConfigKeys.newConfigKeyWithDefault(
            TomcatServer.SERVER_XML_RESOURCE, "tomat8-server.xml");
```
The existing driver will handle naming the file correctly when deploying entities.
* Add a `Tomcat8ServerImpl` class that extends `TomcatServerImpl` and implements `Tomcat8Server`. It doesn't need to override any existing methods.

Having compared the v7 files vs. the v8 files - but not tested! - I think this would work and saves the duplication.

Any thoughts?

sjcorbett:
Alternatively we could follow the same pattern as `CassandraNode` and have logic in the TomcatServer entity itself to distinguish between the files to use for versions 7 and 8.

mikezaccardo:
Thanks for looking this over @sjcorbett.  

Yes, the removal of `JasperListener` from `server.xml` is the main change.

I agree about the dubious creation of a `tomcat8` package.  In fact, I originally attempted to keep them in the same package and made the exact changes to `Tomcat8Server` that you described, with similarly named new xml files, but deployments failed.  Ultimately I made the new package and kept the xml files with the same name, which worked.  I will revisit keeping it the same package because both of our instincts are the same.

mikezaccardo:
I will investigate this `CassandraNode` pattern if the same-package-xml-rename does not work, thanks.

sjcorbett:
@mikezaccardo Happy to collaborate or test things out for you if deployments are failing.

mikezaccardo:
@sjcorbett I've followed your advice about `tomcat` package consolidation; there is no longer a separate `tomcat8` package.  Thanks!

I figured out the reason why my `tomcat8-server.xml` and `tomcat8-web.xml` attempt originally failed: I changed the `String` values of the xml files in lines 77 and 78 of `TomcatSshDriver.java` (renamed from `Tomcat7SshDriver.java`) to their new names which is incorrect.

sjcorbett:
@mikezaccardo Thanks again. A few more points:

* We should include empty `Tomcat7Driver` and `Tomcat7SshDriver` classes that are annotated `@Deprecated` and extend the new sans-7 classes to give leeway to anyone who has subclassed the old versions.
* I've run classes matching `Tomcat8*IntegrationTest`. All passed but for `Tomcat8ServerSimpleIntegrationTest.detectFailureIfTomcatCantBindToPort`, which failed twice after throwing `java.lang.IllegalStateException: Timeout waiting for SERVICE_UP from Tomcat8ServerImpl{id=HH3Gpo2G}`
* Lastly, it's worth squashing your commits into one before we merge.

mikezaccardo:
Thanks @sjcorbett -- I've added the empty classes, added a comment about the integration test failure (on Tomcat 7 and 8), created a JIRA ticket, and squashed the commits.

sjcorbett:
@mikezaccardo Super. Thanks.

mikezaccardo:
:tada:

PR #634, opened 2015-05-08
Jclouds hostnames etc

ahgittin:
bunch of cleanups in jclouds around naming of hosts and domains, and a few misc fixes


ahgittin:
NB: builds on #617 due to merge conflicts; it's the last 6 commits which are new

ahgittin:
unrelated failure:

```
```

kicking off anew

core/src/main/java/brooklyn/catalog/CatalogLoadMode.java
```diff
@@ -20,8 +20,11 @@
 
 import org.slf4j.LoggerFactory;
 
+import brooklyn.catalog.internal.CatalogInitialization;
 import brooklyn.entity.rebind.persister.PersistMode;
 
+/** @deprecated since 0.7.0 replaced by {@link CatalogInitialization} */
```
sjcorbett: I think this class has been added _and_ deprecated in 0.7.0. Can we just get rid of it?

ahgittin: given how long it has been since the `0.6.0` release i tend not to remove things which have been in place for more than a few weeks.  (hopefully soon we *will* be releasing on that cadence...)

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -596,21 +599,35 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
...
+                    Object context = setup.get(CALLER_CONTEXT);
+                    if (context instanceof Entity) {
+                        Entity entity = (Entity)context;
+                        template.getOptions().getUserMetadata().put("brooklyn-app-id", entity.getApplicationId());
+                        template.getOptions().getUserMetadata().put("brooklyn-app-name", entity.getApplication().getDisplayName());
+                        template.getOptions().getUserMetadata().put("brooklyn-entity-id", entity.getId());
+                        template.getOptions().getUserMetadata().put("brooklyn-entity-name", entity.getDisplayName());
+                        template.getOptions().getUserMetadata().put("brooklyn-server-creation-date", Time.makeDateSimpleStampString());
+                    }
                 }
```
andreaturli: I can see why you are doing that but for providers like SoftLayer we are probably (mis-)using this `userMetadata`
Look at [userData.value](http://sldn.softlayer.com/reference/services/SoftLayer_Virtual_Guest/createObject) it is primaily used to pass data to the `postInstallScriptUri` 

I think we need to consider `templateOptions.tags` instead as it looks more portable in jclouds. Thoughts?
Wdyt?

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -596,21 +599,35 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
...
 
-                //FIXME initialCredentials = initUserTemplateOptions(template, setup);
-                for (JcloudsLocationCustomizer customizer : getCustomizers(setup)) {
-                    customizer.customize(this, computeService, template);
-                    customizer.customize(this, computeService, template.getOptions());
+                templateTimestamp = Duration.of(provisioningStopwatch);
+                // "Name" metadata seems to set the display name; at least in AWS
+                // TODO it would be nice if this salt comes from the location's ID (but we don't know that yet as the ssh machine location isn't created yet)
+                // TODO in softlayer we want to control the suffix of the hostname which is 3 random hex digits
+                template.getOptions().getUserMetadata().put("Name", cloudMachineNamer.generateNewMachineUniqueNameFromGroupId(setup, groupId));
```
andreaturli: looks like the `template.getOptions()` offers also `nodeNames` which is used for `aws` api to assign the `name` to the NodeMetadata. so maybe this also answer the comment 

// "Name" metadata seems to set the display name; at least in AWS

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
ahgittin: SL userdata is not metadata.  i think user metadata is different, possibly ignored in softlayer?  it does not seem to end up on a VM disk like userdata does.

in any case user metadata isn't visible in the gui for softlayer, so i added a tweak there to convert user metadata to tags.  i think that's best for clouds which don't support user metadata.  (would be nice if jclouds could communicate this!)

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
ahgittin: good to know

PR #636, opened 2015-05-11
Initialize entity driver on rebind when on-fire

neykov:
Don't create the driver only when `SERVICE_STATE_EXPECTED=ON_FIRE` - the entity is considered permanently failed in this state. When only `SERVICE_STATE_ACTUAL=ON_FIRE` we still should be able to rebind successfully - failure was set by attached enrichers.

### DIFF
--- a/software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
+++ b/software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
@@ -31,11 +31,10 @@ import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.collect.Sets;
-
 import brooklyn.config.ConfigKey;
 import brooklyn.enricher.basic.AbstractEnricher;
 import brooklyn.entity.Entity;
+import brooklyn.entity.basic.Lifecycle.Transition;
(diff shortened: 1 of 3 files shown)


software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
```diff
@@ -350,12 +349,19 @@ public void onManagementStarted() {
...
     @Override
     public void rebind() {
-        Lifecycle state = getAttribute(SERVICE_STATE_ACTUAL);
-        if (state == null || state != Lifecycle.RUNNING) {
-            log.warn("On rebind of {}, not calling software process rebind hooks because state is {}", this, state);
+        //SERVICE_STATE_ACTUAL might be ON_FIRE due to a temporary condition (problems map non-empty)
+        //Only if the expected state is ON_FIRE then the entity has permanently failed.
+        Transition expectedState = getAttribute(SERVICE_STATE_EXPECTED);
+        if (expectedState == null || expectedState.getState() != Lifecycle.RUNNING) {
+            log.warn("On rebind of {}, not calling software process rebind hooks because state is {}", this, expectedState);
```
sjcorbett: Minor but possibly worthwhile: could you clarify that the state referenced is `expectedState`, not `actualState`?

neykov:
Updated comment.

ahgittin:
good fix

PR #638, opened 2015-05-12
Mirror summary for BrooklynEntityMirror

neykov:
Also extract the catalogItemId in a separate attribute for easier access.

### DIFF
--- a/software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynEntityMirror.java
+++ b/software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynEntityMirror.java
@@ -18,6 +18,8 @@
  */
 package brooklyn.entity.brooklynnode;
 
+import java.util.Map;
+
 import brooklyn.catalog.Catalog;
 import brooklyn.config.ConfigKey;
 import brooklyn.entity.Entity;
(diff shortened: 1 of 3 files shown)


software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynEntityMirror.java
```diff
@@ -54,6 +56,9 @@
         "Frequency to poll for client sensors", Duration.FIVE_SECONDS);
     
     public static final AttributeSensor<String> MIRROR_STATUS = Sensors.newStringSensor("brooklyn.mirror.monitoring_status");
+    @SuppressWarnings("rawtypes")
+    public static final AttributeSensor<Map> MIRROR_SUMMARY = Sensors.newSensor(Map.class, "brooklyn.mirror.summary");
+    public static final AttributeSensor<String> MIRROR_CATALOG_ITEM_ID = Sensors.newStringSensor("brooklyn.mirror.catalog_item_id");
```
ahgittin: add description, noting that this is the catalog item id in the remote brooklyn of the mirrored entity, which could be different to the local catalog item id

software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynEntityMirrorImpl.java
```diff
@@ -107,15 +107,25 @@ public Void apply(HttpToolResponse input) {
                 return null;
             }
         };
-
-        String sensorsUri = Urls.mergePaths(mirroredEntityUrl, "sensors/current-state");
+        
+        @SuppressWarnings("rawtypes")
+        Function<HttpToolResponse, Map> mirrorSummary = new Function<HttpToolResponse, Map>() {
+            @Override
+            public Map apply(HttpToolResponse input) {
```
ahgittin: better not to use an anonymous class?

software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynEntityMirror.java
```diff
@@ -54,6 +56,9 @@
         "Frequency to poll for client sensors", Duration.FIVE_SECONDS);
     
     public static final AttributeSensor<String> MIRROR_STATUS = Sensors.newStringSensor("brooklyn.mirror.monitoring_status");
+    @SuppressWarnings("rawtypes")
+    public static final AttributeSensor<Map> MIRROR_SUMMARY = Sensors.newSensor(Map.class, "brooklyn.mirror.summary");
```
ahgittin: add description, to say that this is the json map returned by the entity rest endpoint (ie the `EntitySummary` REST model)

ahgittin:
minor comments, then good to merge

neykov:
Addressed comments, thanks @ahgittin

PR #639, opened 2015-05-12
Allows for lists in templateOptions

nakomis:
Allows more complex templateOptions in brooklyn.properties such as the following:

```
brooklyn.location.named.AWS\ Oregon.templateOptions={ subnetId: subnet-a10e96c4 , securityGroupIds: [['sg-a2d0c2c7']], mapNewVolumeToDeviceName: ["/dev/sda1", 100, true]}
```

### DIFF
--- a/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
@@ -1050,42 +1050,59 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
                     public void apply(TemplateOptions t, ConfigBag props, Object v) {
                         t.networks((String)v);
                     }})
-              .put(TEMPLATE_OPTIONS, new CustomizeTemplateOptions() {
-                  @Override
-                  public void apply(TemplateOptions options, ConfigBag config, Object v) {
-                      if (v == null) return;
-                      @SuppressWarnings("unchecked") Map<String, String> optionsMap = (Map<String, String>) v;
-                      if (optionsMap.isEmpty()) return;
+            .put(TEMPLATE_OPTIONS, new CustomizeTemplateOptions() {
+                @Override
+                public void apply(TemplateOptions options, ConfigBag config, Object v) {
+                    if (v == null) return;
+                    @SuppressWarnings("unchecked") Map<String, Object> optionsMap = (Map<String, Object>) v;
+                    if (optionsMap.isEmpty()) return;
 
-                      Class<? extends TemplateOptions> clazz = options.getClass();
-                      Iterable<Method> methods = Arrays.asList(clazz.getMethods());
-                      for(final Map.Entry<String, String> option : optionsMap.entrySet()) {
-                          Optional<Method> methodOptional = Iterables.tryFind(methods, new Predicate<Method>() {
-                              @Override
-                              public boolean apply(@Nullable Method input) {
-                                  // Matches a method with the expected name, and a single parameter that TypeCoercions
-                                  // can coerce to
-                                  if (input == null) return false;
-                                  if (!input.getName().equals(option.getKey())) return false;
-                                  Type[] parameterTypes = input.getGenericParameterTypes();
-                                  return parameterTypes.length == 1
-                                          && TypeCoercions.tryCoerce(option.getValue(), TypeToken.of(parameterTypes[0])).isPresentAndNonNull();
-                              }
-                          });
-                          if(methodOptional.isPresent()) {
-                              try {
-                                  Method method = methodOptional.get();
-                                  method.invoke(options, TypeCoercions.coerce(option.getValue(), TypeToken.of(method.getGenericParameterTypes()[0])));
-                              } catch (IllegalAccessException e) {
-                                  throw Exceptions.propagate(e);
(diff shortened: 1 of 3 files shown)


andreaturli:
lgtm, maybe @richardcloudsoft can take another look as he knows very well the original code

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1050,42 +1050,59 @@ public void apply(TemplateOptions t, ConfigBag props, Object v) {
...
+                    @SuppressWarnings("unchecked") Map<String, Object> optionsMap = (Map<String, Object>) v;
+                    if (optionsMap.isEmpty()) return;
+
+                    Class<? extends TemplateOptions> clazz = options.getClass();
+                    Iterable<Method> methods = Arrays.asList(clazz.getMethods());
+                    for(final Map.Entry<String, Object> option : optionsMap.entrySet()) {
+                        Optional<Method> methodOptional = Iterables.tryFind(methods, new Predicate<Method>() {
+                            @Override
+                            public boolean apply(@Nullable Method input) {
+                                // Matches a method with the expected name, and a single parameter that TypeCoercions
```
richardcloudsoft: Is the statement "... a single parameter ..." still true, or does this now match methods with multiple parameters?

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1050,42 +1050,59 @@ public void apply(TemplateOptions t, ConfigBag props, Object v) {
...
+                                // can coerce to
+                                if (input == null) return false;
+                                if (!input.getName().equals(option.getKey())) return false;
+                                int numOptionParams = option.getValue() instanceof List ? ((List)option.getValue()).size() : 1;
+                                Type[] parameterTypes = input.getGenericParameterTypes();
+                                if (parameterTypes.length != numOptionParams) return false;
+                                if (numOptionParams == 1 && !(option.getValue() instanceof List) && parameterTypes.length == 1) {
```
richardcloudsoft: A comment to explain this would be helpful

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1050,42 +1050,59 @@ public void apply(TemplateOptions t, ConfigBag props, Object v) {
...
+                            } catch (IllegalAccessException e) {
+                                throw Exceptions.propagate(e);
+                            } catch (InvocationTargetException e) {
+                                throw Exceptions.propagate(e);
+                            }
+                        } else {
+                            LOG.warn("Ignoring request to set template option {} because this is not supported by {}", new Object[] { option.getKey(), clazz.getCanonicalName() });
+                        }
+                    }
+                }
```
richardcloudsoft: Is there any unexpected behaviour here if the template option you want to set here takes a single parameter that is a `List`?

richardcloudsoft:
Please alter the commit message (not the PR message, although you could do that too) to include documentation of this feature. Your example in the pull message...

```
brooklyn.location.named.AWS\ Oregon.templateOptions={ subnetId: subnet-a10e96c4 , securityGroupIds: [['sg-a2d0c2c7']], mapNewVolumeToDeviceName: ["/dev/sda1", 100, true]}
```

... is a bit confusing because it doesn't explain why you are using `[[ ... ]]` in one place and '[ ... ]' in another.

Better still, document it on the [jclouds Config Keys part of the locations page](http://brooklyn.incubator.apache.org/v/latest/ops/locations/index.html#jclouds-config-keys) of the user guide!

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
ahgittin: Agree this case is surprising.  It was ugly in your example that where a method takes a one-arg list i have to pass `[[ "value" ]]` rather than `[ value ]`.

I suggest either changing the logic, both in the `try` and once the method is retrieved, so that either:
* if the first pass fails, i.e. we are unable to match the items in a list against any method's parameters, we try again looking for a `method(Iterable)` (this will do the right thing in more cases, ie for `foo(List)`, we will accept `foo: [ arg ]` *or* `foo: [[ arg ]]`; it is only in the case we want a `foo(List<List>)` that a user has to take care, and in that case they must do `foo: [ [[ arg ]] ]`, which we should document.
* we *always* prefer a one-parameter method, if `option.getValue()` is coercible to that parameter type (however this means you must *not* doubly-wrap lists)

Probably the former is better as it gives preference to multi-argument methods which are less ambiguous.  Make sure you add a test for both the `[ list ]` and `[[ list ]]` cases!

In any case this is a nice feature and nice job with tricky logic.  I've wanted to pass more complex types here before so I'm glad it's now supported.

I also agree with the importance of documenting this in the locations markdown.  You will see instances there where I describe YAML notation (where `templateOptions` are set in a blueprint) in addition to properties file.

ahgittin:
@nakomis - ping

neykov:
The changes are already merged (see https://github.com/apache/incubator-brooklyn/pull/714). @nakomis could you close it manually.

PR #640, opened 2015-05-12
Upgrade nginx to latest stable version (1.8.0)

neykov:
Tested with a `ControlledDynamicWebAppCluster` with both http and https protocols.


sjcorbett:
Looks good. Thanks @neykov.

PR #641, opened 2015-05-12
brooklyn.util.text.Strings adjustments

sjcorbett:
Prompted by the chance observation in VisualVM that way too much CPU time was spent in `Strings.removeFromStart/End ` via `Exceptions.stripBoringPrefixes` and `PropagatedRuntimeException`. Please check a with careful eye.

For more on the time complexity of `String.substring` see http://stackoverflow.com/questions/4679746/time-complexity-of-javas-substring and http://stackoverflow.com/questions/16123446/java-7-string-substring-complexity.

### DIFF
--- a/utils/common/src/main/java/brooklyn/util/text/Strings.java
+++ b/utils/common/src/main/java/brooklyn/util/text/Strings.java
@@ -151,20 +151,24 @@ public class Strings {
         return string;
     }
 
-    /** as removeFromEnd, but repeats until all such suffixes are gone */
-    public static String removeAllFromEnd(String string, String ...suffixes) {
+    /**
+     * As removeFromEnd, but repeats until all such suffixes are gone
+     */
+    public static String removeAllFromEnd(String string, String... suffixes) {
+        if (isEmpty(string)) return string;
+        int index = string.length();
         boolean anotherLoopNeeded = true;
         while (anotherLoopNeeded) {
             if (isEmpty(string)) return string;
             anotherLoopNeeded = false;
             for (String suffix : suffixes)
-                if (string.endsWith(suffix)) {
-                    string = string.substring(0, string.length() - suffix.length());
+                if (!isEmpty(suffix) && string.startsWith(suffix, index - suffix.length())) {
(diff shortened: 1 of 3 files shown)


utils/common/src/main/java/brooklyn/util/text/Strings.java
```diff
@@ -136,26 +150,45 @@ public static String removeFromEnd(String string, String ...suffixes) {
...
-    public static String removeAllFromEnd(String string, String ...suffixes) {
+    /**
+     * As removeFromEnd, but repeats until all such suffixes are gone
+     */
+    public static String removeAllFromEnd(String string, String... suffixes) {
+        if (isEmpty(string)) return string;
+        int index = string.length();
         boolean anotherLoopNeeded = true;
         while (anotherLoopNeeded) {
             if (isEmpty(string)) return string;
```
neykov: `isEmpty(string)` already checked above

neykov:
Good to merge, only minor comments.

utils/common/src/main/java/brooklyn/util/text/Strings.java
sjcorbett: The first check is really for `string != null`, so that `.length()` doesn't throw an exception.

neykov: Yeah, but now `string` isn't modified so one check should be enough. I think that to keep the initial intent (to short circuit the loop if nothing is left of the string) the check should be replaced with `index == 0`.

sjcorbett: Good thinking.

sjcorbett:
By sheer chance I spent yesterday debugging an issue whose root cause turned out to lie in exactly the same area as this PR aims to address. Though the changes here are a minor improvement, they do not solve the root issue, which is a wildly inefficient `Exceptions.collapseText` implementation, creating millions of throwaway `PropagatedRuntimeExceptions`. I'll merge this but will make further adjustments later.

PR #642, opened 2015-05-13
BROOKLYN-143 - Initial support for Hazelcast

ygy:
Provides the basic functionality for cluster of hazelcast nodes

### DIFF
--- a/sandbox/nosql/src/main/java/brooklyn/entity/nosql/hazelcast/HazelcastNode.java
+++ b/sandbox/nosql/src/main/java/brooklyn/entity/nosql/hazelcast/HazelcastNode.java
@@ -27,6 +27,8 @@ import brooklyn.entity.java.UsesJmx;
 import brooklyn.entity.proxying.ImplementedBy;
 import brooklyn.event.basic.BasicAttributeSensorAndConfigKey;
 import brooklyn.event.basic.BasicAttributeSensorAndConfigKey.StringAttributeSensorAndConfigKey;
+import brooklyn.event.basic.PortAttributeSensorAndConfigKey;
+import brooklyn.location.basic.PortRanges;
 import brooklyn.util.flags.SetFromFlag;
 import brooklyn.util.javalang.JavaClassNames;
 
@@ -44,19 +46,40 @@ public interface HazelcastNode extends SoftwareProcess, UsesJava, UsesJmx {
     BasicAttributeSensorAndConfigKey<String> DOWNLOAD_URL = new BasicAttributeSensorAndConfigKey<String>(
             SoftwareProcess.DOWNLOAD_URL, "https://repo1.maven.org/maven2/com/hazelcast/hazelcast/${version}/hazelcast-${version}.jar");
     
-    @SetFromFlag("configFileUrl")
-    ConfigKey<String> TEMPLATE_CONFIGURATION_URL = ConfigKeys.newStringConfigKey(
-            "hazelcast.node.template.configuration.url", "Template file (in freemarker format) for the hazelcast.xml file", 
+    @SetFromFlag("configTemplateUrl")
+    ConfigKey<String> CONFIG_TEMPLATE_URL = ConfigKeys.newStringConfigKey(
+            "hazelcast.node.config.templateUrl", "Template file (in freemarker format) for the Hazelcat config file", 
             JavaClassNames.resolveClasspathUrl(HazelcastNode.class, "hazelcast-brooklyn.xml"));
+    
+    @SetFromFlag("configFileName")
+    ConfigKey<String> CONFIG_FILE_NAME = ConfigKeys.newStringConfigKey(
+            "hazelcast.node.config.fileName", "Name of the Hazelcast config file", "hazelcast.xml");    
     
     @SetFromFlag("nodeName")
     StringAttributeSensorAndConfigKey NODE_NAME = new StringAttributeSensorAndConfigKey("hazelcast.node.name", 
             "Node name (or randomly selected if not set", null);
-    
+
+    @SetFromFlag("nodePort")
+    PortAttributeSensorAndConfigKey NODE_PORT = new PortAttributeSensorAndConfigKey("hazelcast.node.port", "Hazelcast communication port", PortRanges.fromString("5701+"));
+
     /**
      * Specifies the group name in the configuration file. Each Hazelcast cluster has a separate group.
      */ 
     @SetFromFlag("groupName")
-    StringAttributeSensorAndConfigKey GROUP_NAME = new StringAttributeSensorAndConfigKey("hazelcast.group.name", 
-            "Group name", null);
+    ConfigKey<String> GROUP_NAME = ConfigKeys.newStringConfigKey("hazelcast.group.name", 
+            "Group name", "brooklyn");
+  
+    @SetFromFlag("groupPassword")
+    ConfigKey<String> GROUP_PASSWORD = ConfigKeys.newStringConfigKey("hazelcast.group.password", 
+            "Group password", "brooklyn");
+    
+    String getNodeName();
+    
+    Integer getNodePort();
+    
+    String getGroupName();
+    
+    String getGroupPassword();
+    
+    String getHostname();
 }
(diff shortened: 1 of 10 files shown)


sandbox/nosql/src/main/java/brooklyn/entity/nosql/hazelcast/HazelcastNode.java
```diff
@@ -0,0 +1,61 @@
...
+            "hazelcast.node.template.configuration.url", "Template file (in freemarker format) for the hazelcast.xml file", 
+            JavaClassNames.resolveClasspathUrl(HazelcastNode.class, "hazelcast-brooklyn.xml"));
+
+    @SetFromFlag("nodeName")
+    StringAttributeSensorAndConfigKey NODE_NAME = new StringAttributeSensorAndConfigKey("hazelcast.node.name", 
+            "Node name (or randomly selected if not set", null);
+    
+    @SetFromFlag("clusterName")
+    StringAttributeSensorAndConfigKey CLUSTER_NAME = new StringAttributeSensorAndConfigKey("hazelcast.node.cluster.name", 
+            "Cluster name", null);
```
ahgittin: i don't know enough about hazelcast but is the value of this (cluster name at node) the same as the value of the `hazelcast.cluster.name` ?  if so we could have the same key name or use the same constant.

ahgittin:
great start -- will be nice having this.  

i've not tried running this yet, but did have one question:  how do the nodes get told about each others' address?  normally i'd expect to see a list of other cluster addresses in the config template, or some CLI args in the launch step.

sandbox/nosql/src/main/java/brooklyn/entity/nosql/hazelcast/HazelcastNode.java
ygy: Thanks for the remark.

A single JVM can run multiple Hazelcast instances. They are specified in the ```<group>``` element of the configuration file.
```hazelcast.node.cluster.name``` was intended to be used as ```<name>``` value in the ```<group>``` element.
Often group and cluster names names are identical.

Indeed, current constant's name may lead to inconsistency. I will rename it and properly document the new one.

**Comment by ygy** [comment]
This part of the implementation is not finalized yet. It is using TCP/IP autodiscovery. 
By default, TCP/IP autodiscovery requires only one (well known) member. Additional nodes (members) use it to know about each other. Once a new member connect to the "known-member", it will receive the information for all the other members of the cluster. 
Current implementation is suitable for development purposes, but not for real production environment. I will extend it soon, adding some tests as well.

On the other hand side, Hazelcast doesn't provide a cli tool for adding new members. Nodes configuration is described in the config file (hazelcast.xml). Adding a new node (member) should be added in the config on every single node. Additionally, It requires node restart when configuration is changed. I'm still trying to figure out the correct way for editing and applying multiple configurations of already running services in Brooklyn.

File: sandbox/nosql/src/main/java/brooklyn/entity/nosql/hazelcast/HazelcastNodeSshDriver.java [hunk 2]
```diff
@@ -0,0 +1,112 @@
...
+        
+    }
+
+    @Override
+    public void launch() {
+        
+        entity.setAttribute(HazelcastNode.PID_FILE, Os.mergePathsUnix(getRunDir(), PID_FILENAME));
+        
+        StringBuilder commandBuilder = new StringBuilder()
+            .append(format("nohup java -cp ./lib/hazelcast-%s.jar", getVersion()));
```
aledsage: This jar file is referred to by name in `install()` (as resolver.getFilename()), in `customize()` (again calling resolver.getFilename(), and again here as `hazelcast-%s.jar`. Suggest changing this to use same name resolution technique as install and launch.

aledsage: I see you've followed the same pattern as elsewhere for setting the `resolver` field in the superclass. However, I think that might be a (subtly) bad pattern. If we stop and restart Brooklyn, then it restores the entities based on their config+attributes, and calling `rebind()`. I don't think anything will re-set the `resolver` field in the driver.

I suggest you don't worry about that in your pull request. We should refactor + fix that across the board (probably in the super-class) in a separate pull request, rather than doing something specific for hazelcast.

sandbox/nosql/src/main/java/brooklyn/entity/nosql/hazelcast/HazelcastNodeSshDriver.java
```diff
@@ -0,0 +1,112 @@
...
+        	
+        if (entity.getConfig(HazelcastNode.TEMPLATE_CONFIGURATION_URL) != null) {
+            commandBuilder.append(" -Dhazelcast.config=" + Os.mergePaths(getRunDir(), "conf", getConfigFile()));
```
aledsage: I have a preference for using `Os.mergePathsUnix` instead. The `mergePaths` (sort of) takes the settings of the OS running Brooklyn, rather than the target OS that the hazelcast instance is being deployed to. But saying that, it always uses '/' as the separator because that works on both unix and windows.

sandbox/nosql/src/main/java/brooklyn/entity/nosql/hazelcast/HazelcastNodeSshDriver.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @Override
+    public void launch() {
+        
+        entity.setAttribute(HazelcastNode.PID_FILE, Os.mergePathsUnix(getRunDir(), PID_FILENAME));
+        
+        StringBuilder commandBuilder = new StringBuilder()
+            .append(format("nohup java -cp ./lib/hazelcast-%s.jar", getVersion()));
+        	
+        if (entity.getConfig(HazelcastNode.TEMPLATE_CONFIGURATION_URL) != null) {
```
aledsage: The `customize()` has always copied them config file (without checking for null), so presumably would have failed if there was no config set. We should only reach this point if url was non-null?

What would the default be in launch if the `-Dhazelcast.config` was missed out?

sandbox/nosql/src/main/java/brooklyn/entity/nosql/hazelcast/HazelcastNodeSshDriver.java
```diff
@@ -0,0 +1,112 @@
...
+        commandBuilder.append(" com.hazelcast.core.server.StartServer >> ./log/out.log 2>&1 </dev/null &");
+        
+        newScript(MutableMap.of(USE_PID_FILE, true), LAUNCHING)
+            .updateTaskAndFailOnNonZeroResultCode()
+            .body.append(commandBuilder.toString())
+            .execute();
+    }
+       
+    public String getConfigFile() {
+        return "hazelcast.xml";
```
aledsage: Perhaps this should return `Os.mergePathsUnix(getRunDir(), "conf", getConfigFile())`. There are two callers of `getConfigFile()`, and they both do this.

sandbox/nosql/src/main/resources/brooklyn/entity/nosql/hazelcast/hazelcast-brooklyn.xml
```diff
@@ -0,0 +1,58 @@
...
+            Allowed port range when connecting to other nodes.
+            0 or * means use system provided port.
+            -->
+            <ports>0</ports>
+        </outbound-ports>
+        <join>
+            <multicast enabled="false" />
+
+            <tcp-ip enabled="true">
+                <interface>127.0.0.1</interface>
```
aledsage: This looks like a strange ip to use for join? Will it only work for clusters if multiple nodes are running on same machine?

Probably best to inject this using freemarker.

aledsage:
@ygy looking really good. Only very minor low-level comments from me.

Like @ahgittin I'm very interested to see how clustering is handled. A new node would want the address(es) of existing nodes. And presumably existing nodes would want to be updated with the addresses of new nodes in case the process restarted?

ygy:
@aledsage thanks for the feedback. I've adjusted the code based on it.

I'm working on the updating cluster's members configuration. Please see my previous comment https://github.com/apache/incubator-brooklyn/pull/642#issuecomment-102404835

sandbox/nosql/src/main/java/brooklyn/entity/nosql/hazelcast/HazelcastNodeSshDriver.java
```diff
@@ -0,0 +1,159 @@
...
+            }
+            result.add(addressAndPort);
+        }
+        
+        return result;
+    }
+
+    @Override
+    protected String getLogFileLocation() {
+        return Os.mergePathsUnix(getRunDir(),"/log/out.log");
```
aledsage: Personal preference for `console` or `console.log`; the name `out.log` makes it feel like it might be just stdout, but it also has stderr. No strong feelings though.

aledsage:
Thanks @ygy - sorry we've been ignoring this for far too long! 

I'm running the tests now, and will then merge. It would also be good to have integration tests that run locally, assuming that is feasible?

aledsage:
@ygy - can you please close this PR. I created a new PR and squashed your commits in it, and merged that. Therefore all your changes here are now in master. See https://github.com/apache/incubator-brooklyn/pull/767 and https://github.com/apache/incubator-brooklyn/commit/e37f05856c7067081f8e37d3fbdc84408e1bbb9d

PR #645, opened 2015-05-13
Support multiple setup scripts in location config

neykov:
New `setup.scripts` config which accepts an array of scripts to execute on the node.


sjcorbett:
Looks good.

PR #646, opened 2015-05-14
Changes DynamicCluster resize behaviour to favour the location define…

nakomis:
…d on the memberspec (if present)


ahgittin:
good fix

PR #647, opened 2015-05-14
do the node and client deletion as part of stopping a chef node

ahgittin:
without this nodes get left registered at the chef server.  needs testing.


ggrunin:
please add "-y" at the end of knife delete commands in ChefLifecycleEffectorTasks.java to make them executable without confirmation:
 .add("knife node delete "+getNodeName()+" -y")
.add("knife client delete "+getNodeName()+" -y")

ahgittin:
makes sense @ggrunin, thanks -- is it working now?

ggrunin:
node/client delete works. Thanks. Additional small fix needed for bootstrap. Please add "--no-host-key-verify" parameter, otherwise bootstrap is failing if the server was provisioned with IP which was previously used by some other brooklyn server.

ahgittin:
@ggrunin done.  + good spot.  any other tweaks?

ggrunin:
@ahgittin Alex,  "--no-host-key-verify"  parameter has to be added to knife bootstrap command, and not to knife node delete or knife client delete. Sorry for not making it clear the first time. Thanks.

ahgittin:
makes sense -- now applied to `bootstrap` instead

PR #648, opened 2015-05-18
Rebind backwards compatibility

neykov:
Keep anonymous classes order for rebind backwards compatibility. `computing` present in persistence files but not used.


ahgittin:
good fix.  anonymous classes must die!  test failure unrelated (ciphers, fixed in #649)

PR #650, opened 2015-05-19
Support for deployment to Windows

nakomis:


pom.xml
```diff
@@ -205,6 +205,7 @@
         <jsr311-api.version>1.1.1</jsr311-api.version>
         <maxmind.version>0.8.1</maxmind.version>
         <jna.version>4.0.0</jna.version>
+        <winrm4j.version>0.1.0-SNAPSHOT</winrm4j.version>
```
aledsage: Use 0.1.0, which is in the process of being released to maven central.

core/src/main/java/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -588,16 +588,22 @@ protected Entity replaceMember(Entity member, Location memberLoc, Map<?, ?> extr
...
         // choose locations to be deployed to
         List<Location> chosenLocations;
-        if (isAvailabilityZoneEnabled()) {
-            List<Location> subLocations = getNonFailedSubLocations();
-            Multimap<Location, Entity> membersByLocation = getMembersByLocation();
-            chosenLocations = getZonePlacementStrategy().locationsForAdditions(membersByLocation, subLocations, delta);
-            if (chosenLocations.size() != delta) {
-                throw new IllegalStateException("Node placement strategy chose " + Iterables.size(chosenLocations)
-                        + ", when expected delta " + delta + " in " + this);
+        chosenLocations = getMemberSpec() == null ? null : getMemberSpec().getLocations();
```
aledsage: We need to think about this more. Feels wrong to disable the availability zone behaviour entirely.

Similarly would prefer us not to merge with `FIXME: Tidy this up!` in the code. That at least should be fleshed out to a TODO that describes how it could be improved. Or ideally improve the code.

core/src/main/java/brooklyn/event/feed/windows/WindowsPerformanceCounterFeed.java
```diff
@@ -179,50 +186,50 @@ protected WindowsPerformanceCounterFeed(Builder builder) {
...
         Iterable<String> allParams = ImmutableList.<String>builder()
-                .add("typeperf")
-                .addAll(Iterables.transform(performanceCounterNames, QuoteStringFunction.INSTANCE))
-                .add("-sc")
-                .add("1")
+                .add("Get-Counter")
+                .add("-Counter")
+                .add(JOINER_ON_COMMA.join(Iterables.transform(performanceCounterNames, QuoteStringFunction.INSTANCE)))
+                .add("-SampleInterval")
+                .add("2") // TODO: This should be a config key
```
aledsage: What does the "2" mean? Can you add some more description to the TODO

core/src/main/java/brooklyn/event/feed/windows/WindowsPerformanceCounterFeed.java
```diff
@@ -251,10 +258,71 @@ public T call() throws Exception {
...
+        public SendPerfCountersToSensors(EntityLocal entity, List<WindowsPerformanceCounterPollConfig<?>> polls) {
+            this.entity = entity;
+            this.polls = polls;
+        }
+
+        @Override
+        public boolean checkSuccess(WinRmToolResponse val) {
+            if (val.getStatusCode() != 0) return false;
+            String stderr = val.getStdErr();
+            if (stderr == null || stderr.length() != 0) return false;
```
aledsage: Can we not just rely on the statusCode?
Worth a TODO to explain why treating presence of stderr as a failure.

core/src/main/java/brooklyn/event/feed/windows/WindowsPerformanceCounterFeed.java
```diff
@@ -251,10 +258,71 @@ public T call() throws Exception {
...
+            if (out == null || out.length() == 0) return false;
+            return true;
+        }
+
+        @Override
+        public void onSuccess(WinRmToolResponse val) {
+            String[] values = val.getStdOut().split("\r\n");
+            for (int i = 0; i < polls.size(); i++) {
+                Class<?> clazz = polls.get(i).getSensor().getType();
+                Maybe<? extends Object> maybeValue = TypeCoercions.tryCoerce(values[i], TypeToken.of(clazz));
```
aledsage: Does use of `tryCoerce` mean that we'll ignore it if the returned value was of an unexpected format? Would seem good to log at least once to say that (e.g. if get "1.2" when we expect an integer, then we should log that somewhere sensible).

core/src/main/java/brooklyn/event/feed/windows/WindowsPerformanceCounterFeed.java
```diff
@@ -251,10 +258,71 @@ public T call() throws Exception {
...
+                entity.setAttribute(attribute, value);
+            }
+        }
+
+        @Override
+        public void onFailure(WinRmToolResponse val) {
+            log.error("Windows Performance Counter query did not respond as expected. exitcode={} stdout={} stderr={}",
+                    new Object[]{val.getStatusCode(), val.getStdOut(), val.getStdErr()});
+            for (WindowsPerformanceCounterPollConfig<?> config : polls) {
+                entity.setAttribute(Sensors.newSensor(config.getSensor().getClass(), config.getPerformanceCounterName(), config.getDescription()), null);
```
aledsage: Above (for `onSuccess`) we did `polls.get(i).getSensor().getName()`. Why are we using `config.getPerformanceCounterName()` here instead? Same below for `onException`.

core/src/main/java/brooklyn/event/feed/windows/WindowsPerformanceCounterFeed.java
```diff
@@ -251,10 +258,71 @@ public T call() throws Exception {
...
+        public String toString() {
+            return super.toString()+"["+getDescription()+"]";
+        }
+    }
+
     /**
      * A poll handler that takes the result of the <tt>typeperf</tt> invocation and sets the appropriate sensors.
      */
-    private static class SendPerfCountersToSensors implements PollHandler<SshPollValue> {
+    private static class SendPerfCountersToSensors2 implements PollHandler<SshPollValue> {
```
aledsage: Do we need this class? It appears to never be used.

core/src/main/java/brooklyn/event/feed/windows/WindowsPerformanceCounterFeed.java
```diff
@@ -98,10 +104,11 @@
     // This pattern matches CSV line(s) with the date in the first field, and at least one further field.
     protected static final Pattern lineWithPerfData = Pattern.compile("^\"[\\d:/\\-. ]+\",\".*\"$", Pattern.MULTILINE);
     private static final Joiner JOINER_ON_SPACE = Joiner.on(' ');
+    private static final Joiner JOINER_ON_COMMA = Joiner.on(',');
 
     @SuppressWarnings("serial")
-    public static final ConfigKey<Set<WindowsPerformanceCounterPollConfig<?>>> POLLS = ConfigKeys.newConfigKey(
-            new TypeToken<Set<WindowsPerformanceCounterPollConfig<?>>>() {},
+    public static final ConfigKey<List<WindowsPerformanceCounterPollConfig<?>>> POLLS = ConfigKeys.newConfigKey(
```
aledsage: Could/should we use Collection rather than List?

core/src/main/java/brooklyn/location/basic/WinRmMachineLocation.java
```diff
@@ -0,0 +1,266 @@
...
+    public static final ConfigKey<String> WINDOWS_USERNAME = ConfigKeys.newStringConfigKey("windows.username",
+            "Username to use when connecting to the remote machine");
+
+    public static final ConfigKey<String> WINDOWS_PASSWORD = ConfigKeys.newStringConfigKey("windows.password",
+            "Password to use when connecting to the remote machine");
+
+    public static final ConfigKey<Integer> COPY_FILE_CHUNK_SIZE_BYTES = ConfigKeys.newIntegerConfigKey("windows.copy.file.size.bytes",
+            "Size of file chunks (in bytes) to be used when copying a file to the remote server", 1024);
+
+    @SetFromFlag
```
aledsage: Let's use config keys rather than `@SetFromFlag` anywhere here. The old SetFromFlag approach in SshMachineLocation should really be refactored.

Same goes for `address` below.

core/src/main/java/brooklyn/location/basic/WinRmMachineLocation.java
```diff
@@ -0,0 +1,266 @@
...
+
+    @Nullable
+    @Override
+    public String getHostname() {
+        return address.getHostAddress();
+    }
+
+    @Override
+    public Set<String> getPublicAddresses() {
+        return null;
```
aledsage: Why does this return null? Same for `getPrivateAddresses` below?

core/src/main/java/brooklyn/location/basic/WinRmMachineLocation.java
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    public WinRmToolResponse executeScript(String script) {
+        return executeScript(ImmutableList.of(script));
+    }
+
+    public WinRmToolResponse executeScript(List<String> script) {
+        Collection<Throwable> exceptions = Lists.newArrayList();
+        for (int i = 0; i < 10; i++) {
```
aledsage: Worth extracting this as a config key.

Also worth doing some smarter logging. e.g. saying attempt 1 of 10, and not logging "retrying" on the final failure when it's not going to retry.

I wonder if worth creating a wrapper of WinrmTool that gives us any customer behaviour that we want. Having this code here feels quite different from the `SshMachineLocation` implementation. But let's not worry about that in this PR.

core/src/main/java/brooklyn/location/basic/WinRmMachineLocation.java
```diff
@@ -0,0 +1,266 @@
...
+        throw Exceptions.propagate("failed to execute shell script", exceptions);
+    }
+
+    public WinRmToolResponse executePsScript(String psScript) {
+        return executePsScript(ImmutableList.of(psScript));
+    }
+
+    public WinRmToolResponse executePsScript(List<String> psScript) {
+        Collection<Throwable> exceptions = Lists.newArrayList();
+        for (int i = 0; i < 10; i++) {
```
aledsage: See comment about retries in executeScript

core/src/main/java/brooklyn/location/basic/WinRmMachineLocation.java
```diff
@@ -0,0 +1,266 @@
...
+            } catch (Exception e) {
+                LOG.warn("ignoring winrm exception and retrying after 5 seconds:", e);
+                Time.sleep(Duration.FIVE_SECONDS);
+                exceptions.add(e);
+            }
+        }
+        throw Exceptions.propagate("failed to execute powershell script", exceptions);
+    }
+
+    public WinRmToolResponse executePsScriptNoRetry(List<String> psScript) {
```
aledsage: I'd have expected the same pattern of methods to be used for executePsScript and executeScript. Is there a reason not to?

aledsage: Suggest making this protected; same for other methods you don't expect people to call.

core/src/main/java/brooklyn/location/basic/WinRmMachineLocation.java
```diff
@@ -0,0 +1,266 @@
...
+        WinRmTool winRmTool = WinRmTool.connect(getHostname(), getUsername(), getPassword());
+        WinRmToolResponse response = winRmTool.executePs(psScript);
+        return response;
+    }
+
+    public int copyTo(File source, File destination) {
```
aledsage: Not sure you should use `File` as the destination. Problem is that the destination will be a windows path, but File here could be constructed with Brooklyn running on Linux.

Does that matter? Will it have used the right separator for the destination File?!

core/src/main/java/brooklyn/location/basic/WinRmMachineLocation.java
```diff
@@ -0,0 +1,266 @@
...
+            while ((bytesRead = source.read(inputData)) > 0) {
+                byte[] chunk;
+                if (bytesRead == chunkSize) {
+                    chunk = inputData;
+                } else {
+                    chunk = Arrays.copyOf(inputData, bytesRead);
+                }
+                executePsScript(ImmutableList.of("If ((!(Test-Path " + destination.getPath() + ")) -or ((Get-Item '" + destination.getPath() + "').length -eq " +
+                        expectedFileSize + ")) {Add-Content -Encoding Byte -path " + destination.getPath() +
+                        " -value ([System.Convert]::FromBase64String(\"" + new String(Base64.encodeBase64(chunk)) + "\"))}"));
```
aledsage: Worth a comment describing this approach to file-copy

core/src/main/java/brooklyn/location/basic/WinRmMachineLocation.java
```diff
@@ -0,0 +1,266 @@
...
+
+    public String getUsername() {
+        return config().get(WINDOWS_USERNAME);
+    }
+
+    private String getPassword() {
+        return config().get(WINDOWS_PASSWORD);
+    }
+
+    public static String getDefaultUserMetadataString() {
```
aledsage: Worth some javadoc to say what this default user metadata will do.

core/src/main/java/brooklyn/util/flags/TypeCoercions.java
```diff
@@ -709,6 +710,13 @@ public QuorumCheck apply(final String input) {
                 return QuorumChecks.of(input);
             }
         });
+        registerAdapter(ArrayList.class, String[].class, new Function<ArrayList, String[]>() {
```
aledsage: Could we declare this as `List.class`, rather than `ArrayList.class`? Or does it need to be the concrete type?

core/src/main/java/brooklyn/util/flags/TypeCoercions.java
```diff
@@ -709,6 +710,13 @@ public QuorumCheck apply(final String input) {
...
+            @Nullable
+            @Override
+            public String[] apply(@Nullable ArrayList arrayList) {
+                return (String[]) arrayList.toArray(new String[]{});
```
aledsage: If marking as `@Nullable`, then should handle that: return null if `arrayList == null`?

core/src/main/java/brooklyn/location/basic/WinRmMachineLocation.java
```diff
@@ -0,0 +1,266 @@
...
+import com.google.common.base.Charsets;
+import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
+
+public class WinRmMachineLocation extends AbstractLocation implements MachineLocation {
+
+    private static final Logger LOG = LoggerFactory.getLogger(WinRmMachineLocation.class);
+
+    public static final ConfigKey<String> WINDOWS_USERNAME = ConfigKeys.newStringConfigKey("windows.username",
```
aledsage: Why use the name `windows.username`, rather than having it match the names used in `SshMachineLocation` (i.e. "user" and "password")? I expect it will give a more consistent feel to Brooklyn if the names match.

core/src/test/java/brooklyn/location/basic/ByonLocationResolverTest.java
```diff
@@ -307,12 +308,41 @@ public void testEmptySpec() throws Exception {
...
+        Map<String, ?> flags = ImmutableMap.of(
+                "hosts", ImmutableList.of("1.1.1.1", "2.2.2.2"),
+                "osfamily", "windows"
+        );
+        MachineProvisioningLocation<MachineLocation> provisioner = resolve(spec, flags);
+        MachineLocation location = provisioner.obtain(ImmutableMap.of());
+
+        assertEquals(location.config().get(WinRmMachineLocation.WINDOWS_USERNAME), "myuser");
+        assertEquals(location.config().get(WinRmMachineLocation.WINDOWS_PASSWORD), "mypassword");
+        Assert.assertNotNull(provisioner);
```
aledsage: why assertNotNull here? We've already called methods on it. Suggest delete this line?

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -622,6 +624,14 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
             if (node == null)
                 throw new IllegalStateException("No nodes returned by jclouds create-nodes in " + setup.getDescription());
 
+            boolean windows = node.getOperatingSystem().getFamily().equals(OsFamily.WINDOWS);
+
+            if (windows) {
+                // FIXME: Tidy this up and allow for user-specified credentials
```
aledsage: Am I reading this right, that we currently don't support creating a new user or support setting the existing (e.g. administrator) password?

If so, we should include that in the list of things to do for being ready for production.

How does the Brooklyn user discover the user/password of their VM for if they want to subsequently log in? I presume that won't be displayed in the Brooklyn web console at all currently?

aledsage: I'd prefer FIXME/TODO comments to be more specific than "Tidy this up". The second part of comment is very clear. If there is more tidying than that, then please say what specifically you're thinking of.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -635,10 +645,12 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
                 sshHostAndPortOverride = Optional.absent();
             }
 
-            if (waitForSshable && skipJcloudsSshing) {
+            if (waitForSshable && skipJcloudsSshing && !windows) {
                 // once that host:port is definitely reachable, we can create the user
                 waitForReachable(computeService, node, sshHostAndPortOverride, node.getCredentials(), setup);
                 userCredentials = createUser(computeService, node, sshHostAndPortOverride, setup);
+            } else if (windows) {
```
aledsage: We want to be able to defer/never waitForWinRmAvailable. That was intent of `waitForSshable` - to turn that off. 

Historically, reason for that was to let the caller set up port forwarding. However, that is now covered in code above (but not for windows case!). So we generally don't make use of the `waitForSshable` much these days. But given it's there...

I wonder if we should generalise it to a "waitForReachable" or some such.

Don't worry about it for this PR.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -622,6 +624,14 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
...
+            boolean windows = node.getOperatingSystem().getFamily().equals(OsFamily.WINDOWS);
+
+            if (windows) {
+                // FIXME: Tidy this up and allow for user-specified credentials
+                setup.put(USER, node.getCredentials().getUser());
+                setup.put(PASSWORD, node.getCredentials().getOptionalPassword().orNull());
+            }
+
             // Setup port-forwarding, if required
             Optional<HostAndPort> sshHostAndPortOverride;
```
aledsage: We should generalise `sshHostAndPortOverride`, perhaps to `connectionHostAndPortOverride`. And we should ensure it is setup for WinRM port in the windows case.

Can do that in separate PR.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -665,42 +677,55 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
...
             // Wait for the VM to be reachable over SSH
-            if (waitForSshable) {
+            if (waitForSshable && !windows) {
                 waitForReachable(computeService, node, sshHostAndPortOverride, userCredentials, setup);
             } else {
                 LOG.debug("Skipping ssh check for {} ({}) due to config waitForSshable=false", node, setup.getDescription());
             }
             usableTimestamp = Duration.of(provisioningStopwatch);
 
+            JcloudsSshMachineLocation jcloudsSshMachineLocation = null;
```
aledsage: I wonder about just having the `machineLocation` variable, and casting it where required. Rather than having 3 variables where 1 will be null. Not sure what the best pattern is for that. The casting would force us to think about it more, to decide if we can generalise the method to take a machineLocation instead (and thus work with ssh and windows).

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -712,80 +737,105 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
                 List<String> setupScripts = setup.get(JcloudsLocationConfig.CUSTOM_MACHINE_SETUP_SCRIPT_URL_LIST);
                 Collection<String> allScripts = new MutableList<String>().appendIfNotNull(setupScript).appendAll(setupScripts);
                 for (String setupScriptItem : allScripts) {
-                    if (Strings.isNonBlank(setupScriptItem)) {
-                        customisationForLogging.add("custom setup script "+setupScriptItem);
+                    if (Strings.isNonBlank(setupScript)) {
```
aledsage: Why change this to `setupScript`, rather than `setupScriptItem`? See the for loop above.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -712,80 +737,105 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
...
                                 ? Splitter.on(",").withKeyValueSeparator(":").split(setupVarsString)
                                 : ImmutableMap.<String, String>of();
-                        String scriptContent =  ResourceUtils.create(this).getResourceAsString(setupScriptItem);
+                        String scriptContent = ResourceUtils.create(this).getResourceAsString(setupScriptItem);
                         String script = TemplateProcessor.processTemplateContents(scriptContent, getManagementContext(), substitutions);
-                        sshMachineLocation.execCommands("Customizing node " + this + " with script " + setupScriptItem, ImmutableList.of(script));
+                        if (windows) {
+                            winRmMachineLocation.executeScript(ImmutableList.copyOf((script.replace("\r", "").split("\n"))));
+                        } else {
+                            jcloudsSshMachineLocation.execCommands("Customizing node " + this, ImmutableList.of(script));
```
aledsage: This changes the execCommands description. I take it that's deliberate? Was the full script being logged multiple times before?

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -712,80 +737,105 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
...
+                                Arrays.asList("sudo mv /dev/random /dev/random-real", "sudo ln -s /dev/urandom /dev/random"));
+                    }
                 }
 
 
                 if (setup.get(GENERATE_HOSTNAME)) {
-                    customisationForLogging.add("configure hostname");
+                    if (windows) {
+                        // TODO: Generate Windows Hostname
+                        LOG.warn("Ignoring flag GENERATE_HOSTNAME on Windows location {}", winRmMachineLocation);
```
aledsage: We should list (in docs / limitations) the main config options that are not supported (yet) on windows (rather than the user/customer discovering it by trying to configure it and eventually finding this log message, or just it being ignored silently (as would be the case for user/password configuration).

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -795,24 +845,48 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
 
             // Apply any optional app-specific customization.
             for (JcloudsLocationCustomizer customizer : getCustomizers(setup)) {
-                customizer.customize(this, computeService, sshMachineLocation);
+                if (windows) {
+                    LOG.warn("Ignoring customizer {} on Windows location {}", customizer, winRmMachineLocation);
```
aledsage: Tricky; we should update the customizer's interface to take a MachineLocation instead of an ssh machine location.

Let's do that in a separate PR.

Can you keep a list of what we're deferring to later PRs please.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -795,24 +845,48 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
...
+                // FIXME: Remove try-catch!
+                try {
+                    logMessage = "Finished VM " + setup.getDescription() + " creation:"
+                            + " " + winRmMachineLocation.getConfig(WinRmMachineLocation.WINDOWS_USERNAME) + "@" + machineLocation.getAddress()
+                            + " username=" + winRmMachineLocation.getUsername()
+                            + " ready after " + Duration.of(provisioningStopwatch).toStringRounded()
+                            + " (" + template + " template built in " + Duration.of(templateTimestamp).toStringRounded() + ";"
+                            + " " + node + " provisioned in " + Duration.of(provisionTimestamp).subtract(templateTimestamp).toStringRounded() + ";";
+                } catch (Exception e){
+                    logMessage = Arrays.toString(e.getStackTrace());
```
aledsage: What exception are you expecting here? Can we just program more defensively when building the message?

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -795,24 +845,48 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
...
-                    + " and os customized in "+Duration.of(customizedTimestamp).subtract(usableTimestamp).toStringRounded()+" - "+Joiner.on(", ").join(customisationForLogging)+")");
-
-            return sshMachineLocation;
+            String logMessage;
+
+            if (windows) {
+                // TODO: More complete logging for WinRmMachineLocation
+                // FIXME: Remove try-catch!
+                try {
+                    logMessage = "Finished VM " + setup.getDescription() + " creation:"
```
aledsage: Would be good to remove duplication - perhaps having a `String.format("...", ...)` with particular bits of string built differently for windows and linux.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -836,6 +910,52 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
...
+        long delayMs = -1;
+        try {
+            delayMs = Time.parseTimeString(""+waitForWinRmAvailable);
+        } catch (Exception e) {
+            // normal if 'true'; just fall back to default
+        }
+        if (delayMs<0)
+            delayMs = Time.parseTimeString(WAIT_FOR_WINRM_AVAILABLE.getDefaultValue());
+
+        // FIXME: remove this
```
aledsage: Why a FIXME about removing this? What needs removed? Why not just remove it?

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -836,6 +910,52 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
...
+        LOG.info("Address: " + node.getPublicAddresses().iterator().next());
+        LOG.info("User: " + expectedCredentials.getUser());
+        LOG.info("Password: " + expectedCredentials.getOptionalPassword().get());
+        final Session session = WinRMFactory.INSTANCE.createSession(node.getPublicAddresses().iterator().next(),
+                expectedCredentials.getUser(), expectedCredentials.getOptionalPassword().get());
```
aledsage: Previously you have used `getOptionalPassword().orNull()` - can we assume it will exist here?

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -836,6 +910,52 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
         }
     }
 
+    private void waitForWinRmAvailable(final NodeMetadata node, final LoginCredentials expectedCredentials, ConfigBag setup) {
+        // TODO: Reduce / remove duplication between this and waitForReachable
```
aledsage: Agree feels like a fair amount of duplication between this and `waitForReachable`. Worth considering if can have a common impl, e.g. passing in the checker, and the config key for wait time.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -836,6 +910,52 @@ protected JcloudsSshMachineLocation obtainOnce(ConfigBag setup) throws NoMachine
...
+        String waitForWinRmAvailable = setup.get(WAIT_FOR_WINRM_AVAILABLE);
```
aledsage: Perhaps worth merging the `WAIT_FOR_WINRM_AVAILABLE` and the `WAIT_FOR_SSHABLE`. They are conceptually the same thing. Would probably have to deprecate the old key, and generalise the name.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -938,7 +1058,11 @@ public void apply(TemplateOptions t, ConfigBag props, Object v) {
                     public void apply(TemplateOptions t, ConfigBag props, Object v) {
                         if (t instanceof EC2TemplateOptions) {
                             if (v==null) return;
-                            ((EC2TemplateOptions)t).userData(v.toString().getBytes());
+                            String data = v.toString();
+                            if (!(data.startsWith("<script>") || data.startsWith("<powershell>"))) {
+                                data = "<script> " + data + " </script>";
```
aledsage: Do we have a test that this works for linux and for windows? (haven't reached your tests in the PR review yet).

Out of interest, where in jclouds is this documented?

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1071,15 +1195,32 @@ public boolean apply(@Nullable Method input) {
                                   // can coerce to
                                   if (input == null) return false;
                                   if (!input.getName().equals(option.getKey())) return false;
+                                  int numOptionParams = option.getValue() instanceof List ? ((List)option.getValue()).size() : 1;
                                   Type[] parameterTypes = input.getGenericParameterTypes();
-                                  return parameterTypes.length == 1
-                                          && TypeCoercions.tryCoerce(option.getValue(), TypeToken.of(parameterTypes[0])).isPresentAndNonNull();
+                                  if (parameterTypes.length != numOptionParams) return false;
```
aledsage: This looks familiar for another PR (https://github.com/apache/incubator-brooklyn/pull/639) - will assume that is going to be reviewed+merged first, and this PR rebased.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1729,6 +1876,27 @@ protected JcloudsSshMachineLocation createJcloudsSshMachineLocation(ComputeServi
         }
     }
 
+    protected WinRmMachineLocation registerWinRmMachineLocation(String vmHostname, ConfigBag setup, String nodeId) {
+        WinRmMachineLocation winRmMachineLocation = createWinRmMachineLocation(vmHostname, setup);
+        winRmMachineLocation.setParent(this);
```
aledsage: I wonder why we don't set the parent inside `registerJcloudsSshMachineLocation`. Ignore this comment for this PR.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsWinRmMachineLocation.java
```diff
@@ -0,0 +1,25 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package brooklyn.location.jclouds;
+
+import brooklyn.location.basic.WinRmMachineLocation;
+
+public class JcloudsWinRmMachineLocation extends WinRmMachineLocation {
```
aledsage: I wonder if it would be useful to have a `JcloudsMachineLocation` interface, with methods like `getNode()` and `getTemplate()`. That would also extend the `HasSubnetHostname` interface, with `getSubnetIp()` etc.

Ignore for this PR.

software/base/src/main/java/brooklyn/entity/basic/AbstractSoftwareProcessDriver.java
```diff
@@ -251,6 +264,183 @@ public InputStream getResource(String url) {
...
+     * Files and templates to be copied to the server <em>before</em> pre-install. This allows the {@link #preInstall()}
+     * process to have access to all required resources.
+     * <p>
+     * Will be prefixed with the entity's {@link #getInstallDir() install directory} if relative.
+     *
+     * @see SoftwareProcess#PRE_INSTALL_FILES
+     * @see SoftwareProcess#PRE_INSTALL_TEMPLATES
+     * @see #copyRuntimeResources()
+     */
+    public void copyPreInstallResources() {
```
aledsage: I presume these were copied unmodified from AbstractSoftwareProcessSshDriver? Therefore not reviewing thoroughly. If anything (beyond calling abstract methods at the appropriate points) has changed, then let me know and I'll review more thoroughly.

software/base/src/main/java/brooklyn/entity/basic/AbstractSoftwareProcessWinRmDriver.java
```diff
@@ -0,0 +1,192 @@
...
+import brooklyn.util.repeat.Repeater;
+import brooklyn.util.time.Duration;
+import io.cloudsoft.winrm4j.winrm.WinRmToolResponse;
+
+import com.google.api.client.util.Strings;
+import com.google.common.collect.ImmutableList;
+
+public abstract class AbstractSoftwareProcessWinRmDriver extends AbstractSoftwareProcessDriver {
+
+    AttributeSensor<String> WINDOWS_USERNAME = Sensors.newStringSensor("windows.username",
```
aledsage: Is inside the driver the best place to declare these sensors?

software/base/src/main/java/brooklyn/entity/basic/AbstractSoftwareProcessWinRmDriver.java
```diff
@@ -0,0 +1,192 @@
...
+
+    public int executePowerShellNoRetry(List<String> psScript) {
+        return getLocation().executePsScriptNoRetry(psScript).getStatusCode();
+    }
+
+    public int executePowerShell(List<String> psScript) {
+        return getLocation().executePsScript(psScript).getStatusCode();
+    }
+
+    public int copyTo(File source, File destination) {
```
aledsage: As commented in WinRmMachineLocation.java, not sure should use `File destination`, rather than `String destination`, because the OS of the brooklyn server and target VM may well be different. Thoughts?

software/base/src/main/java/brooklyn/entity/basic/AbstractSoftwareProcessWinRmDriver.java
```diff
@@ -0,0 +1,192 @@
...
+        } else {
+            return getLocation().executeScript(ImmutableList.of(regularCommand));
+        }
+    }
+
+    public int execute(List<String> script) {
+        return getLocation().executeScript(script).getStatusCode();
+    }
+
+    public int executePowerShellNoRetry(List<String> psScript) {
```
aledsage: Why the different naming: "executePowerShell" versus "executePsScript"?

software/base/src/main/java/brooklyn/entity/basic/AbstractSoftwareProcessWinRmDriver.java
```diff
@@ -0,0 +1,192 @@
...
+
+    public int copyTo(InputStream source, File destination) {
+        return getLocation().copyTo(source, destination);
+    }
+
+    public void rebootAndWait() {
+        try {
+            executePowerShellNoRetry(ImmutableList.of("Restart-Computer -Force"));
+        } catch (PyException e) {
+            // Restarting the computer will cause the command to fail; ignore the exception and continue
```
aledsage: Can we be more specific - e.g. what exception do you expect. And log if not exactly right. There's a risk we'll pretend the VM is rebooting when actually we got back an error about permissions or something like that.

software/base/src/main/java/brooklyn/entity/basic/VanillaWindowsProcessDriver.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package brooklyn.entity.basic;
+
+public interface VanillaWindowsProcessDriver extends AbstractVanillaProcessDriver {
```
aledsage: Not sure about the name `VanillaWindowsProcessDriver` versus `VanillaSoftwareProcessDriver`. They are both software processes.

But I don't have a better suggestion just now!

software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
```diff
@@ -442,11 +442,12 @@ protected final void callStartHooks() {}
      * plus any ports defined with a config keys ending in .port 
      */
     protected Collection<Integer> getRequiredOpenPorts() {
-        Set<Integer> ports = MutableSet.of(22);
+        // TODO: Should only open 22 *or* 5985. Perhaps a flag / ConfigKey on SoftwareProcessImpl?
+        Set<Integer> ports = MutableSet.of(22, 5985, 3389);
```
aledsage: Hm, tricky to figure out whether it's going to be Windows or Linux from the location. I think the entity impl needs to tell us (e.g. with a method that is overridden by the windows sub-class(es)).

Can you at least add to the TODO what 5985 and 3389 are for.

software/base/src/main/java/brooklyn/entity/basic/AbstractVanillaProcessDriver.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package brooklyn.entity.basic;
+
+public interface AbstractVanillaProcessDriver extends SoftwareProcessDriver {
```
aledsage: What does this interface buy us? Or is it just a marker interface?

software/base/src/main/java/brooklyn/entity/basic/VanillaWindowsProcess.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package brooklyn.entity.basic;
+
+import brooklyn.config.ConfigKey;
+import brooklyn.entity.proxying.ImplementedBy;
+import brooklyn.util.time.Duration;
+
+@ImplementedBy(VanillaWindowsProcessImpl.class)
+public interface VanillaWindowsProcess extends AbstractVanillaProcess {
+    ConfigKey<String> PRE_INSTALL_POWERSHELL_COMMAND = ConfigKeys.newStringConfigKey("pre.install.powershell.command",
```
aledsage: Medium term, would be good to think more about how an entity author (working in yaml) can control the different phases. This is good enough for now, but we'll no doubt hit a use-case where one needs to reboot at a different phase, or where more phases would be useful (particularly for the equivalent of "sub-classing" within the yaml).

@alasdairhodge has some good thoughts on this, as has @ahgittin - worth a bigger discussion on that.

software/base/src/main/java/brooklyn/entity/basic/VanillaWindowsProcess.java
```diff
@@ -0,0 +1,54 @@
...
+            "command to run to stop the process");
+    ConfigKey<String> CUSTOMIZE_COMMAND = ConfigKeys.newStringConfigKey("customize.command",
+            "command to run during the customization phase");
+    ConfigKey<String> CUSTOMIZE_POWERSHELL_COMMAND = ConfigKeys.newStringConfigKey("customize.powershell.command",
+            "powershell command to run during the customization phase");
+    ConfigKey<String> INSTALL_COMMAND = ConfigKeys.newStringConfigKey("install.command",
+            "command to run during the install phase");
+    ConfigKey<String> INSTALL_POWERSHELL_COMMAND = ConfigKeys.newStringConfigKey("install.powershell.command",
+            "powershell command to run during the install phase");
+    ConfigKey<Duration> REBOOT_UNAVAILABLE_TIMEOUT = ConfigKeys.newDurationConfigKey("reboot.unavailable.timeout",
```
aledsage: Same name for config key in REBOOT_UNAVAILABLE_TIMEOUT and REBOOT_AVAILABLE_TIMEOUT (always look out for that - consequences are really bad! it ends up treating both keys as the same thing, which is very hard to debug!)

aledsage: Description is also the same for both.

What does this do? Is there a better name than "unavailable"? e.g. is one a `REBOOT_BEGUN_TIMEOUT`, and the other a `REBOOT_COMPLETED_TIMEOUT`?

software/base/src/main/java/brooklyn/entity/basic/VanillaWindowsProcess.java
```diff
@@ -0,0 +1,54 @@
...
+public interface VanillaWindowsProcess extends AbstractVanillaProcess {
+    ConfigKey<String> PRE_INSTALL_POWERSHELL_COMMAND = ConfigKeys.newStringConfigKey("pre.install.powershell.command",
+            "powershell command to run during the pre-install phase");
+    ConfigKey<Boolean> PRE_INSTALL_REBOOT_REQUIRED = ConfigKeys.newBooleanConfigKey("pre.install.reboot.required",
+            "indicates that a reboot should be performed after the pre-install command is run", false);
+    ConfigKey<Boolean> INSTALL_REBOOT_REQUIRED = ConfigKeys.newBooleanConfigKey("install.reboot.required",
+            "indicates that a reboot should be performed after the install command is run", false);
+    ConfigKey<Boolean> CUSTOMIZE_REBOOT_REQUIRED = ConfigKeys.newBooleanConfigKey("customize.reboot.required",
+            "indicates that a reboot should be performed after the customize command is run", false);
+    ConfigKey<String> LAUNCH_POWERSHELL_COMMAND = ConfigKeys.newStringConfigKey("launch.powershell.command",
```
aledsage: Why do others have `CUSTOMIZE_COMMAND` and `CUSTOMIZE_POWERSHELL_COMMAND`, but launch only has `LAUNCH_POWERSHELL_COMMAND`?

software/base/src/main/java/brooklyn/entity/software/StaticSensor.java
```diff
@@ -0,0 +1,44 @@
...
+import brooklyn.entity.basic.EntityLocal;
+import brooklyn.entity.effector.AddSensor;
+import brooklyn.event.basic.Sensors;
+import brooklyn.util.config.ConfigBag;
+
+public class StaticSensor<T> extends AddSensor<Integer> {
+
+    public static final ConfigKey<Integer> STATIC_VALUE = ConfigKeys.newConfigKey(Integer.class, "static.value");
+
+    private final Integer value;
```
aledsage: Can we generalise this so it is for more than just integer?

software/base/src/main/java/brooklyn/entity/software/StaticSensor.java
```diff
@@ -0,0 +1,44 @@
...
+package brooklyn.entity.software;
+
+import brooklyn.config.ConfigKey;
+import brooklyn.entity.basic.ConfigKeys;
+import brooklyn.entity.basic.EntityLocal;
+import brooklyn.entity.effector.AddSensor;
+import brooklyn.event.basic.Sensors;
+import brooklyn.util.config.ConfigBag;
+
+public class StaticSensor<T> extends AddSensor<Integer> {
```
aledsage: There is no test for `StaticSensor`. Can we add one?

software/base/src/main/java/brooklyn/entity/software/winrm/WindowsPerformanceCounterSensors.java
```diff
@@ -0,0 +1,74 @@
...
+                .entity(entity);
+        for (Map<String, String> sensorConfig : sensors) {
+            String sensorType = sensorConfig.get("sensorType");
+            Class<?> clazz;
+            try {
+                clazz = Strings.isNonEmpty(sensorType) ?
+                ((EntityInternal)entity).getManagementContext().getCatalog().getRootClassLoader().loadClass(sensorType) : String.class;
+            } catch (ClassNotFoundException e) {
+                LOG.warn("Could not load type {} for sensor {}", sensorType, sensorConfig.get("name"));
+                clazz = String.class;
```
aledsage: Do we want to fall back to `String`, or do we want to fail?

aledsage:
Thanks @nakomis !

I've finished reading through the code. A lot of comments, mostly minor. Quite a few things can be deferred to future PRs I think.

Could do with some more tests wherever possible as well (again another PR could potentially do that, as long as we're confident that PR will follow soon after!)

aledsage:
Build failure was because it couldn't find `io.cloudsoft.windows:winrm4j:jar:0.1.0-SNAPSHOT`.

However, version 0.1.0 is now released to maven central: http://search.maven.org/#search%7Cga%7C1%7Cio.cloudsoft.windows%20winrm4j

core/src/main/java/brooklyn/event/feed/windows/WindowsPerformanceCounterFeed.java
aledsage: Means poll every 2 seconds

core/src/main/java/brooklyn/event/feed/windows/WindowsPerformanceCounterFeed.java
aledsage: Status code is unreliable (it seems, empirically) - returns 0 sometimes even when failed (but never returns non-zero on success; but if the connection drops, then we'll get an exception though the command might continue executing on the Windows machine to completion).

software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
aledsage: 3389 is RDP; 5985 is WinRM
3389 isn't used by Brooklyn, but useful for the end-user subsequently.

nakomis:
A sample brooklyn.properties setup for a Windows AWS location is as follows:

```
# WinRM AWS Oregon
brooklyn.location.named.AWS\ Oregon\ Win=jclouds:aws-ec2:us-west-2
brooklyn.location.named.AWS\ Oregon\ Win.displayName = AWS Oregon (Windows)
brooklyn.location.named.AWS\ Oregon\ Win.imageId = us-west-2/ami-8fd3f9bf
brooklyn.location.named.AWS\ Oregon\ Win.hardwareId = m3.medium
brooklyn.location.named.AWS\ Oregon\ Win.useJcloudsSshInit=false
brooklyn.location.named.AWS\ Oregon\ Win.templateOptions={ subnetId: subnet-a10e96c4 , securityGroupIds: [['sg-a2d0c2c7']], mapNewVolumeToDeviceName: ["/dev/sda1", 100, true]}
```

NOTE: The subnetId and securityGroupIds are only required if a specific subnet is required (e.g. if you are to join a domain, and require a subnet where Domain Services have been configured). The mapNewVolumeToDeviceName gives a larger `C:\` drive than the default 30Gb (often needed for MS installations)

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
nakomis: This is documented in the AWS documentation here: http://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/UsingConfig_WinAMI.html#user-data-execution

software/database/src/main/resources/brooklyn/entity/database/mssql/installmssql.ps1
```diff
@@ -0,0 +1,49 @@
...
+$WebClient.DownloadFile( $url, $path )
+
+$mountResult = Mount-DiskImage $Path -PassThru
+$driveLetter = (($mountResult | Get-Volume).DriveLetter) + ":\"
+
+New-Item -ItemType Directory -Force -Path "C:\Program Files (x86)\Microsoft SQL Server\DReplayClient\ResultDir"
+New-Item -ItemType Directory -Force -Path "C:\Program Files (x86)\Microsoft SQL Server\DReplayClient\WorkingDir"
+
+Install-WindowsFeature NET-Framework-Core
+
```
nakomis: Simply calling "D:\setup.exe" at this point will work if the script is run locally, but will fail if the script is run via winrm over http as the credentials used for execution will be marked as remote credentials. To work around this, we re-authenticate within the script to obtain a new set of 'local' credentials and call setup.exe using the 'Invoke-Command scriptlet which allows us to pass in the newly-obtained credentials

software/database/src/main/resources/brooklyn/entity/database/mssql/installmssql.ps1
```diff
@@ -0,0 +1,49 @@
...
+$pass = '${attribute['windows.password']}'
+$secpasswd = ConvertTo-SecureString $pass -AsPlainText -Force
+$mycreds = New-Object System.Management.Automation.PSCredential ($($env:COMPUTERNAME + "\Administrator"), $secpasswd)
+
+Invoke-Command -ComputerName localhost -credential $mycreds -scriptblock {
+    param($driveLetter)
+    Start-Process ( $driveLetter + "setup.exe") -ArgumentList "/ConfigurationFile=C:\ConfigurationFile.ini" -RedirectStandardOutput "C:\sqlout.txt" -RedirectStandardError "C:\sqlerr.txt" -Wait
```
nakomis: NOTE: The $driveLetter defined on line 33 is unavailable in this scriptblock, and so is passed into the scriptblock via the argumentlist on line 47. The param($driveLetter) directive on line 45 indicates that the zeroth argument in the argumentlist on line 47 should be referenced as "$driveLetter" within the script block

docs/guide/yaml/winrm/index.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Windows blueprints using WinRM
+layout: website-normal
+children:
+- about-winrm.md
+- re-authentication.md
+- stdout-and-stderr.md
+---
+
+This guide describes how Brooklyn entities can be easily created from Chef cookbooks.
```
richardcloudsoft: This text and link is c&p from the Chef pages and needs re-written for Windows!

docs/guide/yaml/winrm/re-authentication.md
```diff
@@ -0,0 +1,38 @@
...
+'remote' credentials, which are prohibited from running certain security-related operations. The solution was to obtain
+a new set of credentials within the script and use those credentials to exeute the installer, so this:
+
+```
+( $driveLetter + "setup.exe") /ConfigurationFile=C:\ConfigurationFile.ini
+```
+
+became this:
+
+$pass = '${attribute['windows.password']}'
```
richardcloudsoft: Fence with ``` otherwise the formatting is mangled.

(In fact, fence with ```powershell to get syntax highlighting)

aledsage:
@nakomis please close this PR, in favour of https://github.com/apache/incubator-brooklyn/pull/664

software/base/src/main/java/brooklyn/entity/basic/VanillaWindowsProcessImpl.java
```diff
@@ -0,0 +1,39 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package brooklyn.entity.basic;
+
+public class VanillaWindowsProcessImpl extends SoftwareProcessImpl implements VanillaWindowsProcess {
+    @Override
+    public Class getDriverInterface() {
```
andreaturli: instead of creating an entire new entity that duplicated the VanillaProcessImpl, wouldn't be possibile to load the appropriate driver for the type of location where the entity will be installed? ie SshDriver vs WinRmDriver etc?

PR #651, opened 2015-05-21
allow commands for EmptySoftwareProcess

ahgittin:
if pre- and post- install and launch commands are set for EmptySoftwareProcess, run them

### DIFF
--- a/software/base/src/main/java/brooklyn/entity/basic/EmptySoftwareProcessSshDriver.java
+++ b/software/base/src/main/java/brooklyn/entity/basic/EmptySoftwareProcessSshDriver.java
@@ -18,6 +18,7 @@
  */
 package brooklyn.entity.basic;
 
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import brooklyn.location.basic.SshMachineLocation;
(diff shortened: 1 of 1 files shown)


software/base/src/main/java/brooklyn/entity/basic/EmptySoftwareProcessSshDriver.java
```diff
@@ -43,25 +44,22 @@ public void install() { }
...
-
-    @Override
-    public void runPostInstallCommand(String command) { }
-
-    @Override
-    public void runPreLaunchCommand(String command) { }
+    public void copyInstallResources() { 
+        Map<String, String> installFiles = entity.getConfig(SoftwareProcess.INSTALL_FILES);
+        Map<String, String> installTemplates = entity.getConfig(SoftwareProcess.INSTALL_TEMPLATES);
+        if ((installFiles!=null && !installFiles.isEmpty()) || (installTemplates!=null && !installTemplates.isEmpty())) {
```
neykov: What's the point of guarding the `super` call? To prevent mkdir?
Whatever the reason worth a comment.

neykov:
Nice changes, no reason to deviate from the standard and expected behaviour.

ggrunin:
tested, it works

software/base/src/main/java/brooklyn/entity/basic/EmptySoftwareProcessSshDriver.java
ahgittin: prevent `mkdir`, correct.  adding comment.

PR #653, opened 2015-05-21
ssh command sensor allows configuring the dir where it runs

ahgittin:
fixes bug if ssh sensor used with empty software process entity, where run dir isn't yet created


ggrunin:
tested - it works.

PR #654, opened 2015-05-22
Replaced Jetty6Server download url.

mbarrientos:
Codehaus seems to be terminating its services (http://www.codehaus.org/termination-410.html) and not hosting/proxying distributions anymore.


ahgittin:
great - merging

ahgittin:
and shedding a tear for @codehaus

PR #655, opened 2015-05-22
Jclouds location resolver

sjcorbett:
A handful of disparate additions:
* Adjusts `JcloudsLocation` and its resolver so that it is easier to subclass.
* Exposes the predicate used to find machines in `JcloudsLocation.rebind` for customisation by subclasses.
* Adjusts the web server's logging filter to log requests that take a long time to complete.
* Adjusts EffectorUtils to rethrow `PropagatedRuntimeExceptions` rather than wrapping them in more exceptions.


sjcorbett:
Build server issue: `java.lang.RuntimeException: unable to find a free port at or above 8080`.

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1540,6 +1540,58 @@ public JcloudsSshMachineLocation rebindMachine(Map<?,?> flags, NodeMetadata meta
...
 
     /**
+     * @return a predicate that returns true if a {@link ComputeMetadata} instance is suitable for
+     *      rebinding to given the configuration in {@link ConfigBag config}.
+     */
+    protected Predicate<ComputeMetadata> getComputeMetadataPredicate(ConfigBag config) {
+        final String rawId = (String) config.getStringKey("id");
+        final String rawHostname = (String) config.getStringKey("hostname");
+        final String rawRegion = (String) config.getStringKey("region");
+        return new Predicate<ComputeMetadata>() {
```
ahgittin: if this would ever be used in config then we should avoid anonymous classes

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocationResolver.java
```diff
@@ -50,10 +50,10 @@
 
     public static final Logger log = LoggerFactory.getLogger(JcloudsLocationResolver.class);
     
-    public static final String JCLOUDS = "jclouds";
+    private static final String JCLOUDS = "jclouds";
     
-    public static final Map<String,ProviderMetadata> PROVIDERS = getProvidersMap();
-    public static final Map<String,ApiMetadata> APIS = getApisMap();
+    private static final Map<String,ProviderMetadata> PROVIDERS = getProvidersMap();
+    private static final Map<String,ApiMetadata> APIS = getApisMap();
```
ahgittin: extensions might be accessing these.  if they need to be private then let's go through a deprecation cycle.

locations/jclouds/src/test/java/brooklyn/location/jclouds/BrooklynMachinePoolLiveTest.java
```diff
@@ -98,6 +98,6 @@ public void buildClaimAndDestroy() {
     
 
     private JcloudsLocation resolve(String spec) {
-        return (JcloudsLocation) managementContext.getLocationRegistry().resolve(JcloudsLocationResolver.JCLOUDS+":"+spec);
+        return (JcloudsLocation) managementContext.getLocationRegistry().resolve("jclouds:"+spec);
```
ahgittin: variable reference is better than hard-coded string i think

ahgittin:
lots of good stuff here.  a few minor comments, then good to merge i think.

locations/jclouds/src/test/java/brooklyn/location/jclouds/BrooklynMachinePoolLiveTest.java
sjcorbett: There are so many classes that hardcode "jclouds:..." that I think it's a minor one, but I'll switch it for `new JcloudsLocationResolver().getPrefix()`

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocationResolver.java
sjcorbett: I don't see much reason for them to be public. Protected would make more sense.

sjcorbett:
@ahgittin Thanks

PR #656, opened 2015-05-22
Startup option to exit once the app is deployed and running

neykov:
* Add an option exitAndLeaveAppsRunningAfterStarting to exit the process once the --app entity is up and running
* Force a persist after launching is complete so that the app persistence is guaranteed before exiting
* Pull the startup complete notification before starting the --app entity so that the progress can be monitored in the console.


ahgittin:
good addition, merging

PR #657, opened 2015-05-22
Troubleshoot "no ports available" error

neykov:
Run some system commands whenever we get the dreaded "unable to find a free port" error. Looking for possible causes.

DO NOT merge. For now troubleshooting as a PR build.


neykov:
Problem fixed in #662.

PR #659, opened 2015-05-26
Upgrade nginx addons

neykov:
* pcre too old and version no longer hosted at official url, updated to currently installed version by brew (as tar download only used on OS X)
* sticky module moved development to bitbucket

Sticky module's archive untars to a different folder than the name of the download, updated the ssh driver to force a subfolder name.

Sticky module (any version) doesn't currently compile with nginx 1.8.0 on OS X. See https://bitbucket.org/nginx-goodies/nginx-sticky-module-ng/issue/16/can-not-compile-on-macosx-yosemite. Added a temporary workaround until a fix is released.

### DIFF
--- a/software/webapp/src/main/java/brooklyn/entity/proxy/nginx/NginxSshDriver.java
+++ b/software/webapp/src/main/java/brooklyn/entity/proxy/nginx/NginxSshDriver.java
@@ -96,6 +96,7 @@ public class NginxSshDriver extends AbstractSoftwareProcessSshDriver implements 
     }
 
     /** By default Nginx writes the pid of the master process to {@code logs/nginx.pid} */
+    @Override
     public String getPidFile() {
         return format("%s/%s", getRunDir(), NGINX_PID_FILE);
     }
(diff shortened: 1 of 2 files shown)


neykov:
Tested successfully on SL with images: CENTOS_6_64, CENTOS_7_64, REDHAT_6_64, REDHAT_7_64, UBUNTU_12_64, UBUNTU_14_64.

Failed on:
  * DEBIAN_{6,7}_64 - missing curl
  * CENTOS_5_64, REDHAT_5_64 - curl: option --keepalive-time: is unknown

software/webapp/src/main/java/brooklyn/entity/proxy/nginx/NginxSshDriver.java
```diff
@@ -216,6 +220,12 @@ public void install() {
         String withLdOpt = entity.getConfig(NginxController.WITH_LD_OPT);
         if (isMac) withLdOpt = format("-L%s/pcre-dist/lib", getInstallDir()) + (Strings.isBlank(withLdOpt) ? "" : " " + withLdOpt);
         String withCcOpt = entity.getConfig(NginxController.WITH_CC_OPT);
+        
+        if (isMac) {
+            // TODO Upgrade sticky module as soon as a fix for https://bitbucket.org/nginx-goodies/nginx-sticky-module-ng/issue/16/can-not-compile-on-macosx-yosemite
+            // is released and remove this block.
+            withCcOpt = (Strings.isBlank(withCcOpt) ? "" : (withCcOpt + " ")) + "-Wno-error";
```
andreaturli: would it make sense to move it into the `withCcOpt` configKey? wouldn't that be beneficial for other distros or just mac?

andreaturli:
lgtm, tested on localhost and it works fine for me as well.

software/webapp/src/main/java/brooklyn/entity/proxy/nginx/NginxSshDriver.java
neykov: The problem is just on mac. Wanted to have it applied even if the user supplies custom `withCcOpt` value, making it compatible with existing blueprints.

andreaturli:
merging now

andreaturli:
thanks @neykov

PR #660, opened 2015-05-26
Add REST API Doc Page to Developer Guide

mikezaccardo:
Adds a page which directs the user to the REST API Documentation within the Brooklyn web console.

Previously, it was unclear where this documentation could be found but now it is explicitly mentioned.


sjcorbett:
Looks good.

PR #661, opened 2015-05-27
FIX: library JsonPath was updated to 2.0.0

kiuby88:
The JsonPath was updated to last available version

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -145,7 +145,7 @@
         <mockito.version>1.10.8</mockito.version>
         <swagger.version>1.0.1</swagger.version>
         <jansi.version>1.2.1</jansi.version>
-        <gson.version>2.2.2</gson.version>
+        <gson.version>2.3</gson.version>
         <ivy.version>2.2.0</ivy.version>
         <mx4j.version>3.0.1</mx4j.version>
         <bouncycastle.version>1.49</bouncycastle.version>
@@ -158,7 +158,7 @@
         <freemarker.version>2.3.19</freemarker.version>
         <commons-io.version>2.4</commons-io.version>
         <hazelcast.version>3.0</hazelcast.version>
-        <jsonPath.version>0.9.1</jsonPath.version>
+        <jsonPath.version>2.0.0</jsonPath.version>
         <commons-compress.version>1.4</commons-compress.version>
         <qpid.version>0.20</qpid.version>
         <mongodb.version>2.11.4</mongodb.version>


andreaturli:
According to http://mvnrepository.com/artifact/com.jayway.jsonpath/json-path/2.0.0 jsonPath:2.0.0 depends on `com.google.code.gson:gson:2.3.1` while brooklyn is using `2.2.2` 

we probably need to update the `gson` dep as well. @richardcloudsoft, @grkvlt do you see any problems in that?

richardcloudsoft:
I think that Gson is a dependency of jclouds too. What version does jclouds depend on? Are these two dependencies going to introduce a conflict?

andreaturli:
good point @richardcloudsoft, we need to investigate a little more

ahgittin:
meh dependencies.  we should load jclouds via osgi.

google can be quite harsh in minor version changes so i agree, proceed with caution

@kiuby88 is there a need for 2.0.0 -- apart from wanting to use recent (and non-zero major) versions ?

kiuby88:
Hi all,

Following the PaasLocation proposed in [pull 629] (https://github.com/apache/incubator-brooklyn/pull/629) @mbarrientos and I are defining an approach to allows to carry out the deployment of an application on [CloudFoundry from Brooklyn](https://github.com/kiuby88/brooklyn-location-cloudfoundry/tree/non-machinelocation).

In this project, we need parsing and extract some information from a Json, for this we use the  indefinite expressions management provided by JsonPath, but the current version used by Brooklyn does not support several operators (such as `=`) which are necessary in our CloudFoundry management.

We are using version v2.0.0 because is the last available, of course we could use an older version, but from [v1.0.0](http://mvnrepository.com/artifact/com.jayway.jsonpath/json-path/1.0.0),  gson v2.3.X is always used. Then, I think we could update jsonpath to v2.0.0 but we could exclude the gson v2.3.1 from jsonpath library and use the current version used by jclouds (v2.2.2). I have checked this point and it should works fine (for our project)

WDYT?

andreaturli:
@richardcloudsoft, jclouds 1.9.0 uses [gson 2.3](https://github.com/jclouds/jclouds/blob/master/core/pom.xml#L80-L84) while
brooklyn uses [gson 2.2.2](https://github.com/apache/incubator-brooklyn/blob/321d8879068c1d5a159962c486c3851c2eaa64b8/pom.xml#L148)

I think @ahgittin is right, we should try to use osgi for jclouds and for everything it is possible, but meanwhile I think having the same gson version won't hurt (gson 2.3)

@kiuby88 would be jsonPath [v1.2.0](http://mvnrepository.com/artifact/com.jayway.jsonpath/json-path/1.2.0) ok for you?

ahgittin:
ooh if jclouds 1.9 is on gson 2.3.x but brooklyn is on 2.2 that's bad.  :)

we should increase brooklyn's version in any case.

i'm not worried between various 2.3.x.

@kiuby88 can you resubmit this PR bumping GSON to latest 2.3.x ?  then if tests pass i think we're good to go.

kiuby88:
@andreaturli I think v1.2.0 should work fine, but as @ahgittin I updated gson dependency to 2.3.X

ahgittin:
LGTM

PR #663, opened 2015-05-27
fix and test better rest client

ahgittin:
plus other misc rest cleanups

### DIFF
--- a/usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
+++ b/usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
@@ -20,6 +20,10 @@ package brooklyn.rest.client;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.net.MalformedURLException;
 import java.net.URL;
 
@@ -52,6 +56,8 @@ import brooklyn.rest.api.SensorApi;
 import brooklyn.rest.api.ServerApi;
 import brooklyn.rest.api.UsageApi;
 import brooklyn.rest.api.VersionApi;
+import brooklyn.util.exceptions.Exceptions;
+import brooklyn.util.http.BuiltResponsePreservingError;
 
 /**
  * @author Adam Lowe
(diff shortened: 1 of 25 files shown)


usage/rest-server/src/main/java/brooklyn/rest/security/provider/DelegatingSecurityProvider.java
```diff
@@ -94,10 +97,16 @@ private synchronized SecurityProvider loadDelegate() {
...
+                Object delegateO = constructor.newInstance();
+                if (!(delegateO instanceof SecurityProvider)) {
+                    // if classloaders get mangled it will be a different CL's SecurityProvider
+                    throw new ClassCastException("Delegate is either not a security provider or has an incompatible classloader: "+delegateO);
+                }
+                delegate = (SecurityProvider) delegateO;
             }
         } catch (Exception e) {
             log.warn("REST unable to instantiate security provider " + className + "; all logins are being disallowed", e);
+            e.printStackTrace();
```
neykov: Remove it, probably a leftover.

usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
```diff
@@ -170,7 +169,7 @@ public AccessApi getAccessApi() {
         return proxy(AccessApi.class);
     }
 
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({ "unchecked", "rawtypes" })
```
neykov: Better use a wildcard in the cast, no need for any `@SuppressWarnings` then, i.e.
```
ClientResponse<?> clientResponse = (ClientResponse<?>) response;
```

Same for next method.

usage/rest-server/src/main/java/brooklyn/rest/security/provider/DelegatingSecurityProvider.java
```diff
@@ -94,10 +97,16 @@ private synchronized SecurityProvider loadDelegate() {
                 delegate = constructor.newInstance(mgmt);
             } catch (Exception e) {
                 constructor = clazz.getConstructor();
-                delegate = constructor.newInstance();
+                Object delegateO = constructor.newInstance();
+                if (!(delegateO instanceof SecurityProvider)) {
+                    // if classloaders get mangled it will be a different CL's SecurityProvider
+                    throw new ClassCastException("Delegate is either not a security provider or has an incompatible classloader: "+delegateO);
```
neykov: The reason for the different class loaders is that `BrooklynApiRestClientTest` [configures the `WebAppContext` with `.setExtraClasspath("./target/test-rest-server/");`](https://github.com/apache/incubator-brooklyn/blob/317b69e17338bb541bc903b5ed97e30e58a929a6/usage/rest-client/src/test/java/brooklyn/rest/client/BrooklynApiRestClientTest.java#L70). The additional path contains the server classes, but not the test classes. This causes the test classes to be loaded through the app class loader, while the web app class loader sees the non-test classes.
Why the extra classpath? If there's not a particular reason, then better let the app classloader handle the server classes. This will also help when changing them and running the tests in the client project in Eclipse - currently both projects need to be rebuild for the changes to take effect.
If there's a good reason to keep the extra classpath, then in pom.xml install the `<classifier>tests</classifier>` of `brooklyn-rest-server`.

neykov:
Some minor comments worth addressing.

usage/rest-server/src/main/java/brooklyn/rest/security/provider/DelegatingSecurityProvider.java
ahgittin: I can't see any reason for it.  Makes sense to me to remove all the `target/test-rest-server` references, including in the pom.

It seems to work with them removed.

ahgittin:
however we are getting the `Make sure to release the connection before allocating another one.` error from resteasy in the integration tests.  i've seen this in other uses so i will fix it.

http://meera-subbarao.blogspot.co.uk/2011/03/resteasy-connection-release-problems.html

ahgittin:
okay, i think that's fixed -- the client now extracts the entity and closes the stream so it is free to re-use it.

surprised this isn't an off-the-shelf option!

ahgittin:
failure unrelated:

    java.lang.IllegalArgumentException: Unable to access URL brooklyn.webconsole.security.keystore.url: /home/jenkins/jenkins-slave/workspace/incubator-brooklyn-pull-requests%402/usage/launcher/target/test-classes/server.ks

perhaps due to URL changing `-` to `%40` then treating as a file?

ahgittin:
merging

PR #664, opened 2015-05-27
Support for deployment to windows (strike 2)

aledsage:
Builds on https://github.com/apache/incubator-brooklyn/pull/650, incorporating many of the pull request comments from there.

This also has @nakomis's commit(s) for https://github.com/apache/incubator-brooklyn/pull/639. Would be nice to address those before merge, but am happy if we merge then refactor so we can get the rest of this goodness.

The following have been deferred for this PR, but really must be addressed soon:

* Comments in PR #639 (@nakomis's PR on which this is built included those changes)

* More tests (!), including:

  * WinRmMachineLocation's script execution
    (there is one test, which needs attention!)
  
  * VanillaWindowsProcess
    (there is one test, which needs attention - MssqlBlueprintLiveTest)

  * JcloudsWinRmMachineLocation's public/private/subnet addresses

* Resolve naming differences in config between WinRM location/tool and SshTool/SshMachineLocation, such as SshTool#PROP_SSH_TRIES. Should these be merged int one well-named generic thing, deprecating the previous ssh config key?

* Resolve naming differences in JcloudsLocation, where it's ssh-specific. For example, should WAIT_FOR_WINRM_AVAILABLE and WAIT_FOR_SSHABLE. Would probably have to deprecate the old key, and generalise the name.

* JcloudsLocation:

  * Allow for user-specified credentials: want to support setting up a given user on the VM (or resetting the Administrator password), and to support vCloudDirector use-case where template contains username/password, so don't just use what is returned by jclouds.

  * Support for port-forwarding in JcloudsWinRmMachineLocation.

  * Resolve differences in "waitForReachable" between Windows and ssh code-paths.

  * Can we ensure the returned JcloudsMachineLocation will have the right node.getLoginPort for WinRM? How do we ensure that happens?

  * Generalise the sshHostAndPortOverride (e.g. rename to connectionHostAndPortOverride). Use it for the WinRM port in the windows case.

  * jcloudsLocation: need to release port-forwarding when it's a Windows VM

* WinRmMachineLocation: respect the configured `port` as the one to use when connecting (essential for port-forwarding scenarios).

* We should list (in docs / limitations) the main config options that are not supported (yet) on windows (rather than the user/customer discovering it by trying to configure it and eventually finding this log message, or just it being ignored silently (as would be the case for user/password configuration).

Some unaddressed minor comments:

* WinRmMachineLocation: worth a comment describing the file-copy approach used.

* WinRmMachineLocation.getDefaultUserMetadataString: worth some javadoc to say what this default user metadata will do.

* JcloudsLocation waitForReachable: would catch-block always be for a NumberFormatException? If so, can we be more specific?

* I wonder why we don't set the parent inside JcloudsLocation.registerJcloudsSshMachineLocation.

* AbstractSoftwareProcessWinRmDriver.WINDOWS_USERNAME: is inside the driver the best place to declare these sensors?

* Unaddressed questions from PR #650:

  * in jcloudsLocation, "This changes the execCommands description. I take it that's deliberate? Was the full script being logged multiple times before?"

  * in jcloudsLocation, "What exception are you expecting here? Can we just program more defensively when building the message?"

  * in JcloudsLocation, "Remove try-catch! @Nakomis: why catch exception when doing log message? What exception happened during logging?"

  * AbstractSoftwareProcessWinRmDriver.rebootAndWait: Can we be more specific - e.g. what exception do you expect. And log if not exactly right. There's a risk we'll pretend the VM is rebooting when actually we got back an error about permissions or something like that.

Other comments needing more thought (but less time critical):

* DynamicCluster.grow will not behave right for availability zones if the memberSpec defines a location. It will use that location in preference to the cluster's location.

* WinRmMachineLocation has a retry-loop when calling WinrmTool. This feels quite different from the SshMachineLocation implementation. I wonder if it's worth creating a wrapper of WinrmTool, such as for retries etc.

* Not sure about the name VanillaWindowsProcessDriver versus VanillaSoftwareProcessDriver. They are both software processes.

* Medium term, would be good to think more about how an entity author (working in yaml) can control the different phases. This is good enough for now, but we'll no doubt hit a use-case where one needs to reboot at a different phase, or where more phases would be useful (particularly for the equivalent of "sub-classing" within the yaml).
@alasdairhodge has some good thoughts on this, as has @ahgittin - worth a bigger discussion on that.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,4 +3,4 @@
 *.sh text eol=lf
 *.bat text eol=crlf
 *.ps1 text eol=crlf
-
+*.ini text eol=crlf
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -65,7 +65,11 @@
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>io.cloudsoft.windows</groupId>
+            <artifactId>winrm4j</artifactId>
+            <version>${winrm4j.version}</version>
+        </dependency>
         <dependency>
             <groupId>net.schmizz</groupId>
             <artifactId>sshj</artifactId>
--- a/core/src/main/java/brooklyn/entity/basic/BrooklynConfigKeys.java
+++ b/core/src/main/java/brooklyn/entity/basic/BrooklynConfigKeys.java
@@ -145,6 +145,7 @@ public class BrooklynConfigKeys {
 
     public static final ConfigKey<Boolean> START_LATCH = newBooleanConfigKey("start.latch", "Latch for blocking start until ready");
     public static final ConfigKey<Boolean> SETUP_LATCH = newBooleanConfigKey("setup.latch", "Latch for blocking setup until ready");
+    public static final ConfigKey<Boolean> PRE_INSTALL_RESOURCES_LATCH = newBooleanConfigKey("resources.preInstall.latch", "Latch for blocking pre-install resources until ready");
     public static final ConfigKey<Boolean> INSTALL_RESOURCES_LATCH = newBooleanConfigKey("resources.install.latch", "Latch for blocking install resources until ready");
     public static final ConfigKey<Boolean> INSTALL_LATCH = newBooleanConfigKey("install.latch", "Latch for blocking install until ready");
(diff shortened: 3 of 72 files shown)


aledsage:
Closing, now that #665 has been merged.

PR #665, opened 2015-05-28
Support for deployment to windows (strike 3)

rdowner:
Builds on @aledsage's PR #664, which in turn builds on @nakomis's PR #650. This fixes a couple of regressions that caused live tests to fail, and improves the documentation.

Please see PR #664 for lots more detail!


aledsage:
Looks good; merging.

PR #669, opened 2015-06-01
Don't create install dir in copy-pre-install-resources unless needed

neykov:
Lots of unit tests assume that no ssh connection will be established when testing localhost entities. The addition of `copy-pre-install-resources` breaks the tests if local ssh is not configured properly. Create install dir only if there are resources to copy (and therefore the test expects it will try to ssh).


grkvlt:
Looks simple enough to me, am happy merging :frog:

PR #671, opened 2015-06-02
Remove shared statics causing issues in Clocker

grkvlt:
Clocker has more than one `JcloudsLocation` so the check for exceptions is skipped sometimes, becasue the guard is a static variable shared by all instances. This gives unhelpful error messages occasionally, particularly in the presence of network failures raising `IOException`s.

### DIFF
--- a/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
@@ -46,6 +46,7 @@ import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -204,7 +205,8 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
     private static final Pattern LIST_PATTERN = Pattern.compile("^\\[(.*)\\]$");
     private static final Pattern INTEGER_PATTERN = Pattern.compile("^\\d*$");
 
-    private static boolean loggedSshKeysHint = false;
+    private final AtomicBoolean loggedSshKeysHint = new AtomicBoolean(false);
+    private final AtomicBoolean listedAvailableTemplatesOnNoSuchTemplate = new AtomicBoolean(false);
 
     private final Map<String,Map<String, ? extends Object>> tagMapping = Maps.newLinkedHashMap();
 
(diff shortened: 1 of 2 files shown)


locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1502,8 +1504,7 @@ protected UserCreation createUserStatements(@Nullable Image image, ConfigBag con
                  * if this config key is not set, use a key `brooklyn_id_rsa` and `.pub` in `MGMT_BASE_DIR`,
                  * with permission 0600, creating it if necessary, and logging the fact that this was created.
                  */
-                if (!loggedSshKeysHint && !config.containsKey(PRIVATE_KEY_FILE)) {
-                    loggedSshKeysHint = true;
+                if (loggedSshKeysHint.compareAndSet(false, true) && !config.containsKey(PRIVATE_KEY_FILE)) {
```
ahgittin: `&&` args should be the other way round so that boolean is only set true if message is logged

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -204,7 +205,8 @@
     private static final Pattern LIST_PATTERN = Pattern.compile("^\\[(.*)\\]$");
     private static final Pattern INTEGER_PATTERN = Pattern.compile("^\\d*$");
 
-    private static boolean loggedSshKeysHint = false;
+    private final AtomicBoolean loggedSshKeysHint = new AtomicBoolean(false);
+    private final AtomicBoolean listedAvailableTemplatesOnNoSuchTemplate = new AtomicBoolean(false);
```
ahgittin: `JcloudsLocation` instances do tend to be re-used where it's the same location so this seems like a good idea to me; means we might log more than we should, but that's better than the previous which logged less than we should!

ahgittin:
one comment of note -- the `&&` args order -- then good to merge

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
grkvlt: Yes, I wasn't sure what to do to achieve the actual intent. `ThreadLocal` wouldn't work, maybe a `Map` with the location ID as the key? I wasn't sure whether the IDs would be the same for two `Location` instances generated by the management context for the same underlying location?

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
grkvlt: Right, yes...

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
ahgittin: a guava `Cache` is maybe the hyper-engineered way but i think tying it to the `JcloudsLocation` instance is a pragmatic step.  the `JcloudsLocation` instance should be cached, i *think*, and if not all we do is log some debug/info a bit more.

grkvlt: Can look into that in a further PR, though...

PR #672, opened 2015-06-02
Don't spin BrooklynEntityMirror forever on failure

neykov:
* Fail on non-{2xx, 403} response received
* Fail on BrooklynNode failure

### DIFF
--- a/software/base/src/main/java/brooklyn/entity/brooklynnode/EntityHttpClient.java
+++ b/software/base/src/main/java/brooklyn/entity/brooklynnode/EntityHttpClient.java
@@ -23,16 +23,42 @@ import java.util.Map;
 import brooklyn.util.http.HttpTool;
 import brooklyn.util.http.HttpToolResponse;
 
+import com.google.common.base.Predicate;
+import com.google.common.base.Predicates;
+import com.google.common.collect.Range;
+
 /**
  * Helpful methods for making HTTP requests to {@link BrooklynNode} entities.
  */
 public interface EntityHttpClient {
+    public static class ResponseCodePredicates {
+        private static class ResponseCodeHealthyPredicate implements Predicate<Integer> {
+            @Override
+            public boolean apply(Integer input) {
+                return HttpTool.isStatusCodeHealthy(input);
+            }
+        }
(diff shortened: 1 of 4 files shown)


software/base/src/main/java/brooklyn/entity/brooklynnode/EntityHttpClient.java
```diff
@@ -23,10 +23,29 @@
...
  */
 public interface EntityHttpClient {
+    interface ResponseCodePredicates {
+        static class ResponseCodeHealthyPredicate implements Predicate<Integer> {
+            @Override
+            public boolean apply(Integer input) {
+                return HttpTool.isStatusCodeHealthy(input);
+            }
+        }
+        Predicate<Integer> INFORMATIONAL = Range.closed(100, 199);
```
aledsage: I have a personal preference for methods rather than static constants, which gives us future flexibility. That follows the guava convention, such as the method `Predicates.alwaysTrue()`. In this case I can't see why we'd ever change it, but I'd prefer us sticking to the general guava pattern.

software/base/src/main/java/brooklyn/entity/brooklynnode/EntityHttpClientImpl.java
```diff
@@ -77,6 +81,12 @@ protected EntityHttpClientImpl(Entity entity, ConfigKey<?> urlConfig) {
         return builder;
     }
 
+    @Override
+    public EntityHttpClient responseSuccess(Predicate<Integer> responseSuccess) {
+        responseSuccess = checkNotNull(responseSuccess, "responseSuccess");
```
aledsage: should be `this.responseSuccess = ...`

aledsage:
Looks good. One thing to fix, and another comment to see what you think of. Then good to merge.

software/base/src/main/java/brooklyn/entity/brooklynnode/EntityHttpClientImpl.java
neykov: ouch, good find!

neykov:
Following the guava conventions seems like a good idea. Addressed comments.

PR #674, opened 2015-06-02
Fix Maven plugin versions

grkvlt:


ahgittin:
LGTM. The diff has done some weird suboptimal matching but all the blocks look like straight version bumps and it compiles nicely.

PR #675, opened 2015-06-02
Brooklyn features

sjcorbett:
Adds `BrooklynVersion.getFeatures()` and exposes in response to `GET /v1/server/version`. The downstream-parent pom is updated to automatically trigger projects' inclusion.


core/src/main/java/brooklyn/BrooklynVersion.java
```diff
@@ -46,201 +58,333 @@
...
+     * @return An iterable containing all features found in the management context's classpath and catalogue.
+     */
+    public static Iterable<BrooklynFeature> getFeatures(ManagementContext mgmt) {
+        Iterable<URL> manifests = ResourceUtils.create(mgmt).getResources(MANIFEST_PATH);
+
+        for (CatalogItem<?, ?> catalogItem : mgmt.getCatalog().getCatalogItems()) {
+            OsgiBrooklynClassLoadingContext osgiContext = new OsgiBrooklynClassLoadingContext(
+                    mgmt, catalogItem.getCatalogItemId(), catalogItem.getLibraries());
+            manifests = Iterables.concat(manifests, osgiContext.getResources(MANIFEST_PATH));
+        }
```
sjcorbett: @neykov Is there a more suitable place or way to search catalogue bundles for manifests?

neykov: I needed something similar the other day and used the following snipped:
```
        Maybe<OsgiManager> osgiManager = managementContext.getOsgiManager();
        if (osgiManager.isPresent()) {
            Framework framework = osgiManager.get().getFramework();
            Bundle[] bundles = framework.getBundleContext().getBundles();
            for (Bundle bundle : bundles) {
                if (bundle == framework) continue;
                if (Osgis.isExtensionBundle(bundle)) continue;

                try {
                    //TODO better .start the bundle on install
                    bundle.loadClass("Some.None.Existent.Class.To.Force.BundleWiring.Initialization");
                } catch (ClassNotFoundException e1) {}
                BundleWiring bundleWiring = bundle.adapt(BundleWiring.class);
                if (bundleWiring == null) continue; //bundle not resolved
                ClassLoader bundleClassLoader = bundleWiring.getClassLoader();
                //do something with the CL, i.e.
                bundleClassLoader.getResources(MANIFEST_PATH)
            }
        }
```

It's hackish and not necessarily better than your approach, best way is probably using OSGi services - but not feasible at this point.

One thing to watch out with scanning the catalog items is that the same bundle can be used in multiple catalog items, i.e. one jar providing multiple entities. Then you will get duplicates in the output.

core/src/main/java/brooklyn/management/classloading/BrooklynClassLoadingContextSequential.java
```diff
@@ -101,6 +102,18 @@ public URL getResource(String resourceInThatDir) {
...
     @Override
+    public Iterable<URL> getResources(String name) {
+        List<Iterable<URL>> resources = Lists.newArrayList();
+        for (BrooklynClassLoadingContext target : primaries) {
+            resources.add(target.getResources(name));
+        }
+        for (BrooklynClassLoadingContext target : secondaries) {
+            resources.add(target.getResources(name));
+        }
+        return Iterables.concat(resources);
```
neykov: It's often the case that the application class loader will be contained multiple times in the BCLCs, even in a single one, leading resources.

core/src/main/java/brooklyn/BrooklynVersion.java
neykov: Looks like OSGi bundles already provide the manifest content through `[getHeaders()](https://osgi.org/javadoc/r4v43/core/org/osgi/framework/Bundle.html#getHeaders%28%29)`. So a nice way to fetch them would be to iterate the bundles, similarly to the above code. This would belong in `Osgis` class, on the other hand the goto place for catalog related code which doesn't belong elsewhere is `CatalogUtils`.

core/src/main/java/brooklyn/BrooklynVersion.java
```diff
@@ -46,201 +58,333 @@
...
+                    if (!Strings.isBlank(value)) {
+                        additionalData.put(name.toString(), value);
+                    }
+                }
+            }
+            if (additionalData.isEmpty()) {
+                return Optional.absent();
+            }
+
+            // Name is special cased as it a useful way to indicate a feature without
```
neykov: Without what? Is the comment complete?

core/src/main/java/brooklyn/BrooklynVersion.java
```diff
@@ -46,201 +58,333 @@
...
+            return getClass().getSimpleName() + "{" + symbolicName + (version != null ? ":" + version : "") + "}";
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hashCode(symbolicName, version);
+        }
+
+        @Override
+        public boolean equals(Object other) {
```
neykov: Brooklyn supports loading different bundles with the same symbolicName+version, namespaced in different catalog items. Might not be worth the trouble supporting them here, but definitely it's something to have in mind.

neykov:
Works great for me. Good to merge.
Going straight to scanning OSGi bundles without going through the catalog items will simplify the changes. It will also enable listing duplicate symbolicName+version bundles. The downside is that we don't unregister bundles when removing catalog items, so even unused bundles will be listed - no perfect solution it seems.

core/src/main/java/brooklyn/BrooklynVersion.java
sjcorbett: Perhaps we should incorporate `getCatalogItemId` into the output instead. If I did this and updated the `equals` method in `BrooklynFeature` to include it we'd give a much clearer indication of what features were installed by which catalogue items.

ahgittin:
I will merge. @sjcorbett @neykov @aledsage let's remember the notes in this PR if we choose to do OSGi scanning earlier.

Also note the feature request where we log the version metadata once for each bundle on first read.  This will make the log files more useful if we have to debug.

PR #676, opened 2015-06-03
Fix JcloudsLocation for vCloud Director usage

aledsage:


nakomis:
LGTM, assuming testing works

PR #677, opened 2015-06-03
Fix tests on Windows

neykov:


core/src/main/java/brooklyn/entity/rebind/PeriodicDeltaChangeListener.java
```diff
@@ -278,7 +278,9 @@ public void waitForPendingComplete(Duration timeout, boolean canTrigger) throws
                 while (writeCount.get() <= targetWriteCount) {
                     Duration left = timer.getDurationRemaining();
                     if (left.isPositive()) {
-                        writeCount.wait(left.lowerBound(Duration.millis(10)).toMilliseconds());
+                        synchronized(writeCount) {
+                            writeCount.wait(left.lowerBound(Duration.millis(10)).toMilliseconds());
```
aledsage: wow, that was a bad bug to sneak in! Thanks for fixing!

aledsage:
Looks good; merging.

PR #679, opened 2015-06-04
added type coercions for String to BigDecimal and BigInteger

MichaelSchneiderVirtustream:
Added two type coercions to class brooklyn.util.flags.TypeCoercions in brooklyn-core:
  - String -> java.math.BigDecimal
  - String -> java.math.BigInteger
Also added a corresponding test case to TypeCoercionsTest.


sjcorbett:
@MichaelSchneiderVirtustream Thanks. Looks good.

PR #681, opened 2015-06-05
Provision latch

sjcorbett:
First commit is boring - it just simplifies `JcloudsLocationTest`.

Second commit adds `BrooklynConfigKeys.PROVISION_LATCH`, to block provisioning of locations until some criteria is met.

### DIFF
--- a/core/src/main/java/brooklyn/entity/basic/BrooklynConfigKeys.java
+++ b/core/src/main/java/brooklyn/entity/basic/BrooklynConfigKeys.java
@@ -143,6 +143,7 @@ public class BrooklynConfigKeys {
      * component is up, but this entity does not care about the dependent component's actual config values.
      */
 
+    public static final ConfigKey<Boolean> PROVISION_LATCH = newBooleanConfigKey("provision.latch", "Latch for blocking location provision until ready");
     public static final ConfigKey<Boolean> START_LATCH = newBooleanConfigKey("start.latch", "Latch for blocking start until ready");
     public static final ConfigKey<Boolean> SETUP_LATCH = newBooleanConfigKey("setup.latch", "Latch for blocking setup until ready");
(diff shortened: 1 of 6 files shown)


sjcorbett:
cc @alasdairhodge

core/src/main/java/brooklyn/entity/basic/BrooklynConfigKeys.java
```diff
@@ -143,6 +143,7 @@
      * component is up, but this entity does not care about the dependent component's actual config values.
      */
 
+    public static final ConfigKey<Boolean> PROVISION_LATCH = newBooleanConfigKey("provision.latch", "Latch for blocking location provision until ready");
```
alasdairhodge: `SoftwareProcess` redefines the other lifecycle latches (referencing the ones here in `BrooklynConfigKeys`); should `PROVISION_LATCH` also be aliased there for easier discovery? I'm not convinced this is necessarily a good pattern, but inconsistency is worse.

alasdairhodge: Oh wait, `SoftwareProcess` itself isn't consistent with which keys it aliases. Ignore me.

sjcorbett: I thought about it but figured the provisioning latch doesn't relate to the software-process-ness of an entity.

alasdairhodge: Yeah, I agree. I expect `PROVISION_LATCH` will eventually join `PROVISIONING_PROPERTIES` on some sort of `Provisionable` trait class.

alasdairhodge:
Nice addition. LGTM.

PR #682, opened 2015-06-08
Date parsing and type coercion fixes, require Java 7+

ahgittin:


ahgittin:
this uses named groups in regexes which is only avail in java 7+.

i'm going to suggest we raise the minimum required version to 7 -- also implemented in this PR.  title updated to reflect this.

ahgittin:
Build failure is due to Java 6 being used on Jenkins machine.  If Java 7 minimum is agreed let's update Jenkins and kick off a new build.

utils/common/src/main/java/brooklyn/util/time/Time.java
```diff
@@ -443,36 +472,433 @@ else if (s.equalsIgnoreCase("d") || s.equalsIgnoreCase("day") || s.equalsIgnoreC
...
+    }
+
+    private static String namedGroup(String name, String pattern) {
+        return "(?<"+name+">"+pattern+")";
+    }
+    private static String anyChar(String charSet) {
+        return "["+charSet+"]";
+    }
+    private static String optionally(String pattern) {
+        return "("+pattern+")?";
```
alasdairhodge: Would a non-capturing group be better?

utils/common/src/main/java/brooklyn/util/time/Time.java
```diff
@@ -443,36 +472,433 @@ else if (s.equalsIgnoreCase("d") || s.equalsIgnoreCase("day") || s.equalsIgnoreC
...
+    }
+    private static String options(String ...patterns) {
+        return "("+Strings.join(patterns,"|")+")";
```
alasdairhodge: Would a non-capturing group be better?

sjcorbett:
The regular expressions in `Time` make me wary. Why write date parsing yourself when you could just use joda-time? It's already solved these things for us.

alasdairhodge:
Actually, I'm inclined to agree with @sjcorbett. Better to mandate a standard format (ISO8601 is the obvious choice) and pull in a mature implementation such as joda-time or Apache commons DateUtils.

Core Java has been moving in the right direction: Java 7's SimpleDateFormat supports ISO8601, and Java 8 includes the new java.time API (heavily influenced by joda-time).

Lots of discussions on StackOverflow:
http://stackoverflow.com/questions/2201925/converting-iso-8601-compliant-string-to-java-util-date

alasdairhodge:
I quite like the fluent methods for constructing regexes, although as you mention they work best under Java 7 with named group support; can we pull this into brooklyn.utils.common?

ahgittin:
Joda Time's parser is slightly better than Java's but it too is extremely rigid.  We'll need different formatter specifiers for `.` vs `-` vs ` ` and `MM` vs `MMM` vs `MMMM` which becomes combinatorially tedious to express and slow to execute.  (Apache DateUtils is basically a wrapper around Java's so has even worse drawbacks.)  A flexible off-the-rack parser was my first hunt.  I tried `natty` for this reason (which is extremely cool but has drawbacks as noted in the PR).

The other option is to mandate a single format but that makes things harder to work with as people then have to look up the syntax.  And although ISO 8601 is now a de facto standard for machines it makes for ugly blueprints.

---

Generally in Brooklyn we try to be lenient about parsing and encourage natural-looking blueprints.  Users seem to react well to that.  I think the same is warranted for date support, and I think this custom parser is the best way.  (I'd prefer it more if this could become a standard part of `DateUtils` or similar but people tend to be opinionated on their dates!)

ahgittin:
Re regex conveniences -- these methods aren't strictly fluent; a fluent API for regexes would be lovely but a much bigger task than the few conveniences I've put here.  But thanks for the compliment!

utils/common/src/main/java/brooklyn/util/time/Time.java
ahgittin: It would be permitted but not sure this optimization is worth the add'l regex complexity.

alasdairhodge:
Well I strongly disagree that a single (ubiquitous and standards-defined) format makes things "harder to work with" — just the opposite, actually! ISO's ubiquity easily makes up for it's wrinkles.

I my view, our primary focus should be on an unambiguous, interoperable representation with "friendly" aesthetics a secondary concern. If things like field delimiters are a big issue for users (really?!) then arguably we should simply defer to the locale to do the parsing according to the user's whims (although this will open a huge can of worms when blueprints are shared).

ahgittin:
@alasdairhodge that's the tosca mindset, it's great for systems but not for users.  a single format which many people don't know about (really a `T` in the middle??) is going to make it hard for people.

nearly all the feedback from users has been to the contrary:  blueprints are best when intuitive and forgiving, so long as there isn't a risk associated or leaky complexity.

it *is* good to recommend one way to do things, and we have been guilty of showing too many.  so i'm happy to recommend ISO 8601 but i think almost any human would rather see a space in the middle instead of the ugly `T`.  (of course the weirder your format the riskier it is, such as the weird examples in the test case.)

we definitely can't rely on locale delimiters as the ultimate arbiter as sharing across servers is important.

btw ISO 8601 supports bizarre things like `2015-156` for the 156th day of the year.  and it mandates a `P` prefix for durations `P3H`.  these would not be popular.

OTOH being able to express a stabilization delay as `2m 30s` is extremely popular.  similarly `4g` for RAM instead of `4096` (where you used to have to look up the fact that it is `mB`).

let's not make users' lives difficult!

alasdairhodge:
Oooh, see also: http://yaml.org/type/timestamp.html  Supports "relaxed" ISO8601 with spaces.

ahgittin:
This PR supports that -- except for lower-case `t` which I'll add.

The YAML standard is a good one to recommend -- I'll add that to docs and error messages.

ahgittin:
Updated to recommend the YAML standard, although in practice it supports more, including most of the obvious unambiguous formats.

Also fleshed out docs esp on `TypeCoercions.coerce` which we can start to point to as the canonical way to represent non-custom types from yaml.

(Still need to flip jenkins to java7.)

sjcorbett:
@ahgittin I think your comment that the "other option is to mandate a single format but that makes things harder to work with as people then have to look up the syntax" is erroneous - people will need to look up the syntax accepted by Brooklyn anyway. Why is "10.11-2020-12.01" read as "2020-12-01-1011" and not "2020-11-10 12:01"? Brooklyn shouldn't make the decision, the user should supply the format string for their bonkers date.

We gain some flexibility and we gain a lot of ambiguity. To understand what Brooklyn will do with my date I need to understand about 400 lines of Java. I would prefer to just have to scan a bunch of DateFormatters, even if they are more expensive to express.

ahgittin:
@sjcorbett what?  a user here can use any sensible format and get the expected result without having to look anything up.  do you disagree?  or do you think that's a bad thing?

the point you make is a different one, that to understand really perverse things you'd need to look at the implementation, and i think that's okay.  (although actually here they could just look at the test cases or try it out or infer from the javadoc which recommends date first that date-first is probably preferred.)

ahgittin:
@sjcorbett I think the more general argument you are making is for a feature to have a "strict" mode where a much more restrictive set of type coercions is applied (failing or logging otherwise).  We'd require ISO 8601 here.

I can see this being useful when we get to the point of interoperability and validation.  But for where we are now I think the priority is to remove any barriers towards understanding and writing blueprints.

ahgittin:
yay - bumping our builds.apache.org to java 1.7 and the PR works.

we'll need to do the same for other builds before this is merged.

ahgittin:
@sjcorbett @alasdairhodge do you feel strongly this should not be merged or would that be okay?

utils/common/src/main/java/brooklyn/util/time/Time.java
```diff
@@ -443,36 +500,466 @@ else if (s.equalsIgnoreCase("d") || s.equalsIgnoreCase("day") || s.equalsIgnoreC
...
+        namedGroup("month", anyChar(LETTER)+"+")
+        + anyChar(DATE_SEPARATOR)
+        + namedGroup("day", optionally(DIGIT)+DIGIT)
+        + ",?"+anyChar(DATE_SEPARATOR)+"+"
+        + namedGroup("year", DIGIT+DIGIT+DIGIT+DIGIT);
+    // we could parse NN-NN-NNNN as DD-MM-YYYY always, but could be confusing for MM-DD-YYYY oriented people, so require month named
+    private final static String DATE_WORDS_3 = 
+        namedGroup("day", optionally(DIGIT)+DIGIT)
+        + anyChar(DATE_SEPARATOR)
+        + namedGroup("month", anyChar(LETTER)+"+")
```
sjcorbett: This formatting (starting the line with a +) is a bit weird to me. It's a bit inconsistent between patterns at the moment, too (e.g. `TIME_ONLY_WITH_INNER_SEPARATORS` is at the end of the line).

utils/common/src/main/java/brooklyn/util/time/Time.java
```diff
@@ -443,36 +500,466 @@ else if (s.equalsIgnoreCase("d") || s.equalsIgnoreCase("day") || s.equalsIgnoreC
...
+    }
+    
+    private static Maybe<Matcher> match(String pattern, String input) {
+        Matcher m = Pattern.compile(pattern).matcher(input);
+        if (m.find() && m.start()==0 && m.end()==input.length())
+            return Maybe.of(m);
+        return Maybe.absent();
+    }
+
+//    // TODO https://github.com/joestelmach/natty is cool, but it drags in ANTLR,
```
sjcorbett: What is the TODO? To decide whether ANTLR is useful or not? Would prefer to either include Natty get rid of this commented block.

utils/common/src/main/java/brooklyn/util/time/Time.java
```diff
@@ -443,36 +500,466 @@ else if (s.equalsIgnoreCase("d") || s.equalsIgnoreCase("day") || s.equalsIgnoreC
...
+    private static String options(String ...patterns) {
+        return "("+Strings.join(patterns,"|")+")";
+    }
+    private static String notMatching(String pattern) {
+        return "(?!"+pattern+")";
+    }
+    
+    private static Maybe<Matcher> match(String pattern, String input) {
+        Matcher m = Pattern.compile(pattern).matcher(input);
+        if (m.find() && m.start()==0 && m.end()==input.length())
```
sjcorbett: Could you prepend `^` and append `$` to pattern to achieve the same whole-string match?

sjcorbett:
@ahgittin Thanks for asking. Made a few minor points. I'm alright seeing this merged.

utils/common/src/main/java/brooklyn/util/time/Time.java
ahgittin: agree

utils/common/src/main/java/brooklyn/util/time/Time.java
ahgittin: +1

utils/common/src/main/java/brooklyn/util/time/Time.java
ahgittin: yeah - that occurred to me later.  i'll do it.

PR #683, opened 2015-06-08
BROOKLYN-149: Partial fix for rebind when catalog missing

aledsage:


core/src/main/java/brooklyn/entity/rebind/persister/BrooklynMementoPersisterToObjectStore.java
```diff
@@ -174,11 +176,23 @@ public BrooklynMementoPersisterToObjectStore(PersistenceObjectStore objectStore,
         if (item==null || item.getCatalogItemId()==null) {
             return null;
         }
-        CatalogItem<?, ?> catalogItem = CatalogUtils.getCatalogItemOptionalVersion(lookupContext.lookupManagementContext(), item.getCatalogItemId());
+        // See RebindIteration.BrooklynObjectInstantiator.load(), for handling where catalog item is missing;
+        // similar logic here.
```
neykov: Changes here won't be needed if `catalogItemId` is patched above.

neykov:
Finished review. Patching the `catalogItemId` during load instead of inferring it each time it's needed will make for a more robust failover.

All of the instances where a catalog item is missing during rebind I have seen so far have been due to user error - someone deleting the item without realising it's still in use and needed by the entities on next rebind. A more general fix would be to forbid this case altogether. It will save a lot of head scratching.

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
```diff
@@ -902,45 +902,53 @@ protected void setCatalogItemId(BrooklynObject item, String catalogItemId) {
...
-                    //Try loading from whichever catalog bundle succeeds.
-                    BrooklynCatalog catalog = managementContext.getCatalog();
-                    for (CatalogItem<?, ?> item : catalog.getCatalogItems()) {
-                        BrooklynClassLoadingContext catalogLoader = CatalogUtils.newClassLoadingContext(managementContext, item);
-                        Maybe<Class<?>> catalogClass = catalogLoader.tryLoadClass(jType);
-                        if (catalogClass.isPresent()) {
-                            return (Class<? extends T>) catalogClass.get();
-                        }
+                CatalogItem<?, ?> catalogItem = rebindContext.lookup().lookupCatalogItem(catalogItemId);
+                if (catalogItem == null && BrooklynFeatureEnablement.isEnabled(BrooklynFeatureEnablement.FEATURE_INFER_CATALOG_ITEM_ON_REBIND)) {
```
mikezaccardo: Perhaps change this `if` condition to simply the `null` check and then have another `if` that switches on whether `FEATURE_INFER_CATALOG_ITEM_ON_REBIND` is true.  It's easier to follow the logic this way.

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
```diff
@@ -902,45 +902,53 @@ protected void setCatalogItemId(BrooklynObject item, String catalogItemId) {
...
+                    if (CatalogUtils.looksLikeVersionedId(catalogItemId)) {
+                        String symbolicName = CatalogUtils.getIdFromVersionedId(catalogItemId);
+                        catalogItem = rebindContext.lookup().lookupCatalogItem(symbolicName);
                     }
-                    throw new IllegalStateException("No catalogItemId specified and can't load class from either classpath of catalog items");
+                }
+                if (catalogItem != null) {
+                    BrooklynClassLoadingContext loader = CatalogUtils.newClassLoadingContext(managementContext, catalogItem);
+                    return loader.loadClass(jType, bType);
                 } else {
```
mikezaccardo: This would get moved up (see above).

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
```diff
@@ -902,45 +902,53 @@ protected void setCatalogItemId(BrooklynObject item, String catalogItemId) {
...
+                LOG.warn("Unable to load "+jType+" using reflections; will try standard context");
+            }
 
-        protected BrooklynClassLoadingContext getLoadingContextFromCatalogItemId(String catalogItemId, ClassLoader classLoader, RebindContext rebindContext) {
-            Preconditions.checkNotNull(catalogItemId, "catalogItemId required (should not be null)");
-            CatalogItem<?, ?> catalogItem = rebindContext.lookup().lookupCatalogItem(catalogItemId);
-            if (catalogItem != null) {
-                return CatalogUtils.newClassLoadingContext(managementContext, catalogItem);
+            if (catalogItemId != null) {
+                throw new IllegalStateException("Unable to load catalog item "+catalogItemId+" for "+contextSuchAsId+", or load class from classpath");
```
mikezaccardo: Is this definitely right, an exception if the item is *not* `null`?

core/src/main/java/brooklyn/entity/rebind/persister/BrooklynMementoPersisterToObjectStore.java
```diff
@@ -174,11 +176,23 @@ public BrooklynMementoPersisterToObjectStore(PersistenceObjectStore objectStore,
...
+        String catalogItemId = item.getCatalogItemId();
+        CatalogItem<?, ?> catalogItem = CatalogUtils.getCatalogItemOptionalVersion(lookupContext.lookupManagementContext(), catalogItemId);
+        if (catalogItem == null && BrooklynFeatureEnablement.isEnabled(BrooklynFeatureEnablement.FEATURE_INFER_CATALOG_ITEM_ON_REBIND)) {
```
mikezaccardo: Same `if` structure comment.

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
aledsage: Yes, full context of method is that we've tried all available ways of loading (there are earlier return statements in the method), so will now fail. However, if no catalogItemId was specified and if the backards_compatibility feature was enabled then we'll try to infer a catalog item id.

Not particularly nice logic, having multiple return statements, but is possibly the simplest given the unfortunate number of if-else statements to deal with the different scenarios.

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
aledsage: Can't move this up; we need to do the check after going into the conditional case of trying to infer the different catalogItem.

Really don't like the code, but hard (and probably not worth a lot of refactoring) to avoid all the if-else conditions.

aledsage:
@neykov I've incorporated the comments; can you review again please.

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
```diff
@@ -899,48 +914,64 @@ protected void setCatalogItemId(BrooklynObject item, String catalogItemId) {
...
-                    throw new IllegalStateException("No catalogItemId specified and can't load class from classpath");
+                    LOG.warn("Unable to load catalog item "+catalogItemId+" for "+contextSuchAsId
+                            +" ("+bType.getSimpleName()+"); will try default class loader");
                 }
-
             }
-        }
+            
+            try {
+                return new LoadedClass<T>((Class<T>)reflections.loadClass(jType), catalogItemId);
```
neykov: Shouldn't this return `null` for `catalogItemId`? Why treat it differently from the upgraded-catalog-item case where the id is patched?

Could put the block behind `FEATURE_INFER_CATALOG_ITEM_ON_REBIND` check as this is part of the same logic.

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
```diff
@@ -899,48 +914,64 @@ protected void setCatalogItemId(BrooklynObject item, String catalogItemId) {
...
+            }
 
-        protected BrooklynClassLoadingContext getLoadingContextFromCatalogItemId(String catalogItemId, ClassLoader classLoader, RebindContext rebindContext) {
-            Preconditions.checkNotNull(catalogItemId, "catalogItemId required (should not be null)");
-            CatalogItem<?, ?> catalogItem = rebindContext.lookup().lookupCatalogItem(catalogItemId);
-            if (catalogItem != null) {
-                return CatalogUtils.newClassLoadingContext(managementContext, catalogItem);
+            if (catalogItemId != null) {
+                throw new IllegalStateException("Unable to load catalog item "+catalogItemId+" for "+contextSuchAsId+", or load class from classpath");
+            } else if (BrooklynFeatureEnablement.isEnabled(BrooklynFeatureEnablement.FEATURE_BACKWARDS_COMPATIBILITY_INFER_CATALOG_ITEM_ON_REBIND)) {
```
neykov: Looking at the code in details now it's not obvious why this (old) code repeats the logic in `findCatalogItemId`. Actually the whole catalog item inference could be encapsulated there. Just thinking out loud, no need to change it in this PR.

core/src/main/java/brooklyn/entity/rebind/RebindIteration.java
aledsage: Reason I didn't return null was that it would override the `catalogItemId` to be persisted, so it would then say null. I didn't want to lose the catalogItemId in the situation where it wasn't available. Instead, we'd leave that itemId in and fall back to default class loader.

I wondered about guarding it behind the feature enablement. However, neither of the (renamed) feature options seem quite right:
* `FEATURE_BACKWARDS_COMPATIBILITY_INFER_CATALOG_ITEM_ON_REBIND` is about backwards compatibility, where the `catalogItemId` was missing from old things.
* `FEATURE_AUTO_FIX_CATALOG_REF_ON_REBIND` is for updating the ref (I think that deleting the ref seems extreme there).

We should revisit this in follow-up pull requests. I'd like us to start up such that errors are shown, and can be fixed, rather than aborting entirely or just logging and ignoring.

neykov: This is possible even now if you use `exceptionHandler` to log the errors. Depending on the startup options the process will abort or show a warning in the web console.

neykov:
Good to merge. Solves a frequent pain point around rebinds.

aledsage:
Thanks @neykov - merging.

PR #684, opened 2015-06-10
Add link to GitHub blueprint library

mikezaccardo:
Adds link to https://github.com/brooklyncentral/blueprint-library from http://brooklyn.incubator.apache.org/v/latest/yaml/index.html

### DIFF
--- a/docs/guide/yaml/index.md
+++ b/docs/guide/yaml/index.md
@@ -11,6 +11,7 @@ children:
 - custom-entities.md
 - chef/
 - { path: yaml-reference.md, title: YAML Blueprint Reference }
+- { link: 'https://github.com/brooklyncentral/blueprint-library', title: 'GitHub Blueprint Library' }
 ---
 
 


andreaturli:
I think it makes sense but not sure we consider that `https://github.com/brooklyncentral/blueprint-library` official enough. @grkvlt @ahgittin thoughts?

ahgittin:
Linking to the `brooklyncentral` repo possibly isn't a bad short-term fix -- and adding some text on that site to say it is a collection of community blueprints.  But in the next couple months we need a better way.

docs/guide/yaml/index.md
```diff
@@ -11,6 +11,7 @@ children:
 - custom-entities.md
 - chef/
 - { path: yaml-reference.md, title: YAML Blueprint Reference }
+- { link: 'http://github.com/apache/incubator-brooklyn', title: 'GitHub Blueprint Library' }
```
ahgittin: This isn't the right link.

mikezaccardo:
Fixed; let's never speak of this

PR #686, opened 2015-06-11
Reduces excessively chatty logging on persistence checkpoints

nakomis:


ahgittin:
This is extremely useful information if there is latency in checkpointing which has been a problem so I would be loathe to remove it from the default set.

In terms of impact it is a couple of lines every second which is hugely reduced from where it was.  But if that is still overly chatty the fix I would recommend would be to log debug for the first few writes after any HA change and every 100th checkpoint and if latency is ever more than some threshhold (1s).

We currently do the first of these already for hot standby mode so you should be able to tap into that counter.

ahgittin:
The test failure is almost certainly unrelated but something we should fix -- for posterity it is:

```
	at java.net.PlainSocketImpl.socketBind(Native Method)
	at java.net.AbstractPlainSocketImpl.bind(AbstractPlainSocketImpl.java:376)
	at java.net.ServerSocket.bind(ServerSocket.java:376)
	at java.net.ServerSocket.<init>(ServerSocket.java:237)
	at java.net.ServerSocket.<init>(ServerSocket.java:128)
	at brooklyn.policy.ha.ConnectionFailureDetectorTest.startServerSocket(ConnectionFailureDetectorTest.java:103)
	at brooklyn.policy.ha.ConnectionFailureDetectorTest.setUp(ConnectionFailureDetectorTest.java:92)
```

/cc @aledsage @neykov

nakomis:
Hi @ahgittin, where can I find the implementation of the 'first few writes' in hot standby mode? Also, how can I tell if HA has changed?

ahgittin:
See `RebindIteration.shouldLogRebinding` -- the count here comes from `RebindManagerImpl` but it is only maintained for RO rebind so you probably want an independent count within the `Periodic` class.

It is reset within the `{start,stop}{Persistence,ReadOnly}` methods so you could add a call in each of these to e.g. `periodic.resetCount()`.

sjcorbett:
@nakomis has a point about excessively noisy logs - ran a simple test with brooklyn-ambari and there are so many "Checkpointing delta of memento" messages that it took one minute for my debug log to roll over. It was more like once per millisecond than once per second.

aledsage:
I agree we need to massively reduce the logging. It fills up one of our customer's rolling log files way too much.

@nakomis can you please take another look at @ahgittin 's suggestion of how to reduce logging (and let us know if you agree with that)?

nakomis:
PR comments addressed

core/src/main/java/brooklyn/entity/rebind/PeriodicDeltaChangeListener.java
```diff
@@ -494,5 +496,8 @@ public synchronized void onChanged(BrooklynObject instance) {
     public PersistenceExceptionHandler getExceptionHandler() {
         return exceptionHandler;
     }
-    
+
+    protected boolean shouldLogCheckpoint() {
+        return (checkpointLogCount.get() < 5) || (checkpointLogCount.get() % 1000 == 0);
```
sjcorbett: Did you intend to use incrementAndGet? Nothing increments the counters at the moment so they'll always be 0.

Also suggest getting the value of the integer once so we don't pay synchronisation penalty twice.

sjcorbett: Why the magic number 5?

core/src/main/java/brooklyn/entity/rebind/PeriodicDeltaChangeListener.java
```diff
@@ -78,6 +78,8 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(PeriodicDeltaChangeListener.class);
 
+    protected final AtomicInteger checkpointLogCount = new AtomicInteger();
```
sjcorbett: I think you should use an `AtomicLong` instead. If `shouldLogCheckpoint` were called once per millisecond the integer would overflow in about three and a half weeks: `(2^31 - 1) / (1000 * 60 * 60 * 24 * 7) = 3.55`. So `shouldLogCheckpoint` will always return true until `checkpointLogCount.get() < 5` is false. With a Long it would take  292,277 millenia to overflow.

core/src/main/java/brooklyn/entity/rebind/PeriodicDeltaChangeListener.java
```diff
@@ -383,7 +385,7 @@ protected void persistNowInternal(boolean alreadyHasMutex) {
                 deltaCollector = new DeltaCollector();
             }
             
-            if (LOG.isDebugEnabled()) LOG.debug("Checkpointing delta of memento: "
+            if (shouldLogCheckpoint() && LOG.isDebugEnabled()) LOG.debug("Checkpointing delta of memento: "
```
sjcorbett: Swap the order of the if conditions for cheaper short-circuiting when `isDebugEnabled` is false.

nakomis:
Comments addressed

nakomis:
Failed test passes locally, and appears to be unrelated to this PR

sjcorbett:
@nakomis Thanks. Will merge.

PR #687, opened 2015-06-11
Tweak PasswordHasher to avoid potentially misleading use of ByteBuffer.array()

alasdairhodge:
Using `ByteBuffer.array()` to convert a salted password to its constituent bytes frequently results in an indeterminate number of spurious bytes to be included, due to the fact that the buffer's raw backing array is usually larger than the space consumed just by the values written to it. This causes Brooklyn-generated hashes to deviate from the expected `sha256(salt + password)` values.

Even though Brooklyn itself can generate internally-consistent hashes, it is desirable for the algorithm to be sufficiently predictable to permit generation by simple external mechanisms, for example:
````
echo -n "saltpassword" | openssl dgst -sha256
echo -n "saltpassword" | sha256sum
echo -n "saltpassword" | shasum -a 256
````

### DIFF
--- a/usage/rest-server/src/main/java/brooklyn/rest/security/PasswordHasher.java
+++ b/usage/rest-server/src/main/java/brooklyn/rest/security/PasswordHasher.java
@@ -24,8 +24,9 @@ import com.google.common.hash.Hashing;
 
 public class PasswordHasher {
     public static String sha256(String salt, String password) {
-        byte[] salted = Charsets.UTF_8.encode((salt == null ? "" : salt) + password).array();
-        HashCode hash = Hashing.sha256().hashBytes(salted);
+        if (salt == null) salt = "";
+        byte[] bytes = (salt + password).getBytes(Charsets.UTF_8);
+        HashCode hash = Hashing.sha256().hashBytes(bytes);
         return hash.toString();
     }
 }
--- a/usage/rest-server/src/test/java/brooklyn/rest/security/PasswordHasherTest.java
+++ b/usage/rest-server/src/test/java/brooklyn/rest/security/PasswordHasherTest.java
@@ -26,6 +26,6 @@ public class PasswordHasherTest {
 
     @Test
     public void testHashSha256() throws Exception {
-        assertEquals(PasswordHasher.sha256("mysalt", "mypassword"), "e1c2390613b1beff83420c15d6ceca3b2e77e66e3f4be4e45186032120a30f22");
+        assertEquals(PasswordHasher.sha256("mysalt", "mypassword"), "d02878b06efa88579cd84d9e50b211c0a7caa92cf243bad1622c66081f7e2692");
     }
 }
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by bernd-clemenz** [comment]
Great, very helpfull, if we could build valid SHA hashes with out calling Brooklyn on the command line. This change should be accepted as soon as possible!

**Comment by alasdairhodge** [comment]
Whoops, forgot to push the updated test. Try again, @asfbot!

File: usage/rest-server/src/main/java/brooklyn/rest/security/PasswordHasher.java [hunk 1]
```diff
@@ -24,8 +24,9 @@
 
 public class PasswordHasher {
     public static String sha256(String salt, String password) {
-        byte[] salted = Charsets.UTF_8.encode((salt == null ? "" : salt) + password).array();
-        HashCode hash = Hashing.sha256().hashBytes(salted);
+        if (salt == null) salt = "";
+        byte[] bytes = (salt + password).getBytes(Charsets.UTF_8);
```
sjcorbett: Appreciate this is existing behaviour but calling `sha256(null, null)` will mean we hash "null" (so always `74234e98afe7498fb5daf1f36ac2d78acc339464f950703b8c019892f982b90b`). I think this is a bit surprising and potentially revealing. Should we throw if `password` is null?

alasdairhodge: Throwing would seem appropriate, or defaulting to the empty string, as is done for the salt. Perverse usage, but should at least be consistent.

bernd-clemenz:
I like the idea of not accepting an null or empty password. Soemthing like an IllegalArgumentException should be thrown in that case.

usage/rest-server/src/main/java/brooklyn/rest/security/PasswordHasher.java
sjcorbett: On looking at the context I think the existing behaviour is ok - it will only be acting on user input in a request. If it were being used elsewhere I think the existing comment would stand.

alasdairhodge:
Have added trivial tests for empty salt values, but left the null password issue as-is, in line with latest comment from @sjcorbett. Good to merge?

sjcorbett:
@alasdairhodge Thanks. Will merge. Nice find, by the way.

grkvlt:
We should change `ExplicitUsersSecurityProvider` and `GeneratePasswordCommand` to handle missing salts by using the username, rather than relying on `Passwordhasher` to handle null by replacong with an empty string. This will increase security, since using an empty string would result in an unsalted and therefore rainbow-table-lookup possibile hash.

grkvlt:
@alasdairhodge @aledsage @ahgittin as this will have the effect of preventing users from logging in when they upgrade, we need to be very clear about what to do in the release notes if hashed passwords are configured. Perhaps we should write an upgrade tool?

ahgittin:
Is it possible to also compare against the legacy poorly-salted encrypted value, and for 0.7.0 accept but log warning if that is used?

ahgittin:
i'm okay with the straightforward (breaking) change, as long as we update the release notes.  afaik the people using this feature can live with the manual overhead.

PR #688, opened 2015-06-15
Put user metadata into SL notes.

neykov:
Will be NOP until the jclouds version is bumped to include https://github.com/jclouds/jclouds/pull/756, probably `1.9.1`.

### DIFF
--- a/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
@@ -207,6 +207,8 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
     private static final Pattern LIST_PATTERN = Pattern.compile("^\\[(.*)\\]$");
     private static final Pattern INTEGER_PATTERN = Pattern.compile("^\\d*$");
 
+    private static final int NOTES_MAX_LENGTH = 1000;
+
     private final AtomicBoolean loggedSshKeysHint = new AtomicBoolean(false);
     private final AtomicBoolean listedAvailableTemplatesOnNoSuchTemplate = new AtomicBoolean(false);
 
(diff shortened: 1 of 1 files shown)


locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -207,6 +207,8 @@
     private static final Pattern LIST_PATTERN = Pattern.compile("^\\[(.*)\\]$");
     private static final Pattern INTEGER_PATTERN = Pattern.compile("^\\d*$");
 
+    private static final int NOTES_MAX_LENGTH = 1000;
```
sjcorbett: Will this limit ever change? Is it worth allowing making it configurable?

neykov: It's highly unlikely that it will ever decrease as it's part of the official API description. Since it's only used for the user-metadata generated notes I don't think it's worth having it configurable (in the case the limit increases).

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1222,8 +1224,16 @@ protected void customizeTemplate(ConfigBag setup, ComputeService computeService,
                     tags.add(AbstractCloudMachineNamer.sanitize(entry.getKey())+":"+AbstractCloudMachineNamer.sanitize(entry.getValue()));
                 }
                 slT.tags(tags);
+
+                if (!md.containsKey("notes")) {
+                    String notes = "User Metadata\n=============\n\n  * " + Joiner.on("\n  * ").withKeyValueSeparator(": ").join(md);
+                    if (notes.length() > NOTES_MAX_LENGTH) {
+                        String truncatedMsg = "...\n<truncated>";
```
mikezaccardo: Perhaps it would be useful to indicate how long the entire message was / how much was truncated? Never hurts to give more info with only a few more characters.

neykov: Agree, but then the truncate point depends on the length of the message which depends on the truncate point :). Not worth the effort of getting it right. I can easily add the total length so don't mind doing it.

mikezaccardo:
(LGTM)

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
neykov: Addressed.

aledsage:
Looks good; merging.

PR #689, opened 2015-06-15
move all camp code to EntityManagementUtils

andreaturli:
first baby-step to remove CAMP dependencies from `core`

@neykov, @ahgittin could you guys please have a look?

Next steps:
- extract a common interface called `SpecCreator` to replace the CAMP specific-calls inside `EntityManagementUtils`. This will use SPI to plug different implementation of the `SpecCreator`
- move `camp.brooklyn.api` to `/usage/camp`

### DIFF
--- a/core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
+++ b/core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
@@ -18,19 +18,19 @@
  */
 package brooklyn.catalog.internal;
 
+import static brooklyn.management.internal.EntityManagementUtils.LOCATIONS_KEY;
+import static brooklyn.management.internal.EntityManagementUtils.POLICIES_KEY;
+import static brooklyn.management.internal.EntityManagementUtils.createEntitySpec;
+import static brooklyn.management.internal.EntityManagementUtils.createLocationSpec;
+import static brooklyn.management.internal.EntityManagementUtils.createPolicySpec;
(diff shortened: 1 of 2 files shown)


neykov:
The changes look straightforward. As the methods are static anyway, I'd scope them inside another class so it's clear the additions to `EntityManagementUtils` are strictly catalog related - but not important.

As a next step, before moving to extracting the interface, you should make sure that all calls to `PdpProcessor` and `AssemblyTemplateInstantiator` methods are moved to `EntityManagementUtils` (or similar).

andreaturli:
good points, @neykov I'll take care of those. thanks!

mikezaccardo:
(LGTM)

neykov:
I'd say we leave this for after we cut the release (if it doesn't take too long), what do you think @andreaturli? It is still a work in progress?

andreaturli:
thanks @mikezaccardo and @neykov 

I think it makes sense to post-pone this PR until we cut the release as it is still in-progress

aledsage:
@andreaturli does this replace PR #545? If so, should that be closed? Any comments there that are worth summarising here (rather than any viewer having to follow that link as well)?

aledsage:
@andreaturli do you think this is ready to merge into master (which is now 0.8.0-SNAPSHOT) or is there anything you want to change before it's merged?

Also, does this replace #545, so should that be closed?

andreaturli:
@aledsage I think this PR needs to be closed as #760 supersedes it.

PR #690, opened 2015-06-15
Determine type of entity in rest response from api annotation

robertgmoss:
Previously the Response was read into a String, which makes it difficult
to convert back into the domain object.

### DIFF
--- a/usage/rest-client/pom.xml
+++ b/usage/rest-client/pom.xml
@@ -114,14 +114,14 @@
             <groupId>org.apache.brooklyn</groupId>
             <artifactId>brooklyn-core</artifactId>
             <version>${project.version}</version>
+            <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.brooklyn</groupId>
             <artifactId>brooklyn-core</artifactId>
             <version>${project.version}</version>
-            <classifier>tests</classifier>
-            <scope>test</scope>
+            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.brooklyn</groupId>
--- a/usage/rest-client/pom.xml
+++ b/usage/rest-client/pom.xml
@@ -119,12 +119,6 @@
         </dependency>
         <dependency>
             <groupId>org.apache.brooklyn</groupId>
-            <artifactId>brooklyn-core</artifactId>
-            <version>${project.version}</version>
-            <scope>compile</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.brooklyn</groupId>
             <artifactId>brooklyn-rest-server</artifactId>
             <version>${project.version}</version>
             <scope>test</scope>
--- a/usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
+++ b/usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
@@ -34,7 +34,6 @@ import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.Credentials;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.impl.client.DefaultHttpClient;
-import org.eclipse.jetty.util.log.Log;
 import org.jboss.resteasy.client.ClientExecutor;
 import org.jboss.resteasy.client.ClientRequest;
 import org.jboss.resteasy.client.ClientResponse;
@@ -62,7 +61,6 @@ import brooklyn.rest.api.UsageApi;
 import brooklyn.rest.api.VersionApi;
 import brooklyn.util.exceptions.Exceptions;
 import brooklyn.util.http.BuiltResponsePreservingError;
-import brooklyn.util.http.HttpTool;
 
 import com.wordnik.swagger.core.ApiOperation;
 
@@ -130,7 +128,7 @@ public class BrooklynApi {
                     Class<?> type = String.class;
                     if (result1 instanceof Response) {
                         Response resp = (Response)result1;
-                        if(HttpTool.isStatusCodeHealthy(resp.getStatus()) && method.isAnnotationPresent(ApiOperation.class)) {
+                        if(isStatusCodeHealthy(resp.getStatus()) && method.isAnnotationPresent(ApiOperation.class)) {
                            type = getClassFromMethodAnnotationOrDefault(method, type);
                         }
                         // wrap the original response so it self-closes
@@ -146,6 +144,8 @@ public class BrooklynApi {
                 }  
             }
             
+            private boolean isStatusCodeHealthy(int code) { return (code>=200 && code<=299); }
+            
             private Class<?> getClassFromMethodAnnotationOrDefault(Method method, Class<?> def){
                 Class<?> type;
                 try{
@@ -153,7 +153,7 @@ public class BrooklynApi {
                     type = Class.forName(responseClass);
                 } catch (Exception e) {
                     type = def;
-                    LOG.info("Unable to get class from annotation: {}.  Defaulting to {}", e.getMessage(), def.getName());
+                    LOG.debug("Unable to get class from annotation: {}.  Defaulting to {}", e.getMessage(), def.getName());
                     Exceptions.propagateIfFatal(e);
                 }
                 return type;
--- a/usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
+++ b/usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
@@ -34,6 +34,7 @@ import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.Credentials;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.impl.client.DefaultHttpClient;
+import org.eclipse.jetty.util.log.Log;
 import org.jboss.resteasy.client.ClientExecutor;
 import org.jboss.resteasy.client.ClientRequest;
 import org.jboss.resteasy.client.ClientResponse;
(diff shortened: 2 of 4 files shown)


usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
```diff
@@ -116,22 +119,26 @@ public BrooklynApi(URL endpoint, ClientExecutor clientExecutor) {
...
             @Override
-            public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
-                Object result1;
+            public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {                 
                 try {
-                    result1 = method.invoke(result0, args);
+                    Object result1 = method.invoke(result0, args);
+                    Class<?> type = String.class;
+                    if (result1 instanceof Response) {
+                        if(((Response)result1).getStatus()/100 == 2) {
```
neykov: Better use `HttpTool.isStatusCodeHealthy`

usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
```diff
@@ -116,22 +119,26 @@ public BrooklynApi(URL endpoint, ClientExecutor clientExecutor) {
...
+                           String responseClass = method.getAnnotation(ApiOperation.class).responseClass();
```
neykov: Looks like `responseClass` is used purely for documentation, can't rely on it for the exact return type. Better get the return type from the method handle.
For example:
  * `CatalogApi` doesn't include the fully qualified names
  * Some methods return collections, indicated by `multiValueResponse` in the annotation
  * Some have atomic values, not including a `responseClass`
  * Some methods have complex return types, but don't have `responseClass`
  * etc.

robertgmoss: @neykov I've changed it to best effort now.  We can't use the return type because in this case it is wrapped in a `Response` object.  The other cases you list are also not a problem because we are only looking at those that return a `Response`.

robertgmoss:
This failure appears not to be to do with this PR (failure to access ports)

usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
neykov: Good point. Then there's only one case where I am not clear what happens - in `brooklyn.rest.api.CatalogApi.create(String)` the annotation is `responseClass = "Response"`, it will probably fall back?

usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
```diff
@@ -116,22 +120,37 @@ public BrooklynApi(URL endpoint, ClientExecutor clientExecutor) {
...
-            public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
-                Object result1;
+            public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {                 
                 try {
-                    result1 = method.invoke(result0, args);
+                    Object result1 = method.invoke(result0, args);
+                    Class<?> type = String.class;
+                    if (result1 instanceof Response) {
+                        Response resp = (Response)result1;
+                        if(HttpTool.isStatusCodeHealthy(resp.getStatus()) && method.isAnnotationPresent(ApiOperation.class)) {
```
mikezaccardo: Is a healthy response the only case in which we care about the specific type?

usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
robertgmoss: @neykov That will fall back to String because it isn't a FQN

mikezaccardo:
LGTM

usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
robertgmoss: @mikezaccardo yes, because an unhealthy one cannot be converted to a domain object so should just be serialized as a String

mikezaccardo: Roger that, thanks

neykov:
LGTM as well.

robertgmoss:
Can this be merged?

usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
```diff
@@ -116,22 +120,37 @@ public BrooklynApi(URL endpoint, ClientExecutor clientExecutor) {
...
+                }  
+            }
+            
+            private Class<?> getClassFromMethodAnnotationOrDefault(Response resp, Method method, Class<?> def){
+                Class<?> type;
+                try{
+                    String responseClass = method.getAnnotation(ApiOperation.class).responseClass();
+                    type = Class.forName(responseClass);
+                } catch (Exception e) {
+                    type = def;
```
aledsage: Worth a log statement, rather than swallowing the exception, to explain (in the debug log) why it defaulted to a String?

I also like to do `Exceptions.propagateIfFatal()`. This mainly is for handing InterruptedExceptions and RuntimeInterruptedExceptions right, which should never happen in the calls within the try block (so not important here).

usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
```diff
@@ -197,6 +216,11 @@ public AccessApi getAccessApi() {
     }
 
     public static <T> T getEntity(Response response, Class<T> type) {
```
aledsage: It looks like this method isn't called (or is it reflective)?

If it's not useful, but are concerned about API compatibility, then could mark as deprecated and delete in 0.8.0-SNAPSHOT?

aledsage:
Looks good. Only a couple of very minor points.

usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
robertgmoss: This is called later by consumers of the api, which still returns a Response Object (which has had its contents read from one Response into another reflectively).

usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
robertgmoss: Agree, will do

robertgmoss:
Thanks @aledsage, comments addressed

usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
```diff
@@ -34,12 +34,16 @@
 import org.apache.http.auth.Credentials;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.impl.client.DefaultHttpClient;
+import org.eclipse.jetty.util.log.Log;
```
ahgittin: wrong `Log` no?

usage/rest-client/src/main/java/brooklyn/rest/client/BrooklynApi.java
```diff
@@ -116,22 +124,39 @@ public BrooklynApi(URL endpoint, ClientExecutor clientExecutor) {
...
+            }
+            
+            private Class<?> getClassFromMethodAnnotationOrDefault(Method method, Class<?> def){
+                Class<?> type;
+                try{
+                    String responseClass = method.getAnnotation(ApiOperation.class).responseClass();
+                    type = Class.forName(responseClass);
+                } catch (Exception e) {
+                    type = def;
+                    LOG.info("Unable to get class from annotation: {}.  Defaulting to {}", e.getMessage(), def.getName());
```
ahgittin: debug might be better?

usage/rest-client/pom.xml
```diff
@@ -114,14 +114,14 @@
...
+            <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.brooklyn</groupId>
             <artifactId>brooklyn-core</artifactId>
             <version>${project.version}</version>
-            <classifier>tests</classifier>
-            <scope>test</scope>
+            <scope>compile</scope>
```
ahgittin: why do we need to add `<scope>compile</scope>`?

previously both dependencies, regular and `<classifier>tests</classifier>` were only brought in for `test` scope I think which lowers the footprint for someone using the REST client.  also the ordering of the regular classifier first is cleaner i think.

robertgmoss: was brought in to use `HttpTool.isStatusCodeHealthy()`.  It wouldn't
compile without the `compile` scope.

On 19 June 2015 at 13:15, Alex Heneveld <notifications@github.com> wrote:

> In usage/rest-client/pom.xml
> <https://github.com/apache/incubator-brooklyn/pull/690#discussion_r32823340>
> :
>
> >              <scope>test</scope>
> >          </dependency>
> >          <dependency>
> >              <groupId>org.apache.brooklyn</groupId>
> >              <artifactId>brooklyn-core</artifactId>
> >              <version>${project.version}</version>
> > -            <classifier>tests</classifier>
> > -            <scope>test</scope>
> > +            <scope>compile</scope>
>
> why do we need to add <scope>compile</scope>?
>
> previously both dependencies, regular and <classifier>tests</classifier>
> were only brought in for test scope I think which lowers the footprint
> for someone using the REST client. also the ordering of the regular
> classifier first is cleaner i think.
>
> —
> <https://github.com/apache/incubator-brooklyn/pull/690/files#r32823340>.
>

-- 
 
message after it was sent.

ahgittin: i see.  just bring in the method as private static rather than add `core` as a dependency?

robertgmoss:
@ahgittin done

PR #691, opened 2015-06-15
Riak uses ulimit configuration recommended in Basho's docs

ygy:
Riak has a specific requirement for the number of the allowed open files. It is described at http://docs.basho.com/riak/latest/ops/tuning/open-files-limit/.

Riak generates a warning message if the limit is bellow 65536.
```
# /etc/init.d/riak start
Starting riak: !!!!
!!!! WARNING: ulimit -n is 1024; 65536 is the recommended minimum.
!!!!

```
This PR provides configuration parameter for the number of the open files. The number of open files will default to the recommended value (65536) in case provided value is less than it.

### DIFF
--- a/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
+++ b/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
@@ -72,7 +72,12 @@ public interface RiakNode extends SoftwareProcess {
 
     ConfigKey<String> RIAK_CONF_ADDITIONAL_CONTENT = ConfigKeys.newStringConfigKey(
             "riak.riakConf.additionalContent", "Template file (in freemarker format) for setting up additional settings in the riak.conf file", "");
-
+    
+    // maxOpenFiles' default value (65536) is based on the Basho's recommendation - http://docs.basho.com/riak/latest/ops/tuning/open-files-limit/ 
+    @SetFromFlag("maxOpenFiles")
+    ConfigKey<Integer> RIAK_MAX_OPEN_FILES = ConfigKeys.newIntegerConfigKey(
+            "riak.max.open.files", "Number of the open files required by Riak", 65536);
+    
     @SetFromFlag("downloadUrlRhelCentos")
     AttributeSensorAndConfigKey<String, String> DOWNLOAD_URL_RHEL_CENTOS = ConfigKeys.newTemplateSensorAndConfigKey("download.url.rhelcentos",
             "URL pattern for downloading the linux RPM installer (will substitute things like ${version} automatically)",
--- a/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
+++ b/software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
@@ -314,7 +314,7 @@ public class RiakNodeSshDriver extends AbstractSoftwareProcessSshDriver implemen
 
         if (isPackageInstall()) {
             commands.add(addSbinPathCommand());
-            commands.add(sudo("service riak start"));
+            commands.add(sudo(format("sh -c \"ulimit -n %s && service riak start\"", maxOpenFiles())));
         } else {
             // NOTE: See instructions at http://superuser.com/questions/433746/is-there-a-fix-for-the-too-many-open-files-in-system-error-on-os-x-10-7-1
             // for increasing the system limit for number of open files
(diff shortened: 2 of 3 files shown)


software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -314,6 +314,7 @@ public void launch() {
 
         if (isPackageInstall()) {
             commands.add(addSbinPathCommand());
+            commands.add(sudo(format("ulimit -n %s", getNumberOfOpenFiles())));
```
neykov: This will increase the soft limit, but what about the hard limit? Distributions usually have a hard limit < 65536 so this will fail.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeDriver.java
```diff
@@ -45,4 +45,6 @@
     void bucketTypeUpdate(String bucketTypeName, String bucketTypeProperties);
 
     void bucketTypeActivate(String bucketTypeName);
+    
+    Integer getNumberOfOpenFiles();
```
neykov: No need to add it to the interface, usually used when accessed from freemarker templates.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
```diff
@@ -72,7 +72,12 @@
 
     ConfigKey<String> RIAK_CONF_ADDITIONAL_CONTENT = ConfigKeys.newStringConfigKey(
             "riak.riakConf.additionalContent", "Template file (in freemarker format) for setting up additional settings in the riak.conf file", "");
-
+    
+    // numberOpenFiles' default value (65536) is based on the Basho's recommendation - http://docs.basho.com/riak/latest/ops/tuning/open-files-limit/ 
+    @SetFromFlag("numberOpenFiles")
```
neykov: Better name it **max**OpenFiles? Setting this will change the limit.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNode.java
```diff
@@ -72,7 +72,12 @@
...
+    AttributeSensorAndConfigKey<Integer, Integer> RIAK_NUM_OPEN_FILES = ConfigKeys.newIntegerSensorAndConfigKey(
```
neykov: Using a plain `ConfigKey<Integer>` is fine here. You need an `AttributeSensor` only if reading the value back from the machine.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
```diff
@@ -594,4 +595,16 @@ private void addRiakOnPath(ScriptHelper scriptHelper) {
...
         scriptHelper.environmentVariablesReset(newPathVariable);
     }
+
+    @Override
+    public Integer getNumberOfOpenFiles() {
+        Integer defaultNumberOfFiles = RiakNode.RIAK_NUM_OPEN_FILES.getConfigKey().getDefaultValue();
+        Integer numberOfFiles = entity.getConfig(RiakNode.RIAK_NUM_OPEN_FILES);
+        if (numberOfFiles.compareTo(defaultNumberOfFiles) < 0) {
+            log.warn("Specified number of open files : {} : less than the required minimum. Using defaults : {}.", numberOfFiles, defaultNumberOfFiles);
+            entity.setAttribute(RiakNode.RIAK_NUM_OPEN_FILES, defaultNumberOfFiles);
```
neykov: Warning the user is a good idea, but I think it's surprising if the value he sets is not applied. He must have a good reason to change the default value.

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
ygy: There will be no problem with setting the hard limit since the ulimit command is executed with elevated privileges (sudo). 

RedHat based distro
```
[root@localhost ~]# cat /etc/redhat-release 
CentOS release 6.6 (Final)
[root@localhost ~]# ulimit -n
1024
[root@localhost ~]# ulimit -n 65536
[root@localhost ~]# ulimit -Sn
65536
[root@localhost ~]# ulimit -Hn
65536
```

Ubuntu
```
# cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.2 LTS"
# ulimit -n 65536
# ulimit -Sn
65536
# ulimit -Hn
65536
```

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakNodeSshDriver.java
ygy: Completely agree. It would be better to notify the user and stop the installation.

neykov:
Looks good after the changes, can be merged.
Only note is that the image used might set ulimit > 65536, so the command will actually decrease it in this case, but I don't think it's worth supporting it right now.

mikezaccardo:
(LGTM)

PR #692, opened 2015-06-16
Change to install or run dir before executing configured pre and post commands

grkvlt:
This means that pre-install and post-install commands run in `$INSTALL_DIR` and pre-launch and post-launch run in `$RUN_DIR`


grkvlt:
This won't actually work, due to the ordering of the startup events. Closing.

PR #694, opened 2015-06-16
[BROOKLYN-99] add support for NovaSecurityGroupExtension

andreaturli:


aledsage:
Thanks @andreaturli !

Could you add a few additional lines of javadoc to `JcloudsLocationSecurityGroupCustomizer.java` that describes what this class does, where it can be used (e.g. for any jclouds that has the security group extension, or is it more limited - I see explicit mention of nova)?

It would also be useful to have an example of how it should be used, e.g. for someone who wants to use security groups in OpenStack.

andreaturli:
thanks @aledsage!

I've added some javadoc, trying to explain how to use it. HTH

locations/jclouds/src/main/java/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupCustomizer.java
```diff
@@ -225,41 +238,47 @@ void addPermissionsToLocation(Iterable<IpPermission> permissions, final String n
...
-                    new Object[]{applicationId, expectedSharedName, groupsOnNode});
-            return null;
-        }
-        // Shared entry might be missing if Brooklyn has rebound to an application
-        SecurityGroup old = sharedGroupCache.asMap().putIfAbsent(shared.getLocation(), shared);
-        LOG.info("Loaded unique security group for node {} (in {}): {}",
-                new Object[]{nodeId, applicationId, unique});
-        if (old == null) {
-            LOG.info("Proactively set shared group for app {} to: {}", applicationId, shared);
+        if (locationId.equals("aws-ec2")) {
```
mikezaccardo: It may be nice to extract this AWS-specific functionality into a new method.

locations/jclouds/src/main/java/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupCustomizer.java
```diff
@@ -225,41 +238,47 @@ void addPermissionsToLocation(Iterable<IpPermission> permissions, final String n
...
+            if (groupsOnNode.size() != 2) {
+                LOG.warn("Expected to find two security groups on node {} in app {} (one shared, one unique). Found {}: {}",
+                        new Object[]{nodeId, applicationId, groupsOnNode.size(), groupsOnNode});
+                return null;
+            }
+            String expectedSharedName = getNameForSharedSecurityGroup();
+            Iterator<SecurityGroup> it = groupsOnNode.iterator();
+            SecurityGroup shared = it.next();
+            if (shared.getName().endsWith(expectedSharedName)) {
```
mikezaccardo: This could be simplified a bit using `Iterables.tryFind(Iterable<T> iterable, Predicate<? super T> predicate)`

mikezaccardo:
Minor suggestions; LGTM

mikezaccardo:
Official PR image: http://www.eggplante.com/wp-content/uploads/2015/01/Brooklyn-Nine-Nine-Logo1.jpg

andreaturli:
thanks @mikezaccardo!

I'll try to address your suggestions and re-test asap, unless we want to merge it as-is and address it in a subsequent PR. wdyt @richardcloudsoft ?

Really like the official PR image ;)

PR #697, opened 2015-06-18
Fix GeoscalingWebClient create primary domain REST call

mikezaccardo:


mikezaccardo:
The subdomain component of the tests work so they both pass now -- thanks @alasdairhodge

software/webapp/src/test/java/brooklyn/entity/dns/geoscaling/GeoscalingWebClientTest.java
```diff
@@ -42,8 +42,8 @@
     
     private final static String USERNAME = "cloudsoft";
     private final static String PASSWORD = "cl0uds0ft";
-    private final static String PRIMARY_DOMAIN = "domain"+((int)(Math.random()*10000))+".test.org";
-    private final static String SUBDOMAIN = "subdomain"+((int)(Math.random()*10000));
+    private final static String PRIMARY_DOMAIN = "domain"+((int)(Math.random()*10000) + 10000)+".test.org";
+    private final static String SUBDOMAIN = "subdomain"+((int)(Math.random()*10000) + 10000);
```
alasdairhodge: Why the constant offset? Trying to keep the lengths consistent?

alasdairhodge:
Nice catch, Mike; thanks for digging into this :o)

software/webapp/src/test/java/brooklyn/entity/dns/geoscaling/GeoscalingWebClientTest.java
mikezaccardo: To guarantee that they don't collide with the existing domains (as unlikely as it is).

alasdairhodge: Hmm, I don't think that's necessary: the existing (what, four?) `*.test.org` domains look like remnants of previous failed test runs and should be deleted.

alasdairhodge: Actually, that makes me think we should review the domain cleanup logic in the test; looks like tear-down code wasn't executed following previous test failures...

software/webapp/src/test/java/brooklyn/entity/dns/geoscaling/GeoscalingWebClientTest.java
```diff
@@ -42,8 +42,8 @@
     
     private final static String USERNAME = "cloudsoft";
     private final static String PASSWORD = "cl0uds0ft";
-    private final static String PRIMARY_DOMAIN = "domain"+((int)(Math.random()*10000))+".test.org";
-    private final static String SUBDOMAIN = "subdomain"+((int)(Math.random()*10000));
+    private final static String PRIMARY_DOMAIN = "domain"+((int)(Math.random()*10000) + 10000)+".test.org";
```
ahgittin: use `Strings.makeRandomId()` or `Strings.makeZeroPaddedString()` instead ?

mikezaccardo:
Will improve cleanup logic, use better name generation, and remove vestigial domains from GeoScaling.

mikezaccardo:
^ Done x3

aledsage:
Thanks @mikezaccardo - looks good; merging.

PR #699, opened 2015-06-19
Fix JBoss6Server Integration Test

bostko:
Remove unnecessary HOSTNAME vs ROOT_URL comparison


aledsage:
Thanks @bostko - makes sense. Merging.

PR #700, opened 2015-06-19
Fix integration tests (core and software-base)

neykov:


core/src/main/java/brooklyn/util/internal/ssh/ShellAbstractTool.java
```diff
@@ -296,9 +296,10 @@ public ToolAbstractAsyncExecScript(Map<String,?> props) {
          * The executed command will return immediately, but the output from the script
          * will continue to be written 
          * note that some modes require \$RESULT passed in order to access a variable, whereas most just need $ */
+        @Override
         protected List<String> buildRunScriptCommand() {
             String touchCmd = String.format("touch %s %s %s %s", stdoutPath, stderrPath, exitStatusPath, pidPath);
-            String cmd = String.format("( %s > %s 2> %s < /dev/null ; echo $? > %s ) & disown", scriptPath, stdoutPath, stderrPath, exitStatusPath);
```
neykov: This command was working fine with the tests, but using `disown` while ending the ssh session doesn't seem right, so I converted it to use `nohup`. Any reason it shouldn't be changed? cc @aledsage

aledsage: Your change makes sense, I think. It was probably fine before, because stdin is /dev/null and and stdout is the file.

As it says in [1], "if the terminal is destroyed..., the program will fail as soon as it tries to read from standard input or write from standard output." Other than that, it's very similar behaviour (from a user's perspective) to `nohup ... &`.

@richardcloudsoft anything to add?

[1] http://unix.stackexchange.com/questions/3886/difference-between-nohup-disown-and

core/src/main/java/brooklyn/location/basic/SshMachineLocation.java
```diff
@@ -876,7 +877,8 @@ public boolean isSshable() {
         String cmd = "date";
         try {
             try {
-                Socket s = new Socket(getAddress(), getPort());
+                Socket s = new Socket();
+                s.connect(new InetSocketAddress(getAddress(), getPort()), SSHABLE_CONNECT_TIMEOUT);
```
aledsage: Makes sense. Could overload `isSshable()` to allow a custom timeout to be passed in, but not sure where we'd make use of that. So happy to stick with your change.

core/src/test/java/brooklyn/event/feed/http/HttpFeedTest.java
```diff
@@ -300,24 +300,11 @@ public void testStartSuspended() throws Exception {
...
-     * <p>  
-     * 127.0.0.1  unresolvable_hostname_or_one_with_no_webserver_on_port_80
-    // or some other IP which won't resolve
-     */
+    @Test
     public void testPollsAndParsesHttpErrorResponseWild() throws Exception {
         feed = HttpFeed.builder()
                 .entity(entity)
-                .baseUri("http://unresolvable_hostname_or_one_with_no_webserver_on_port_80")
+                .baseUri("http://0.0.0.0")
```
aledsage: Nice!

core/src/test/java/brooklyn/location/basic/SshMachineLocationIntegrationTest.java
```diff
@@ -63,7 +63,14 @@ public void tearDown() throws Exception {
         mgmt = null;
     }
 
-    // Note: requires `named:localhost-passphrase` set up with a key whose passphrase is "localhost"    
+    // Note: requires `named:localhost-passphrase` set up with a key whose passphrase is "localhost"
+    // * create the key with:
```
aledsage: Excellent! We need more of this kind of thing, and to get in the habit of writing it at the same time as the test (me included!).

core/src/test/java/brooklyn/location/geo/HostGeoLookupIntegrationTest.java
```diff
@@ -34,6 +35,7 @@
 
     public static final Logger log = LoggerFactory.getLogger(HostGeoLookupIntegrationTest.class);
     
+    // Needs fast network connectivity to figure out the external IP. If response not returned in 2s fails.
```
aledsage: Interesting. Longer term do you think it's right to fail? Is it possible to differentiate between "still waiting" and failed? Could we wait longer? Or could we throw `SkipException` if we know that it's not a true failure?

Certainly happy with as-is for this release.

software/base/src/test/java/brooklyn/entity/basic/lifecycle/StartStopSshDriverTest.java
```diff
@@ -113,10 +116,10 @@ public void testExecuteDoesNotLeaveRunningStreamGobblerThread() {
             @Override
             public void run() {
               List<ThreadInfo> currentThreads = getThreadsCalling(StreamGobbler.class);
-              List<Long> currentThreadIds = getThreadId(currentThreads);
+              Collection<Long> currentThreadIds = MutableSet.copyOf(getThreadId(currentThreads));
```
aledsage: Personal preference to declare as `Set<Long>`, given line 122 will compare it with a set.

But that's really minor, given that 3 lines earlier we can see it being instantiated!

aledsage:
@neykov thanks, great improvements/fixes!

Merging this now, so we can get the benefits of the fixes in the next test runs. Any review comments that you think deserve code changes can be done later.

aledsage:
asfgit build failure is for an unrelated test: `brooklyn.event.feed.http.HttpFeedTest.testPollsAndParsesHttpErrorResponseWild`. We should fix that in a separate PR.

neykov:
This is caused by a "fix" in this PR. Will correct in a new PR.
Looks like it took 5 hours for the build, and completed only a couple of minutes after the merge :).

aledsage:
Ah, yes. It works for me with your `http://0.0.0.0` change (which makes perfect sense). I guess asfgit's environment behaves differently.

Very happy to leave this with you to investigate @neykov !

core/src/test/java/brooklyn/location/geo/HostGeoLookupIntegrationTest.java
neykov: Can't wait longer/differentiate the result here. It's designed to avoid long blocks on purpose and fail gracefully.
We are already testing all implementations separately so it's worth investigating if we can use stubs here to test the higher level logic.

neykov:
Comments addressed and failure fixed in https://github.com/apache/incubator-brooklyn/pull/702

PR #701, opened 2015-06-19
sensors at riak cluster level and other minor tidies for riak and couchbase

ahgittin:
@bostko hi can you review this?


bostko:
Very good changes, thank you!
I tested the cluster avarage values with https://github.com/apache/incubator-brooklyn/blob/master/examples/simple-web-cluster/src/main/resources/brooklyn/demo/nodejs-riak-todo.yaml and it looks good

P.S. I think it would be better if changes in doStart and those for cluster stats were in separate commits

software/nosql/src/main/java/brooklyn/entity/nosql/riak/RiakClusterImpl.java
```diff
@@ -112,6 +122,38 @@ public URI apply(Collection<URI> input) {
                  .fromMembers()
                  .build();
         addEnricher(first);
+        
+        Map<? extends AttributeSensor<? extends Number>, ? extends AttributeSensor<? extends Number>> enricherSetup = 
+            ImmutableMap.<AttributeSensor<? extends Number>, AttributeSensor<? extends Number>>builder()
+                .put(RiakNode.NODE_PUTS, RiakCluster.NODE_PUTS_1MIN_PER_NODE)
```
aledsage: Would be good if the Riak cluster tests were updated to assert that the `NODE_PUTS_1MIN_PER_NODE` etc are being populated (e.g. that the attribute is eventually non-null).

aledsage:
Looks good; would be nice to improve our test coverage for things like this (even if it is just that the attribute values are non-null). Merging this now.

PR #702, opened 2015-06-19
Remove redundant (and failing) test

neykov:
Functionality covered with next test (without it being integration). Turns out IP 0.0.0.0 points to localhost ports. Ad blockers should reconsider using it :).
Also use port which is known to not be bound, instead of hard coding.


aledsage:
Nice and simple; looks good. Will wait for CI to confirm before merging this time.

PR #703, opened 2015-06-20
Jsgui tweaks

ahgittin:
* config+sensors tooltip renders nicely
* catalog items listed in accordion view don't wrap
* yaml preview window focuses at the top not the bottom (for long blueprints)

The second one needs tested on windows to ensure scrollbars don't clutter.

You can test 2 and 3 with:

```
brooklyn.catalog:
  items:
  - id: foo-bar-baz-foo-bar-baz-foo-bar-baz-foo-bar-baz-foo-bar-baz-foo-bar-baz-foo-bar-baz_boo
    itemType: template
    item:
      location: localhost
      services:
      - type:  brooklyn.entity.basic.BasicEntity
        brooklyn.config:
         x:
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
          blah
```

### DIFF
--- a/usage/jsgui/src/main/webapp/assets/css/base.css
+++ b/usage/jsgui/src/main/webapp/assets/css/base.css
@@ -1114,6 +1114,11 @@ tr.app-add-wizard-config-entry {
     margin-bottom: 6px;
     background-color: #F0F0F0;
     border-radius: 5px;
+    overflow-y: scroll;
+}
+.catalog-accordion-wrapper .accordion-nav-row {
+    white-space: pre;
+    word-wrap: normal;
 }
 .accordion-item {
     -webkit-border-radius: 5px;
--- a/usage/jsgui/src/main/webapp/assets/tpl/apps/config-name.html
+++ b/usage/jsgui/src/main/webapp/assets/tpl/apps/config-name.html
@@ -23,7 +23,7 @@ under the License.
     title="<% 
          if (description) { %><b><%- description %></b><br/><% } 
          if (type) { %>(<% 
-           if (type.startsWith("java.lang.") || type.startsWith("util.")) type = type.substring(10);
+           if (type.startsWith("java.lang.") || type.startsWith("java.util.")) type = type.substring(10);
          %><%- type %>)<% } %>">
     <%- name %>
 </span>
--- a/usage/jsgui/src/main/webapp/assets/tpl/apps/config-name.html
+++ b/usage/jsgui/src/main/webapp/assets/tpl/apps/config-name.html
@@ -18,7 +18,17 @@ specific language governing permissions and limitations
 under the License.
 -->
 
+<% if (description || type) { %>
 <span class="config-name" rel="tooltip" data-placement="left"
-    title="<% if (description) { %><b><%- description %></b><br/><% } %>(<%- type %>)">
+    title="<% 
+         if (description) { %><b><%- description %></b><br/><% } 
+         if (type) { %>(<% 
+           if (type.startsWith("java.lang.") || type.startsWith("util.")) type = type.substring(10);
+         %><%- type %>)<% } %>">
     <%- name %>
 </span>
+<% } else { %>
+<span class="config-name">
+    <%- name %>
+</span>
+<% } %>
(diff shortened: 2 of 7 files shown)


aledsage:
@sjcorbett could you review this please?

@ahgittin asfbot failure is unrelated - fixed in #702, introduced in #700 

@ahgittin do you want this merged for 0.7.0 release, or can it be deferred?

ahgittin:
Yes, I'd like this in 0.7.0, it fixes several glitches in the GUI.

And I have one more on catalog / user-start experience which will come later today which I'd love to get in.  Unfortunately that one isn't such a small one.  :(

usage/jsgui/src/main/webapp/assets/tpl/apps/config-name.html
```diff
@@ -18,7 +18,17 @@
 under the License.
 -->
 
+<% if (description || type) { %>
 <span class="config-name" rel="tooltip" data-placement="left"
-    title="<% if (description) { %><b><%- description %></b><br/><% } %>(<%- type %>)">
+    title="<% 
+         if (description) { %><b><%- description %></b><br/><% } 
+         if (type) { %>(<% 
+           if (type.startsWith("java.lang.") || type.startsWith("util.")) type = type.substring(10);
```
sjcorbett: Why `substring(10)` if `type.startsWith("util.")`?

usage/jsgui/src/main/webapp/assets/tpl/apps/sensor-name.html
```diff
@@ -18,7 +18,17 @@
 under the License.
 -->
 
+<% if (description || type) { %>
 <span class="sensor-name" rel="tooltip" data-placement="left"
-    title="<% if (description) { %><b><%- description %></b><br/><% } %>(<%- type %>)">
+    title="<% 
+         if (description) { %><b><%- description %></b><br/><% } 
+         if (type) { %>(<% 
+           if (type.startsWith("java.lang.") || type.startsWith("util.")) type = type.substring(10);
```
sjcorbett: As before.

neykov:
Horizontal scroll looks fine on Windows:
windows catalog

There's a vertical scrollbar though, caused by a `scrollbar-y: scroll` rule.

usage/jsgui/src/main/webapp/assets/tpl/apps/config-name.html
ahgittin: good spot.  should be `java.util.` of course!

ahgittin:
@neykov thanks for the screen shot -- i'd rather the scrollbar wasn't there but i think worst case we can live with it; wdyt?

the only CSS rule intro'd here applies to the individual rows -- and on mac the vert scroll bar is per "accordion-item" (e.g. applications, or entities; but not all for) -- so i'm wondering if that is introduced here?

ahgittin:
@sjcorbett fix pushed for the `util.` issue (in two places)

good to merge?

sjcorbett:
Fire away.

usage/jsgui/src/main/webapp/assets/css/base.css
```diff
@@ -1114,6 +1114,11 @@ tr.app-add-wizard-config-entry {
     margin-bottom: 6px;
     background-color: #F0F0F0;
     border-radius: 5px;
+    overflow-y: scroll;
```
neykov: This causes the vertical scrollbar on Windows. It's on the element that contains all accordion elements. Looks like it's not really needed? It's not visible on OS X due to the transparent scrollbars.

Not important of course, can be merged as is.

ahgittin: Why the hell did I add this?!?

ahgittin: I'm gonna sneak a remove in to master directly.

PR #704, opened 2015-06-21
clean up how locations are rendered in the js gui, and expand "delete"

ahgittin:
there was (is) confusion between catalog name and location displayName and a named location.
it deserves a major sort-out, i think treating location defs as catalog items and location instances as entities;
and only using the catalog api, not a dedicated locations api (or if there is, it delegates to catalog),
also removing all the LocationManager blah.

but until then we needed something which avoided confusion where catalog item names for locations
aren't rendered, and magic display names from location-metadata were being showed so it looked
like nothing was being added, though the location worked.

for locations, a catalog item name now does nothing, but the location's name (i.e. human-readable ID aka catalog item ID, not the display name)
is shown in the catalog accordion list (the "IdentifierName"), better info (but not the full yaml yet) is shown in the detail,
and versions are at least handled without bugs, even if only one version for location is really supported in the gui.

also adds delete for entity and apps and policies, and fixes delete for locations.

@nakomis this should help prevent the user-confusion which you experienced

### DIFF
--- a/core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
+++ b/core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
@@ -355,6 +355,10 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
                 default: throw new RuntimeException("Only entity, policy & location catalog items are supported. Unsupported catalog item type " + item.getCatalogItemType());
             }
             ((AbstractBrooklynObjectSpec<?, ?>)spec).catalogItemId(item.getId());
+            
+            if (Strings.isBlank( ((AbstractBrooklynObjectSpec<?, ?>)spec).getDisplayName() ))
+                ((AbstractBrooklynObjectSpec<?, ?>)spec).displayName(item.getDisplayName());
+            
             return spec;
         }
 
--- a/usage/jsgui/src/main/webapp/assets/js/view/catalog.js
+++ b/usage/jsgui/src/main/webapp/assets/js/view/catalog.js
@@ -113,9 +113,16 @@ define([
             // Could use wait flag to block removal of model from collection
             // until server confirms deletion and success handler to perform
             // removal. Useful if delete fails for e.g. lack of entitlement.
-            this.activeModel.destroy();
-            var displayName = $(event.currentTarget).data("name");
-            this.renderEmpty(displayName ? "Deleted " + displayName : "");
+            var that = this;
+            var displayName = $(event.currentTarget).data("name") || "item";
+            this.activeModel.destroy({
+                success: function() {
+                    that.renderEmpty("Deleted " + displayName);
+                },
+                error: function(info) {
+                    that.renderEmpty("Unable to permanently delete " + displayName+". Deletion is temporary, client-side only.");
+                }
+            });
         }
     });
 
@@ -259,6 +266,11 @@ define([
     }
 
     var Catalog = Backbone.Collection.extend({
+        modelX: Backbone.Model.extend({
+          url: function() {
+            return "/v1/catalog/" + this.name + "/" + this.id + "?allVersions=true";
+          }
+        }),
         initialize: function(models, options) {
             this.name = options["name"];
             if (!this.name) {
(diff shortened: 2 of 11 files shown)


aledsage:
Code looks sensible - haven't had a chance to test it though.

core/src/main/java/brooklyn/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -355,6 +355,10 @@ public void load() {
                 default: throw new RuntimeException("Only entity, policy & location catalog items are supported. Unsupported catalog item type " + item.getCatalogItemType());
             }
             ((AbstractBrooklynObjectSpec<?, ?>)spec).catalogItemId(item.getId());
+            
+            if (Strings.isBlank( ((AbstractBrooklynObjectSpec<?, ?>)spec).getDisplayName() ))
+                ((AbstractBrooklynObjectSpec<?, ?>)spec).displayName(item.getDisplayName());
```
robertgmoss: Would be nicer to see this cast only once

usage/rest-server/src/main/java/brooklyn/rest/resources/LocationResource.java
```diff
@@ -82,7 +86,22 @@ public LocationSummary apply(LocationDefinition l) {
...
+    private static NaturalOrderComparator COMPARATOR = new NaturalOrderComparator();
+    private static Comparator<LocationSummary> nameOrSpecComparator() {
+        return new Comparator<LocationSummary>() {
+            @Override
+            public int compare(LocationSummary o1, LocationSummary o2) {
+                return COMPARATOR.compare(getNameOrSpec(o1).toLowerCase(), getNameOrSpec(o2).toLowerCase());
+            }
+        };
+    }
+    private static String getNameOrSpec(LocationSummary o) {
```
robertgmoss: would this be better in the comparator as a non static method? Or is it more widely useful?

usage/jsgui/src/main/webapp/assets/js/view/catalog.js
```diff
@@ -113,9 +113,16 @@ define([
...
-            var displayName = $(event.currentTarget).data("name");
-            this.renderEmpty(displayName ? "Deleted " + displayName : "");
+            var that = this;
+            var displayName = $(event.currentTarget).data("name") || "item";
+            this.activeModel.destroy({
+                success: function() {
+                    that.renderEmpty("Deleted " + displayName);
+                },
+                error: function(info) {
+                    that.renderEmpty("Unable to permanently delete " + displayName+". Deletion is temporary, client-side only.");
```
robertgmoss: Reverse the order of these two sentences - I had to do a double take because it looks like a failure message but the location is removed.

robertgmoss: In fact, didn't realise the item would reappear as soon as refresh is clicked.  Perhaps it would be better not to remove it from the GUI and just say it failed (and why)?

usage/jsgui/src/main/webapp/assets/js/view/catalog.js
```diff
@@ -267,7 +279,12 @@ define([
             //this.model is a constructor so it shouldn't be _.bind'ed to this
             //It actually works when a browser provided .bind is used, but the
             //fallback implementation doesn't support it.
-            var model = this.model;
+            var that = this;
```
mikezaccardo: Is `that` variable necessary?

usage/rest-server/src/main/java/brooklyn/rest/resources/LocationResource.java
```diff
@@ -152,6 +171,15 @@ public Response create(LocationSpec locationSpec) {
     @Override
     @Deprecated
     public void delete(String locationId) {
-        brooklyn().getCatalog().deleteCatalogItem(locationId);
+        // TODO make all locations be part of the catalog, then flip the JS GUI to use catalog api
+        if (deleteAllVersions(locationId)>0) return;
```
mikezaccardo: Personal preference: make the condition == 0 then throw, no return necessary.

usage/jsgui/src/main/webapp/assets/js/view/catalog.js
ahgittin: yes - else if we did `this.name` in line 285 it might not resolve.  (whereas the `this.id` refers to the model's `id`.)

usage/jsgui/src/main/webapp/assets/js/view/catalog.js
ahgittin: We need to overhaul this anyway to treat locations like other catalog items so not going to address.

ahgittin:
Good nit-picking, I don't disagree with most, but this code needs overhaul so not worth addressing.

PR #705, opened 2015-06-21
Catalog default niceness

ahgittin:
set up a default catalog which includes:
* four YAML application templates to get started with, teaching most of what people should know
* scanned items plus some convenient short-names for common entities

makes the Brooklyn code so that the four sample templates can give a pretty nice user experience with sensors and enrichers, and they all run without warnings in the log or ugliness in the activities view (in prep for 0.7.0 release):
* resolution of values is more thorough, respecting timeouts, and run as transient
* new enrichers for time-weighted-delta and rolling averages from YAML

also the default catalog comes from the CLI project, and is copied to the dist, so that we don't have to maintain two variants but can test from the IDE of the CLI project


duncangrant:
The Resilient Load-Balanced Web Cluster fails for me with a VPCResourceNotSpecified error.

ahgittin:
@duncangrant That sounds like it isn't due to this PR, but this class of error from AWS should be a showstopper for a GA release.  Is it one component or all of them?  Did the other items work?  Can you give any more details?  (If you're busy fixing this then ignore me for now but if not I want to make sure someone can pick this up.)

@aledsage @rdowner do you agree this should block a GA release?

rdowner:
Definitely worth holding back the release until we know if this is an issue affecting `master` or not. I wouldn't like to make a GA release with a known problem affecting the world's favourite compute cloud...

duncangrant:
@ahgittin All the components fail.  
I can easily replicate with the following using master so you're right :
name: Empty large System
location: jclouds:aws-ec2:eu-west-1
services:
- type: brooklyn.entity.basic.EmptySoftwareProcess
  provisioning.properties:
    minRam: 2gb

A simple fix for the above is to ask for 8gb.  Doing that does seem to fix the issue for the nginx node but the node in the cluster fails.  I'm not entirely sure why it fails but the node seems to stop and then it seems to retry waiting for main.uri a few times.  And then the node is torn down.

rdowner:
Ok that's pretty clear, the release will have to be paused until we understand this problem and have a solution. As it is unrelated to this PR let's continue the discussion on the dev@brooklyn list instead of here.

core/src/main/java/brooklyn/enricher/basic/AbstractAggregator.java
```diff
@@ -187,6 +187,24 @@ protected void setEntityAfterSubscribingProducerChildrenEvents() {
         }
     }
 
+    /** true if this should aggregate members */
+    protected boolean isAggregatingMembers() {
+        if (Boolean.TRUE.equals(fromMembers)) return true;
```
aledsage: remind me: is setEntity(EntityLocal) called on rebind? OR could `fromMembers` potentially be null after rebeind?

ahgittin:
confirmed for me too, specifying `minRam: 2gb` in AWS results in:

```
2015-06-22 19:33:29,968 ERROR Failed to start VM for aws-ec2:eu-west-1@NginxControllerImpl{id=m8aiWnk1}: request POST https://ec2.eu-west-1.amazonaws.com/ HTTP/1.1 failed with code 400, error: AWSError{requestId='4e534c1d-1884-4776-a3e6-5efd470c9c79', requestToken='null', code='VPCResourceNotSpecified', message='The specified instance type can only be used in a VPC. A subnet ID or network interface ID is required to carry out the request.', context='{Response=, Errors=}'}
```

We hit issues like this when writing https://github.com/jclouds/jclouds/pull/732 and temporary local Brooklyn patch/copy https://github.com/apache/incubator-brooklyn/pull/610 -- these focussed on fixing the issue with selecting incompatible images, so it's possible we need more fixes from the former PR added to Brooklyn.  Or possibly AWS have made more restrictive which instance types can run outwith a VPC.  More anon.

core/src/main/java/brooklyn/enricher/basic/AbstractAggregator.java
ahgittin: yes, rebind calls `Entity.addEnricher` which calls `Enricher.setEntity`

ahgittin:
According to http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-vpc.html#vpc-only-instance-types these instance types -- C4, M4, T2 -- are only supported in VPC, not in "classic".  The jclouds PR includes code to prefer instance types available in both, and so would fix the problem, but the Brooklyn cut only takes the image half of that PR.  So one option would be to copy more of the PR.  But I'm still investigating...

ahgittin:
Right, it looks like non-VPC use ("EC2-Classic") is pretty much deprecated, only available to people with older accounts and for them in regions used before or during 2013.

So I think the pragmatic thing to do is give a nice error in this case, directing people to use `eu-central-1` which launched in 2014 (and so only has VPC) or to specify a specific `networkName` e.g. `subnet-041c8373` for a subnet which auto-assigns public IP's or where Brooklyn has access or is configured to use a jump host.

I will add this nice error message and change the default location to `eu-central-1`.

For more info:

* http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-vpc.html#vpc-only-instance-types
* http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/default-vpc.html
* http://aws.amazon.com/vpc/faqs/#Default_VPCs

ahgittin:
Okay, I've added a more helpful error if this situation is encountered, and that seems to be working:

| 2015-06-22 20:30:33,643 ERROR Detected that your EC2 account is a legacy 'classic' account, but the recommended instance type requires VPC. You can specify the 'eu-central-1' region to avoid this problem, or you can specify a classic-compatible instance type, or you can specify a subnet to use (taking care that the subnet auto-assigns public IP's, or Brooklyn can see the subnet or is configured to use a jump host for this location). For more information see http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-vpc.html.

However the "fixes" aren't working:

* `networkName: subnet-ID` gives the same `VPCResourceNotSpecified` error; maybe the syntax is off?

* `eu-central-1` (VPC) gives `com.google.common.util.concurrent.UncheckedExecutionException: java.lang.RuntimeException: security group eu-central-1/jclouds#brooklyn-bxza-alex-eu-central-shoul-u2jy-nginx-ielm is not available after creating` -- I saw one of those pages suggested max security groups was 5, so maybe that's it

I'm investigating.

ahgittin:
Got to the bottom of it:

* `networkName` sets `TemplateOptions.networks(...)` which it looks like is only supported for CloudStack and OpenStack.  @andreaturli can you confirm?  i've mapped this to `subnetId` for AWS, and log warning in SoftLayer.  we might want other warnings in future.

* the `not available after creating` message appears to be a very occasional race behaviour; on a retry it was fine.  I've upped the timeout in hopes of reducing it.

* however for *non-default* VPC's (i.e. if you set `networkId: subnet-ID`) jclouds does not set up security groups, so that suggestion now says to configure the default security group for that VPC to allow all the access you want.  not ideal, but advanced-networking is wanted when working with this so we can live ith it.

the message for "classic" users who get the VPC error now says:

                LOG.error("Detected that your EC2 account is a legacy 'classic' account, but the recommended instance type requires VPC. "
                    + "You can specify the 'eu-central-1' region to avoid this problem, or you can specify a classic-compatible instance type, "
                    + "or you can specify a subnet to use with 'networkName' "
                    + "(taking care that the subnet auto-assigns public IP's and allows ingress on all ports, "
                    + "as Brooklyn does not currently configure security groups for non-default VPC's; "
                    + "or setting up Brooklyn to be in the subnet or have a jump host or other subnet access configuration). "
                    + "For more information on VPC vs classic see http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-vpc.html.");

ahgittin:
@duncangrant - can you let me know if this fixes it?

all - anyone else had the chance to work through this and want to give input?

@aledsage - you've done the most on the auto-scaler; any objections to the change just made to set the default min to 1 ?

tbouron:
@ahgittin Regarding VPC and security groups configuration, here is what works for me:
 * As you said, subnet ID is configured through `subnetId` for AWS
 * To define my own security groups, I add a `templateOptions.securityGroupIds: [[abc, xyz]]`. While this works, the specified security groups have to created manually first. They also have to be assigned to the same VPC as the `subnetId` you choose.

andreaturli:
@ahgittin I think your analysis is very good and complete!
Personally I think jclouds needs some clean up in this area: the generic `TemplateOptions` contains `networks` so I think for this needs to be used in most of the cases, instead of adding a custom option.
There are already TODOs or @deprecated in the code so it is ongoing.

a couple of comments:
- Cloudstack uses custom `securityGroupIds` and `networkIds` not the generic `networks`
- Openstack Nova uses `securityGroupNames` which is marked as @deprecated (suggests to use `groups` which is maybe even more confusing)
- `Azure` supports the custom `networkSecurityGroupName` 

so maybe we should warn the Openstack, Cloudstack and Azure users as you did for SoftLayer.

-  `GCE` seems the only one using `templateOptions.networks()` correctly

duncangrant:
@ahgittin that change certainly seems to fix both issues I was having

usage/dist/pom.xml
```diff
@@ -98,9 +132,14 @@
...
                             <descriptors>
-                                <descriptor>src/main/config/build-distribution-archive.xml</descriptor>
+                                <descriptor>src/main/config/build-distribution.xml</descriptor>
                             </descriptors>
                             <finalName>brooklyn-${project.version}</finalName>
+                            <formats>
+                                <format>tar.gz</format>
+                                <format>zip</format>
+                            </formats>
+                            <attach>false</attach>
```
neykov: I think we should stick to the default behaviour of attaching the artifacts here and remove the explicit `attach-artifact` goal instead, see comments in https://github.com/apache/incubator-brooklyn/pull/709.

usage/dist/pom.xml
```diff
@@ -63,6 +63,35 @@
     <build>
         <plugins>
             <plugin>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <!-- copy the config file from the CLI project -->
+                <executions>
+                    <execution>
+                        <id>copy</id>
```
neykov: Better use `unpack` goal with `includes` filter. This will make the bom artifact type on cli unnecessary.

software/base/src/main/java/brooklyn/entity/software/StaticSensor.java
```diff
@@ -36,9 +43,17 @@ public StaticSensor(ConfigBag params) {
         value = params.get(STATIC_VALUE);
     }
 
+    @SuppressWarnings("unchecked")
     @Override
     public void apply(EntityLocal entity) {
         super.apply(entity);
-        entity.setAttribute(sensor, (T) TypeCoercions.coerce(value, sensor.getType()));
+        
+        Maybe<T> v = Tasks.resolving(value).as((Class<T>)sensor.getType()).timeout(Duration.millis(200)).getMaybe();
```
neykov: Why not set the raw value and resolve & coerce at read time?

The resolve timeouts seem to be arbitrarily chosen - I've seen 50ms at other places. Better use a global constant instead?

ahgittin:
Thanks @tbouron @andreaturli .  I've updated the warnings and code comments to reflect this sad state of affairs with `TemplateOptions.networks(...)`.  @tbouron your confirmation was helpful; you can now use `networkName` directly, it is mapped to `subnetId` for AWS.

usage/dist/pom.xml
ahgittin: I wondered if there was such a thing.  That's cool.  I'd have used it if I'd known.

However I do like having an artifact for the `bom` as it is easier to access directly.  Maybe it makes sense for that to be a standard pattern for bundles.  We'll see.  For now I'll leave it but add a comment about your suggestion.

software/base/src/main/java/brooklyn/entity/software/StaticSensor.java
ahgittin: This is an important distinction which I don't think we make explicit.

Config supports tasks/suppliers which are resolved whenever read.  Sensors are values published at a point in time so don't normally support resolution (except for this one).

I've added javadoc:

```
/** 
 * Provides an initializer/feed which simply sets a given value.
 * <p>
 * {@link Task}/{@link Supplier} values are resolved when written,
 * unlike config values which are resolved on each read.
 * <p>
 * This supports a {@link StaticSensor#SENSOR_PERIOD} 
 * which can be useful if the supplied value is such a function.
 * However when the source is another sensor,
 * consider using {@link Propagator} which listens for changes instead. */
```

ahgittin: 100% agree re arbitrary numbers in code.  i've replace them with defined constants whose name and javadoc give a bit of reasoning behind the arbitrariness.  lemme know if you think it's an improvement @neykov .

usage/dist/pom.xml
ahgittin: Yes, better to `<attach>false</attach>` only for the `dir`.  I've removed `attach-artifact` and it works well -- though it was still archiving twice.  Changing to `<phase>package</phase>` seems to fix that, and runs it after tests.

ahgittin:
@neykov I've addressed your excellent observations.  Do the last few commits look alright?

ahgittin:
Rebased to fix conflicts with #709, and merged to prevent the build error I was getting from it.  The last few commits could do with a quick scan-after-commit, though it's really just addressing comments here so not crucial.

However @rdowner if the `apache-` prefix in `usage/dist/pom.xml` was needed for something we're going to have to find a different way to achieve it I think.

core/src/main/java/brooklyn/enricher/basic/YamlTimeWeightedDeltaEnricher.java
```diff
@@ -0,0 +1,81 @@
...
+    
+    public static ConfigKey<Duration> DELTA_PERIOD = ConfigKeys.newConfigKey(Duration.class, "enricher.delta.period",
+        "Duration that this delta should compute for, default per second", Duration.ONE_SECOND);
+    
+    @Override
+    protected Function<SensorEvent<T>, Double> getTransformation() {
+        return new Function<SensorEvent<T>, Double>() {
+            @Override
+            public Double apply(SensorEvent<T> event) {
+                Number current = TypeCoercions.coerce(event.getValue(), Double.class);
```
neykov: Is this necessary, when you are already using `.doubleValue()` further down?

core/src/main/java/brooklyn/enricher/basic/YamlTimeWeightedDeltaEnricher.java
```diff
@@ -0,0 +1,81 @@
...
+                        // cannot calculate time-based delta with a single value
+                        if (LOG.isTraceEnabled()) LOG.trace("{} received event but no last value so will not emit, null -> {} at {}", new Object[] {this, current, eventTime}); 
+                    } else {
+                        double duration = eventTime - lastTime;
+                        result = (current.doubleValue() - lastValue.doubleValue()) / (duration / unitMillis);
+                    }
+                }
+                
+                lastValue = current;
+                lastTime = eventTime;
```
neykov: Synchronize reads/writes in pairs?

policy/src/main/java/brooklyn/enricher/TimeWeightedDeltaEnricher.java
```diff
@@ -41,6 +42,8 @@
  * presented as a units/timeUnit based on the event timing.
  * <p>
  * NB for time (e.g. "total milliseconds consumed") use {@link TimeFractionDeltaEnricher}
+ * <p>
+ * See also {@link YamlTimeWeightedDeltaEnricher} designed for use from YAML.
```
neykov: If both this and the YAML alternative are going to be used long-term, why not re-use the code? Alternatively deprecate this one?

Analogous comment for `RollingTimeWindowMeanEnricher` would be nice.

usage/rest-server/src/main/java/brooklyn/rest/transform/CatalogTransformer.java
```diff
@@ -59,21 +59,28 @@
...
 
+            for (ConfigKey<?> x: type.getConfigKeys())
+                config.add(EntityTransformer.entityConfigSummary(x, typeMap.getConfigKeyField(x.getName())));
+            for (Sensor<?> x: type.getSensors())
+                sensors.add(SensorTransformer.sensorSummaryForCatalog(x));
+            for (Effector<?> x: type.getEffectors())
+                effectors.add(EffectorTransformer.effectorSummaryForCatalog(x));
+            
+        } catch (Exception e) {
+            // templates with multiple entities can't have spec created in the manner above; just ignore
```
neykov: Rethrow if type==Entity (or at least log an error), also propagateIfFatal.

utils/common/src/main/java/brooklyn/util/yaml/Yamls.java
```diff
@@ -350,7 +351,11 @@ public String getMatchedYamlTextOrWarn() {
                 if (e instanceof KnownClassVersionException) {
                     log.debug("Known class version exception; no yaml text being matched for "+this+": "+e);
                 } else {
-                    log.warn("Unable to match yaml text in "+this+": "+e, e);
+                    if (e instanceof UserFacingException) {
+                        log.warn("Unable to match yaml text in "+this+": "+e.getMessage());
```
neykov: nitpick: The stack trace would still be useful in the debug log.

core/src/main/java/brooklyn/entity/basic/BrooklynTaskTags.java
```diff
@@ -267,7 +267,12 @@ public static WrappedStream stream(Task<?> task, String streamType) {
     
     public static void setInessential(Task<?> task) { addTagDynamically(task, INESSENTIAL_TASK); }
     public static void setTransient(Task<?> task) { addTagDynamically(task, TRANSIENT_TASK_TAG); }
-    public static boolean isTransient(Task<?> task) { return hasTag(task, TRANSIENT_TASK_TAG); }
+    public static boolean isTransient(Task<?> task) { 
+        if (hasTag(task, TRANSIENT_TASK_TAG)) return true;
+        if (hasTag(task, NON_TRANSIENT_TASK_TAG)) return true;
```
neykov: should be `return false`

core/src/main/java/brooklyn/util/task/BasicExecutionContext.java
```diff
@@ -169,6 +169,11 @@ public T call() {
 
         taskTags.addAll(tags);
         
+        if (Tasks.current()!=null && BrooklynTaskTags.isTransient(Tasks.current()) && !taskTags.contains(BrooklynTaskTags.NON_TRANSIENT_TASK_TAG)) {
```
neykov: Check for `!taskTags.contains(TRANSIENT_TASK_TAG)` to prevent multiple entries (in case already set by the user).

core/src/main/java/brooklyn/enricher/basic/Transformer.java
```diff
@@ -24,25 +24,20 @@
...
 
 import com.google.common.base.Function;
 import com.google.common.reflect.TypeToken;
 
 //@Catalog(name="Transformer", description="Transforms attributes of an entity; see Enrichers.builder().transforming(...)")
 @SuppressWarnings("serial")
-public class Transformer<T,U> extends AbstractEnricher implements SensorEventListener<T> {
+public class Transformer<T,U> extends AbstractTransformer<T,U> {
 
+    @SuppressWarnings("unused")
```
neykov: What's the point of keeping the logger, better delete it.

core/src/main/java/brooklyn/util/flags/TypeCoercions.java
```diff
@@ -723,7 +727,31 @@ public Class apply(final String input) {
...
+                return Sensors.newSensor(Object.class, input);
+            }
+        });
+        registerAdapter(String.class, Sensor.class, new Function<String,Sensor>() {
+            @Override
+            public AttributeSensor apply(final String input) {
+                Entity entity = BrooklynTaskTags.getContextEntity(Tasks.current());
+                if (entity!=null) {
+                    Sensor<?> result = null;
+                    if (entity!=null) {
```
neykov: Already checked != null above.

usage/cli/src/main/resources/brooklyn/default.catalog.bom
```diff
@@ -1,23 +1,359 @@
...
+    description: |
+      Sample YAML to provision a server in a cloud with illustrative VM properties
+    item:
+      name: Server (Brooklyn Example)
+      
+      # this basic example shows how Brooklyn can provision a single raw VM
+      # in the cloud or location of your choice
+      
+      services:
+      - type:           server
```
neykov: personal preference: The indentation looks confusing.

usage/cli/src/main/resources/brooklyn/default.catalog.bom
```diff
@@ -1,23 +1,359 @@
...
+          - reqs.per_sec.windowed_30s.per_node
+      
+      location:
+        jclouds:aws-ec2:
+          # edit these (or delete if credentials specified in brooklyn.properties)      
+          identity:     <REPLACE>
+          credential:   <REPLACE>
+          
+          region:       eu-central-1
+          minRam:       2gb
```
neykov: Excellent examples! A comprehensive showcase of the power of Brooklyn.

neykov:
Finished review (a bit late :) ). At least a couple of the comments worth addressing in a new PR.

policy/src/main/java/brooklyn/enricher/TimeWeightedDeltaEnricher.java
ahgittin: I've marked the older classes `@Beta` and added a reference.  I think they should go in time, but not sure yet...

utils/common/src/main/java/brooklyn/util/yaml/Yamls.java
ahgittin: not sure it would actually -- possibly trace but an error here should be easy to reproduce

usage/rest-server/src/main/java/brooklyn/rest/transform/CatalogTransformer.java
ahgittin: +1 and +1

core/src/main/java/brooklyn/entity/basic/BrooklynTaskTags.java
ahgittin: great spot!!!

core/src/main/java/brooklyn/util/task/BasicExecutionContext.java
ahgittin: +1

core/src/main/java/brooklyn/util/flags/TypeCoercions.java
ahgittin: who wrote this mess?  :)

usage/cli/src/main/resources/brooklyn/default.catalog.bom
ahgittin: the *values* are aligned -- it's a format @alasdairhodge showed me, i quite like it ... but we can revert it at any point if people don't!

mikezaccardo:
I get this output when running master now:

`2015-06-24 13:52:20,214 WARN  Location resolution failed for 'brooklyn.entity.proxy.nginx.NginxController' (will fail shortly): known resolvers are: [id, named, brooklyn.catalog, localhost, byon, single, host, multi, portForwardManager, pool, jclouds, jcloudsByon]
2015-06-24 13:52:20,219 WARN  Error importing catalog from classpath:/brooklyn/default.catalog.bom: brooklyn.util.exceptions.CompoundRuntimeException: Could not resolve item load-balancer:
type: brooklyn.entity.proxy.nginx.NginxController
name: Load Balancer (nginx)


brooklyn.util.exceptions.CompoundRuntimeException: Could not resolve item load-balancer:
type: brooklyn.entity.proxy.nginx.NginxController
name: Load Balancer (nginx)


	at brooklyn.util.exceptions.Exceptions.create(Exceptions.java:278) ~[classes/:na]`

ahgittin:
Thanks @neykov .  Great comments.  Addressed in b5cc2779dc9de9d44093a8041c8cbd5591c23f47 pushed directly to master as it is just PR tweaks.

ahgittin:
This is bizarre. It's trying to load it as a location. And I can't
reproduce. Is this with a distinctive build and no other catalog? Can you
paste the full stack trace from the debug log?
On 24 Jun 2015 13:53, "Mike Zaccardo" <notifications@github.com> wrote:

> I get this output when running master now:
>
> `2015-06-24 13:52:20,214 WARN Location resolution failed for
> 'brooklyn.entity.proxy.nginx.NginxController' (will fail shortly): known
> resolvers are: [id, named, brooklyn.catalog, localhost, byon, single, host,
> multi, portForwardManager, pool, jclouds, jcloudsByon]
> 2015-06-24 13:52:20,219 WARN Error importing catalog from
> classpath:/brooklyn/default.catalog.bom:
> brooklyn.util.exceptions.CompoundRuntimeException: Could not resolve item
> load-balancer:
> type: brooklyn.entity.proxy.nginx.NginxController
> name: Load Balancer (nginx)
>
> brooklyn.util.exceptions.CompoundRuntimeException: Could not resolve item
> load-balancer:
> type: brooklyn.entity.proxy.nginx.NginxController
> name: Load Balancer (nginx)
>
> at brooklyn.util.exceptions.Exceptions.create(Exceptions.java:278) ~[classes/:na]`
>
> —
> <https://github.com/apache/incubator-brooklyn/pull/705#issuecomment-115009336>
> .
>

-- 
 
message after it was sent.

ahgittin:
I've found the problem @mikezaccardo referred to.  The logging re "location" is spurious, I've downgraded that message in 6648495077669378a6540d939e9d8f44a68a1ff6 .  Currently when running `Main` in an IDE you don't have the load-balancer project and so that type does not load, and that fails the catalog load.

It's an IDE-only issue, but at least now we get sensible warnings.  The dist will be fine.

The commit above also fixes the two `YamlTime...` comments @neykov made above.

PR #706, opened 2015-06-21
Fix various integration tests

aledsage:


software/base/src/test/java/brooklyn/entity/brooklynnode/BrooklynNodeIntegrationTest.java
```diff
@@ -361,6 +362,18 @@ protected static void waitForApps(String webConsoleUri) {
         HttpTestUtils.assertHttpStatusCodeEventuallyEquals(webConsoleUri+"/v1/applications", 200);
     }
 
+    protected void waitForApps(final URI webConsoleUri, final int num) {
+        waitForApps(webConsoleUri.toString());
+        
+        // e.g. [{"id":"UnBqPcqg","spec":{"name":"Application (UnBqPcqg)","type":"brooklyn.entity.basic.BasicApplication","locations":["pOL4NtiW"]},"status":"RUNNING","links":{"self":"/v1/applications/UnBqPcqg","entities":"/v1/applications/UnBqPcqg/entities"}}]
+        Asserts.succeedsEventually(new Runnable() {
```
neykov: I think the test correctly fails in this case. Once the API returns non-403 response all apps are supposed to be rebinded successfully. Clients shouldn't see inconsistent state.

We should fix the underlying code rather than mask the problem in the test.

neykov: Scratch that, the change is for startup (non-rebind) cases only, so good to have.

neykov:
LGTM

PR #707, opened 2015-06-22
BROOKLYN-137 stopIptables and openIptables don't work on CentOS7

ygy:
Implemented to retain full backwards compatibility before the release

- Adding preliminary support for firewalld
- Starting/stopping/restrting of firewalld service
- Adding rules via firewalld's direct interface
- Tests

### DIFF
--- a/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
@@ -364,7 +364,15 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
                 ? (OsFamily.WINDOWS == os.getFamily()) 
                 : (OsFamily.WINDOWS == confFamily);
     }
-
+    
+    public boolean isNodeFirewalldEnabled(NodeMetadata node) {
+        String OS = node.getOperatingSystem().getFamily().toString();
+        String version = node.getOperatingSystem().getVersion();
+        return node.getOperatingSystem().getVersion().startsWith("7") &&
+             (node.getOperatingSystem().getFamily().equals(OsFamily.RHEL) ||
+              node.getOperatingSystem().getFamily().equals(OsFamily.CENTOS));
+    }
+    
     protected Semaphore getMachineCreationSemaphore() {
         return checkNotNull(getConfig(MACHINE_CREATION_SEMAPHORE), MACHINE_CREATION_SEMAPHORE.getName());
     }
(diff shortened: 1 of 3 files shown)


bostko:
Looks good. @ygy Can you remove sudo from the firewall-cmd related methods in IptablesCommands.
This will improve the readability of `IptablesCommandsFirewalldTest`

andreaturli:
@ygy thanks for that! great start!

Do you think it is possible to refactor a bit your implementation? It would be nice to rename `IptablesCommands` a more generic `FirewallCommands` interface, which is able to use the correct implementation (`iptables` vs `firewalld` and possibly `windows firewall`) based on the OsFamily to make it more portable?

If so, we should probably deprecate `IptablesCommands` instead of removing it immediately.

tbouron:
@andreaturli +1. Actually, this PR can also convert that: https://issues.apache.org/jira/browse/BROOKLYN-151

andreaturli:
thanks @tbouron it makes a lot of sense!

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -364,7 +364,15 @@ public boolean isWindows(NodeMetadata node, ConfigBag config) {
...
                 : (OsFamily.WINDOWS == confFamily);
     }
-
+    
+    public boolean isNodeFirewalldEnabled(NodeMetadata node) {
+        String OS = node.getOperatingSystem().getFamily().toString();
+        String version = node.getOperatingSystem().getVersion();
+        return node.getOperatingSystem().getVersion().startsWith("7") &&
+             (node.getOperatingSystem().getFamily().equals(OsFamily.RHEL) ||
+              node.getOperatingSystem().getFamily().equals(OsFamily.CENTOS));
```
neykov: The check is not future proof. Instead of checking fixed OS versions is it possible to check for the existence of firewall-cmd instead? This will allow the blueprint to work on future versions of the OS without any changes.

ygy:
Thanks for the feedback.

@bostko: adding rules with firewall-cmd requires root privileges. it follows the implementation of addIptablesRule

@andreaturli @tbouron: actually, the initial implementation was just as your proposal. i've discussed it with @bostko already.

I've decided to go with the current one, because we are days before a major release. I didn't wanted to change the interface or introduce big changes which will require comprehensive testing. It will be good to have firewalld support in 0.7 so we can get Riak running on CentOS 7 too.
If you agree, I can get it implemented with proper abstraction post 0.7 release?

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -365,6 +365,16 @@ public boolean isWindows(NodeMetadata node, ConfigBag config) {
                 : (OsFamily.WINDOWS == confFamily);
     }
 
+    public boolean isLocationFirewalldEnabled(SshMachineLocation location) {
+        int result = location.execCommands("checking if firewalld is active", 
+                ImmutableList.of(IptablesCommands.firewalldServiceIsActive()));
+        if (result == 0) {
+            return true;
+        }
+
```
ygy: @neykov here is addressed your previous comment regarding the firewall-cmd existence

utils/common/src/main/java/brooklyn/util/ssh/IptablesCommands.java
```diff
@@ -89,6 +89,38 @@ public static String iptablesServiceStatus() {
         return iptablesService("status");
     }
 
+    @Beta // implementation not portable across distros
+    public static String firewalldService(String cmd) {
+        return sudo(alternatives(
+                BashCommands.ifExecutableElse1("systemctl", "systemctl " + cmd + " firewalld"),
+                "/usr/bin/systemctl " + cmd + " firewalld"));
```
neykov: Does this mean that `/usr/bin` might not be on the `PATH`? In what cases?

neykov:
LGTM. +1 for @andreaturli's suggestions which can be addressed in a future PR.

utils/common/src/main/java/brooklyn/util/ssh/IptablesCommands.java
ygy: It's just as a precaution in case someone changed or removed environment variables for non-interactive shells

PR #708, opened 2015-06-22
Ignore entity.addFeed(...) call with same feed object

neykov:
Adding the same feed object a second time would stop it, ignore the call instead. Adding the same object twice is very common when `FEATURE_FEED_REGISTRATION_PROPERTY` is enabled, because blueprints already add it manually.

Fixes `NginxRebindWithHaIntegrationTest.testChangeModeFailureStopsTasksButHappyUponResumption`, which enables `FEATURE_FEED_REGISTRATION_PROPERTY`.

### DIFF
--- a/core/src/test/java/brooklyn/event/feed/function/FunctionFeedTest.java
+++ b/core/src/test/java/brooklyn/event/feed/function/FunctionFeedTest.java
@@ -19,6 +19,7 @@
 package brooklyn.event.feed.function;
 
 import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotEquals;
 import static org.testng.Assert.assertTrue;
 
 import java.util.List;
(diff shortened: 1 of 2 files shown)


core/src/test/java/brooklyn/event/feed/function/FunctionFeedTest.java
```diff
@@ -248,6 +261,26 @@ public void testFunctionPollConfigBuilding() throws Exception {
...
+        final Integer val = entity.getAttribute(SENSOR_INT);
+        Asserts.succeedsEventually(new Runnable() {
+            @Override
+            public void run() {
+                assertNotEquals(val, entity.getAttribute(SENSOR_INT));
+            }
+        });
+    }
+    
+    private void assertFeedIsPollingContiniously() {
```
sjcorbett: `Continiously` should be `Continuously`.

sjcorbett:
Looks good.

neykov:
Fixed spelling.

PR #709, opened 2015-06-23
Fixes for release blockers

rdowner:
A couple of small issues which would nevertheless cause a release built on a clean build machine to fail.

### DIFF
--- a/usage/camp/pom.xml
+++ b/usage/camp/pom.xml
@@ -59,7 +59,13 @@
             <groupId>org.apache.brooklyn.camp</groupId>
             <artifactId>camp-base</artifactId>
             <version>${project.version}</version>
-        </dependency>
+            <exclusions>
+                <!-- Dependency versions mismatch between transitive dependencies, declare explicitly -->
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
+            </exclusions>        </dependency>
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
--- a/usage/dist/pom.xml
+++ b/usage/dist/pom.xml
@@ -118,12 +118,12 @@
                         <configuration>
                             <artifacts>
                                 <artifact>
-                                    <file>${project.build.directory}/brooklyn-${project.version}-dist.tar.gz</file>
+                                    <file>${project.build.directory}/apache-brooklyn-${project.version}-dist.tar.gz</file>
                                     <type>tar.gz</type>
                                     <classifier>dist</classifier>
                                 </artifact>
                                 <artifact>
-                                    <file>${project.build.directory}/brooklyn-${project.version}-dist.zip</file>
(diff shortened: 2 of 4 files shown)


usage/camp/pom.xml
```diff
@@ -59,7 +59,13 @@
...
             <artifactId>camp-base</artifactId>
             <version>${project.version}</version>
-        </dependency>
+            <exclusions>
+                <!-- Dependency versions mismatch between transitive dependencies, declare explicitly -->
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
+            </exclusions>        </dependency>
```
neykov: Formatting

ahgittin:
I will look to address the `attach-artifact` goal in #705

usage/dist/pom.xml
```diff
@@ -118,12 +118,12 @@
...
                             <artifacts>
                                 <artifact>
-                                    <file>${project.build.directory}/brooklyn-${project.version}-dist.tar.gz</file>
+                                    <file>${project.build.directory}/apache-brooklyn-${project.version}-dist.tar.gz</file>
                                     <type>tar.gz</type>
                                     <classifier>dist</classifier>
                                 </artifact>
                                 <artifact>
-                                    <file>${project.build.directory}/brooklyn-${project.version}-dist.zip</file>
+                                    <file>${project.build.directory}/apache-brooklyn-${project.version}-dist.zip</file>
```
ahgittin: Good addition.  However this will break documentation.

ahgittin:
Fixed the formatting, and merging so I can resolve conflicts with #705 around apache name addition.

Note @rdowner we must update the docs to point at `apache-*` for the dist builds.

ahgittin:
I don't understand how this PR managed to build, unless it only does `mvn clean compile test`.  When I try a `mvn clean install` (after having merged :( ) I get:

    [ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:2.5.2:install (default-install) on project brooklyn-dist: Failed to install artifact org.apache.brooklyn:brooklyn-dist:tar.gz:dist:0.7.0-SNAPSHOT: /Users/alex/Data/cloudsoft/dev/gits/brooklyn/usage/dist/target/apache-brooklyn-0.7.0-SNAPSHOT-dist.tar.gz (No such file or directory) -> [Help 1]

And this makes sense as I don't think maven allows us to change the name of the files attached, they always follow `${projectName}-${version}-${classifier}.${ext}` .

I'm going to remove the nice-but-maybe-impossible `apache-` prefix to the dist archives from #705 and merge it else I worry the build might be breaking.

ahgittin:
If the intention is to make a file `apache-brooklyn-${version}-dist.${ext}` (which would be a good thing IMO) maybe we should just change the project name from `brooklyn-dist` to `apache-brooklyn`?

neykov:
I believe the file name can be changed directly in the assembly plugin with the `finalName` config. Setting it to `apache-${project.build.finalName}` should do the trick.

PR #710, opened 2015-06-23
Remove obsolete Tomcat fail binding test

bostko:
Basically the behavior was that it was expecting `LocalhostMachineProvisioningLocation.obtainSpecificPort` to throw an exception "port for httpPort is null" but instead currently it is silent so it causes other exceptions which breaks the test (Freemarker is complaining for missing template value driver.httpPort)
I discussed it with Svet and we decided to delete those.
Please share your thoughts and tell if you think we can add any other relevant tests.


neykov:
Good to merge. Even when created the test wasn't exercising Tomcat behaviour, but config to attribute convertion + localhost specifics.

mikezaccardo:
LGTM as well

PR #711, opened 2015-06-23
Describes AWS VPC issue

duncangrant:
This describes the issue discussed on #705
https://github.com/apache/incubator-brooklyn/pull/705


ahgittin:
Great adding this to the docs, thanks @duncangrant .

This does clutter the main page so I think we ought to do a further refactoring adding some cloud-specific pages: this, and openstack as there's a lot of info in the template `brooklyn.properties` which would be better described there, and softlayer where there's a lot of useful hidden capabilities (privateNetworkOnlyFlag, primaryNetworkComponentNetworkVlanId, primaryBackendNetworkComponentNetworkVlanId, postInstallScriptUri -- all in the `SoftlayerTemplateOptions`).

Not needed for this PR though - merging.

PR #713, opened 2015-06-24
Fixing Integration test - RubyRepIntegrationTest

ygy:


neykov:
@ygy Looks like `MySqlIntegrationTest.CREATION_SCRIPT` is used in a number of places, apart from this test, not all of which set the table name. Could you add the config key there as well. Then good to merge.

ygy:
@neykov Config added where ```MySqlIntegrationTest.CREATION_SCRIPT``` is used

PR #714, opened 2015-06-24
Clarify operation of `templateOptions` config key

rdowner:
Follow on for PR #639; that PR has not been merged, but it's only commit snuck into another PR and has actually been merged. The intention of this PR is to address the review comments.

Resolves the ambiguity around single parameters of type list. Extracts the bulk of the code into a new standalone class and adds tests. Adds documentation.

Tested with this blueprint:

```yaml
location: AWS_eu-west-1
services:
- type: brooklyn.entity.basic.EmptySoftwareProcess
  provisioningProperties:
    templateOptions:
      subnetId: subnet-041c8373
      securityGroupIds: ['sg-4db68928']
      mapNewVolumeToDeviceName: ["/dev/sda1", 100, true]
```

...and verified that all three parameters (each of which exercise a different code path) had been applied.

### DIFF
--- a/core/src/main/java/brooklyn/util/flags/MethodCoercions.java
+++ b/core/src/main/java/brooklyn/util/flags/MethodCoercions.java
@@ -1,0 +1,183 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package brooklyn.util.flags;
+
+import brooklyn.util.exceptions.Exceptions;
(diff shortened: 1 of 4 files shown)


core/src/main/java/brooklyn/util/flags/MethodCoercions.java
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * Returns a predicate that matches a method with the given name, and a single parameter that
+     * {@link brooklyn.util.flags.TypeCoercions#tryCoerce(Object, com.google.common.reflect.TypeToken)} can process
+     * from the given argument.
+     *
+     * @param methodName name of the method
+     * @param argument argument that is intended to be given
+     * @return a predicate that will match a compatible method
+     */
+    public static Predicate<Method> matchSingleParameterMethod(final String methodName, final Object argument) {
```
mikezaccardo: Could be good to add `@Nonnull` to the params.

core/src/main/java/brooklyn/util/flags/MethodCoercions.java
```diff
@@ -0,0 +1,183 @@
...
+     */
+    public static Maybe<?> tryFindAndInvokeBestMatchingMethod(final Object instance, final String methodName, final Object argument) {
+        if (argument instanceof List) {
+            List<?> arguments = (List<?>) argument;
+
+            // ambiguous case: we can't tell if the user is using the multi-parameter syntax, or the single-parameter
+            // syntax for a method which takes a List parameter. So we try one, then fall back to the other.
+
+            Maybe<?> maybe = tryFindAndInvokeMultiParameterMethod(instance, methodName, arguments);
+            if (maybe.isAbsent())
```
mikezaccardo: Personal preference: I like the conciseness of a ternary, so I would do something like:

`return (maybe.isPresent()) ? maybe : tryFindAndInvokeSingleParameterMethod(instance, methodName, argument);`

mikezaccardo:
All tests pass, LGTM

PR #717, opened 2015-06-25
Fixes issue where port was not being opened if port range was overridden

nakomis:
The following YAML was failing to open the port (8080 or 9090) in the AWS security group

```
location: aws-ec2:eu-west-1
services:
  brooklyn.config:
    http.port: 9090+
```

### DIFF
--- a/software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
+++ b/software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
@@ -18,6 +18,8 @@
  */
 package brooklyn.entity.basic;
 
+import brooklyn.util.flags.TypeCoercions;
+import brooklyn.util.guava.Maybe;
 import groovy.time.TimeDuration;
 
 import java.util.Collection;
@@ -66,6 +68,7 @@ import com.google.common.base.Functions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
+import com.google.common.reflect.TypeToken;
 
 /**
  * An {@link Entity} representing a piece of software which can be installed, run, and controlled.
@@ -448,15 +451,27 @@ public abstract class SoftwareProcessImpl extends AbstractEntity implements Soft
         Set<ConfigKey<?>> configKeys = Sets.newHashSet(allConfig.keySet());
         configKeys.addAll(getEntityType().getConfigKeys());
 
+        /* TODO: This won't work if there's a port collision, which will cause the corresponding port attribute
+           to be incremented until a free port is found. In that case the entity will use the free port, but the
+           firewall will open the initial port instead. Mostly a problem for SameServerEntity, localhost location.
+         */
         for (ConfigKey<?> k: configKeys) {
-            if (PortRange.class.isAssignableFrom(k.getType())) {
-                PortRange p = (PortRange)getConfig(k);
+            Object value;
+            if (PortRange.class.isAssignableFrom(k.getType()) || k.getName().matches(".*\\.port")) {
+                value = config().get(k);
+            } else {
+                // config().get() will cause this to block until all config has been resolved
+                // using config().getRaw(k) means that we won't be able to use e.g. 'http.port: $brooklyn:component("x").attributeWhenReady("foo")'
+                // but that's unlikely to be used
+                Maybe<Object> maybeValue = config().getRaw(k);
+                value = maybeValue.isPresent() ? maybeValue.get() : null;
+            }
+
+            Maybe<PortRange> maybePortRange = TypeCoercions.tryCoerce(value, new TypeToken<PortRange>() {});
+
+            if (maybePortRange.isPresentAndNonNull()) {
+                PortRange p = maybePortRange.get();
                 if (p != null && !p.isEmpty()) ports.add(p.iterator().next());
-            } else if(k.getName().matches(".*\\.port")){
-                Object value = getConfig(k);
-                if (value instanceof Integer){
-                    ports.add((Integer)value);
-                }
             }
         }        
         
--- a/software/base/src/main/java/brooklyn/entity/basic/SameServerDriverLifecycleEffectorTasks.java
+++ b/software/base/src/main/java/brooklyn/entity/basic/SameServerDriverLifecycleEffectorTasks.java
@@ -29,6 +29,7 @@ import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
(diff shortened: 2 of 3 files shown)


software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
```diff
@@ -456,6 +456,9 @@ protected final void callStartHooks() {}
                 Object value = getConfig(k);
                 if (value instanceof Integer){
                     ports.add((Integer)value);
+                } else if (PortRange.class.isAssignableFrom(value.getClass())) {
```
robertgmoss: Why isn't this check picked up by line 452, would it be better to fix that?

nakomis: 452 checks the class of the key's type, then checks assignability of the value from getConfig. If you use `entityspec.configure(Attributes.HTTP_PORT, "9999+")` then it will be picked up on line 452, however if you use `entityspec.configure("http.port"), which is effectively what happens when you put "http.port" in the YAML, then it creates a new `ConfigKey<Object>` which is not picked up on line 452

robertgmoss:
ok.  But still better to pick it up on all keys that are a `PortRange`, not just those ending `.port`, right?

software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
robertgmoss: ok. But still better to pick it up on all keys that are a PortRange, not just those ending .port, right?

nakomis: Ah, good point. My `else if` is in the wrong place

nakomis:
Comments incorporated

robertgmoss:
LGTM

software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
```diff
@@ -456,6 +456,9 @@ protected final void callStartHooks() {}
...
+                    PortRange p = (PortRange)value;
+                    if (p != null && !p.isEmpty()) ports.add(p.iterator().next());
```
neykov: This won't work if there's a port collision, which will cause the corresponding port attribute to be incremented until a free port is found. In that case the entity will use the free port, but the firewall will open the initial port instead. Mostly a problem for `SameServerEntity`, `localhost` location. As you've copied the existing logic I am fine with keeping it as is, but worth adding a comment.

There's identical logic in `SameServerDriverLifecycleEffectorTasks` so better update it there as well.

software/base/src/main/java/brooklyn/entity/basic/SoftwareProcessImpl.java
```diff
@@ -449,11 +449,14 @@ protected final void callStartHooks() {}
         configKeys.addAll(getEntityType().getConfigKeys());
 
         for (ConfigKey<?> k: configKeys) {
+            Object value = getConfig(k);
             if (PortRange.class.isAssignableFrom(k.getType())) {
                 PortRange p = (PortRange)getConfig(k);
                 if (p != null && !p.isEmpty()) ports.add(p.iterator().next());
+            } else if (value != null && PortRange.class.isAssignableFrom(value.getClass())) {
```
robertgmoss: nitpick: add the first if condition as an or to the second to remove duplicate code

nakomis: Thinking about it, I'm not sure that the first if condition is needed at all. If the key's type is assignable to PortRange, then the value must be too (or null)

nakomis:
The failing tests are related to these changes. The call to getRequiredOpenPorts blocks until all config is resolvable. Will switch to config().getRaw(k)

nakomis:
Ok, comments incorporated, and tests are passing. I've had to switch to `config().getRaw(k)`, but I've added a comment to explain why

neykov:
The previous implementation had better behaviour where it would resolve only port config keys. You can keep the old behaviour, while introducing a new case where the value is `PortRange` and doesn't need resolving. Also for the case where the value is already available, but wrapped in a future, you can use ValueResolver with a 0 timeout.

Also the `SameServerDriverLifecycleEffectorTasks` doesn't use `getRaw`.

nakomis:
Using `http.port: 9999+` in the YAML (akin to using `entity.configure("http.port", "9999+")` as opposed to `entity.configure(Attributes.HTTP_PORT, "9999+")`, like I used in the unit test that I added) will create a `ConfigKey<Object>` and the previous implementation did recognise it as a port range. To determine that it is a port range, you must either resolve the value (which causes the thread to hang if any of them are futures etc, breaking `SoftwareProcessEntityLatchTest`), or use `getRaw`

I'd intended to update `SameServerDriverLifecycleEffectorTasks` to use `getRaw`, but overlooked that. If you agree the right approach is to use `getRaw`, I'll update `SameServerDriverLifecycleEffectorTasks`

neykov:
I agree that the changes are useful, probably didn't explain clearly what I mean in the previous comment.
The current implementation will not resolve the value even if we know that it is a port range (from the config key type). Something in the spirit of your initial commit will keep existing behaviour, while still getting us the benefit from your changes:
```
if config key type is port range {
    getConfig(k) - resolve value, we know it's a port range
} else if config key contains ".port" {
   getConfig(k) - resolve value
} else {
  v = getConfigRaw(k) - doesn't resolve
  if (v instanceof PortRange) {
  
  }
}
```

Not important for this PR, but in future we could go one step further and try to resolve the value with a 0 or a very short timeout so even more cases are covered.


wdyt?

nakomis:
Right, I'd misunderstood your previous comment. I'll make the changes as you suggested, and update `SameServerDriverLifecycleEffectorTasks` too

ahgittin:
@nakomis @neykov why not just   `TypeCoercions(getRaw(PORT_KEY), PortRange.class)` ?

this should convert `int` to `PortRange` and then iterate, which might be slightly wasteful but worth it as it's simpler.

also @nakomis note uses of `Tasks.resolving(...)` -- resolving config is normally non-blocking but if blocking is a problem you can do `Tasks.resolving(<raw config>).timeout(ValueResolver.REAL_QUICK_WAIT)...`

rdowner:
@nakomis what's next for this PR? Are you still adding changes, or do you think it's ready and it should be reviewed again? Thanks :smile:

nakomis:
I think it's ready to be reviewed again

neykov:
Looks great, can be merged.

robertgmoss:
LGTM too

rdowner:
Thanks all! :+1:

PR #718, opened 2015-06-25
Prevents Location Events being reported to metering for temporary locations

nakomis:


ahgittin:
yuck, we should revisit this -- remove the whole idea of the registry and temporary locations and treat locations as entities or else something transient -- but while we have this bad model this seems a fair enough workaround.  merging.

PR #719, opened 2015-06-25
More integration test fixes

neykov:
* Don't unmanage `BrooklynNode` until the STOP effector and all parent entity tasks complete. Prevents `CancellationException: null` error when the STOP effector is called from another effector (i.e. `STOP_NODE_AND_KILL_APPS`)
* AutoScaler now has a default min of 1 member
* Wait for all apps to be running before continuing tests in `BrooklynNodeIntegrationTest` (not just managed)
* Don't use remote services when a local one will do, avoids quota exceeded failures


ahgittin:
nice - merging

PR #721, opened 2015-06-26
Feature: byon customization

aledsage:
Allows a BYON location to have more complex configuration (e.g. each machine having its own separate map of config), including in YAML.

Also supports a MACHINE_CHOOSER config. This can be used to customize the choice of which available machine should be returned, when obtain() is called.

The motivation is to make BYON more useful for multi-VM deployments, where you want to pick particular machines for particular software components.

### DIFF
--- a/core/src/main/java/brooklyn/location/basic/ByonLocationResolver.java
+++ b/core/src/main/java/brooklyn/location/basic/ByonLocationResolver.java
@@ -18,6 +18,8 @@
  */
 package brooklyn.location.basic;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 import java.net.InetAddress;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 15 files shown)


hzbarcea:
+1 good to go

ahgittin:
@aledsage Do we want to add any documentation for this?

And it looks like any attempt to use `MachineChooser` is going to require writing java, is that right?  Is there a way we could encourage YAML for this instead?  We could define `tags` as config on BYON and then if `MACHINE_CHOOSER` is a map with no type the default instance will expect a `requiredTags` string or list.  (This could also allow us to define a "FYON" -- find your own nodes -- where we point it at all cloud instances with some tags, and it imports other tags which we can use to give specific machines to specific roles.)

aledsage:
@ahgittin good point about docs - will add some!

For `MachineChooser`, you're right that it would currently either have to write Java, or use existing YAML mechanisms such as `$brooklyn:object`.

In the spirit of getting the Java API right, and then ensuring that we can cleanly express things in YAML, do you think the Java API is right? Should we use `Function<Iterable<? extends MachineLocation>, MachineLocation>`, or would we be better using a specific type (e.g. so there is more opportunity to supply `TypeCoercion` code)? If the Java API is ok, and won't make YAML support hard to add, then shall we merge this PR?

Your suggested YAML behaviour sounds good and interesting. However, I'd be hesitant to implement a default behaviour (which we'd then have to support for quite a while) in this PR. There are other valid options, such as matching on a config key (rather than on a tag's value), which is what the customer driving this feature is proposing. I do like the way the tags approach can translate to jclouds-byon, where the tags are on the VMs themselves.

How about we take it to the dev list to discuss what we'd want to support as "native" yaml.

aledsage:
I've added docs to `docs/guide/ops/locations/index.md`.

aledsage:
Merging now, into 0.8.0-SNAPSHOT. We can revisit this, to make it more easily used by yaml. We can change the Java interface for this if required.

PR #724, opened 2015-06-27
Remove unnecessary use of snapshot repositories

hzbarcea:


hzbarcea:
While this fix is necessary, I am not sure if it's sufficient.

hzbarcea:
This pull request included a patch that changed the parent in usage/downstream-parent/pom.xml to change the parent from oss.sonatype to apache.

Unfortunately, while it is something that needs to be done, it's not as simple as just changing the parent. I'll have to come back with a better, separate patch for that.

This PR can be merged though as it does not depend on that fix.

ahgittin:
@hzbarcea what's the problem with changing the parent pom of "downstream"?

and when you say "this is necessary", do you mean it's an apache requirement for a release?  (that makes sense...)

finally -- can you put the snapshot repos into the archetype's sample pom, commented out, but at least this gives an easy way for people to enable them if required?  (it is a common desire.)

ahgittin:
BTW once the repos are in a comment in the archetype sample pom, i think this is good to merge.

And I think we want to **backport to the 0.7.0** release branch, with versions adjusted as appropriate if required (possibly not required).

hzbarcea:
I see the point about the snapshots repo added as a comment in the archetypes. 

Depending on what we decide to do with the downstream-parent however, it may not be necessary, because if the generated archetype ultimately derives from org.apache:apache:17 there is a snapshot repository <id>apache.snapshots.https</id> in there.

Do you prefer to add the snapshot repo as a comment anyway?

hzbarcea:
I added the ref to the snapshot repository, although after fixing the poms it'll be unnecessary.

sjcorbett:
@hzbarcea Thanks.

PR #726, opened 2015-06-29
Adds sensor path to performance counter output

nakomis:


core/src/main/java/brooklyn/event/feed/windows/WindowsPerformanceCounterFeed.java
```diff
@@ -275,20 +286,30 @@ public boolean checkSuccess(WinRmToolResponse val) {
 
         @Override
         public void onSuccess(WinRmToolResponse val) {
```
aledsage: Is it possible to write a unit test for this?

nakomis:
Test added, as requested

aledsage:
Looks good; merging.

PR #727, opened 2015-06-30
[BROOKLYN-132] Add versions dropdown when creating applications from catalog template

tbouron:
This PR adds a versions dropdown on the second step modal, when an application is created from a template. This dropdown is always visible for consistency, even if there is only one version available.

Please note that as I was not sure that the `catalog-item` model could apply to any object returned by the `/v1/catalog/*` endpoints so I created a specific one for the applications only: `catalog-application`. It also makes sense as one backbone model should have only one endpoint.

### DIFF
--- a/usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
+++ b/usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
@@ -22,7 +22,7 @@
  */
 define([
     "underscore", "jquery", "backbone", "brooklyn-utils", "js-yaml",
-    "model/entity", "model/application", "model/location",
+    "model/entity", "model/application", "model/location", "model/catalog-application",
     "text!tpl/app-add-wizard/modal-wizard.html",
     "text!tpl/app-add-wizard/create.html",
     "text!tpl/app-add-wizard/create-step-template-entry.html", 
@@ -30,14 +30,15 @@ define([
     "text!tpl/app-add-wizard/required-config-entry.html",
     "text!tpl/app-add-wizard/edit-config-entry.html",
     "text!tpl/app-add-wizard/deploy.html",
+    "text!tpl/app-add-wizard/deploy-version-option.html",
     "text!tpl/app-add-wizard/deploy-location-row.html",
     "text!tpl/app-add-wizard/deploy-location-option.html",
     "bootstrap"
     
-], function (_, $, Backbone, Util, JsYaml, Entity, Application, Location,
+], function (_, $, Backbone, Util, JsYaml, Entity, Application, Location, CatalogApplication,
              ModalHtml, CreateHtml, CreateStepTemplateEntryHtml, CreateEntityEntryHtml,
              RequiredConfigEntryHtml, EditConfigEntryHtml, DeployHtml,
-             DeployLocationRowHtml, DeployLocationOptionHtml
+             DeployVersionOptionHtml, DeployLocationRowHtml, DeployLocationOptionHtml
 ) {
 
     /** Special ID to indicate that no locations will be provided when starting the server. */
@@ -56,7 +57,7 @@ define([
     function oldSpecToCamp(spec) {
         var services;
         if (spec.type) {
-            services = [entityToCamp({type: spec.type, config: spec.config})];
+            services = [entityToCamp({type: spec.type, version: spec.version, config: spec.config})];
         } else if (spec.entities) {
             services = [];
             var entities = spec.entities;
@@ -80,6 +81,7 @@ define([
         var result = {};
         if (entity.name && (!options || !options.exclude_name)) result.name = entity.name;
         if (entity.type) result.type = entity.type;
+        if (entity.type && entity.version) result.type += ":" + entity.version;
         if (entity.config && _.size(entity.config)) result["brooklyn.config"] = entity.config;
         return result;
     }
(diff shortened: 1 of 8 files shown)


ahgittin:
good fix -- showing a tile for each version in the wizard was awful

however i don't like adding dropdowns etc unless we really need them.  arguably the wizard interface never asks you (use yaml if you want a non-standard version) -- but anything is an improvement over the one-tile-per-version!

can we make the dropdown so it only shows if there is more than one (that should be easy, no?), and it's populated with the default?

also note that the default is the highest **non-snapshot** version -- `CatalogItemComparator` sorts this, so you should be able just to take the first in the list.

usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -105,15 +109,15 @@ define([
...
+                              view:new ModalWizard.StepCreate({ model:this.model, wizard: this, catalog: this.catalog })
                           },
                           {
                               // TODO rather than make this another step -- since we now on preview revert to the yaml tab
                               // this should probably be shown in the catalog tab, replacing the other contents.
                               step_id:'name-and-locations',
                               title:'<%= appName %>',
-                              instructions:'Specify the locations to deploy to and any additional configuration',
-                              view:new ModalWizard.StepDeploy({ model:this.model })
+                              instructions:'Specify the application version, locations to deploy to and any additional configuration',
```
ahgittin: it's not the version of the actual *application* -- it's the version of the blueprint -- so this wording would be confusing.

i suggest leave as it was, simplify -- we don't always show version, and if we do it's obvious enough what it is.  in time we could have a hover-help if it's an issue.

usage/jsgui/src/main/webapp/assets/tpl/app-add-wizard/deploy-version-option.html
```diff
@@ -0,0 +1,23 @@
...
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<option value="<%= version %>">
+    <span class="provider"><%= version %></span>
```
sjcorbett: Versions can come from humans so both of these should be escaped with `<%- .. %>`.

usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
tbouron: True, I'll revert that to what it was.

usage/jsgui/src/main/webapp/assets/tpl/app-add-wizard/deploy-version-option.html
tbouron: True that, my old JSP habit gets in a way.

tbouron:
@ahgittin The version sorting is already done as per as what you want. Here is an example:

screen shot 2015-06-30 at 10 49 26

I can easily hide the dropdown if there is only version. I'll push these changes ASAP

sjcorbett:
To bikeshed a little, how about disabling the dropdown rather than hiding it when there's only one entry?

tbouron:
@sjcorbett: I don't have any preference in that matter. But between having a disabled dropdown and no having one, I would prefer the second option. Does it make sense to add a `select` on this tiny modal if you cannot interact with it?

sjcorbett:
Personally I think it's more consistent to always keep the element. Will users understand why it has disappeared? Anyway, just an opinion.

tbouron:
@sjcorbett A valuable one! @ahgittin, I guess you have the final say on that?

ahgittin:
most people aren't using versions and it would confuse/distract them so i think better to hide it in that vast majority case

ps - @tbouron i don't have any special say in the matter

ahgittin:
@neykov @tbouron following from IRC discussion i like the idea that:

* for multiple versions, show a drop down
* for any single version, the user cannot interact but it *might* be useful to show the version
  * for a single version `0.0.0-SNAPSHOT` (the default) it isn't useful, so don't show anything
  * for any other single version, *do* show the version, for information

the last one is optional in my mind but useful

tbouron:
@ahgittin I really like this idea too. I'll make the changes and commit ASAP

tbouron:
Done.

neykov:
Tried to test the functionality, but couldn't see the version selector.
In the app selection window the template appears, tagged with the latest version:
screen shot 2015-07-09 at 13 36 52

After selecting the template next window doesn't have version info:
screen shot 2015-07-09 at 13 37 01

Once done either squash the commits or add subjects/description to each logically separate commit.

alasdairhodge:
I'd suggest always including the version, even the uninteresting default. This communicates to the user that versioning is suppoprted (and encouraged), that a default is provided if none is specified, and reinforces the location of version info in the UI. It needn't be prominent, but should be present in a consistent place.

tbouron:
@neykov: I tried again my branch https://github.com/tbouron/incubator-brooklyn/tree/app-wizard and it works as expected. Are you sure you checked out the correct branch?

neykov:
Looks like I got bitten by https://github.com/apache/incubator-brooklyn/pull/753. After a handful of rebuilds I can finally see the dropdown.

Good to merge, but better squash the commits first (see comment from above).

tbouron:
Commit squashed

aledsage:
Thanks @tbouron - and thanks to all the reviewers! Merging now.

PR #729, opened 2015-06-30
[BROOKLYN-153] Update the REST API documentation for the catalog endpoints

tbouron:


sjcorbett:
Is "type" the right word for the documentation? It has too many meanings to be immediately understandable in this context. The methods the arguments are passed to use "symbolicName" (in brooklyn.catalog.BrooklynCatalog). Are either better than just "id"?

It's also worth changing the names of the parameters of the implementing methods in `CatalogResource` to match the interface.

tbouron:
`type` is for me the best word here, that's why I used it. I didn't really understand `symbolicName` at first, not to mention that symbolicName represents something unique to me, like an ID. 

FYI, `symbolicName` is actually redundant as the API returns both `type` and `symbolicName` attributes (which have the same value every time)
 
That being said, I do agree that the `brooklyn.catalog.BrooklynCatalog` method's parameters need to reflect whatever name we choose for the API.

neykov:
The user-visible term for catalog item is `id`. This is what the user sets in the catalog spec. The technical term is `symbolicName` with `type` kept for backwards compatibility with existing clients (it was removed initially but we had to put it back in).

tbouron:
@neykov: Right, I was not aware of that. I guess I'm confused because we use two different terms internally and externally.

Anyway, is everyone ok if I change `type` by `symbolicName` then?

neykov:
`symbolicName` makes sense as then the urls will become `.../{symbolicName}/{version}` which is consistent with the internal representation.

tbouron:
Done. I also updated parameters' names within `CatalogResource` as @sjcorbett suggested.

ahgittin:
`symbolicName` is the right unambiguous terminology -- good improvement

usage/rest-api/src/main/java/brooklyn/rest/api/CatalogApi.java
```diff
@@ -180,15 +180,15 @@ public CatalogEntitySummary getEntity(
...
     @ApiErrors(value = {
         @ApiError(code = 404, reason = "Entity not found")
     })
     public CatalogEntitySummary getEntity(
-        @ApiParam(name = "entityId", value = "The ID of the entity or template to retrieve", required = true)
-        @PathParam("entityId") String entityId,
-        
+        @ApiParam(name = "symbolicName", value = "The symbolic name of the entity or template to retrieve", required = true)
+        @PathParam("symbolicName") String symbolicName,
+
```
sjcorbett: Surplus line

usage/rest-api/src/main/java/brooklyn/rest/api/CatalogApi.java
```diff
@@ -202,18 +202,18 @@ public CatalogEntitySummary getEntity(
...
     @ApiErrors(value = {
         @ApiError(code = 404, reason = "Entity not found")
     })
     public CatalogEntitySummary getApplication(
-        @ApiParam(name = "applicationId", value = "The ID of the application to retrieve", required = true)
-        @PathParam("applicationId") String entityId,
-        
+        @ApiParam(name = "symbolicName", value = "The symbolic name of the application to retrieve", required = true)
+        @PathParam("symbolicName") String symbolicName,
+
```
sjcorbett: And here

sjcorbett: I guess you're just following the existing format? Doesn't matter, will merge.

tbouron: Strange, probably my Intellij not setup as it should be.

neykov:
good to merge

neykov:
@tbouron Could you close this.

tbouron:
Done

PR #730, opened 2015-07-01
BROOKLYN-154: Don't call methods that might not be there

rdowner:
This code attempted to call non-standard extensions on the `javax.ws.rs.core.Response` class. This change does the appropriate `instanceof` checks to ensure that no non-standard methods are called if
we have a plain old Response object.

### DIFF
--- a/usage/rest-client/src/main/java/brooklyn/util/http/BuiltResponsePreservingError.java
+++ b/usage/rest-client/src/main/java/brooklyn/util/http/BuiltResponsePreservingError.java
@@ -22,6 +22,7 @@ import java.lang.annotation.Annotation;
 
 import javax.ws.rs.core.Response;
 
+import org.jboss.resteasy.client.core.BaseClientResponse;
 import org.jboss.resteasy.core.Headers;
 import org.jboss.resteasy.specimpl.BuiltResponse;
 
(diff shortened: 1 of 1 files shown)


usage/rest-client/src/main/java/brooklyn/util/http/BuiltResponsePreservingError.java
```diff
@@ -62,7 +64,8 @@ public BuiltResponsePreservingError(int status, Headers<Object> headers, Object
             Exceptions.propagateIfFatal(e);
             return new BuiltResponsePreservingError(status, headers, entity, new Annotation[0], e);
         } finally {
-            source.close();
+            if (source instanceof BaseClientResponse)
```
aledsage: Why only call `close()` if it's an instance of `BaseClientResponse`? The close method is on `javax.ws.rs.core.Response`, so should we always call close?

aledsage:
Merging this - @rdowner fixed this during testing of a downstream project. I presume there is good reason to guard the `close()` call in the finally block. We can add it back in if desired!

ahgittin:
What a mess that these deps include their own javax classes, and that they are different!  I'd love to get a fix for that root problem, as even with this fix we seem at risk of similar errors, but I don't see how to do that easily.

This PR helps, and while we could reflectively looked for a `close()` method that seems like a worse band-aid, if we know that the `BaseClientResponse` is always what the resteasy API impl supplies then I'm fine with this.

usage/rest-client/src/main/java/brooklyn/util/http/BuiltResponsePreservingError.java
rdowner: @aledsage the thing is, there *isn't* a `close()` method on `Response` - at least not officially. There are multiple different versions of `Response` on the classpath and sometimes the "wrong" one (one without a `close()` method) ends up at the front. See the Jira issue for full details.

PR #731, opened 2015-07-01
Count the null values for the average enrichers

bostko:
This fixes TomcatAutoScalerPolicyTest

Basically the problem is that after making a request to Tomcat, update event for REQUEST_COUNT_PER_NODE is triggered immediately.
This causes Enrichers.average() to have a null value for the requests of the newly created Tomcat Entity.
This leads to average of 2 requests per node which exceeds the allowed metric of 0 to 1 which triggers scaling.

It is discuss-able whether we should apply the fix here, but I think in the general case it is good to include count the null when calculating average enrichers.

### DIFF
--- a/software/webapp/src/main/java/brooklyn/entity/webapp/DynamicWebAppClusterImpl.java
+++ b/software/webapp/src/main/java/brooklyn/entity/webapp/DynamicWebAppClusterImpl.java
@@ -117,6 +117,7 @@ public class DynamicWebAppClusterImpl extends DynamicClusterImpl implements Dyna
                     .publishing(average)
                     .fromMembers()
                     .computingAverage()
+                    .defaultValueForUnreportedSensors(0)
                     .build());
         }
     }


bostko:
Svet suggested that everywhere we should use 0 as a default null value for the average enricher because we could often have situations where some of the entities isn't up.
One way I can think of for doing this is if we change the argument `valueToReportIfNoSensors` to `0` in Enrichers.java:268 but if we do so we have to call `defaultValueForUnreportedSensors(null)` on the enricher to get to the old behavior.

Basically we should consider how the null values will be treated especially when binding a Scaling policy to aggregated sensor.

sjcorbett:
One for the mailing list?

software/nosql/src/test/java/brooklyn/entity/nosql/cassandra/CassandraDatacenterTest.java
```diff
@@ -74,6 +74,7 @@ public void testPopulatesInitialSeeds() throws Exception {
                 .configure(CassandraDatacenter.MEMBER_SPEC, EntitySpec.create(EmptySoftwareProcess.class)));
 
         app.start(ImmutableList.of(loc));
+        Thread.sleep(Duration.seconds(1).toMilliseconds());
```
neykov: Why the sleep?

neykov:
Having looked at this in more detail, I still think that the averaging enricher should count the null values by default. Perhaps it could also look at the entity state and exclude stopped entities.

I agree with @sjcorbett  that the place to discuss this is the dev mailing list, @bostko could you start a thread there.

neykov:
LGTM - can be merged regardless of what we decide on the default value for the averaging enricher.

aledsage:
Thanks all - merging now.

PR #732, opened 2015-07-03
MongoDB fixes

ygy:
Probably resolves problems in local and/or cloud environments.

It seems that ```BrooklynAccessUtils``` is the preferred way to determine Brooklyn's accessible address. Now it uses hostname when used with locations in local or cloud environments. On some occasions it should use subnet_hostname instead of the hostname.
```MongoDBShardedDeploymentIntegrationTest``` demonstrates the scenario.

### DIFF
--- a/core/src/main/java/brooklyn/location/basic/LocalhostMachineProvisioningLocation.java
+++ b/core/src/main/java/brooklyn/location/basic/LocalhostMachineProvisioningLocation.java
@@ -238,7 +238,7 @@ public class LocalhostMachineProvisioningLocation extends FixedListMachineProvis
             releasePort(null, p);
     }
     
-    public static class LocalhostMachine extends SshMachineLocation {
+    public static class LocalhostMachine extends SshMachineLocation implements HasSubnetHostname {
         // declaring this here (as well as on LocalhostMachineProvisioningLocation) because:
         //  1. machine.getConfig(key) will not inherit default value of machine.getParent()'s key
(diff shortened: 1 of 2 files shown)


bostko:
I'd rather change the address in the `MongoDBShardClusterImpl`.
Also I think the getBrooklynAccessibleAddress is used for defining the public url so changing it to return SUBNET_HOSTNAME would return the opposite - the private hostname

neykov:
After discussion:

The problem is that `addShard` and `--configdb` are supposed to have the same values, matching textually. Can't have hostname at one place, IP at the other. Atm `--configdb` uses `getAttribute(MongoDBConfigServer.SUBNET_HOSTNAME)`, while `addShard` uses the public `MONGO_SERVER_ENDPOINT` initialized with `BrooklynAccessUtils.getBrooklynAccessibleAddress`. The solution would be to use `addShard` with `SUBNET_HOSTNAME` as well. Shouldn't matter for cloud machines (where the values match), but for localhost they are different.

Optional improvements:
  * update documentation of `MONGO_SERVER_ENDPOINT` and `HTTP_INTERFACE_URL` to indicate that they are public IPs (better not be used for local connections)
  * Implement `HasSubnetHostname` for `LocalhostMachineProvisioningLocation` returning `localhost`, so that `Machines.getSubnetHostname` doesn't return `127.0.0.1` (for consistence).

neykov:
Good fixes, can be merged.

aledsage:
Thanks @ygy - merging.

PR #734, opened 2015-07-03
Typo fix.

conet:
The withespace at the end-of-line should also be unnecessary. How often is the site updated? I'm asking because another typo fixed on Mon Feb 16 16:05:43 2015 in 716a8923b2355a9085610dd12b8ba7089b331ae5 is still on the site.

### DIFF
--- a/docs/guide/start/running.md
+++ b/docs/guide/start/running.md
@@ -33,7 +33,7 @@ This will create a `brooklyn-{{ site.brooklyn-version }}` folder.
 
 **Note**: You'll need a Java JRE or SDK installed (version 1.7 or later), as Brooklyn is Java under the covers.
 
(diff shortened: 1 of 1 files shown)


sjcorbett:
@conet Thanks. Not sure about website update regularity. @ahgittin?

ahgittin:
@conet The website is normally updated every month or two.  But the *guide* section of the website is tied to specific versions and is only updated for important (and backported) docs -- ie less often.  The fix in https://github.com/apache/incubator-brooklyn/commit/716a8923b2355a9085610dd12b8ba7089b331ae5 is shown in the *SNAPSHOT* guide at https://brooklyn.incubator.apache.org/v/0.7.0-SNAPSHOT/start/running.html and will be in the next release version.

conet:
@ahgittin Thanks for the explanation, that makes sense.

PR #735, opened 2015-07-07
SoftwareProcess exposes LifecycleEffectorTasks implementation as config

sjcorbett:
Thus it is much easier for subclasses of `SoftwareProcess` to customise which effectors they expose and how those effectors behave.

### DIFF
--- a/software/webapp/src/test/java/brooklyn/test/entity/TestJavaWebAppEntity.java
+++ b/software/webapp/src/test/java/brooklyn/test/entity/TestJavaWebAppEntity.java
@@ -18,9 +18,20 @@
  */
 package brooklyn.test.entity;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import brooklyn.config.ConfigKey;
+import brooklyn.entity.basic.Attributes;
+import brooklyn.entity.basic.ConfigKeys;
+import brooklyn.entity.basic.Lifecycle;
+import brooklyn.entity.basic.ServiceStateLogic;
+import brooklyn.entity.basic.SoftwareProcess;
+import brooklyn.entity.basic.SoftwareProcessDriverLifecycleEffectorTasks;
 import brooklyn.entity.java.VanillaJavaApp;
 import brooklyn.entity.proxying.ImplementedBy;
 import brooklyn.entity.webapp.WebAppService;
+import brooklyn.location.Location;
 
 /**
  * Mock web application server entity for testing.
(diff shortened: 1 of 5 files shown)


software/webapp/src/test/java/brooklyn/test/entity/TestJavaWebAppEntityImpl.java
```diff
@@ -91,4 +73,25 @@ public int getB() {
...
+
+            public void stop() {
+                ServiceStateLogic.setExpectedState(entity(), Lifecycle.STOPPING);
+                LOG.trace("Stopping {}", this);
+                entity().setAttribute(Attributes.SERVICE_UP, false);
+                entity().setAttribute(SERVICE_PROCESS_IS_RUNNING, false);
+                ServiceStateLogic.setExpectedState(entity(), Lifecycle.STOPPED);
+            }
+        };
+    }
```
alasdairhodge: This override ignores any confgured lifecycle-tasks, which could be surprising-in-a-bad-way. I guess we'd really like to override the config key's *default* (which isn't possible).

sjcorbett: I thought about it but figured it was a not-too-widely used test class. Happy to update if you think fit.

alasdairhodge: Workaround: check *raw* config in `init()`, manually call `setConfig(LIFECYCLE_EFFECTOR_TASKS, getLifecycleEffectorTasks())` if unset. Pretty ugly.

sjcorbett: Actually in truth I did this in two steps - the first was to make getLifecycleEffectorTasks `protected` and the change above, then I added the config option without updating this case.

sjcorbett: Or just alter `TestJavaWebAppEntity` to set the default to this subclass?

alasdairhodge: Even though it's just a "minor" class, it nonetheless serves as an example of how we recommend navigating this maze of lifecycle methods/effectors/overrides/fields/annotations; doubly so in the absence of decent documentation.

software/webapp/src/test/java/brooklyn/test/entity/TestJavaWebAppEntityImpl.java
```diff
@@ -46,28 +46,10 @@ public TestJavaWebAppEntityImpl() {}
...
-            entity().setAttribute(SERVICE_PROCESS_IS_RUNNING, false);
-            ServiceStateLogic.setExpectedState(entity(), Lifecycle.STOPPED);
-        }
-    };
-
     @Override
     public void init() {
         super.init();
-        LIFECYCLE_TASKS.attachLifecycleEffectors(this);
+        getLifecycleEffectorTasks().attachLifecycleEffectors(this);
```
alasdairhodge: This is redundant, given `super.init()`

sjcorbett:
@alasdairhodge Thanks for comments. Would you like to double check?

software/webapp/src/test/java/brooklyn/test/entity/TestJavaWebAppEntity.java
```diff
@@ -28,8 +39,38 @@
 @ImplementedBy(TestJavaWebAppEntityImpl.class)
 public interface TestJavaWebAppEntity extends VanillaJavaApp, WebAppService {
 
+    /**
+     * Injects the test entity's customised version of the class.
+     */
```
alasdairhodge: Nitpick over phrasing: "Injects the test entity's customised lifecycle tasks."

alasdairhodge:
Trivial nitpick, otherwise looks grand.

software/webapp/src/test/java/brooklyn/test/entity/TestJavaWebAppEntity.java
sjcorbett: Got an improvement?

alasdairhodge: Ha! My suggested improvement is right there in quotes ;o)

sjcorbett: Err, no explanation for how I missed that.

sjcorbett:
Closing since merged.

PR #737, opened 2015-07-07
Release 08

hzbarcea:
It's starting to take shape. I still have an issue with release:prepare dryRun, but while I sort it comments are welcome and encouraged.


hzbarcea:
Looks better, the dryRun for the release looks good. There is still an issue left with the archetypes, the version doesn't get updated, so if one wipes out the local maven repo and does a release dryRun, there will be a: "[ERROR]     Non-resolvable parent POM". 

Hoping to resolve this tomorrow. I believe this is the last thing that needs fixing.

hzbarcea:
This is ready for review+merge, other improvements will come as separate PRs.

CMoH:
Please pull cmoh@deb2cf37de2383bfe11fef446b15ecf8d43ec609 as well.

hzbarcea:
The quickstart archetype should not depend on the downstream parent, but it does rely on it being present in the repo when an archetype is built. The real problem was that because there is no apache brooklyn release yet (and brooklyn.version points to current snapshot version) such version was not found in snapshot repos. Adding the apache.snapshots repo to archetype solved the problem.

At this point, both `mvn install` and `mvn release:prepare -DdryRun=true` work fine for me (even after removing all brooklyn artifacts from local maven repo). With my current karma I cannot test beyond this point, but I think there are no more blockers for a release.

Ready to review+merge. I think this PR is complete now and there will be no additional commits coming.

software/nosql/pom.xml
```diff
@@ -295,4 +295,26 @@
...
+	<property><name>!brooklyn.cloudtest</name></property>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <excludes>
+                <exclude>**/CassandraFabricTest.java</exclude>
```
neykov: Is the test failing? If that's the case, then better to exclude it in the release branch only.

hzbarcea: Only if executed on localhost, as one would expect.

neykov: `CassandraFabricTest` is designed to execute on localhost. It starts an `EmptySoftwareProcess` instead of real instances, so no commands get executed in reality. If the test is failing there is some problem within the test. It passes for me so probably is timing sensitive.
If the test is disabled it will go into oblivion and rot, that's why I think it's better to leave it enabled on master so it is fixed eventually, and disable it just for the release.

hzbarcea: Fair enough. I assume however this is not a blocker for merging the PR, right? We can take a look later (me or somebody more experienced in that particular area).

From what I recall the issue was related to host seeds for the cassandra cluster when running on just localhost. After all the refactoring it looks like that hack is no longer necessary. Will add a commit shortly.

rdowner:
Thanks @hzbarcea, merging. No need to backport to release branch, I've done that.

PR #738, opened 2015-07-08
Bundle rebind transformers with blueprints

neykov:
Add means for blueprints to register persistence state transformers, applied on rebind. This will shift the responsibility of creating the transforms from the Brooklyn operators to the blueprints developers.

Very useful when managing brooklyn of brooklyns or lots of instances (transparent for the upgrade effector). Works with partial rebind which allows upgrading of entities needing state transforms without restarting the instance.

As a next step I would like to be able to declare the version of the blueprint for which a transform is written, so it is executed only against those versions. I still don't see how to introduce versions in the blueprint classes though - probably best is to link them to the bundle version, needs further investigation.

I've been working on the changes now and then for some time, so there are merge conflicts now, needs rebasing against master.


neykov:
Resolved merge conflicts.

PR #739, opened 2015-07-08
Changes launch command to redirect nohup to /dev/null

nakomis:
This follows a customer issue where the Brooklyn server ran out of disk space as nohup.out was > 3GB


sjcorbett:
Looks good.

PR #742, opened 2015-07-09
Fix KafkaIntegrationTest tests

bostko:
- updated kafka version
- installing from binary instead of compiling from source

Note on `KafkaIntegrationTest.testTwoBrokerCluster`:
The `getMessage()` return value is hardcoded since the new kafka client is still [not implemented](https://github.com/apache/kafka/blob/0.8.2/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L553).
I tried other version kafka_2.10, 0.8.2-beta  etc, but it is unfortunately it is still not supported, it is available only in trunk.


software/messaging/src/test/java/brooklyn/entity/messaging/kafka/KafkaIntegrationTest.java
```diff
@@ -128,8 +128,13 @@ public Void call() {
         KafkaSupport support = new KafkaSupport(cluster);
 
         support.sendMessage("brooklyn", "TEST_MESSAGE");
-        String message = support.getMessage("brooklyn");
-        assertEquals(message, "TEST_MESSAGE");
+        try {
+            Thread.sleep(Duration.seconds(5).toMilliseconds());
```
mikezaccardo: Probably should let this propagate with `Time.sleep()` since it's in a test?

software/messaging/src/test/java/brooklyn/entity/messaging/kafka/KafkaSupport.java
```diff
@@ -55,15 +53,39 @@ public KafkaSupport(KafkaCluster cluster) {
...
-        Properties props = new Properties();
-        props.put("zk.connect", String.format("%s:%d", zookeeper.getAttribute(Attributes.HOSTNAME), zookeeper.getZookeeperPort()));
-        props.put("serializer.class", "kafka.serializer.StringEncoder");
-        ProducerConfig config = new ProducerConfig(props);
-
-        Producer<String, String> producer = new Producer<String, String>(config);
-        ProducerData<String, String> data = new ProducerData<String, String>(topic, message);
-        producer.send(data);
-        producer.close();
+        for(Entity e : cluster.getCluster().getChildren()) {
```
mikezaccardo: Can this loop be omitted?

software/messaging/src/test/java/brooklyn/entity/messaging/kafka/KafkaSupport.java
```diff
@@ -55,15 +53,39 @@ public KafkaSupport(KafkaCluster cluster) {
...
+
+            props.put("metadata.broker.list", format("%s:%d", broker.getAttribute(KafkaBroker.HOSTNAME), broker.getKafkaPort()));
+            props.put("bootstrap.servers", format("%s:%d", broker.getAttribute(KafkaBroker.HOSTNAME), broker.getKafkaPort()));
+            props.put("key.serializer", "org.apache.kafka.common.serialization.StringSerializer");
+            props.put("value.serializer", "org.apache.kafka.common.serialization.StringSerializer");
+
+            Producer<String, String> producer = new KafkaProducer<>(props);
+            try {
+                ((KafkaZooKeeper)cluster.getZooKeeper()).createTopic(topic);
+                Thread.sleep(Duration.seconds(1).toMilliseconds());
```
mikezaccardo: Same comment as above wrt propagating.

software/messaging/src/main/java/brooklyn/entity/messaging/kafka/Kafka.java
```diff
@@ -30,11 +30,11 @@
  */
 public interface Kafka {
 
-    ConfigKey<String> SUGGESTED_VERSION = ConfigKeys.newConfigKeyWithDefault(SoftwareProcess.SUGGESTED_VERSION, "0.7.2-incubating");
+    ConfigKey<String> SUGGESTED_VERSION = ConfigKeys.newConfigKeyWithDefault(SoftwareProcess.SUGGESTED_VERSION, "2.9.2-0.8.2.1");
 
     @SetFromFlag("downloadUrl")
     BasicAttributeSensorAndConfigKey<String> DOWNLOAD_URL = new BasicAttributeSensorAndConfigKey<String>(
-            Attributes.DOWNLOAD_URL, "http://mirror.catn.com/pub/apache/incubator/kafka/kafka-${version}/kafka-${version}-src.tgz");
+            Attributes.DOWNLOAD_URL, "http://apache.cbox.biz/kafka/0.8.2.1/kafka_${version}.tgz");
```
mikezaccardo: Can the hardcoded `0.8.2.1` be avoided?

mikezaccardo:
LGTM -- KafkaIntegrationTest passes for me.

aledsage:
Thanks @bostko - merging.

neykov:
A bit late, but was this ready for merging @bostko? I am referring to the hard-coded message in `getMessage()`.

bostko:
Yes it is just for the integration test, the java client doesn't implement getting message, poll() is not implemented so I hardcoded this in the integration test.

software/messaging/src/main/java/brooklyn/entity/messaging/kafka/KafkaZooKeeperSshDriver.java
```diff
@@ -60,4 +67,12 @@ public Integer getZookeeperPort() {
         return getEntity().getAttribute(KafkaZooKeeper.ZOOKEEPER_PORT);
     }
 
+    @Override
+    public void createTopic(String topic) {
+        String zookeeperUrl = getEntity().getAttribute(Attributes.HOSTNAME) + ":" + getZookeeperPort();
+        newScript(CUSTOMIZING)
+                .failOnNonZeroResultCode()
+                .body.append(String.format("./bin/%s  --create --zookeeper %s --replication-factor 1 --partitions 1 --topic %s", getTopicsScriptName(), zookeeperUrl, topic))
```
neykov: `getTopicsScriptName()` is coming from the user, right? Should be bash-escaped, see `BashStringEscapes`. Otherwise will let the user execute arbitrary commands (or fail when special characters are used).

software/messaging/src/test/java/brooklyn/entity/messaging/kafka/KafkaIntegrationTest.java
```diff
@@ -128,8 +128,8 @@ public Void call() {
         KafkaSupport support = new KafkaSupport(cluster);
 
         support.sendMessage("brooklyn", "TEST_MESSAGE");
+        Thread.sleep(Duration.seconds(5).toMilliseconds());
```
neykov: Is the sleep added on purpose or a leftover from testing?

software/messaging/src/test/java/brooklyn/entity/messaging/kafka/KafkaSupport.java
```diff
@@ -54,16 +52,33 @@ public KafkaSupport(KafkaCluster cluster) {
...
```
neykov: Is the sleep needed?

software/messaging/src/test/java/brooklyn/entity/messaging/kafka/KafkaSupport.java
```diff
@@ -54,16 +52,33 @@ public KafkaSupport(KafkaCluster cluster) {
...
+
+                ProducerRecord<String, String> data = new ProducerRecord<>(topic, message);
+                producer.send(data);
+                producer.close();
+            } catch (InterruptedException e) {
+                e.printStackTrace();
```
neykov: If the sleep really needs to stay can use `Time.sleep(Duration.ONE_SECOND)` which doesn't declare `InterruptedException` so you can remove the try-catch. At the very least replace the `printStackTrace` with a call to a logger or `Excpetions.propagate`.

neykov:
Finished review (after the fact...). Worth addressing the bash escape in another PR.

PR #743, opened 2015-07-13
Fix SUSE on aws-ec2

aledsage:
And some other minor fixes - see separate commits for "Fix jclouds createUser..." and "Fix REST api location returning config"


richardcloudsoft:
+1

PR #744, opened 2015-07-14
Update brooklyn-node.yaml example

bostko:
Include the more reasonable `brooklynnode.distro.uploadurl` attribute in the `brooklyn-node.yaml` example

### DIFF
--- a/software/base/src/main/resources/brooklyn/entity/brooklynnode/brooklyn-node.yaml
+++ b/software/base/src/main/resources/brooklyn/entity/brooklynnode/brooklyn-node.yaml
@@ -22,8 +22,8 @@ name: Example Brooklyn Node
 services:
 - type: brooklyn.entity.brooklynnode.BrooklynNode
 
-  ## to use a local file, specify something such as the following:   [BROOKLYN_VERSION_BELOW] 
-  # downloadUrl: file:///tmp/brooklyn-dist-0.8.0-SNAPSHOT-dist.tar.gz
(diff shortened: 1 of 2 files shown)


mikezaccardo:
LGTM

software/base/src/main/resources/brooklyn/entity/brooklynnode/brooklyn-node.yaml
```diff
@@ -22,8 +22,8 @@ name: Example Brooklyn Node
 services:
 - type: brooklyn.entity.brooklynnode.BrooklynNode
 
-  ## to use a local file, specify something such as the following:   [BROOKLYN_VERSION_BELOW] 
-  # downloadUrl: file:///tmp/brooklyn-dist-0.8.0-SNAPSHOT-dist.tar.gz
+  ## URL for uploading the brooklyn distro (retrieved locally and pushed to remote install location)
```
neykov: `BROOKLYN_VERSION_BELOW` is a marker, pointing out that the version on the line below should be updated when changing the project version. To avoid having a stale version you can either get the marker back in or use some generic text in the url (i.e. `brooklyn-dist-<version>-dist.tar.gz`).

bostko: Fixed. Thank you!

aledsage:
Thanks @bostko - merging.

PR #745, opened 2015-07-14
Update BrooklynNode default expand folder

neykov:
The expand folder name changed to `brooklyn-dist-<version>`.
Also check the healthy status instead of the root URL so we know the instance has warmed up and is ready to service requests.

### DIFF
--- a/software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeSshDriver.java
+++ b/software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeSshDriver.java
@@ -112,7 +112,7 @@ public class BrooklynNodeSshDriver extends JavaSoftwareProcessSshDriver implemen
                 }
             }
         }
-        if (subpath==null) subpath = format("brooklyn-%s", getVersion());
+        if (subpath==null) subpath = format("brooklyn-dist-%s", getVersion());
         setExpandedInstallDir(Os.mergePaths(getInstallDir(), resolver.getUnpackedDirectoryName(subpath)));
     }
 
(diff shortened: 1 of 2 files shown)


software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeSshDriver.java
```diff
@@ -127,8 +127,8 @@ public void install() {
         String uploadUrl = entity.getConfig(BrooklynNode.DISTRO_UPLOAD_URL);
         
         // Need to explicitly give file, because for snapshot URLs you don't get a clean filename from the URL.
-        // This filename is used to generate the first URL to try: 
-        // file://$HOME/.brooklyn/repository/BrooklynNode/0.6.0-SNAPSHOT/brooklyn-0.6.0-SNAPSHOT-dist.tar.gz
+        // This filename is used to generate the first URL to try: [BROOKLYN_VERSION_BELOW]
+        // file://$HOME/.brooklyn/repository/BrooklynNode/0.8.0-SNAPSHOT/brooklynnode-0.8.0-snapshot.tar.gz
```
aledsage: Are we using lower-case "snapshot" in the filename? Should we use "brooklynnode-0.8.0-SNAPSHOT.tar.gz" instead (for consistency, even if it is a bit ugly)? What will BROOKLYN_VERSION_BELOW do with the lower-case snapshot?

aledsage:
@neykov One question about something inside a comment. Won't let that hold up merging this now.

software/base/src/main/java/brooklyn/entity/brooklynnode/BrooklynNodeSshDriver.java
neykov: Yes, the filename is lowercase (I copied the URL from stdin). See https://github.com/apache/incubator-brooklyn/blob/master/core/src/main/java/brooklyn/entity/drivers/downloads/DownloadProducerFromLocalRepo.java#L51.

Good point about `BROOKLYN_VERSION_BELOW`, the lower case part won't get replaced, so I think best is to just remove the marker and leave the version as is.

PR #746, opened 2015-07-14
Fix dist to have suitable name, and suitably-named directory inside

rdowner:
In commit 56285ad89cc2ba7b9ea540cfb0de25f411d14e29, which modifies the way that Brooklyn dist is built, @ahgittin notes: *"finalName affects name in `target/` but we cannot influence name when it is attached/installed, so `apache-` prefix would be lost there. to keep it consistent this is commented out, but would be nice to have if there is a way!"*

However I am not certain why this is an issue, and why we need to insist that the built artifact name must be consistent with the attached-artifact name. The primary concern of the distribution is a single downloadable binary archive. This archive should be named appropriately for the product, and convention dictates that the archive contains a single directory with the same name as the archive, and that directory contains all the files. maven-assembly-plugin will also upload them to the repository, but forces them to be a particular name. However, *that isn't important* - users aren't going to get the binary download by digging through Maven repositories, they are going to get it from the Apache download page. The Maven artifact of the distribution is merely a side effect, so does it really matter what its name is?

This PR changes the name of the archives to `apache-brooklyn-VERSION-bin.tar.gz` and `.zip`. Each contains a single folder, `apache-brooklyn-VERSION-bin`, which contains all the files for the distribution.

This corresponds to the release process, which is producing names and formats identical except that `bin` is substituted for `src`.

### DIFF
--- a/usage/dist/pom.xml
+++ b/usage/dist/pom.xml
@@ -134,11 +134,8 @@
                             <descriptors>
                                 <descriptor>src/main/config/build-distribution.xml</descriptor>
                             </descriptors>
-                          <!-- finalName affects name in `target/` but we cannot influence name when it is attached/installed,
-                               so `apache-` prefix would be lost there. to keep it consistent this is commented out, 
-                               but would be nice to have if there is a way!
-                            <finalName>apache-brooklyn-${project.version}</finalName>
-                          -->
+                            <finalName>apache-brooklyn-${project.version}-bin</finalName>
+                            <appendAssemblyId>false</appendAssemblyId>
                             <formats>
                                 <format>tar.gz</format>
                                 <format>zip</format>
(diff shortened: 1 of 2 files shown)


rdowner:
Once #745 is merged, this PR will need an additional change to ensure `BrooklynNode` continues to work.

neykov:
If we don't need the tar.gz artifact in the repo let's just not attach it. But if attached it'd be nice to have the same extracted folder as the file name.

One instance where it's important is `BrooklynNode` - it will download the dist from the repository by default and infer the extract folder from the download file name.

Alternatively we could have two tar.gz files - one attached with the default name and one with apache prefix which is not uploaded to the repo. This will increase the build time, so not ideal.

rdowner:
@neykov I'm perfectly happy to not attach the artifacts. I can't see many use cases that would require downloading the artifact from deep inside Nexus.

hzbarcea:
> However, that isn't important - users aren't going to get the binary download by digging through Maven repositories, they are going to get it from the Apache download page.

I wouldn't be so sure about that.

rdowner:
Ok. Rather than proceeding with this and its potential complications (including breaking `BrooklynNode` again), I'm closing this PR and will instead modify the release script to post-process the dist archives to have the expected names.

hzbarcea:
I believe we can iron out the release process via a series of releases.

Hadrian

PR #747, opened 2015-07-15
Don't use hostname when obtaining machines

neykov:
By default Softlayer machines will have a hostname of the form `xxxx.local.brooklyncentral.com` which resolves to a parked domain, not the public IP of the machine. When the machine has privte IP only the hostname was used instead of the private IP which caused a failure. This commit removes the hostname fallback - I don't see a valid reason why it should be included.


aledsage:
Looks good. Happy to merge once asfbot confirms build works.

The previous attempt failed. Can you close/reopen (or equivalent) to kick off the asfbot again please?

neykov:
New build is successful.

hzbarcea:
In addition to ASF bot reporting a successful build, I fetched, built, ran all tests successfully. I confirm that this PR can be merged.

PR #749, opened 2015-07-16
Add EntityDynamicType.removeEffector(Effector)

sjcorbett:
Not sure if this is controversial.

### DIFF
--- a/core/src/main/java/brooklyn/entity/basic/EntityDynamicType.java
+++ b/core/src/main/java/brooklyn/entity/basic/EntityDynamicType.java
@@ -130,15 +130,36 @@ public class EntityDynamicType extends BrooklynDynamicType<Entity, AbstractEntit
             instance.emit(AbstractEntity.EFFECTOR_ADDED, newEffector.getName());
     }
 
-    /** Adds an effector with an explicit body */
+    /**
+     * Adds an effector with an explicit body to this entity.
+     */
     @Beta
     public <T> void addEffector(Effector<T> effector, EffectorTaskFactory<T> body) {
         addEffector(new EffectorAndBody<T>(effector, body));
     }
-    /** Adds an effector with an explicit body */
+
+    /**
+     * Adds an effector with an explicit body to this entity.
+     */
     @Beta
     public <T> void addEffector(Effector<T> effector, EffectorBody<T> body) {
         addEffector(effector, new EffectorBodyTaskFactory<T>(body));
+    }
+
+    /**
+     * Removes the given {@link Effector} from this entity.
+     * <p>
+     * Note that if the argument is an instance of {@link EffectorWithBody} it will
+     * still be possible to invoke the effector on the entity by calling
+     * <code>entity.invoke(effector, argumentsMap)</code>.
+     */
+    @Beta
+    public void removeEffector(Effector<?> effector) {
+        Effector<?> removed = effectors.remove(effector.getName());
+        invalidateSnapshot();
+        if (removed != null) {
+            instance.emit(AbstractEntity.EFFECTOR_REMOVED, removed.getName());
+        }
     }
 
     // --------------------------------------------------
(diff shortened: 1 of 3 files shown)


alasdairhodge:
Awesome. Not controversial, IMO; the symmetry makes perfect sense.

alasdairhodge:
Controversial lack of a unit test, however.

sjcorbett:
@alasdairhodge Thanks for comments. Would you cast your eye over a second time? Note the comment on `removeEffector` and the assertions of the new test. Brooklyn will still invoke removed effectors because it knows better than you and gets Groovy to look up the method dynamically.

core/src/main/java/brooklyn/entity/basic/EntityDynamicType.java
```diff
@@ -130,17 +130,38 @@ public void addEffector(Effector<?> newEffector) {
...
         addEffector(effector, new EffectorBodyTaskFactory<T>(body));
     }
 
+    /**
+     * Removes the given {@link Effector} from this entity.
+     * <p>
+     * Note that if the argument is an instance of {@link EffectorWithBody} it will
+     * still be possible to invoke the effector on the entity by calling
+     * <code>entity.invoke(effector, argumentsMap)</code>.
+     */
```
alasdairhodge: The ability to invoke removed efectors is a bug, this comment should reflect that. Is it trivial to fix `entity.invoke()`?

alasdairhodge:
Unfortunate behaviour in `entity.invoke()`; I'd call that a bug, and suggest a fix in the form of an explicit check in `Effectors.invocation()`, which already throws `UnsupportedOperationException` for missing effectors.

sjcorbett:
Is it a bug? We can remove the effector but for example `Startable` entities are still going to have a `stop` method the can be invoked programatically. Brooklyn turns those method calls into effector invocations. Should we not do this if the related effector has been removed?

I'm wary of modifying `Effectors.invocation` (see https://github.com/apache/incubator-brooklyn/blob/master/core/src/main/java/brooklyn/entity/effector/Effectors.java#L153-159) without looking more into what is relying on this at the moment.

neykov:
I don't see it as a bug. Effectors are like enrichers - the entity might not be aware of it at all. The entity could override an effector body by adding it to the `entityType`, but otherwise the caller's implementation will be used.
Take for example an `SshEffector` which executes a shell command on the entity's machine. It can be executed on any entity which uses `SshMachineLocation`, but it's not feasible to add it to all such entities.

alasdairhodge:
Good points, @sjcorbett and @neykov, thanks. I agree that "fixing" `Effectors` is too risky to consider here. I would still consider it a bug if effectors that have been *explicitly removed in code* can still be invoked.

alasdairhodge:
Interesting example, @svet. Just out of curiosity: this specific usage feels like a band-aid for lacking the right place to define a (static) `executeCommand` effector. Would you dynamically add the mixin effector if an explicit `SshMachine` entity was present in the mgmt tree? (Not that I'm wanting to reopen the "locations as entities" discussion...)

neykov:
Well I wouldn't add it dynamically at all, just invoke it on the entity (not sure if I got your question right). 
There are other examples of cross-cutting concerns which would work on a large number of entities, but none would add them explicitly to the "favoured" list. Like an effector which reports for potential rebind problems on running entities.

alasdairhodge:
@neykov: Hmm, interesting and powerful. The (undeclared) "cross-cutting concern" effectors won't be shown in the UI, right? And can't be invoked via REST?

alasdairhodge:
@sjcorbett: this thread has been a real eye-opener. I'm coming round to the elegance of "unattached" effectors for cross-cutting concerns, but this does call into question the notion of removing effectors, which is what this PR is all about.

In our specific case, can we instead prevent the unwanted effector being added in the first place?

neykov:
>  The (undeclared) "cross-cutting concern" effectors won't be shown in the UI, right? And can't be invoked via REST?

That's right. We could add functionality to invoke undeclared effectors to the REST API but it's not supported now.

On the other hand the effector task will appear in the entity's Activity tab for inspection.

neykov:
Removing effectors could still be useful for UI reasons.

aledsage:
Thanks all - merging now.

PR #750, opened 2015-07-16
Boolean effector parameters are rendered as checkboxes.

sjcorbett:
.. in jsgui.

### DIFF
--- a/usage/jsgui/src/main/webapp/assets/js/view/effector-invoke.js
+++ b/usage/jsgui/src/main/webapp/assets/js/view/effector-invoke.js
@@ -71,6 +71,9 @@ define([
                 // select the body of the table we just rendered and append params
                 var $tbody = this.$("tbody")
                 _(params).each(function (param) {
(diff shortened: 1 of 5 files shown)


neykov:
What if we want to pass a null value? Should we use a tri-state checkbox?

sjcorbett:
Is there really a valid case for requiring true/false/null? If someone does that they need to revisit how they design things. If there's an existing case that relies on a Boolean being null I will revisit.

neykov:
I don't know of any, just wanted to point out this (theoretical) possibility. Since it's a UI only convenience and doesn't prevent one from passing null through other means I am all for it.

sjcorbett:
It's a good one. Thanks for raising it. I'll ask the mailing list first.

sjcorbett:
Screenshot:

screen shot 2015-07-17 at 12 10 42

neykov:
No objections on the mailing list. Good to merge.

aledsage:
Thanks all; merging.

PR #752, opened 2015-07-17
Minor improvement on an ajax call

ZJaffee:


hzbarcea:
This PR breaks the build. Tested locally as well and got the error below. Could you please take a look again?

```
[ERROR] Caused by: com.gargoylesoftware.htmlunit.ScriptException: missing } after property list (http://localhost:55424/src/js/view/policy-new.js#66)
[ERROR] at com.gargoylesoftware.htmlunit.javascript.JavaScriptEngine$HtmlUnitContextAction.run(JavaScriptEngine.java:684)
```

sjcorbett:
@ZJaffee Could you describe the improvement here? From what I can see you've swapped fail/done for success/error. Personally I prefer the former. Is there good reason to use the latter?

tbouron:
@ZJaffee On top of @sjcorbett comment, I would rather use the ` jqXHR.fail()` and `jqXHR.done()` (which can be chained) instead of the configuration keys `error` and `success`.

PR #753, opened 2015-07-17
Fix build order - launcher has build-time depency on jsgui

neykov:
Also make the ordering more natural - almost all projects depend on `utils`. `all`, `dist`, `qa` are includes-all modules.

The build would use an older version of `jsgui` (not the one being compiled with the project), or fail if none is available because `launcher` was built before `jsgui`.  Since `launcher` has only build time dependency on `jsgui` it's not convenient to include it in the dependencies section because it would have to be excluded manually in every project that depends on `launcher`.

### DIFF
--- a/usage/launcher/pom.xml
+++ b/usage/launcher/pom.xml
@@ -77,6 +77,14 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.brooklyn</groupId>
+            <artifactId>brooklyn-jsgui</artifactId>
+            <type>war</type>
+            <!-- Needed during build time only -->
+            <scope>provided</scope>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.felix</groupId>
             <artifactId>org.apache.felix.framework</artifactId>
         </dependency>


hzbarcea:
In general, build order is not fixed via the order of listing <module>s in the pom.xml but via the dependency mechanism. Modules were initially listed in a sorta alphabetical order to make them easier to find. This is a workable hack, but the builds will require more love for the next releases.

neykov:
I don't agree it's a hack - it relies on documented maven behaviour.

1) Maven doesn't provide a mechanism to declare a build-time dependency. Adding a dependency to the `jsgui` war file means that it will be included in all project which include `launcher` as a (transitive) dependency. Also it will appear everywhere dependencies are copied (i.e. when using assembly plugin, as the dist project does). To avoid this `jsgui` would have to be excluded from each dependency to `launcher` which is far from ideal (this includes downstream projects).

2) From http://maven.apache.org/guides/mini/guide-multiple-modules.html#Reactor_Sorting on build order sorting:
```
the order declared in the <modules> element (if no other rule applies)
```

Having in mind 1 & 2 I decided that defining the build order in the reactor is the best option.

The solution is a nuisance of course - it shifts the responsibility to keep the build order on the developers, but it's not that big of a deal. Any suggestions to improve it are welcome.

As for breaking the alphabetical sorting - I find it easier to find the projects if they follow build order, but that's a bike shedding discussion and not the reason for the reorder. Some of the modules had to be pulled down to let jsgui build on time. As for examples and QA - it's nice to have dist built as soon as possible when developing, while having the leaf projects built afterwards to confirm the successful changes.

rdowner:
@neykov in response to your comment *"Maven doesn't provide a mechanism to declare a build-time dependency"*, does `<scope>provided</scope>` have that effect?

neykov:
Works perfectly and much more elegant. Thanks @rdowner - updated the PR.

hzbarcea:
@neykov, I suspect you missed my point. If the build system is defined correctly, it should succeed regardless of the order of the modules. That is ensured via the dependency mechanism. The alphabetical order is as good as any, should be completely irrelevant to the build system, but hopefully useful for a human reader.

@rdowner, thanks for clarifying.

neykov:
Build failure not related, kicked another build.

hzbarcea:
Good to merge.

aledsage:
Thanks all - merging.

PR #754, opened 2015-07-17
Unescapes sensor and config values before copying to clipboard

nakomis:
Previously `foo&bar` was being copied to the clipboard as `foo&amp;bar`

### DIFF
--- a/usage/jsgui/src/main/webapp/assets/js/view/entity-config.js
+++ b/usage/jsgui/src/main/webapp/assets/js/view/entity-config.js
@@ -180,11 +180,11 @@ define([
                     // the zeroClipboard instance is a singleton so check our scope first
                     if (!$(this).closest("#config-table").length) return;
                     var text = $(this).attr('copy-value');
-                    if (!text) text = $(this).closest('.floatGroup').find('.value').html();
+                    if (!text) text = $(this).closest('.floatGroup').find('.value').text();
                     
 //                    log("Copying config text '"+text+"' to clipboard");
                     client.setText(text);
-                    
+
                     // show the word "copied" for feedback;
                     // NB this occurs on mousedown, due to how flash plugin works
                     // (same style of feedback and interaction as github)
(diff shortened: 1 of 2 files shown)


hzbarcea:
This PR causes a build problem for me:

```
[INFO] Brooklyn Policies .................................. FAILURE [01:31 min]
```

Will look into it later.

nakomis:
Hi @hzbarcea. This PR only changes a couple of .js files, so I'd be surprised if it caused an issue with the Brooklyn Policies project. Can you post the details of the failure?

hzbarcea:
@nakomis yes, I know, that's why it's surprising. Looks like the asfbot was slower than me, but reports success. The asfbot builds in a clean environment, I did an `mvn clean install`, but I fear the build picked up something from the local maven repo from a previous build. I'd have to look into it, but can only do that later today.

usage/jsgui/src/main/webapp/assets/js/view/entity-config.js
```diff
@@ -181,7 +181,7 @@ define([
                     if (!$(this).closest("#config-table").length) return;
                     var text = $(this).attr('copy-value');
                     if (!text) text = $(this).closest('.floatGroup').find('.value').html();
```
neykov: Would be easier to use .text() instead of trying to unescape the html. Also `_.unescape` works for a limited html escapes.

tbouron: +1, it will make thing easier and shorter.

tbouron:
@hzbarcea, @nakomis: Fetched on my local machine, all tests passed. :+1:

aledsage:
Thanks @nakomis - do you want to switch to the suggestion from @neykov and then we can merge?

usage/jsgui/src/main/webapp/assets/js/util/brooklyn-utils.js
```diff
@@ -33,6 +33,11 @@ define([
         return _.escape(s);
     };
 
+    Util.unescape = function (s) {
+        if (s == undefined || s == null) return "";
```
ahgittin: The preferred way is to use `typeof` and compare against the string `'undefined'`:

http://stackoverflow.com/questions/3390396/how-to-check-for-undefined-in-javascript

ahgittin:
+1 to @neykov 's suggestion to use `text()` which I think should remove the need for unescaping.  Worth testing with a sensor whose value is something perverse like `I use Brooklyn & &amp; <lol>` .

aledsage:
Thanks @nakomis - did you manually test this with a perverse sensor value?

Merging.

PR #756, opened 2015-07-17
Kafka - fix escaping

bostko:
Addressed issues fixed

### DIFF
--- a/software/messaging/src/test/java/brooklyn/entity/messaging/kafka/KafkaIntegrationTest.java
+++ b/software/messaging/src/test/java/brooklyn/entity/messaging/kafka/KafkaIntegrationTest.java
@@ -125,11 +125,16 @@ public class KafkaIntegrationTest {
 
         Entities.dumpInfo(cluster);
 
-        KafkaSupport support = new KafkaSupport(cluster);
+        final KafkaSupport support = new KafkaSupport(cluster);
 
         support.sendMessage("brooklyn", "TEST_MESSAGE");
-        Thread.sleep(Duration.seconds(5).toMilliseconds());
-        String message = support.getMessage("brooklyn");
-        assertEquals(message, "TEST_MESSAGE");
+
+        Asserts.succeedsEventually(MutableMap.of("timeout", Duration.FIVE_SECONDS), new Runnable() {
+            @Override
+            public void run() {
+                String message = support.getMessage("brooklyn");
+                assertEquals(message, "TEST_MESSAGE");
+            }
+        });
     }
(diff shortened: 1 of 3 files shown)


software/messaging/src/test/java/brooklyn/entity/messaging/kafka/KafkaIntegrationTest.java
```diff
@@ -125,11 +125,17 @@ public Void call() {
...
 
-        KafkaSupport support = new KafkaSupport(cluster);
+        final KafkaSupport support = new KafkaSupport(cluster);
 
         support.sendMessage("brooklyn", "TEST_MESSAGE");
-        Thread.sleep(Duration.seconds(5).toMilliseconds());
-        String message = support.getMessage("brooklyn");
-        assertEquals(message, "TEST_MESSAGE");
+
+        Asserts.succeedsEventually(MutableMap.of("timeout", Duration.FIVE_SECONDS), new Callable<Void>() {
```
neykov: Can use `Runnable` in this case, won't need `return null` further down.

neykov:
Looks good for merge.

neykov:
@bostko could you trigger a rebuild

software/messaging/src/test/java/brooklyn/entity/messaging/kafka/KafkaIntegrationTest.java
```diff
@@ -125,11 +125,16 @@ public Void call() {
...
 
-        KafkaSupport support = new KafkaSupport(cluster);
+        final KafkaSupport support = new KafkaSupport(cluster);
 
         support.sendMessage("brooklyn", "TEST_MESSAGE");
-        Thread.sleep(Duration.seconds(5).toMilliseconds());
-        String message = support.getMessage("brooklyn");
-        assertEquals(message, "TEST_MESSAGE");
+
+        Asserts.succeedsEventually(MutableMap.of("timeout", Duration.FIVE_SECONDS), new Runnable() {
```
aledsage: My personal preference is to not bother configuring an explicit timeout (so it defaults to 30 seconds). Given they are not performance tests, I don't see the point in having a short(ish) timeout. On low-powered asf builds we've seen some unit tests take several seconds.

Happy to merge as-is. You've improved it over the 5 second sleep, and left the time as it was. And this is an integration test, so shouldn't fail on a low-powered asf build server.

aledsage:
Thanks @bostko - and thanks @neykov for reviewing. Merging now.

PR #757, opened 2015-07-17
Fix SshMachineLocation unable to connect

CMoH:
Details in commit message.

### DIFF
--- a/core/src/main/java/brooklyn/location/basic/ByonLocationResolver.java
+++ b/core/src/main/java/brooklyn/location/basic/ByonLocationResolver.java
@@ -96,7 +96,11 @@ public class ByonLocationResolver extends AbstractLocationResolver {
         String user = (String) config.getStringKey("user");
         Integer port = (Integer) TypeCoercions.coerce(config.getStringKey("port"), Integer.class);
         Class<? extends MachineLocation> locationClass = OS_TO_MACHINE_LOCATION_TYPE.get(config.get(OS_FAMILY));
-        
+
+        MutableMap<String, Object> defaultProps = MutableMap.of();
+        defaultProps.addIfNotNull("user", user);
+        defaultProps.addIfNotNull("port", port);
+
         List<String> hostAddresses;
         
         if (hosts instanceof String) {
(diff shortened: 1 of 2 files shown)


neykov:
What are the conditions that lead to the problem? Why does the location put a null value in the config? 
Could there be some bigger issue hiding?

CMoH:
The null value comes from ByonLocationResolver.extractConfig(), at https://github.com/apache/incubator-brooklyn/blob/master/core/src/main/java/brooklyn/location/basic/ByonLocationResolver.java#L97

    Integer port = (Integer) TypeCoercions.coerce(config.getStringKey("port"), Integer.class);

When the port is missing from the config bag, null gets coerced to null and transitions through until the call to getOptionalVal().

In my understanding, an "optional" value should provide the default when missing - which is the exact meaning of null. Whether it's missing from the property map, or it is mapped to null should make no difference.

The opposite interpretation would be to not put nulls in maps, which means the ByonLocationResolver (and all similar client code) must implement such null checks.

neykov:
The changes leading to this were merged recently - https://github.com/apache/incubator-brooklyn/pull/721. It's surprising to have null values in the map, so might be worth fixing the new code as well. Regardless your change is valid and agree that null values should cause the default being used. What I'd like to see is the same guards to the other `getOptionalVal` overloading implementation.

CMoH:
You are right. I update the PR. 

The second commit avoids adding the null port value to the config map.  I've tested it and both fixes perform as intended.

core/src/main/java/brooklyn/location/basic/ByonLocationResolver.java
```diff
@@ -96,7 +96,11 @@ protected ConfigBag extractConfig(Map<?,?> locationFlags, String spec, brooklyn.
         String user = (String) config.getStringKey("user");
         Integer port = (Integer) TypeCoercions.coerce(config.getStringKey("port"), Integer.class);
         Class<? extends MachineLocation> locationClass = OS_TO_MACHINE_LOCATION_TYPE.get(config.get(OS_FAMILY));
-        
+
+        MutableMap<String, Object> defaultProps = MutableMap.of();
+        if (user != null) defaultProps.add("user", user);
```
neykov: `MutableMap` supports `addIfNotNull` to avoid the check.

neykov:
Good fixes, can be merged.

core/src/main/java/brooklyn/location/basic/ByonLocationResolver.java
CMoH: Thanks - I'll update again! :)

core/src/main/java/brooklyn/util/internal/ssh/ShellAbstractTool.java
```diff
@@ -105,7 +105,7 @@ protected static Boolean hasVal(Map<String,?> map, ConfigKey<?> keyC) {
     public static <T> T getOptionalVal(Map<String,?> map, ConfigKey<T> keyC) {
         if (keyC==null) return null;
         String key = keyC.getName();
-        if (map!=null && map.containsKey(key)) {
+        if (map!=null && map.containsKey(key) && map.get(key) != null) {
```
aledsage: I'll go with this and merge it.

However, I'm slightly hesitant: in some places, it's nice to be able to explicitly supply null so that it clears the value rather than it using the default. For example, in the use of JcloudsLocation I want to be able to set the privateKeyFile to none, rather than it picking the default of "~/.ssh/id_rsa". But I could probably still do that with a blank string. So not sure if there are any valid use-cases for explicitly setting null.

neykov: `ConfigBag` returns the default if the value is null, so I think it's nice to have the same behaviour here.

PR #758, opened 2015-07-18
License files are not needed for tests

hzbarcea:


CMoH:
Only affects text files. Looks good for merging.

tbouron:
+1, looks good

rdowner:
Before this can be merged we need to confirm that the boilerplate mechanism is still adding LICENSE/NOTICE/DISCLAIMER to the Maven `-test.jar` artifact. (I'll try and check this later)

rdowner:
We **do** need `LICENSE`/`NOTICE`/`DISCLAIMER` for tests, as the test artifacts are released by the project and published to Maven. This PR causes these files to be missing from the test artifact, against Apache release policy.

Because the "main" parts of these two projects require custom LICENSE files to account for bundled dependencies, the module's pom.xml disables maven-remote-resources-plugin to stop it copying the boilerplate files, and instead we supply our own files. However we're not able to disable the plugin for "main" and enable it for "test" - we have to disable it for both - therefore we also need to add `LICENSE`/`NOTICE`/`DISCLAIMER` for "test", even though in this case it's just the unmodified Apache version of these files.

For the remaining projects, they do not have `LICENSE`/`NOTICE`/`DISCLAIMER` in the source code, but maven-remote-resources-plugin automatically installs those files into artifacts (main and test) so they do not appear in the source folders.

See commit 430fb33 to see how this change was made.

(Apologies it's taken so long to review this.)

hzbarcea:
@rdowner , thanks for the review. I agree with your statement that the notices are required in the -test jars. Consolidation is needed though. I will revisit my PR, looks like it's not that simple. I will close this one for now and come back with a better solution later. I think there are a couple of comments from the release vote on general@ that should be addressed. Thanks!

PR #759, opened 2015-07-21
Fix httpclient version mismatch in pom

rdowner:
In commit 3fb4d9a @andreaturli bumped the httpclient version in the parent pom, but he didn't update the version in `brooklyn-downstream-parent` pom. This was causing dependency convergence errors in downstream projects.

### DIFF
--- a/usage/downstream-parent/pom.xml
+++ b/usage/downstream-parent/pom.xml
@@ -58,7 +58,7 @@
     <jackson.version>1.9.13</jackson.version>  <!-- codehaus jackson, used by brooklyn rest server -->
     <fasterxml.jackson.version>2.4.2</fasterxml.jackson.version>  <!-- more recent jackson, but not compatible with old annotations! -->
     <jersey.version>1.18.1</jersey.version>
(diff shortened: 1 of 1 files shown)


CMoH:
Should be merged.

However, are there any specific reasons why brooklyn-downstream-parent was not parented with brooklyn-parent? That would eliminate such problems.

neykov:
Good to merge. Adding a pairing `dependencyManagement/dependency` to the `xxx.version` properties would be nice, but can be addressed in a separate PR.

rdowner:
Thanks for the review everyone.

@CMoH there has been extensive discussion about the root pom, parent pom, and downstream pom on the ML in recent weeks. To answer your question in short, downstream-parent is intended for our downstream users to use in their pom. If it inherited from parent pom, they would pick up a lot of stuff that is needed for our build and irrelevant (and possibly conflicting) with our user's requirements.

The correct solution is to put this common stuff into the root pom that both parent pom and downstream pom inherit from. As @neykov says, this can be addressed in a separate PR.

PR #763, opened 2015-07-21
Support suspending machines

sjcorbett:
Locations can indicate that they can suspend machines by implementing a new `SuspendsMachines` interface. 

A use of this is given in `MachineLifecycleEffectorTasks`, which provides a `suspend` effector implementation. Attaching this effector to entities is left to subclasses of MLET.

An interface for locations that can resume machines is also given but unused. I began an implementation but realised that we need to discuss what it means to "resume" a machine. Eventually `JcloudsLocation` should implement `SuspendResumeLocation` rather than `SuspendsMachines`.

### DIFF
--- a/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
@@ -126,6 +126,7 @@ import brooklyn.location.MachineLocation;
 import brooklyn.location.MachineLocationCustomizer;
 import brooklyn.location.MachineManagementMixins.MachineMetadata;
 import brooklyn.location.MachineManagementMixins.RichMachineProvisioningLocation;
+import brooklyn.location.MachineManagementMixins.SuspendsMachines;
 import brooklyn.location.NoMachinesAvailableException;
 import brooklyn.location.access.PortForwardManager;
 import brooklyn.location.access.PortMapping;
@@ -187,7 +188,9 @@ import io.cloudsoft.winrm4j.pywinrm.WinRMFactory;
  * Configuration flags are defined in {@link JcloudsLocationConfig}.
  */
 @SuppressWarnings("serial")
-public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation implements JcloudsLocationConfig, RichMachineProvisioningLocation<MachineLocation>, LocationWithObjectStore {
+public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation implements
+        JcloudsLocationConfig, RichMachineProvisioningLocation<MachineLocation>,
+        LocationWithObjectStore, SuspendsMachines {
 
     // TODO After converting from Groovy to Java, this is now very bad code! It relies entirely on putting
     // things into and taking them out of maps; it's not type-safe, and it's thus very error-prone.
@@ -1020,6 +1023,35 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
             }
 
             throw Exceptions.propagate(e);
+        }
+    }
+
+    // ------------- suspend and resume ------------------------------------
+
+    /**
+     * Suspends the given location.
+     * <p>
+     * Note that this method does <b>not</b> call the lifecycle methods of any
+     * {@link #getCustomizers(ConfigBag) customizers} attached to this location.
+     */
+    @Override
+    public void suspendMachine(MachineLocation rawLocation) {
+        String instanceId = vmInstanceIds.remove(rawLocation);
+        if (instanceId == null) {
+            LOG.info("Attempt to suspend unknown machine " + rawLocation + " in " + this);
+            throw new IllegalArgumentException("Unknown machine " + rawLocation);
+        }
+        LOG.info("Suspending machine {} in {}, instance id {}", new Object[]{rawLocation, this, instanceId});
+        Exception toThrow = null;
+        try {
+            getComputeService().suspendNode(instanceId);
+        } catch (Exception e) {
+            toThrow = e;
+            LOG.error("Problem suspending machine " + rawLocation + " in " + this + ", instance id " + instanceId, e);
+        }
+        removeChild(rawLocation);
+        if (toThrow != null) {
+            throw Exceptions.propagate(toThrow);
         }
     }
 
@@ -2159,7 +2191,7 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
             throw new IllegalArgumentException("Unknown machine "+rawMachine);
         }
         JcloudsMachineLocation machine = (JcloudsMachineLocation) rawMachine;
-        
+
         LOG.info("Releasing machine {} in {}, instance id {}", new Object[] {machine, this, instanceId});
 
         Exception tothrow = null;
--- a/software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
+++ b/software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
@@ -63,6 +63,7 @@ import brooklyn.entity.trait.StartableMethods;
 import brooklyn.event.feed.ConfigToAttributes;
 import brooklyn.location.Location;
 import brooklyn.location.MachineLocation;
+import brooklyn.location.MachineManagementMixins.SuspendsMachines;
 import brooklyn.location.MachineProvisioningLocation;
 import brooklyn.location.NoMachinesAvailableException;
 import brooklyn.location.basic.AbstractLocation;
@@ -72,7 +73,6 @@ import brooklyn.location.basic.Machines;
 import brooklyn.location.basic.SshMachineLocation;
 import brooklyn.location.cloud.CloudLocationConfig;
 import brooklyn.management.Task;
-import brooklyn.management.TaskFactory;
 import brooklyn.util.collections.MutableMap;
 import brooklyn.util.config.ConfigBag;
 import brooklyn.util.exceptions.Exceptions;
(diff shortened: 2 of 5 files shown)


sjcorbett:
@neykov I think the existing stop effector behaviour is your work so would appreciate your wisdom on these additions.

sjcorbett:
Seems we've got unpredictable tests again..

software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
```diff
@@ -151,8 +151,18 @@ public void attachLifecycleEffectors(Entity entity) {
...
 
+    /** @see {@link #newStartEffector()} */
+    public Effector<Void> newSuspendEffector() {
+        return Effectors.effector(Void.class, "suspend")
+                .description("Suspend the process/service represented by an entity")
+                .parameter(StopSoftwareParameters.STOP_PROCESS_MODE)
+                .parameter(StopSoftwareParameters.STOP_MACHINE_MODE)
+                .impl(newSuspendEffectorTask())
+                .build();
+    }
```
alasdairhodge: What are the semantics of `stopMachineMode` here? Would `IF_NOT_STOPPED` really mean `IF_NOT_SUSPENDED`? How does this interact with `stop`? Is a suspended machine stoppable via `stop(IF_NOT_STOPPED)`?

software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
```diff
@@ -749,17 +791,70 @@ public String toString() {
...
+        }
+
+        if (!(provisioner instanceof SuspendsMachines)) {
+            log.debug("Location provisioner ({}) cannot suspend machines", provisioner);
+            throw new UnsupportedOperationException("Location provisioner cannot suspend machines: " + provisioner);
+        }
+
+        clearEntityLocationAttributes(machine);
+        try {
+            SuspendsMachines.class.cast(provisioner).suspendMachine(MachineLocation.class.cast(machine));
```
neykov: What's the advantage of calling `.cast` on the class rather than doing an old school cast? I thought the only usage is when one has an instance of the class in a variable (so it's dynamic).

software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
```diff
@@ -749,17 +791,70 @@ public String toString() {
...
+        } catch (Throwable t) {
```
neykov: Don't see what the point of the catch is here. Yes it's like this in the analogous stop method, but still...

locations/jclouds/src/main/java/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1008,6 +1011,24 @@ protected MachineLocation obtainOnce(ConfigBag setup) throws NoMachinesAvailable
...
+     * {@link #getCustomizers(ConfigBag) customizers} attached to this location.
+     */
+    @Override
+    public void suspendMachine(MachineLocation rawLocation) {
+        String instanceId = vmInstanceIds.remove(rawLocation);
+        if (instanceId == null) {
+            LOG.info("Attempt to suspend unknown machine " + rawLocation + " in " + this);
+            throw new IllegalArgumentException("Unknown machine " + rawLocation);
+        }
+        getComputeService().suspendNode(instanceId);
```
neykov: Why skip `removeChild(location)`? Shouldn't the location be unmanaged?

software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
```diff
@@ -206,6 +216,21 @@ public Void call(ConfigBag parameters) {
         };
     }
 
+    /**
+     * Calls {@link #suspend(ConfigBag)}.
+     *
+     * @see {@link #newStartEffectorTask()}
+     */
+    public EffectorBody<Void> newSuspendEffectorTask() {
+        return new EffectorBody<Void>() {
```
neykov: Effectors are persisted, avoid anonymous classes.

software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
```diff
@@ -550,6 +575,27 @@ public void stop() {
      * If no errors were encountered call {@link #postStopCustom()} at the end.
      */
     public void stop(ConfigBag parameters) {
+        doStop(parameters, new Callable<StopMachineDetails<Integer>>() {
```
neykov: Now that `MachineLifecycleEffectorTasks` is configurable it could end up being persisted. Don't introduce new anonymous classes even if no references to them are kept, so that existing anonymous classes keep their indexes.
From what I see the anonymous classes are task bodies, so they are not supposed to get persisted, but better be safe than sorry.

neykov:
Finished review, a couple of comments worth addressing.

ResumeLocation will tie nicely with your proposal for binding JcloudsLocation to existing machines.

software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
aledsage: It probably ends up as the same byte code either way.
I'm guessing it's to avoid the layering of many brackets, such as:

    ((SuspendsMachines)provisioner).suspendMachine((MachineLocation)machine);

No strong feelings from me either way.

sjcorbett: Personal preference. I find `.class.cast` easier to read.

software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
sjcorbett: Suspending a machine has the same semantics as stopping a machine. They both use `MachineLifecycleEffectorTasks#doStop`. The former calls `MachineProvisioningLocation#release` (which becomes `JcloudsLocation#releaseNode`), the latter `MachineManagementMixins.SuspendsMachines#suspendMachine` (which becomes `JcloudsLocation#suspendMachine`). A suspended machine cannot subsequently be stopped if `IF_NOT_STOPPED` is true because `MachineLifecycleEffectorTasks#doStop` sets the entity's state to `STOPPED`.

sjcorbett: .. except of course that when suspending a machine `JcloudsLocation` does not call the pre- and post-release methods on any `JcloudsLocationCustomizers` attached to the location. I think this is fair given that the machine may be resumed.

sjcorbett:
I think I've addressed the comments. I took the opportunity to name all the anonymous classes in `MachineLifecycleEffectorTasks` and its subclasses.

alasdairhodge:
We have no way to trigger the `SUSPENDED` to `STOPPED` transition within Brooklyn, but if this is fine (and it is for my customer project) then I'd be delighted to see this merged.

sjcorbett:
Thanks for the comments. I will merge this.

software/base/src/main/java/brooklyn/entity/software/MachineLifecycleEffectorTasks.java
```diff
@@ -155,29 +155,42 @@ public void attachLifecycleEffectors(Entity entity) {
...
+     * Returns the {@link EffectorBody} which supplies the implementation for the start effector.
      * <p>
      * Calls {@link #start(Collection)} in this class.
      */
     public EffectorBody<Void> newStartEffectorTask() {
-        return new EffectorBody<Void>() {
-            @Override
-            public Void call(ConfigBag parameters) {
-                Collection<? extends Location> locations  = null;
+        return new StartEffectorBody();
```
neykov: This will break rebinding as the anonymous classes will be needed when reconstructing the entities. The approach used to keep backwards compatibility in similar cases is to keep the methods creating the anonymous classes but not use them, i.e. rename to newStartEffectorTaskDeprecated and change its visibility to private. Only needed for the newXXXEffectorTask methods. Once apache#738 is merged one could instead add code to do the rename on the fly.
Apologies for writing this late, was just able to review this.

sjcorbett: It will affect anybody who ran a 0.8.0-SNAPSHOT server with persistence since https://github.com/apache/incubator-brooklyn/pull/735 was closed, on the tenth of July, right? What do you think? Revert fcb1af0cbd7eb9799e48471e37934b046330baf3 and only name the new suspend effector? How many users are using the latest snapshots?

neykov: I think both re-introducing the effector anonymous classes (in inaccessible methods) or reverting is a valid way to go.
The effect should be quite limited (I'd say non-existent) because users have to explicitly re-configure the key with a `LifecycleEffectorTasks` instance. Using it as a default value doesn't get it serialized.

aledsage: @sjcorbett @neykov This is a big scary. I just hit it in my testing for a customer release. But that's because I tested a release candidate a day ago and then again today with this change.

Looking at what is persisted (with non-static classes etc), it feels scary for trying to support (or transform) that in future releases.

I'm going to change SoftwareProcess. LIFECYCLE_EFFECTOR_TASKS so that if the config is null it defaults to as before; if it's non-null then it will use the config value. That means we can take our time over whether we want to have a SoftwareProcessDriverLifecycleEffectorTasks instances persisted by pretty much every customer.

aledsage: Actually, I see that this is kind-of separate from LIFECYCLE_EFFECTOR_TASKS. 

@sjcorbett Why will it not affect previous users, where their code did `LIFECYCLE_TASKS.attachLifecycleEffectors(this)` so will have populated the effector body?

PR #764, opened 2015-07-21
Fix warnings in eclipse

hzbarcea:


aledsage:
Test failure is unrelated: `testIsPortAvailableReportsTrueWhenPortIsFree` "requires that at least some ports near 58769+ not be in use." It's either an environment problem, or it's too strict an assertion when testing `Networking.isPortAvailable`. It assumes that at least 5 of the 10 ports in the range 58769-58779 will be available. If they are not, the assumption was that `isPortAvailable` may well not be working.

I'll look at relaxing that in a separate PR. If it keeps failing (with no ports available) then we may have to move that to be an integration test, rather than asfbot running it in the normal unit test environment.

aledsage:
Thanks @hzbarcea - change looks sensible. Merging.

hzbarcea:
@aledsage,

Thanks. Fwiw, that was in Kepler, in Luna I don't have that problem and Mars (relatively fresh) doesn't have a stable Groovy plugin yet, so I didn't try. I am also testing with Oracle JDK 7 and 8 and OpenJDK 1.7.0_79 with no issues.

PR #769, opened 2015-07-24
Add convenience ManagementContext accessor.

alasdairhodge:
Simple utility method to obtain a given entity's `ManagementContext`, hiding the cast to the non-public `EntityInternal` class.


aledsage:
Thanks @alasdairhodge - merging.

PR #770, opened 2015-07-27
Externalised configuration

alasdairhodge:
**Bare-bones implementation** of a proposed scheme for YAML blueprints to delegate selected configuration values to runtime-injected suppliers.

[Rationale, description and discussion in Google docs.](https://docs.google.com/document/d/1_cId6tVZ79ycmvn-DAmFBZDdtQxiHRm9sOilCfMigBc/edit#)


core/src/main/java/brooklyn/config/external/ExternalConfigSupplier.java
```diff
@@ -0,0 +1,29 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package brooklyn.config.external;
+
+/**
+ * Provider of "externalised" entity configuration that is resolved at runtime.
+ */
+public interface ExternalConfigSupplier {
```
aledsage: Do we want to include `@Beta`?

Worth including `@since 0.8.0` in the javadoc. Same for other classes that a user might touch.

core/src/main/java/brooklyn/config/external/InPlaceExternalConfigSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+ * </pre>
+ *
+ * This will instantiate an <code>InPlaceExternalConfigSupplier</code> populated with values for <code>key1</code>
+ * and <code>key2</code>. Note that the <code>brooklyn.external.&lt;name&gt;</code> prefix is stripped.
+ */
+public class InPlaceExternalConfigSupplier extends AbstractExternalConfigSupplier {
+
+    private final Map<?,?> config;
+
+    public InPlaceExternalConfigSupplier(ManagementContext managementContext, String name, Map<?, ?> config) {
```
aledsage: Should the constructor's map arg be `Map<String, String>`? We look up using string keys and we cast the returned value to String.

core/src/main/java/brooklyn/config/external/PropertiesFileExternalConfigSupplier.java
```diff
@@ -0,0 +1,68 @@
...
+ * brooklyn.external.foo.propertiesUrl = http://brooklyn.example.com/config/foo.properties
+ * </pre>
+ */
+public class PropertiesFileExternalConfigSupplier extends AbstractExternalConfigSupplier {
+
+    public static final String PROPERTIES_URL = "propertiesUrl";
+
+    private final Properties properties;
+
+    public PropertiesFileExternalConfigSupplier(ManagementContext managementContext, String name, Map<?, ?> config) throws IOException {
```
aledsage: Why take a `Map<?, ?> config` in the constructor, rather than passing the properties URL as an explicit parameter (very groovy'ish!)? Is there a special way you envisage this being used?

core/src/main/java/brooklyn/config/external/PropertiesFileExternalConfigSupplier.java
```diff
@@ -0,0 +1,68 @@
...
+        super(managementContext, name);
+        this.properties = loadProperties((String) config.get(PROPERTIES_URL));
```
aledsage: Should there be a way to force a reload of the properties?

core/src/main/java/brooklyn/config/external/PropertiesFileExternalConfigSupplier.java
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Instances are populated from a plain java properties file named in the passed <code>config</code> map
+ * under the <code>propertiesUrl</code> key:
+ *
+ * <pre>
+ * brooklyn.external.foo = brooklyn.management.config.external.PropertiesFileExternalConfigSupplier
+ * brooklyn.external.foo.propertiesUrl = http://brooklyn.example.com/config/foo.properties
+ * </pre>
+ */
+public class PropertiesFileExternalConfigSupplier extends AbstractExternalConfigSupplier {
```
aledsage: Simple test case for this, and for `InPlaceExternalConfigSupplier`?

core/src/main/java/brooklyn/management/internal/BasicExternalConfigSupplierRegistry.java
```diff
@@ -0,0 +1,108 @@
...
+            try {
+                Class<? extends ExternalConfigSupplier> providerClass = (Class<? extends ExternalConfigSupplier>) classloader.loadClass(providerClassname);
+                Constructor<? extends ExternalConfigSupplier> constructor = providerClass.getConstructor(
+                    ManagementContext.class, String.class, Map.class);
+                ExternalConfigSupplier configSupplier = constructor.newInstance(this, name, config);
+                addProvider(name, configSupplier);
+                LOG.info("Added external config supplier named '" + name + "': " + configSupplier);
+
+            } catch (Exception e) {
+                LOG.error("Failed to instantiate external config supplier named '" + name + "': " + e, e);
```
aledsage: Should we just log and continue? Or should we propagate? Given these values came from brooklyn.properties (rather than persisted state etc), then it feels reasonable to fail startup of the management context.

Thoughts?

core/src/main/java/brooklyn/management/internal/BasicExternalConfigSupplierRegistry.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(BasicExternalConfigSupplierRegistry.class);
+
+    private final Map<String, ExternalConfigSupplier> providersByName = Maps.newLinkedHashMap();
+
+    public BasicExternalConfigSupplierRegistry(ManagementContext mgmt) {
+        updateFromBrooklynProperties(mgmt);
+    }
+
+    synchronized public void addProvider(String name, ExternalConfigSupplier provider) {
```
aledsage: Personal preference to not synchronize on `this`. I'd prefer synchronizing on a `private final Object` field. If you also want to support more advanced sub-classing, then can have a protected getter for the mutex.

core/src/main/java/brooklyn/config/external/PropertiesFileExternalConfigSupplier.java
aledsage: Ah, I see it's because this will be reflectively instantiated!

core/src/main/java/brooklyn/management/internal/BasicExternalConfigSupplierRegistry.java
```diff
@@ -0,0 +1,108 @@
...
+            throw new IllegalArgumentException("Provider already registered with name '" + name + "'");
+        providersByName.put(name, provider);
+    }
+
+    synchronized public void removeProvider(String name) {
+        providersByName.remove(name);
+    }
+
+    /**
+     * Searches the named {@link ExternalConfigSupplier} for the config value associated with the specified key.
```
aledsage: Why repeat the implemented interface's method's javadoc verbatim?

Also, I'd include an `@Override` on the method.

aledsage:
A good start to the implementation of your doc (which I really like).

Quoting from your doc, will the impl achieve this (or is that still to be done):

    Brooklyn will ensure that every request (i.e. call to entity.getConfig()) is delegated to the supplier, not just the first

Next step I think is unit tests.

core/src/main/java/brooklyn/config/external/ExternalConfigSupplier.java
```diff
@@ -0,0 +1,29 @@
...
+
+    String getName();
+    String get(String key);
```
aledsage: As stated in your doc:

    The supplier’s get()-semantics should be clearly described in the javadoc of the implementation class.

Can you add javadoc to that effect please.

core/src/main/java/brooklyn/config/external/PropertiesFileExternalConfigSupplier.java
alasdairhodge: Yes, good suggestion. It could be a specific local mechanism (e.g. JMX operation), or we could extend the interface and wire it all back through Brooklyn; any preference?

ahgittin:
Very useful.  Test and documentation are needed, but I'll merge it.

PR #771, opened 2015-07-27
Don't leak machines on shutdown

neykov:
STOPping an app and then doing a shutdown-and-stop-apps could leak machines because the second stop (from the shutdown) can complete faster as it's not re-releasing the machines, then unmanage the app and cancel the first stop's tasks. Cancelling the tasks will not release the machines properly.

This PR works around the problem in the shutdown code. Double stopping an app will still keep the same behaviour.

### DIFF
--- a/software/base/src/test/java/brooklyn/entity/basic/SoftwareProcessEntityTest.java
+++ b/software/base/src/test/java/brooklyn/entity/basic/SoftwareProcessEntityTest.java
@@ -18,39 +18,17 @@
  */
 package brooklyn.entity.basic;
 
-import brooklyn.config.ConfigKey;
-import brooklyn.entity.BrooklynAppUnitTestSupport;
-import brooklyn.entity.Entity;
-import brooklyn.entity.basic.SoftwareProcess.RestartSoftwareParameters;
-import brooklyn.entity.basic.SoftwareProcess.RestartSoftwareParameters.RestartMachineMode;
-import brooklyn.entity.basic.SoftwareProcess.StopSoftwareParameters;
-import brooklyn.entity.basic.SoftwareProcess.StopSoftwareParameters.StopMode;
-import brooklyn.entity.effector.Effectors;
-import brooklyn.entity.proxying.EntitySpec;
-import brooklyn.entity.proxying.ImplementedBy;
-import brooklyn.entity.software.MachineLifecycleEffectorTasksTest;
-import brooklyn.entity.trait.Startable;
-import brooklyn.event.basic.PortAttributeSensorAndConfigKey;
-import brooklyn.location.Location;
-import brooklyn.location.LocationSpec;
-import brooklyn.location.basic.FixedListMachineProvisioningLocation;
-import brooklyn.location.basic.Locations;
-import brooklyn.location.basic.SshMachineLocation;
-import brooklyn.management.Task;
-import brooklyn.management.TaskAdaptable;
-import brooklyn.util.collections.MutableMap;
-import brooklyn.util.config.ConfigBag;
-import brooklyn.util.exceptions.PropagatedRuntimeException;
-import brooklyn.util.net.UserAndHostAndPort;
-import brooklyn.util.os.Os;
-import brooklyn.util.task.DynamicTasks;
-import brooklyn.util.task.Tasks;
-import brooklyn.util.text.Strings;
-import brooklyn.util.time.Duration;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
 
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 
 import org.jclouds.util.Throwables2;
 import org.slf4j.Logger;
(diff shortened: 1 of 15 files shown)


software/base/src/test/java/brooklyn/entity/basic/SoftwareProcessEntityTest.java
```diff
@@ -568,4 +706,85 @@ protected String getInstallLabelExtraSalt() {
...
     }
+
+    public static class ReleaseLatchLocation extends SimulatedLocation {
+        private static final long serialVersionUID = 1L;
+        
+        private AtomicBoolean lock = new AtomicBoolean();
+        private boolean isBlocked;
+
+        public void unblock() {
+            synchronized (lock) {
```
aledsage: Given it's for a test, I don't really care but...

I much prefer using higher-level synchronization utils from java.util.concurrent whenever possible, rather than synchronize.
An object.wait() can wake up prematurely, which is why one should always wrap it in a while block rather than an if block.

software/base/src/test/java/brooklyn/entity/basic/SoftwareProcessEntityTest.java
```diff
@@ -568,4 +706,85 @@ protected String getInstallLabelExtraSalt() {
...
+            }
+        }
+
+        public boolean isBlocked() {
+            return isBlocked;
+        }
+
+    }
+
+    public static class GenericSoftwareProcessTestDriver implements SoftwareProcessDriver {
```
aledsage: For the future (not this PR)...
Feels like this is useful for more than just here. For example, we already have:
* `DoNothingSoftwareProcessDriver`
* `SoftwareProcessEntityTest.SimulatedDriver`
* `SoftwareProcessSshDriverIntegrationTest.SimulatedDriver`
* `MockSshDriver` (which is a stub rather than mock)
* `RestMockSimpleEntity.MockSshDriver`

usage/cli/src/main/java/brooklyn/cli/Main.java
```diff
@@ -964,4 +979,35 @@ public ToStringHelper string() {
     protected Class<? extends BrooklynCommand> cliDefaultInfoCommand() {
         return DefaultInfoCommand.class;
     }
+    
+    public static class AppShutdownRequestListener implements ShutdownRequestListener {
+        private AtomicBoolean mutex = new AtomicBoolean();
+
+        @Override
+        public void onShutdownRequest() {
+            synchronized (mutex) {
```
aledsage: I prefer use of CountDownLatch - slightly simpler code, which is always good when dealing with multi-threading.

aledsage: FYI (completely different topic really!) I'm toying with a new heuristic that use of "synchronized" should be considered a warning sign in most code; it could use things like `CountDownLatch`, `ConcurrentHashMap`, etc; or one could extract the concurrency-handling logic into a wrapping class or into a different threading model (e.g. actor-based, or "channels").

I like Go's slogan: "Do not communicate by sharing memory; instead, share memory by communicating."

Obviously `synchronized` is the right thing to do sometimes, but it's used far too often.

usage/launcher/src/main/java/brooklyn/launcher/BrooklynLauncher.java
```diff
@@ -772,6 +782,7 @@ protected void startWebApps() {
             webServer.setPublicAddress(publicAddress);
             if (port!=null) webServer.setPort(port);
             if (useHttps!=null) webServer.setHttpsEnabled(useHttps);
+            webServer.setShutdownRequestListener(shutdownRequestListener);
```
aledsage: I was initilly a little worried about nulls being passed through here (but looks like it's ok at the end, where it's finally used).

If nulls are allowed, then worth including `@Nullable` on the parameter.

usage/rest-server/src/main/java/brooklyn/rest/resources/ServerResource.java
```diff
@@ -171,17 +207,22 @@ public void run() {
...
-                        //give the http request a chance to complete gracefully
+                        //give the http request a chance to complete gracefully, the server will be stopped in a shutdown hook
                         Time.sleep(delayForHttpReturn);
-                        
-                        System.exit(0);
+
+                        if (shutdownRequestListener != null) {
+                            shutdownRequestListener.onShutdownRequest();
+                        } else {
+                            log.warn("ShutdownRequestListener not set, existing process");
```
aledsage: Should be "exiting process", instead of "existing process"?

usage/rest-server/src/main/java/brooklyn/rest/util/ShutdownRequestListener.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package brooklyn.rest.util;
+
+public interface ShutdownRequestListener {
```
aledsage: Should this be called "listener" or "handler"?

I see in `ServerResource.java` that it does not do `System.exit(0)` if there is a listener set, so it feels like it is not just a listener.

usage/rest-server/src/main/java/brooklyn/rest/resources/ServerResource.java
```diff
@@ -191,6 +232,15 @@ public void run() {
                 }
             }
 
+            private boolean hasStoppingApps(ManagementContext mgmt) {
```
aledsage: Should this be called `hasStoppableApps`?

usage/rest-server/src/test/java/brooklyn/rest/BrooklynRestApiLauncher.java
```diff
@@ -235,6 +238,7 @@ private ContextHandler servletContextHandler(ManagementContext managementContext
         ResourceConfig config = new DefaultResourceConfig();
         for (Object r: BrooklynRestApi.getAllResources())
             config.getSingletons().add(r);
+        config.getSingletons().add(new ShutdownRequestListenerProvider(new TestShutdownRequestListener()));
```
aledsage: How does this `new TestShutdownRequestListener` related to the `shutdownListener` field? Does anything use the field?

usage/rest-server/src/test/java/brooklyn/rest/resources/ServerResourceTest.java
```diff
@@ -134,4 +131,52 @@ void testGetMissingConfigThrowsException() throws Exception {
...
+                lock.set(true);
+                lock.notifyAll();
+            }
+        }
+
+        @Override
+        protected void postStop() {
+            super.preStop();
+            synchronized (lock) {
+                if (!lock.get()) {
```
aledsage: See previous comment in ReleaseLatchLocation.release about `synchronized`. Could change to `CountDownLatch`.

usage/rest-server/src/test/java/brooklyn/rest/util/TestShutdownRequestListener.java
```diff
@@ -0,0 +1,39 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package brooklyn.rest.util;
+
+import brooklyn.rest.util.ShutdownRequestListener;
+
+public class TestShutdownRequestListener implements ShutdownRequestListener {
+    boolean isRequested;
```
aledsage: Should declare this volatile, given it is multi-threaded usage.

aledsage:
Thanks @neykov - looks really good!

I've added a bunch of minor comments. Please ping me when you've been through those, and incorporated those that you think should be addressed.

software/base/src/test/java/brooklyn/entity/basic/SoftwareProcessEntityTest.java
neykov: Agree that some code deleting is in order. In this case I explicitly didn't want an SshDriver implementation though (as most of the above are).

usage/cli/src/main/java/brooklyn/cli/Main.java
neykov: Agree, updated the code.

usage/launcher/src/main/java/brooklyn/launcher/BrooklynLauncher.java
neykov: It's a simple setter which for me assumes `@Nullable`. I've already marked the provider constructor. Added the annotation here as well though, can't hurt.

usage/rest-server/src/main/java/brooklyn/rest/resources/ServerResource.java
neykov: All apps are stopping by now, but you are right the suggested name will better reflect the code in it.

usage/rest-server/src/test/java/brooklyn/rest/BrooklynRestApiLauncher.java
neykov: Update code to use the field. The field itself is not used elsewhere, but added it anyway for consistency.

neykov:
Thanks for the great comments @aledsage, should all be addressed now.

aledsage:
Great, thanks @neykov - merging now.

PR #776, opened 2015-07-28
Feature: BYON port mapping

aledsage:
And a few other things! See individual commits.


ygy:
@aledsage  LGTM

Successfully tested for :
- BYON (RHEL) for iptables and sudo
- iptables/sudo with CentOS 6/7, RHEL 6 on AWS

Have to make some more tests in a customer specific environments to fully cover the port mapping/nat scenario

aledsage:
Thanks @ygy - merging.

PR #779, opened 2015-07-29
Fix REST api app’s status as ERROR when ON_FIRE

aledsage:
Previously, if the app’s status was ON_FIRE, the rest api translated
that into “UNKNOWN”. Now it will report it as “ERROR”.


aledsage:
@rdowner can you take a look please? I'm going to do a commit-then-review as this is small and low risk, and I really need it!

rdowner:
+1

PR #780, opened 2015-07-30
Fix LDAP dc parameter

bostko:
Fix LDAP connection parameter.
I tested it locally If the parameters are not formatted in like this `cn=John Smith,ou=Users,dc=example,dc=com` then it fails to recognize the login format `error code 34 - Incorrect DN given`
@grkvlt @aledsage Can you review it please?


aledsage:
Thanks @bostko - looks good. Merging.

PR #781, opened 2015-07-30
BYON: machine gets extra conf for duration of use

aledsage:
See more detailed description in each commit.


bostko:
Good changes in both commits.
We will test it Live to confirm.

aledsage:
Thanks @bostko - merging now.

PR #782, opened 2015-07-30
Add basic sensors to RedisCluster

mikezaccardo:
RedisCluster did not expose any sensors, so this adds basic information about the master node like host address and Redis port.

### DIFF
--- a/software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
+++ b/software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
@@ -21,21 +21,22 @@ package brooklyn.entity.nosql.redis;
 import java.util.Collection;
 
 import brooklyn.enricher.Enrichers;
-import brooklyn.entity.Entity;
 import brooklyn.entity.basic.AbstractEntity;
 import brooklyn.entity.basic.Lifecycle;
 import brooklyn.entity.basic.ServiceStateLogic;
+import brooklyn.entity.basic.ServiceStateLogic.ComputeServiceIndicatorsFromChildrenAndMembers;
 import brooklyn.entity.basic.ServiceStateLogic.ServiceProblemsLogic;
 import brooklyn.entity.group.DynamicCluster;
 import brooklyn.entity.proxying.EntitySpec;
-import brooklyn.entity.trait.Startable;
 import brooklyn.event.AttributeSensor;
 import brooklyn.event.basic.Sensors;
 import brooklyn.location.Location;
+import brooklyn.util.collections.QuorumCheck.QuorumChecks;
 import brooklyn.util.exceptions.Exceptions;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 
 public class RedisClusterImpl extends AbstractEntity implements RedisCluster {
 
@@ -55,6 +56,7 @@ public class RedisClusterImpl extends AbstractEntity implements RedisCluster {
         return getAttribute(SLAVES);
     }
 
+    @Override
     public void init() {
         super.init();
 
@@ -70,6 +72,16 @@ public class RedisClusterImpl extends AbstractEntity implements RedisCluster {
                 .from(master)
                 .build());
     }
+
+    @Override
+    protected void initEnrichers() {
+        super.initEnrichers();
+        ServiceStateLogic.newEnricherFromChildrenUp().
+            checkChildrenOnly().
+            requireUpChildren(QuorumChecks.all()).
+            configure(ComputeServiceIndicatorsFromChildrenAndMembers.IGNORE_ENTITIES_WITH_THESE_SERVICE_STATES, ImmutableSet.<Lifecycle>of()).
+            addTo(this);
+    }
     
     @Override
     public void start(Collection<? extends Location> locations) {
@@ -82,8 +94,6 @@ public class RedisClusterImpl extends AbstractEntity implements RedisCluster {
             ServiceProblemsLogic.updateProblemsIndicator(this, START, "Start failed with error: "+e);
             ServiceStateLogic.setExpectedState(this, Lifecycle.ON_FIRE);
             throw Exceptions.propagate(e);
-        } finally {
-            setAttribute(Startable.SERVICE_UP, calculateServiceUp());
         }
     }
 
@@ -105,8 +115,6 @@ public class RedisClusterImpl extends AbstractEntity implements RedisCluster {
             ServiceProblemsLogic.updateProblemsIndicator(this, STOP, "Stop failed with error: "+e);
             ServiceStateLogic.setExpectedState(this, Lifecycle.ON_FIRE);
             throw Exceptions.propagate(e);
-        } finally {
-            setAttribute(Startable.SERVICE_UP, calculateServiceUp());
         }
     }
 
@@ -118,14 +126,5 @@ public class RedisClusterImpl extends AbstractEntity implements RedisCluster {
     @Override
     public void restart() {
         throw new UnsupportedOperationException();
-    }
-
-    private boolean calculateServiceUp() {
-        if (!Boolean.TRUE.equals(getMaster().getAttribute(SERVICE_UP))) return false;
-
-        for (Entity member : getSlaves().getMembers())
-            if (!Boolean.TRUE.equals(member.getAttribute(SERVICE_UP))) return false;
-
-        return true;
     }
 }
--- a/software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
+++ b/software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
@@ -20,49 +20,73 @@ package brooklyn.entity.nosql.redis;
 
 import java.util.Collection;
 
+import brooklyn.enricher.Enrichers;
 import brooklyn.entity.Entity;
 import brooklyn.entity.basic.AbstractEntity;
 import brooklyn.entity.basic.Entities;
 import brooklyn.entity.group.DynamicCluster;
 import brooklyn.entity.proxying.EntitySpec;
 import brooklyn.entity.trait.Startable;
+import brooklyn.event.AttributeSensor;
+import brooklyn.event.basic.Sensors;
 import brooklyn.location.Location;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 
 public class RedisClusterImpl extends AbstractEntity implements RedisCluster {
 
-    protected RedisStore master;
-    protected DynamicCluster slaves;
+    private AttributeSensor<RedisStore> MASTER = Sensors.newSensor(RedisStore.class, "redis.master");
+    private AttributeSensor<DynamicCluster> SLAVES = Sensors.newSensor(DynamicCluster.class, "redis.slaves");
 
     public RedisClusterImpl() {
     }
 
     @Override
     public RedisStore getMaster() {
-        return master;
+        return getAttribute(MASTER);
     }
     
     @Override
     public DynamicCluster getSlaves() {
-        return slaves;
+        return getAttribute(SLAVES);
     }
     
     @Override
     public void start(Collection<? extends Location> locations) {
-        master = addChild(EntitySpec.create(RedisStore.class));
-        Entities.manage(master);
-        master.start(locations);
+        RedisStore master = getMaster();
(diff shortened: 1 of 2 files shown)


software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
```diff
@@ -80,4 +84,26 @@ protected boolean calculateServiceUp() {
         return up;
     }
 
+    @Override
+    public void init() {
+        super.init();
+
+        subscribe(master, RedisStore.HOSTNAME, new MasterAttributePropagator());
```
aledsage: You could instead use:

    addEnricher(Enrichers.builder()
        .propagating(RedisStore.HOSTNAME, RedisStore.ADDRESS, RedisStore.SUBNET_HOSTNAME, RedisStore.SUBNET_ADDRESS, RedisStore.REDIS_PORT)
        .from(master)
        .build());

aledsage: However, at this point (in `init`), `master` will be null. The reason it seems to work is that `subscribe(null, ...)` is taken to mean a wild card to subscribe to all entities for that sensor! Therefore when master does have its sensor published, we receive it here.

Perhaps better to do this subscription / addEnricher in `start` where the `master` is being set.

software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
```diff
@@ -23,9 +23,13 @@
...
+import brooklyn.entity.basic.EntityInternal;
 import brooklyn.entity.group.DynamicCluster;
 import brooklyn.entity.proxying.EntitySpec;
 import brooklyn.entity.trait.Startable;
+import brooklyn.event.AttributeSensor;
+import brooklyn.event.SensorEvent;
+import brooklyn.event.SensorEventListener;
 import brooklyn.location.Location;
 
 public class RedisClusterImpl extends AbstractEntity implements RedisCluster {
```
aledsage: Also, looking at the existing `RedisClusterImpl` code, the use of fields for `master` and `slaves` is going to cause problems on Brooklyn restart (or HA failover). The fields won't get reset. Those need to be stored as sensors, or the getter methods need to look up the values by searching the children. Caching it in a field is fine, obviously.

Note that HA matters a lot to the customer who is driving this, so we need to test that before we give it to him for his lab.

mikezaccardo:
**This is not yet working properly!**

Stopping and starting the `DynamicCluster` of slaves fails.  When testing on `localhost`, a `LocalhostMachineProvisioningLocation` is created (with an ID value we'll call X) upon the first time that the cluster is started.  I then invoke the "stop" effector in the Brooklyn GUI which stops the master and slave cluster.  I then invoke "start" in the GUI, select "localhost" when prompted for a location, and the slave cluster fails to start.  This occurs because another `LocalhostMachineProvisioningLocation` is created with a new ID (we'll call Y).  `DynamicClusterImpl` line 262 attempts to add this new location to the `locations` parent member variable.  `addLocation()` checks to see if the supplied location already exists, but since the ID of the new localhost location does not match the old one (X != Y), it is added to `locations`.  When the location to deploy to is attempted to be retrieved on `DynamicClusterImpl` line  264, an exception is thrown because more than one location now exists.

@aledsage @neykov or anyone else, do you have any thoughts about how to fix this?

mikezaccardo:
I added a test which validates the error when performing [start / stop / start in new location] on `DynamicClusterImpl`.

neykov:
@mikezaccardo when starting for a second time use `None` location, meaning that it should use the existing one.

software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
```diff
@@ -20,49 +20,73 @@
...
-        master = addChild(EntitySpec.create(RedisStore.class));
+        RedisStore master = getMaster();
+        if (master == null) {
+            master = addChild(EntitySpec.create(RedisStore.class));
+            setAttribute(MASTER, master);
+        }
+
         Entities.manage(master);
-        master.start(locations);
+        master.invoke(RedisStore.START, ImmutableMap.<String, Object>of("locations", ImmutableList.copyOf(locations))).blockUntilEnded();
```
neykov: @mikezaccardo I mislead you to use `blockUntilEnded`. It will not throw exceptions if there was an error while executing the effector. Better use `getUnchecked`.

software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
```diff
@@ -20,49 +20,73 @@
...
     }
 
     @Override
     public void stop() {
-        if (slaves != null) slaves.stop();
-        if (master != null) master.stop();
+        DynamicCluster slaves = getSlaves();
+        RedisStore master = getMaster(); 
+        
+        if (slaves != null) slaves.invoke(DynamicCluster.STOP, ImmutableMap.<String, Object>of()).blockUntilEnded();
```
neykov: Same here, you could invoke both effectors and then check for the return value.

neykov:
Looks good once `getUnchecked` is used.

mikezaccardo:
Thanks @neykov 

Re: `None` location, I did try that before and saw that it worked, so I guess it's just a user experience issue, then.  I thought it was odd that it was using the existing location when `None` was selected, but in this context it really determines which *additional* location.

mikezaccardo:
Addressed

software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
```diff
@@ -20,49 +20,73 @@
...
 import brooklyn.location.Location;
 
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
 public class RedisClusterImpl extends AbstractEntity implements RedisCluster {
 
-    protected RedisStore master;
-    protected DynamicCluster slaves;
+    private AttributeSensor<RedisStore> MASTER = Sensors.newSensor(RedisStore.class, "redis.master");
```
aledsage: We generally declare the config + sensors as `public static final`, and usually on the interface unless it is definitely just internal.

software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
```diff
@@ -20,49 +20,73 @@
...
     }
     
     @Override
     public DynamicCluster getSlaves() {
-        return slaves;
+        return getAttribute(SLAVES);
     }
     
     @Override
     public void start(Collection<? extends Location> locations) {
```
aledsage: Also see the code in things like `DynamicCluster.start(locations)`, which does things like `ServiceStateLogic.setExpectedState(this, Lifecycle.STARTING);`. That gives more visibility of what the entity is doing and better error reporting. The finally block also ensures that we'll correctly report failures, and will stop the green-spinning in the web-console that was shown while starting.

aledsage: If you go down that road of setting starting/running, then I'd probably also separate it out into the two methods - delegate to something like `doStart()` to actually invoke start on the entities.

I think you're free to instantiate the `slave` cluster before calling master.start. That will mean the web-console will show the master + slaves in the tree immediately, rather than waiting for the master to start before the cluster appears.

You could also move the instantiation of the master + slave entities into init. Then you wouldn't need to guard for null either (and don't need to call `Entities.manage()` when inside init - on completion of init, any child entities will immediately get managed).

software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
```diff
@@ -20,49 +20,73 @@
...
         setAttribute(Startable.SERVICE_UP, calculateServiceUp());
+
+        addEnricher(Enrichers.builder()
+                .propagating(RedisStore.HOSTNAME, RedisStore.ADDRESS, RedisStore.SUBNET_HOSTNAME, RedisStore.SUBNET_ADDRESS, RedisStore.REDIS_PORT)
+                .from(master)
+                .build());
     }
 
     @Override
     public void stop() {
```
aledsage: See the pattern used in `DynamicCluster.stop()`, for setting stopping/stopped/on_fire.

software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
```diff
@@ -72,12 +129,12 @@ public void restart() {
...
     }
 
-    protected boolean calculateServiceUp() {
-        boolean up = false;
-        for (Entity member : slaves.getMembers()) {
-            if (Boolean.TRUE.equals(member.getAttribute(SERVICE_UP))) up = true;
-        }
-        return up;
-    }
+    private boolean calculateServiceUp() {
```
mikezaccardo: Is this the appropriate method for calculating `SERVICE_UP`?

Before it would return `true` if one or more slave nodes were up, with no consideration of master.

Now it requires master and all slaves to be up.  Perhaps that is too strict, so perhaps the correct way is somewhere in the middle?

software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
mikezaccardo: In this case the attributes are used exclusively internally in lieu of storing master and slaves as non-persistent member variables, so I have left them as `private static`.  However, if you feel that they should be `public` anyway, I will make that change.

software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
```diff
@@ -72,12 +120,12 @@ public void restart() {
...
+        if (!Boolean.TRUE.equals(getMaster().getAttribute(SERVICE_UP))) return false;
 
+        for (Entity member : getSlaves().getMembers())
```
neykov: You can check the status of the `getSlaves()` entity directly, it will update based on the members state. It's at [least one](https://github.com/apache/incubator-brooklyn/blob/master/core/src/main/java/brooklyn/entity/group/DynamicClusterImpl.java#L171) by default.

neykov: Also can replace the manual checks with enricher (in `initEnrichers`)
```
ServiceStateLogic.newEnricherFromChildrenUp().checkChildrenOnly().requireUpChildren(QuorumChecks.all()).addTo(this);
```

Where `QuorumChecks` is in `brooklyn.util.collections.QuorumCheck`.

software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
```diff
@@ -20,64 +20,111 @@
...
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 
 public class RedisClusterImpl extends AbstractEntity implements RedisCluster {
 
-    protected RedisStore master;
-    protected DynamicCluster slaves;
+    private static AttributeSensor<RedisStore> MASTER = Sensors.newSensor(RedisStore.class, "redis.master");
```
aledsage: I'd declare it `final` as well, given it is a constant.

software/nosql/src/main/java/brooklyn/entity/nosql/redis/RedisClusterImpl.java
```diff
@@ -20,64 +20,111 @@
...
+        } catch (Exception e) {
+            ServiceProblemsLogic.updateProblemsIndicator(this, STOP, "Stop failed with error: "+e);
+            ServiceStateLogic.setExpectedState(this, Lifecycle.ON_FIRE);
+            throw Exceptions.propagate(e);
+        }
+    }
 
-        setAttribute(Startable.SERVICE_UP, false);
+    private void doStop() {
+        getSlaves().invoke(DynamicCluster.STOP, ImmutableMap.<String, Object>of()).getUnchecked();
```
aledsage: This is fine for the PR, but I suspect if there is an error stopping the slaves then the master won't be stopped either. Not sure off-hand what best approach is for that.

aledsage:
Looks good; merging now.

PR #783, opened 2015-07-31
MySql master-slave cluster implementation

neykov:


software/database/src/main/java/brooklyn/entity/database/mysql/MySqlClusterImpl.java
```diff
@@ -0,0 +1,445 @@
...
+        return EntitySpec.create(MySqlNode.class)
+                .displayName("MySql Slave")
+                .configure(MySqlNode.MYSQL_SERVER_ID, serverIdSupplier.get())
+                .configure(MySqlNode.TEMPLATE_CONFIGURATION_URL, SLAVE_CONFIG_URL)
+                // block inheritance, only master should execute the creation script
+                .configure(MySqlNode.CREATION_SCRIPT_URL, (String) null)
+                .configure(MySqlNode.CREATION_SCRIPT_CONTENTS, (String) null);
+    }
+
+    private EntitySpec<?> applyDefaults(EntitySpec<?> spec, Supplier<Integer> serverId, String configUrl, boolean resetCreationScript) {
```
mikezaccardo: Might be nice to have `applyMasterDefaults` and `applySlaveDefaults` methods which call a third helper method with most of this content.

I like the clarity of that over just the `resetCreationScript` variable which is 1:1 with master / slave.

software/database/src/main/java/brooklyn/entity/database/mysql/MySqlClusterImpl.java
```diff
@@ -0,0 +1,445 @@
...
+
+    @Override
+    protected Entity createNode(Location loc, Map<?, ?> flags) {
+        Entity node = super.createNode(loc, flags);
+        if (!IS_MASTER.apply(node)) {
+            ServiceNotUpLogic.updateNotUpIndicator((EntityLocal)node, MySqlSlave.SLAVE_HEALTHY, "Replication not started");
+
+            FunctionFeed.builder()
+                .entity((EntityLocal)node)
+                .period(Duration.FIVE_SECONDS)
```
mikezaccardo: Perhaps extract to `private static` variable in this class to avoid magic number.

software/database/src/main/java/brooklyn/entity/database/mysql/MySqlClusterImpl.java
```diff
@@ -0,0 +1,445 @@
...
+            final MySqlNode node = (MySqlNode) event.getSource();
+            if (Boolean.TRUE.equals(event.getValue()) &&
+                    // We are interested in SERVICE_PROCESS_IS_RUNNING only while haven't come online yet.
+                    // Probably will get several updates while replication is initialized so an additional
+                    // check is needed whether we have already seen this.
+                    Boolean.FALSE.equals(node.getAttribute(MySqlNode.SERVICE_UP)) &&
+                    !Boolean.TRUE.equals(node.getAttribute(NODE_REPLICATION_INITIALIZED))) {
+
+                // Events executed sequentially so no need to synchronize here.
+                if (Boolean.TRUE.equals(node.getAttribute(NODE_REPLICATION_INITIALIZED))) {
```
mikezaccardo: Shouldn't this condition never be true given the opposite condition in the if-statement above? If so then it can be removed.

software/database/src/main/java/brooklyn/entity/database/mysql/MySqlClusterImpl.java
```diff
@@ -0,0 +1,445 @@
...
+            MySqlNode master = (MySqlNode) Iterables.find(cluster.getMembers(), IS_MASTER);
+            String username = validateSqlParam(cluster.getConfig(SLAVE_USERNAME));
+            executeScriptOnNode(master, String.format("DROP USER '%s'@'%s';", username, slaveAddress));
+        }
+
+    }
+
+    // Can't call node.executeScript directly, need to change execution context, so use an effector task
+    private static String executeScriptOnNode(MySqlNode node, String commands) {
+        return node.invoke(MySqlNode.EXECUTE_SCRIPT, ImmutableMap.of("commands", commands)).getUnchecked();
```
mikezaccardo: Perhaps "commands" could live in a `public static String` variable in `MySQLNode` specifying the effector parameter value.

mikezaccardo:
LGTM, integ test passes

software/database/src/main/java/brooklyn/entity/database/mysql/MySqlClusterImpl.java
```diff
@@ -0,0 +1,445 @@
...
+
+            node.addEnricher(Enrichers.builder().updatingMap(Attributes.SERVICE_NOT_UP_INDICATORS)
+                    .from(MySqlSlave.SLAVE_HEALTHY)
+                    .computing(Functionals.ifNotEquals(true).value("Slave replication status is not healthy") )
+                    .build());
+        }
+        return node;
+    }
+
+    public class SlaveStateCallable implements Callable<String> {
```
aledsage: Can we make this inner-class static?

Will the `FunctionFeed` that uses this get persisted? I don't think so because it's not using `AbstractEntity`'s `addFeed(feed)`. Does that mean it will stop populating `SLAVE_HEALTHY` after a brooklyn reboot?

software/database/src/main/java/brooklyn/entity/database/mysql/MySqlClusterImpl.java
```diff
@@ -0,0 +1,445 @@
...
+            if (Boolean.TRUE.equals(slave.getAttribute(MySqlNode.SERVICE_PROCESS_IS_RUNNING))) {
+                return slave.invoke(MySqlNode.EXECUTE_SCRIPT, ImmutableMap.of("commands", "SHOW SLAVE STATUS \\G")).asTask().getUnchecked();
+            } else {
+                return null;
+            }
+        }
+
+    }
+
+    public class SlaveStateParser implements Function<String, Boolean> {
```
aledsage: Again, would it be better if this was static?

software/database/src/main/java/brooklyn/entity/database/mysql/MySqlClusterImpl.java
neykov: In the slave case I'll need to know if the third helper method cloned the spec or not which means duplicating the same checks, essentially having the same method twice.

software/database/src/main/java/brooklyn/entity/database/mysql/MySqlClusterImpl.java
neykov: good catch

software/database/src/main/java/brooklyn/entity/database/mysql/MySqlClusterImpl.java
neykov: +1 for static

Good point re `addFeed`. I thought it's done on `feed.setEntity`, but it's behind an `isEnabled(FEATURE_FEED_REGISTRATION_PROPERTY)` check, disabled by default.

software/database/src/main/java/brooklyn/entity/database/mysql/MySqlClusterImpl.java
neykov: I don't see how it would be less magical, unless we make it configurable, but not worth it at this point.

neykov:
Addressed comments, thanks @mikezaccardo, @aledsage.

neykov:
Build failure caused by package name changes, updated PR.

PR #784, opened 2015-07-31
LDAP authentication doc

bostko:
Added an example brooklyn.properties for ldap


aledsage:
Thanks @bostko - merging.

@neykov @grkvlt what are implications of us searching in the LDAP tree in LDAP://cn=John Smith,ou=Users,dc=example,dc=com

Do we risk that not working for some people's LDAP/AD etc?

Longer term, I think we want to change entirely how we do this authentication (e.g. using Apache Shiro), which I know @grkvlt as strong opinions about. That is one for the brooklyn mailing list.

PR #768, opened 2015-07-24
Docs troubleshooting

ahgittin:
Fix merge conflicts from #765 (moving @nakomis 's "exceptions" troubleshooting) and attempt to update titles so menu is more intuitive.  @aledsage feedback welcome.  BTW merging this should close #765 due to the commit history.

PR #785, opened 2015-07-31
Fix lots of broken links in the troubleshooting section

rdowner:
Most from PR #768 which moved a Markdown page but didn't move the images it depended on. A couple more broken links and typos fixed also (thanks to @sjcorbett and @nakomis)


aledsage:
Looks good; merging.

PR #786, opened 2015-07-31
Fix/rebind machine lifecycle effector tasks

aledsage:
As discussed in https://github.com/apache/incubator-brooklyn/pull/763.

Fixes backwards compatibility failure in persistence by re-adding the anonymous inner classes to `MachineLifecycleEffectorTasks` (but they are no longer instantiated, so new things will use the new class names).


neykov:
Looks good.

aledsage:
Thanks @neykov - merging  now.

PR #788, opened 2015-08-04
Rename base package from brooklyn to org.apache.brooklyn - 3

CMoH:
This time starting from the root projects.


hzbarcea:
Tested and reviewed. Good to merge.

PR #792, opened 2015-08-05
PostgreSQL lifecycle improvements

ygy:
- Moving database's initialization to the install phase
- Skipping install phase when node was provisioned, stopped via its 
  effectors and then started
- Only Postgres service is stopped when stop effector is invoked
- VM stays up so the stopped Postgres service can be started again
  on the same machine

### DIFF
--- a/software/database/src/main/java/brooklyn/entity/database/postgresql/PostgreSqlNode.java
+++ b/software/database/src/main/java/brooklyn/entity/database/postgresql/PostgreSqlNode.java
@@ -90,4 +90,6 @@ public interface PostgreSqlNode extends SoftwareProcess, HasShortName, Datastore
 
     String executeScript(String commands);
 
+    ConfigKey<StopSoftwareParameters.StopMode> STOP_MACHINE_MODE = ConfigKeys.newConfigKeyWithDefault(StopSoftwareParameters.STOP_MACHINE_MODE, StopSoftwareParameters.StopMode.NEVER);
+
 }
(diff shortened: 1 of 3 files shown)


aledsage:
Background context for others... A Brooklyn user has a number of entities in the dev/test environment, including `PostgreSqlNode`. For each, they stopped the process (e.g. using the `stop` effector with `stopProcessMode=ALWAYS` and `stopMachineMode=NEVER`). They then called the `start` effector, expecting that to restart the process. For most types of entities it worked (because the `install` and `customize` phases were idempotent. However, for `PostgreSqlNode` the customize phase executed `initdb` (which you don't want to do on an existing database).

The user should have called the `restart` effector with parameters `restartMachine=false`. This would have just done the `launch` phase.

However, the `restart` effector has the downside of first stopping the process (if it is currently running). What the user really wants is an effector for "launch to process if it's not already running; if it is running then do nothing".

aledsage:
Thanks @ygy 

I have a few concerns with these changes:

* The behaviour of `PostgreSqlNode` is now different from that of all other `SoftwareProcess` entities. We should strive for consistency - particularly for the semantics of the operations/attributes, and for the default behaviours.
* The PostgreSQL `STOP_MACHINE_MODE ` default is changed to NEVER. However, that means when an app is terminated (recursively calling `stop` on the entities within it), the VM will be left running (with all references to that VM being removed).
* The PostgreSQL "install" phase now does more than just install the software. It feels to me like the execution of `initdb` is part of customising the install.

For the last point (difference between install, customize and launch), this is particularly important for entities written to work well on localhost for integration tests - for many entities you could have multiple instances running simultaneously, or you could configure + launch a new test instance without it being affected by previous runs (assuming you hadn't changed version numbers etc).

This difference becomes more vague as we move to installing things in conventional places (e.g. `/opt/apache-tomcat-8.0.24`, rather than `~/brooklyn-managed-processes/...`).

aledsage:
@ygy I therefore propose that we close this PR. If we want to be able to parameterise an effector so that it has the behaviour "launch to process if it's not already running; if it is running then do nothing", then let's propose that on the Apache Brooklyn mailing list.

Thoughts?

ygy:
Agree.

Closing this PR

PR #793, opened 2015-08-05
Rename base package from brooklyn to org.apache.brooklyn

CMoH:
Refs BROOKLYN-162

Update package names of brooklyn-cli. Also fix brooklyn-qa after new code was merged.

mvn clean install works. Also deployed some blueprints successfully.


hzbarcea:
Looks good, tested, reviewed, fixed conflicts due to renames.

PR #794, opened 2015-08-05
Address remaining comments from #782 (Redis cluster sensors)

mikezaccardo:
By catching any exceptions thrown by stopping the slaves or by stopping master, both actions are guaranteed to be attempted.

`MultiException` is used because it allows us to capture and propagate both exceptions in the event that both the slaves and master fail to stop.


mikezaccardo:
See #795

PR #797, opened 2015-08-05
package names: usage/cli and usage/camp

aledsage:


hzbarcea:
Tested, looks good, merging.

PR #798, opened 2015-08-05
BasicStartable entities transition through expected lifecycle states

sjcorbett:
And adds useful test class `RecordingSensorEventListener`.

### DIFF
--- a/core/src/main/java/brooklyn/entity/basic/BasicStartableImpl.java
+++ b/core/src/main/java/brooklyn/entity/basic/BasicStartableImpl.java
@@ -30,6 +30,7 @@ import brooklyn.entity.trait.Startable;
 import brooklyn.entity.trait.StartableMethods;
 import brooklyn.location.Location;
 import brooklyn.management.Task;
+import brooklyn.util.exceptions.Exceptions;
 
 import com.google.common.base.Predicates;
 import com.google.common.collect.Iterables;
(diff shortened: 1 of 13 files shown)


core/src/main/java/brooklyn/entity/basic/BasicStartableImpl.java
```diff
@@ -38,38 +39,47 @@
...
+            ServiceStateLogic.setExpectedState(this, Lifecycle.ON_FIRE);
+            throw Exceptions.propagate(t);
         }
-        for (Task<?> t: tasks) t.getUnchecked();
     }
 
     @Override
     public void stop() {
+        ServiceStateLogic.setExpectedState(this, Lifecycle.STOPPING);
         StartableMethods.stop(this);
```
aledsage: should this be wrapped in a try-catch, so that we can do `ServiceStateLogic.setExpectedState(this, Lifecycle.ON_FIRE);`? For example, see `DynamicClusterImpl.stop()` and `AbstractApplication.stop()`.

core/src/test/java/org/apache/brooklyn/entity/basic/RecordingSensorEventListener.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    /**
+     * @return An immutable iterable of the recorded events.
+     */
+    public Iterable<SensorEvent<T>> getEvents() {
+        return ImmutableList.copyOf(events);
+    }
+
+    public Iterable<T> getEventValues() {
```
aledsage: Deserves a comment that this is a live (read-only) view of the event values. At first I didn't see how this worked with `Suppliers.ofInstance(listener.getEventValues())`, but after looking at what a `FluentIterable`'s `iterator()` method does, this is an elegant solution!

aledsage:
Looks great. Only a couple of very minor comments. Good to merge once you've looked at those.

sjcorbett:
@aledsage Thanks for your comments. In #790 @alasdairhodge raised the idea of having the listener in the test-support module. I'm happy to make that change if considered appropriate.

aledsage:
Thanks @sjcorbett - I'm happy to merge now. We can discuss separately if that listener should be in test-support.

PR #800, opened 2015-08-06
[BROOKLYN-162] Renaming of the NoSQL packages

ygy:
- From brooklyn.entity.nosql to org.apache.brooklyn.entity.nosql


examples/simple-nosql-cluster/src/main/java/org/apache/brooklyn/demo/CumulusRDFApplication.java
```diff
@@ -138,8 +139,8 @@ public void initApp() {
             cassandra = addChild(EntitySpec.create(CassandraFabric.class)
                     .configure(CassandraDatacenter.CLUSTER_NAME, "Brooklyn")
                     .configure(CassandraDatacenter.INITIAL_SIZE, getConfig(CASSANDRA_CLUSTER_SIZE)) // per location
-                    .configure(CassandraDatacenter.ENDPOINT_SNITCH_NAME, "brooklyn.entity.nosql.cassandra.customsnitch.MultiCloudSnitch")
-                    .configure(CassandraNode.CUSTOM_SNITCH_JAR_URL, "classpath://brooklyn/entity/nosql/cassandra/cassandra-multicloud-snitch.jar")
+                    .configure(CassandraDatacenter.ENDPOINT_SNITCH_NAME, "org.apache.brooklyn.entity.nosql.cassandra.customsnitch.MultiCloudSnitch")
```
aledsage: Note that this will require us re-building the jar for MultiCloudSnitch (which shouldn't even be in brooklyn!).

See the comment at https://github.com/apache/incubator-brooklyn/pull/802#issuecomment-128424637

Also see docs/_extra/big_examples/nosql-cassandra/cassandra.include.md lines 191 and 205, where the `MultiCloudSnitch` is mentioned. We should do a global search for "MultiCloudSnitch".

Don't worry about it here. Let's fix it by moving MultiCloudSnitch somewhere more sensible in a completely different PR!

aledsage:
Merging now.

PR #802, opened 2015-08-06
package rename to org.apache.brooklyn: sandbox

aledsage:


aledsage:
Note that `MultiCloudSnitch`'s package has been renamed, but the jar has not been rebuilt (it is at `software/nosql//src/main/resources/brooklyn/entity/nosql/cassandra/cassandra-multicloud-snitch.jar` - not good!).

Really this code needs to be updated and contributed to Apache Cassandra, and deleted from brooklyn!
Note that the code is in sandbox, so is not part of the Brooklyn release.

In the short term, I have left references to the `MultiCloudSnitch` fully qualified classname with the old name.

hzbarcea:
Reviewed, tested, GTM, merging.

hzbarcea:
Reviewed, tested, GTM, merging.

PR #804, opened 2015-08-07
Fix Cassandra multicloud snitch following package renaming to org.apache

ygy:
- cassandra-multicloud-snitch.jar rebuilt with the new packages
- updating the pom.xml to exclude commons-codec, because of dependency
  convergence issues


hzbarcea:
Tested, reviewed, GTM, merging. This will probably have to be redone as package renaming progresses. Ideally we shouldn't have jars in the sources, a task we may want to tackle after better osgi support is in place.

PR #811, opened 2015-08-11
brooklyn-software-network: add org.apache package prefix

CMoH:
Tested; 1 integration test fails (BindDnsServerIntegrationTest.testRebindDns), but it does so also on master.

Refs BROOKLYN-162


aledsage:
Thanks @CMoH - merging.

PR #814, opened 2015-08-12
brooklyn-software-osgi: add org.apache package prefix

CMoH:
Refs BROOKLYN-162

All tests that are part of the Integration profile are marked WIP.
I did not run any live tests.

There are 2 more issues for this project:

1. src/test/resources contains a hello-world.jar that is used as test OSGi bundle. Shouldn't this be part of the osgi-test-bundles processed by hadrian?

2. Talking of split packages: the sources also contain the classes org.osgi.jmx.Item and org.osgi.jmx.JmxConstants, each with the following note:

```java
/*
 * BROOKLYN NOTE: This is a verbatim copy of the original in org.osgi.enterprise-4.2.0, 
 * reproduced here to avoid pulling in the rest of that bundle and its chain of dependencies.
 * The only difference is this comment and the suppress warnings.
 */
```

As far as I can see org.osgi.enterprise-4.2.0 has no dependencies, though it does bundle a lot more classes.

### DIFF
--- a/software/osgi/src/main/java/org/apache/brooklyn/entity/osgi/karaf/KarafContainer.java
+++ b/software/osgi/src/main/java/org/apache/brooklyn/entity/osgi/karaf/KarafContainer.java
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package brooklyn.entity.osgi.karaf;
+package org.apache.brooklyn.entity.osgi.karaf;
 
 import java.net.URISyntaxException;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


CMoH:
I've added a commit with my proposal to remove the org.osgi.enterprise copies, and add it as dependency. Merge if OK, or I can remove it from the PR.

hzbarcea:
Looks good. For org.osgi.enterprise the version should have been a property actually (easier to manage) and the scope should be <scope>provided</provided>. That said, the second commit would fit  better in the context of the karaf upgrades, so my preference would be to remove the second commit for now and just leave the first one in this PR.

CMoH:
I removed the extra-commit.

hzbarcea:
GTM, tested, merging.

PR #818, opened 2015-08-13
[BROOKLYN-162] Rename ./api/event; use the no split package convention

hzbarcea:


hzbarcea:
@asfbot tests pass, merging.

PR #819, opened 2015-08-13
[BROOKLYN-162] Rename ./api/management; use the no split package conv…

hzbarcea:
…ention


hzbarcea:
@asfbot tests pass, merging.

PR #823, opened 2015-08-13
license automation

ahgittin:
update the LICENSE with add'l reqs, and automate the process

@rdowner does the new format look good to you?


ahgittin:
BTW one reason for adding the preface to LICENSE is that otherwise people will think the doc is the ASL (the first 300 lines look the same!), not the Brooklyn LICENSE

software/database/pom.xml
```diff
@@ -107,18 +107,10 @@
...
-            <groupId>com.beust</groupId>
-            <artifactId>jcommander</artifactId>
-        </dependency>
-        <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.google.code.findbugs</groupId>
-            <artifactId>jsr305</artifactId>
```
aledsage: Why remove this? It's used for `@Nullable`, which is used in `DatastoreMixins` etc. It's also a transitive dependency from brooklyn-api etc.

I believe the convention that @rdowner implemented was that if a package directly uses a class then it includes the dependency as an explicit dependency.

I can't remember whether that is still the convention - I vaguely remember a discussion and some disagreement about it previously. Personally, I have no strong feelings except that we are consistent.

@rdowner can you comment please?

rdowner: I prefer to see explicit dependencies rather than transitive dependencies. It makes intent clearer, it means there's no surprises if the transitive dependencies change, and it reduces the noise from `mvn dependency:analyze`.

In the past I did go through and make all dependencies explicit - it made the initial process of dependency license discovery with `mvn dependency:analyze` much easier back when we were doing due diligence ahead of joining the incubator. I've not been following it up with further checks, though, so it probably has slipped.

Other transitive dependencies supply `@Nullable` however - jclouds has an implementation - so perhaps `DatastoreMixins` is referring to a non-JSR305 `@Nullable`? If so that is where the fix should be made.

rdowner:
@ahgittin sorry I'm a bit late reviewing this. I'm looking through it just now and more comments may follow.

First off there's an issue with the root `LICENSE` file. This file is the LICENSE for the source distribution, and the source distribution does *not* include the big list of dependencies - only a small number of dependencies are directly bundled into the source distribution.

See https://brooklyn.incubator.apache.org/v/latest/dev/code/licensing.html#about-license-and-notice-files for some guidance.

rdowner:
Could SPDX[1][2] be any use in handling this problem?

The state of license analysis Maven plugins is pretty awful, so writing a new plugin to help with this is a good idea. However it'd be preferable to be standards-based so I'd been keen to know if your Maven plugin could handle SPDX metadata instead of making its own format?

[1]https://spdx.org/about-spdx
[2]https://lwn.net/Articles/644876/

usage/dist/licensing/README.md
```diff
@@ -0,0 +1,57 @@
...
+        -DoverridesFile=licensing/overrides.yaml \
+        -DextrasFile=licensing/extras.yaml
+
+
+To create the LICENSE file in the root and in the dist:
+
+    pushd licensing > /dev/null
+    ./make-license.sh > LICENSE.autogenerated || ( echo 'FAILED!!!' && rm LICENSE.autogenerated )
+    cp LICENSE.autogenerated ../../../LICENSE
+    cp LICENSE.autogenerated ../src/main/license/LICENSE
```
rdowner: See my comment on the main PR. This is an issue. The source and binary distributions have *different* `LICENSE` files, because they bundle different dependencies. Each distribution needs to have a `LICENSE` file that is accurate for that distribution, without superfluous entries.

This approach is great for the binary distribution, but the source distribution will need to be handled differently. Given the relatively small number of dependencies bundled into the source distribution, I'm happy for that to be done manually, if that is the simplest approach.

usage/dist/licensing/extras.yaml
```diff
@@ -0,0 +1,179 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+# extras file for org.heneveld.license-audit-maven-plugin
+# supplying info for addl licenses included with brooklyn
```
rdowner: So this file is for supplying information on non-Maven dependencies? If so could you state that here?

usage/dist/licensing/README.md
```diff
@@ -0,0 +1,57 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+This uses a special maven plugin to create license and notification information,
+based on metadata in overrides.yaml and extras.yaml.
+
+First install  https://github.com/ahgittin/license-audit-maven-plugin
```
rdowner: Any chance of getting this into Maven Central? I'd prefer to avoid instructions like "download this and install it, then..." (although since this plugin is only loosely bound into the project, it's not a big issue.)

rdowner:
Aside from the change to the root `LICENSE` and the unrelated `pom.xml` changes, I'm happy with this PR. Handling license compliance in a project with as many dependencies as Brooklyn is a pain and adding this automation simplifies part of that process.

I have a good wishlist of what a license-helping tool would do! As I said earlier, established tooling is thin on the ground and I can easily see this and your Maven plugin developing into a comprehensive tool that could be re-used by many projects. The questions is do you want to consider that just now, or stick to the simplest thing that could possible work for Brooklyn :smile:

software/database/pom.xml
ahgittin: i agree, with richard (and maven recommendation) that projects declare their compile-time deps -- will restore.  somehow missed the fact this was being pulled in from other projects.

usage/dist/licensing/extras.yaml
ahgittin: will do

usage/dist/licensing/README.md
ahgittin: eventually - don't think this is a high priority atm though

ahgittin:
@rdowner thanks for detailed review.  i will:

* sort out the source from binary dist LICENSE files (thanks for detailed notes)
* comment in `extras.yaml`
* restore findbugs
* add TODO about investigating SPDX -- i don't see easy wins but it's certainly more comprehensive than what maven or i have done here

as for making this a comprehensive tool, the best thing i think would be to contribute it to the maven-dependency-plugin.  but that would take some major refactoring; i'm content to leave that for the time being.  if there is interest in this then maybe.

ahgittin:
@rdowner do we need the custom NOTICE files in `jsgui` and `cli`?  they are the same as the base one.

rdowner:
Yes, due to a funny in the way that the Maven resource copier plugin works. By overriding it's configuration we're actually *replacing* it's configuration, so we need to configure both files - otherwise LICENSE gets copied and NOTICE goes missing.

ahgittin:
@rdowner thanks - makes sense

so what i think the process should be:

* the script in `dist` will build *four* licenses:
  * root source license
  * `dist` binary license
  * source licenses for `jsgui` and `cli` JARs
* the dist's `overrides.yaml` contains all the metadata for JS (so it just needs to be in one place)
* there will be an `extras.yaml` file for `jsgui` and `cli` listing the deps
* `dist` will not have an `extras.yaml` file but the script will be able to take multiple such (the jsgui and cli ones)

the source build will omit maven dependencies (new `onlyExtras` flag or similar on the `notices` target)

BTW i'm opposed to doing it manually because already we've drifted (e.g. bloodhound is removed)

i'll also update the `licensing.md` with instructions/pointers, and look through your dist's LICENSE (which I hadn't noticed was different) for anything I've left out.

any comments on this?

rdowner:
Sounds perfect :smile: 
Will it be easy to add another project like `jsgui` and `cli` - e.g. if a different artifact in our project wants to start bundling a dependency?

ahgittin:
yes - that's important

ahgittin:
This now generates source licenses differently to that of the binary.  The /LICENSE contains a superset of notices put in individual JARs (from `source-inclusion.yaml`).  The dist's LICENSE also includes all the maven dependencies.

We don't currently list the files against the licenses they reference; this may be a regression but I think there is enough information that this shouldn't be a problem, and having it automated is going to help prevent drift in future.  Note that JAR's and JS file contain license/attribution in most cases (so license/notices may be included twice -- but that's good I think).

@rdowner you happy?

rdowner:
Tested and ensured that source, binary-dist, and jsgui and cli Maven artifacts all contain individualised licenses.

The database `pom.xml` change is still there. If you can drop that - then I will be a happy bunny and you can go ahead and merge :rabbit:

ahgittin:
well spotted on findbugs -- git messed up, fixed now.

i will merge.  :)

ahgittin:
merged/closed against an out-of-date branch :( -- but i've reverted those and applied the correct ones

PR #826, opened 2015-08-14
[BROOKLYN-162] Move location api to the right package location

hzbarcea:
Last packages in ./api


hzbarcea:
@asfbot tests pass, merging.

PR #830, opened 2015-08-17
Fix missing org.apache.brooklyn imports

aledsage:
- In sandbox and in groovy files


aledsage:
Merging now before there are more merge conflicts. Only changes sandbox and a groovy file that is in examples.

hzbarcea:
The @asfbot has some issues and is currently unreliable. I tested and reviewed, it's a good merge. Because the  sandbox is not part of the build, it may be easier to fix all at the end, but incremental fixes are good too. A big of extra work, but limit the scope of bisect. Thanks @aledsage.

PR #832, opened 2015-08-17
add the creation of default database and user with generated password.

robertgmoss:
Also publishes these credentials as sensors


software/database/src/main/java/brooklyn/entity/database/postgresql/PostgreSqlSshDriver.java
```diff
@@ -294,7 +295,24 @@ public void customize() {
 
         // Wait for commands to complete before running the creation script
         DynamicTasks.waitForLast();
-
+        String createUserCommand = String.format(
+                "\"CREATE USER %s WITH PASSWORD '%s'; \"",
+                entity.getConfig(PostgreSqlNode.USERNAME), getUserPassword()
```
sjcorbett: We need to escape everything that's user configurable.

robertgmoss:
@sjcorbett added escapes using StringEscapes#escapeSql is this sufficient?

sjcorbett:
@robertgmoss Could you resolve the merge conflicts please?

software/database/src/main/java/brooklyn/entity/database/postgresql/PostgreSqlSshDriver.java
```diff
@@ -294,7 +296,25 @@ public void customize() {
...
+        String createUserCommand = String.format(
+                "\"CREATE USER %s WITH PASSWORD '%s'; \"",
+                StringEscapes.escapeSql(entity.getConfig(PostgreSqlNode.USERNAME)), 
+                StringEscapes.escapeSql(getUserPassword())
+        );
+        String createDatabaseCommand = String.format(
+                "\"CREATE DATABASE %s OWNER %s\"",
+                StringEscapes.escapeSql(entity.getConfig(PostgreSqlNode.DATABASE)),
+                StringEscapes.escapeSql(entity.getConfig(PostgreSqlNode.USERNAME)));
+        newScript("initializing user and database")
```
sjcorbett: This should be optional.

software/database/src/main/java/brooklyn/entity/database/postgresql/PostgreSqlSshDriver.java
```diff
@@ -307,6 +327,17 @@ public void customize() {
         // on port 5432?" error then the port is probably closed. Check that the firewall allows external TCP/IP
         // connections (netstat -nap). You can open a port with lokkit or by configuring the iptables.
     }
+    
+    protected String getUserPassword() {
+        String password = entity.getConfig(PostgreSqlNode.PASSWORD);
+        if (Strings.isEmpty(password)) {
+            log.debug(entity + " has no password specified for " + PostgreSqlNode.PASSWORD + "; using a random string");
+            password = brooklyn.util.text.Strings.makeRandomId(8);
+            entity.setAttribute(PostgreSqlNode.PASSWORD, password);
```
sjcorbett: Why set the attribute too? It would be quite confusing to `setAttribute(password)` but have it overwritten when `getUserPassword` is called. Perhaps `USERNAME` and `PASSWORD` are just `ConfigKeys` that are set for you if not given?

robertgmoss:
@sjcorbett I've made this optional as you have suggested, and I take your point about the attribute, but my client program (Brooklyn Service Broker) expects to get credentials through the sensors.  I've taken out the redundant setting of the config though.

ahgittin:
looks good

software/database/src/main/java/org/apache/brooklyn/entity/database/postgresql/PostgreSqlSshDriver.java
```diff
@@ -294,7 +294,9 @@ public void customize() {
 
         // Wait for commands to complete before running the creation script
         DynamicTasks.waitForLast();
-
+        if(entity.getConfig(PostgreSqlNode.INITIALIZE_DB)){
```
sjcorbett: This will throw a null pointer exception if `INITIALIZE_DB` is null. You can avoid this with the slightly ungainly `if (Boolean.TRUE.equals(getConfig(..))`.

robertgmoss:
@sjcorbett addressed in #836

PR #834, opened 2015-08-17
Adding --replacepkgs parameter to ensure proper epel install

ygy:
- rpm -Uvh exits with 1 if it is already installed
- replacepkgs ensures that the package will be properly updated
  or replaced if already installed

The proposed fix is needed for compatibility between CentOS and RHEL.
RedHat doesn't have EPEL repo enabled by default, but CentOS has.

### DIFF
--- a/software/messaging/src/main/java/brooklyn/entity/messaging/rabbit/RabbitSshDriver.java
+++ b/software/messaging/src/main/java/brooklyn/entity/messaging/rabbit/RabbitSshDriver.java
@@ -100,7 +100,7 @@ public class RabbitSshDriver extends AbstractSoftwareProcessSshDriver implements
 
         List<String> commands = ImmutableList.<String>builder()
                 // EPEL repository for erlang install required on some Centos distributions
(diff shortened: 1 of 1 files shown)


aledsage:
Thanks @ygy 

Just to clarify, does it work previously on RHEL (which does not have EPEL), but fails on CentOS (which does have EPEL repo enabled)?

I'm slightly hesitant about use of `--replacepkgs`. Any implications for if we are upgrading/downgrading the EPEL packages, for other stuff installed on the OS?

On balance, I think it's fine to merge but would appreciate other's opinions: @grkvlt @richardcloudsoft @ahgittin

ahgittin:
Share slight hesitation but agree better than not doing it.

This is a good argument for using YAML to describe this, so people have increased visibility into what is happening.

ygy:
@ahgittin thanks for merging

@aledsage yes, it worked on RHEL

There will be no implications during upgrades/downgrades. The rpm packages is installed with parameters "-Uvh" (upgrade or install), but not "ivh" (only install).

aledsage:
This fixes https://issues.apache.org/jira/browse/BROOKLYN-159

PR #836, opened 2015-08-18
add check to avoid null pointer exception

robertgmoss:


aledsage:
Looks good - nice and simple! merging.

PR #845, opened 2015-08-19
Fix launch classpath on Windows

alasdairhodge:
Adds missing 'lib' segment to library 'patch' and 'dropins' dirs


sjcorbett:
@alasdairhodge Your commit hasn't touched dropins. Have you pushed everything?

alasdairhodge:
Thanks, @sjcorbett; now pushed!

ahgittin:
LGTM.  And this shouldn't conflict with the package renaming :) so good to merge.

PR #760, opened 2015-07-21
Pluggable YAML parsers

neykov:
* Changes brooklyn-core not to depend on camp-base. Instead plugins (in this case offered by brooklyn-camp) are responsible for parsing the input and creating EntitySpecs from it. The idea is to use the approach for different types of inputs, not just camp yaml, while avoiding pulling in dependencies into core.
* Extract all the code that uses PdpMatcher.parseDeploymentPlan + AssemblyTemplateSpecInstantiator.createSpec/instantiate pairs in a common place - a lot of code repetition

Supersedes #100, #436, #545, #689.

PR #846, opened 2015-08-19
YAML pluggable camp

ahgittin:
rebases #760 with new package structure. builds on #844 so merge that first!
(it's just the last commit here)


ahgittin:
Now that #844 is merged this is clean.  @neykov any comments?

core/src/main/java/org/apache/brooklyn/core/plan/PlanNotRecognizedException.java
```diff
@@ -0,0 +1,40 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.core.plan;
+
+public class PlanNotRecognizedException extends RuntimeException {
+    private static final long serialVersionUID = 1L;
+
+    public PlanNotRecognizedException() {
```
aledsage: Delete the no-arg constructor.

core/src/main/java/org/apache/brooklyn/core/plan/PlanNotRecognizedException.java
```diff
@@ -0,0 +1,40 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.core.plan;
+
+public class PlanNotRecognizedException extends RuntimeException {
+    private static final long serialVersionUID = 1L;
```
aledsage: If going to include a serialVersionUID, then might as well let IDE generate a good one?

core/src/main/java/org/apache/brooklyn/core/plan/PlanToSpecTransformer.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.apache.brooklyn.api.catalog.CatalogItem;
+import org.apache.brooklyn.api.entity.Application;
+import org.apache.brooklyn.api.entity.EntitySpec;
+import org.apache.brooklyn.api.internal.AbstractBrooklynObjectSpec;
+import org.apache.brooklyn.core.mgmt.ManagementContextInjectable;
+
+import com.google.common.annotations.Beta;
+
+public interface PlanToSpecTransformer extends ManagementContextInjectable {
```
aledsage: Worth adding some javadoc

core/src/main/java/org/apache/brooklyn/core/plan/PlanToSpecTransformer.java
```diff
@@ -0,0 +1,35 @@
...
+    String getName();
+    @Beta
+    boolean accepts(String mime);
+    <T extends Application> EntitySpec<T> createApplicationSpec(String plan);
```
aledsage: I never really like the use of generics like this - it basically allows the caller to avoid a cast by pretending that the method actually does something with the generic type passed in.

But no strong feelings.

core/src/main/java/org/apache/brooklyn/core/plan/PlanToSpecTransformer.java
```diff
@@ -0,0 +1,35 @@
...
+    AbstractBrooklynObjectSpec<?, ?> createCatalogSpec(CatalogItem<?, ?> item);
```
aledsage: Can we use generics here, such as:

    <T,SpecT> AbstractBrooklynObjectSpec<T, SpecT> createCatalogSpec(CatalogItem<T, SpecT> item);

Or does it get much more complicated (e.g. needing something horrendous like `<T, SpecT extends AbstractBrooklynObjectSpec<T, SpecT>>`)?

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/CampCatalogUtils.java
```diff
@@ -0,0 +1,244 @@
...
+import org.apache.brooklyn.util.stream.Streams;
+import org.apache.brooklyn.util.text.Strings;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+
+public class CampCatalogUtils {
+
+    public static AbstractBrooklynObjectSpec<?, ?> createSpec(ManagementContext mgmt, CatalogItem<?, ?> item) {
```
aledsage: Why put all of this in `CampCatalogUtils`, rather than in `CampToSpecTransformer`?

aledsage:
Looks really good - have taken an initial look at the overall structure. My comments are mostly minor. Happy for you to merge once you've read through my comments.

ahgittin:
Have addressed, and will push.  Agree with all comments, except:

* generics, I applied this but it doesn't help much as we abandon the generics in the impl; the Transformer interface is not marked `@Beta` as a class so we can experiement; also, we might want to support ZIPs and other items in future
* use of `CampCatalogUtils` - no strong feelings, this is how @neykov did it originally.  it can move back and forth.  (this project should not really be part of the public API.  when we do OSGi we can make it `export-package: none` due to `ServiceLoader`!)

pushing and merging now

PR #847, opened 2015-08-19
Adding docs - stop/start/restart behaviour

ygy:


ahgittin:
Good stuff @ygy .  Does `index.md` in the same dir need to be updated to point to this file?  And should it have the jekyll "front matter" like the other files?

(IOW have you tested this by running `jekyll serve` in the docs dir?!)

ygy:
Thanks @ahgittin 
All suggestions have been addressed

PR #851, opened 2015-08-19
Skip Maven install for the OSGi test bundles

rdowner:
These cannot be installed into a repository and uploaded to Nexus as they do not use the same version number as Brooklyn. They have been breaking our Jenkins CI builds as Nexus refuses to accept non-SNAPSHOT-versioned artifacts into the snapshot repository. We would also likely encounter versioning conflicts for release, too.

This change sets `maven.install.skip` on the affected modules so they are not installed into the local repository, therefore Jenkins will not attempt to deploy them into the remote repository.


ahgittin:
Makes sense.  Will merge.  See #810 for background and the long-term fix.

Sorry for this trouble!!!

PR #852, opened 2015-08-19
Resuming locations

sjcorbett:
* `JcloudsLocation` implements `SuspendResumeLocation` (rather than just `SuspendsLocations`).
* The various `rebindMachine` methods in `JcloudsLocation` are deprecated in favour of `registerMachine`, which returns a `MachineLocation` rather than a `JcloudsSshMachineLocation`. This means that it can also be used to register Windows-based systems.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -1903,9 +1903,9 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
      * <p>
      * Required fields are:
      * <ul>
-     *   <li>id: the jclouds VM id, e.g. "eu-west-1/i-5504f21d" (NB this is @see JcloudsSshMachineLocation#getJcloudsId() not #getId())
+     *   <li>id: the jclouds VM id, e.g. "eu-west-1/i-5504f21d" (NB this is {@see JcloudsMachineLocation#getJcloudsId()} not #getId())
      *   <li>hostname: the public hostname or IP of the machine, e.g. "ec2-176-34-93-58.eu-west-1.compute.amazonaws.com"
-     *   <li>userName: the username for ssh'ing into the machine
+     *   <li>userName: the username for sshing into the machine if it is not a Windows system
      * <ul>
      */
     public MachineLocation registerMachine(ConfigBag setup) throws NoMachinesAvailableException {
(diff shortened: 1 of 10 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1861,16 +1891,41 @@ protected UserCreation createUserStatements(@Nullable Image image, ConfigBag con
...
-        return rebindMachine(MutableMap.of(), metadata);
+        return (JcloudsSshMachineLocation) registerMachine(metadata);
     }
+
+    protected MachineLocation registerMachine(NodeMetadata metadata) throws NoMachinesAvailableException {
+        return registerMachine(MutableMap.of(), metadata);
+    }
+
+    /**
+     * @deprecated since 0.8.0 use {@link #registerMachine(Map, NodeMetadata)} instead.
```
aledsage: Worth including in the javadoc about the return type, and that will get a `ClassCastException` if try to register a windows VM with this method? No strong feelings.

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/JcloudsLocationRegisterMachineLiveTest.java
```diff
@@ -63,12 +63,12 @@ public void testRebindWithIncorrectId() throws Exception {
     }
     
     @Test(groups = { "Live" })
-    public void testRebindVm() throws Exception {
+    public void testRegisterVm() throws Exception {
```
aledsage: Also worth testing registering a pre-existing Windows VM.

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/JcloudsLocationSuspendResumeMachineLiveTest.java
```diff
@@ -0,0 +1,63 @@
...
+        super.setUp();
+        jcloudsLocation = (JcloudsLocation) managementContext.getLocationRegistry()
+                .resolve(AWS_EC2_PROVIDER + ":" + AWS_EC2_EUWEST_REGION_NAME);
+    }
+
+    @Test(groups = "Live")
+    public void testObtainThenSuspendThenResumeMachine() throws Exception {
+        MachineLocation machine = obtainMachine(ImmutableMap.of(
+                "imageId", EUWEST_IMAGE_ID));
+        assertTrue(machine instanceof JcloudsSshMachineLocation);
```
aledsage: I have personal preference for just doing the cast, rather than first asserting. The cast will fail with an exception saying what type it actually was. However, the `assertTrue` doesn't have a message so will just say something like "false".

aledsage:
Thanks @sjcorbett - looks great. Only a few very minor comments.

Before rebasing, you might want to wait for https://github.com/apache/incubator-brooklyn/pull/854 to be merged. Otherwise if https://github.com/apache/incubator-brooklyn/pull/854 gets merged before this, you'll probably end up with yet more import conflicts.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
sjcorbett: Good idea. It doesn't throw a `ClassCastException`, it throws an `IllegalStateException` in `JcloudsUtil#getFirstReachableAddress`.

aledsage:
@sjcorbett ping me when you've rebased, and I'll merge this.

sjcorbett:
@aledsage Thanks for your comments. I've rebased. Unit tests pass locally. Thanks

sjcorbett:
I'll merge this.

PR #853, opened 2015-08-19
Refactor: move packages

aledsage:
As discussed on #844 , this PR does the subsequent changes that were agreed in the doc https://docs.google.com/document/d/1un7SwvoE43lWaIlD5CRdIdza4XJkHji-OhTKXpkr0NA (but not yet everything in that doc).

See individual commits descriptions for the actual changes being done.


ahgittin:
@aledsage great stuff here

two changes to this I'd like to see, if you're okay with them:

* `software/base/src/main/java/org/apache/brooklyn/sensor/ssh/SshEffectorTasks.java` now in `core`, `org/apache/brooklyn/util/core/task/ssh/`:  both seem wrong; unlike the rest of `util/core` which has only minimal dependencies on the rest of core, this is designed to create effectors; so maybe `core`, `org/apache/brooklyn/effector/core` is best?

* in `utils/groovy/src/main/java/org/apache/brooklyn/util/groovy/internal/` can we drop the `internal` while we're doing this?


i'd also like to try to draw a line under refactoring as much as possible soon, as it is disruptive for people.  (should be far less so now, but even so!)  i like the way `o.a.b.entity.*` is entities and same for `location` and `policy`, whereas supporting classes are `o.a.b.core.{entity,location,policy}`.  with this in mind i'm no thinking:

* it makes sense with effectors, since everything is support, for them to be under `o.a.b.core.effector`
* with sensors likewise, supporting things going in `o.a.b.core.sensor` makes sense; the feeds and enrichers i'm on the fence.  `o.a.b.core.sensor.feed.*` is long-winded but okay.  or any of `o.a.b.feed` or `o.a.b.feed` or `o.a.b.sensor.feed` or `o.a.b.core.feed` seem fine (and same for `enricher`), no strong feelings, maybe a slight preference for the last ones.

any other areas which you think need major attention?

(btw i see that you've changed references in yaml etc wrt this PR, all looks fine.  i'm happy for this to be merged before the above suggestions or after, as you see fit.  big improvement.)

aledsage:
Thanks for the comments @ahgittin - I'll merge this PR, and then do the changes in another PR tonight. They will be additional changes, except for moving `SshEffectorTasks`.

PR #854, opened 2015-08-19
Refactor: move more packages

aledsage:
See individual commit descriptions for explanation of what has changed.

See @ahgittin comment in https://github.com/apache/incubator-brooklyn/pull/853 and also the doc https://docs.google.com/document/d/1un7SwvoE43lWaIlD5CRdIdza4XJkHji-OhTKXpkr0NA that was originally written #844


aledsage:
For other package renames / class moves that I'd like to see (or at least think about more):

* More review of `org.apache.brooklyn.util.core.*` for what we can/should move out of core.  
  What are your thoughts on brooklyn-util-common getting bigger, and ending up with lots more dependencies  
  (e.g. we could move .osgis and .crypto, but that would give osgi.jar and bouncycastle.jar dependencies).

* Think about org.apache.brooklyn.util.core.task (and other task classes):  
  Do we want to move those into their own brooklyn-task module?

* core.mgmt.persist versus core.mgmt.rebind: figure out exactly what is the difference, and move classes accordingly.

* org.apache.brooklyn.camp.brooklyn.api.HasBrooklynManagementContext:  
  can we move this, and get rid of the package?

* org.apache.brooklyn.enricher.stock: I wonder if these should be in brooklyn-policy instead of brooklyn-core.  
  But the o.a.b.entity.stock makes use of it, so probably not.

* brooklyn-rest-server: no common prefix (except o.a.b.rest, which is also used by rest-api and rest-swagger).

* Initializers such as AddEffector etc: not sure if these should be gathered somewhere, or if it's good to leave them in the package most related to what they do.  
  There are still some anomalies such as `SshCommandEffector` in `o.a.b.sensor.ssh`.

* `org.apache.brooklyn.entity.system_service`: do we want to remove the underscore?

* .internal package names.
  Would be nice if that meant what OSGi expects - i.e. package is not accessible to any other modules (except "friend", but use that sparingly).

* A definition of our "public" api, and thus what our deprecation policy applies to.  
  I completely agree that YAML is most important moving forwards.
  But if are free to refactor back-end stuff (without deprecating) then it will make our lives much simpler.

ahgittin:
looks great.  merging.  to your other questions:

* `org.apache.brooklyn.camp.brooklyn.api.HasBrooklynManagementContext` - let's move this to be alongside `ManagementContextInjectable`
* underscore - this is becoming more common with python ascending but no strong feelings
* `SshCommandEffector` move to `o.a.b.effector.core.ssh`
* enrichers to `brooklyn-policy` - i think not

the others i think are less important and/or less clear, and for now we should do nothing with them.  when we osgify things hopefully many of the packages can avoid being exported, and these we can move to `internal` sub-packages.  (and if any `internal` ones are used, we fix those.)  i am less concerned with doing that later on as it will not be very disruptive.

PR #855, opened 2015-08-20
Skip Maven deploy for the OSGi test bundles

rdowner:
Like #851, but uses a different property (that one didn't work).

These cannot be installed into a repository and uploaded to Nexus as they do not use the same version number as Brooklyn. They have been breaking our Jenkins CI builds as Nexus refuses to accept non-SNAPSHOT-versioned artifacts into the snapshot repository. We would also likely encounter versioning conflicts for release, too.

This change sets `maven.deploy.skip` on the affected modules so they are not installed into the local repository, therefore Jenkins will not attempt to deploy them into the remote repository.

### DIFF
--- a/utils/test-bundles/entities/pom.xml
+++ b/utils/test-bundles/entities/pom.xml
@@ -41,6 +41,7 @@
 
     <properties>
         <maven.install.skip>true</maven.install.skip>
+        <maven.deploy.skip>true</maven.deploy.skip>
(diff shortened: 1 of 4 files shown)


aledsage:
Thanks @rdowner for fixing!

@hzbarcea can you take a look at this as well please.

It looks sensible to me - merging.

PR #858, opened 2015-08-20
Adds test for ssh streams being captures in VanillaSoftwareProcess

aledsage:
Also adds `TaskPredicates`.

And adds `TaskBuilder.displayName()` (deprecating the old `TaskBuilder.name()`) so that it is consistent with `Task.getDisplayName()`.


ahgittin:
LGTM.  i like the change to `displayName`.

is there a failure which prompted this?  or just adding good integration tests?

ahgittin:
noticed `TaskPredicatesTest` is missing the RAT header; i'll add it when i push

aledsage:
@ahgittin thanks.

What prompted it is us adding similar support for `VanillaWindowsProcess`, which currently does not capture the streams. It made sense to write the test for the ssh-case while @bostko works on the windows side of things.

PR #859, opened 2015-08-20
Put OSGi test deps back

ahgittin:
This updates all the OSGi and JAR tests which were moved and disabled in #810.  They are in the original places for want of anything better as discussed there.  @rdowner this should save you any more build-fix-version-mismatch woes (though now likely you've solved them)!

Tests not yet passing, for early review only


ahgittin:
tests seem fixed now, for me.  @hzbarcea does this fix things for you?

ahgittin:
the cassandra failure in #1721 is concerning, it doesn't look remotely relevant to this work /cc @aledsage 

the one after it is happy so this PR is good to go

aledsage:
@ahgittin Thanks - looks good; merging.

hzbarcea:
I reviewed and tested too, @aledsage beat me to it.

PR #860, opened 2015-08-20
Illustration and test for PlanToSpecTransformer, and tidies

ahgittin:
It's pretty straightforward to write a new plan-to-spec transformer.  This paves the way for TOSCA support, among other things.

I wonder in time about this being bi-directional, i.e. also usable for persisting and rebinding.

/cc @aledsage @neykov

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/EntityManagementUtils.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/EntityManagementUtils.java
@@ -18,8 +18,6 @@
  */
 package org.apache.brooklyn.core.mgmt;
 
-import java.util.ArrayList;
-import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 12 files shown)


ahgittin:
BTW @neykov @andreaturli excellent job on the original work; very easy to work with

Great to have CAMP entirely pluggable, through a tiny interface!

hzbarcea:
There is no problem with this PR, that is a test issue that occurs rarely. I'll take a look at it.
Reviewed, tested, merging.

neykov:
Looks good - a nice improvement.
I don't see the value of using `attemptWithLoaders` vs passing the string directly, but it's just a personal preference (don't like the syntax overhead of anonymous classes).
Also the `OVERRIDE` test property is a bit smelly, but agree it's the best we can do without introducing even more abstraction.

PR #861, opened 2015-08-20
Adds Entity.sensors() to group the sensor methods

aledsage:
- The attribute/sensor methods are all in Entity.sensors().
- Deprecates the previous things in Entity.* and EntityLocal.*,
  except for getAttribute(AttributeSensor) which is left as a convenience.
- In EntityInternal.sensors(), it returns SensorSupportInternal
  which has more methods on it.

There are three big motivations for this:
- Consistency with entity.config()
- Simplification of the Entity class (which has way too many methods)
  by grouping them.
- Exposing the methods of EntityLocal on Entity, so we can 
  deprecate EntityLocal (this commit deals with just the attribute/sensor
  methods of EntityLocal)

Note that I haven't changed any of the uses of the old attribute methods. I wanted to get feedback first as to whether people agreed this was the right thing to do!

Do people agree that `sensors()` is the best name? We could call it `attributes()`?

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/entity/Entity.java
+++ b/api/src/main/java/org/apache/brooklyn/api/entity/Entity.java
@@ -27,15 +27,21 @@ import org.apache.brooklyn.api.effector.Effector;
 import org.apache.brooklyn.api.location.Location;
 import org.apache.brooklyn.api.mgmt.Task;
 import org.apache.brooklyn.api.objs.BrooklynObject;
+import org.apache.brooklyn.api.objs.Configurable.ConfigurationSupport;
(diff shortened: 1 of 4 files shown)


hzbarcea:
Reviewed, tested, I was thinking of merging last night, but waited for more community input per @aledsage request. Thinking about it I think I am a bit more biased towards ```attributes()```, that express the ```what``` vs ```sensors()``` referring to the ```how```. Slight semantic difference. I am fine with either though.

sjcorbett:
Sticking with autonomic terminology I am in favour of `sensors`.

hzbarcea:
Lazy consensus. Merging.

PR #864, opened 2015-08-23
Entity proxy unwraps InvocationTargetException

alasdairhodge:
Propagates *original* exception thrown by actual entity.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/objs/proxy/EntityProxyImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/core/objs/proxy/EntityProxyImpl.java
@@ -20,6 +20,7 @@ package org.apache.brooklyn.core.objs.proxy;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


alasdairhodge:
Test failure appears unrelated.

alasdairhodge:
Motivated by a desire to write test assertions that my (custom) entities were correctly throwing `IllegalStateException` and other undeclared runtime exceptions in the right situations.

ahgittin:
I usually use `Exceptions.getFirstInteresting(...)` or `e.toString().contains("...")`.  Also note `Exceptions.collapse` will prevent `InvocationTargetException`s and similar from being shown.

In general I think we're better off with that approach (caller cleans up) rather than attempting to clean up intermediate.  (Note if execution is remote, when that's available, this "clean-up" approach will be brittle.)  But if you really want this no objections.

ahgittin:
@alasdairhodge please confirm if you'd like this or close

aledsage:
@alasdairhodge as discussed offline, I don't like it when the exception thrown in one thread is then rethrown by another thread, without wrapping it.

If that is done, the exception thrown in the user's thread won't include the user's stacktrace, so the user won't be able to tell what in their code was executing when the exception happened. The avoidance of that confusion justifies keeping the `InvocationTargetException`.

alasdairhodge:
Yes, you've convinced me that there's value in keeping the ITE, confusing as it may be in naive use-cases. Happy to close this PR.

PR #865, opened 2015-08-23
[BROOKLYN-166] add codemirror, yaml syntax-highlighting, autocompletion)

azbarcea:
I was looking into the `codemirror` integration.

I didn't integrate it in all the views, only in `application-add-wizard.js`, and I look forward to your feedback.

I tried to improve the `scroll down` and now, there aren't 2 scrollers. There is only one for the editor.

The following things are still open:
* I excluded `rat-plugin` for the files added (actually, right now excludes the whole `src`).
* I had to modify the original codemirror addons and modes to include `codemirror` instead of `../../lib/codemirror`
* The way `angular.js` views and `require.js` is used (using the `config.js`), make the presumptions the `javascript` libraries are static in `libs`. Similar to `mvn`, I propose `grunt` and `npm` to be used, so that the whole `libs` directory to disappear.
* I have no `yaml-hint.js`, or even `camp-hint.js`. I need to study further the CAMP spec for it. I think also that `brooklyn` has a little different spec for `.yaml` files, extending the language. I am right? Right now, I used the `anyword-hint.js`, which adds to the hints (auto-completion), the words that already exist in the document.
* There are many themes that can be added, I will also add a configuration for: addons, modes, themes to be used globally. Maybe this should be at user profile level ?!?


azbarcea:
Testing from `usage/dist/target/brooklyn-dist/brooklyn/bin/brooklyn launch` doesn't work.

Tested with 

```shell
org.apache.brooklyn.rest.jsgui.BrooklynJavascriptGuiLauncher -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005
```

I'll take a look at what's missing, tomorrow

tbouron:
Tested, it works great. There are indeed two scroll but that should be easily fixable with a bit of CSS. Also, the editor is only one line height, could we extend to the height of the pop-up by default?

Regarding the autocompletion, my view on that is first to pull out the catalogue entities when we are to a `type:` line. We can even imagine to have specific hints based on where we are, like having all the provisioning properties after a `provisioning.properties:` and so on.

hzbarcea:
@tbouron, regarding extending to the hight of the pop-up, we could integrate the tree view you proposed a while back. I don't know if that could be done without enlarging the pop-up which will impact the other tabs. Unless we make it a different pop-up or view. WDYT?

ahgittin:
A few comments:

* If I have nothing in the catalog, it shows me YAML editor and it looks nice.  Surprising at first how lines are added but it's nice.  Cut and paste and other things work as expected.
* If I then click on the "catalog" tab, with nothing there, i have option to write YAML blueprint.  I click that button and it returns me to the YAML tab, but a *new* editor is added.  This looks like a bug.
* If I leave the wizard, go to the catalog and add some YAML (e.g. default.catalog.bom from CLI project), then open the add wizard, I have the catalog tab selected (good), but then when I go to the YAML editor I get the *old* YAML view.

Of course we need RAT re-enabled.

Code completion will require some REST API enhancements or better understanding of the model in the client.  I'd also like to see this as a separate main tab, rather than a wizard.  But we can leave these for a follow-on PR, as with the bug fixes before this paragraph resolved, this is a nice improvement.

azbarcea:
@ahgittin - I will take a look to the bug. Also keep in mind that is not integrated in all the views, neither the _events_ aren't propagated.

@tbouron, is there a documentation for the CAMP syntax/semantics (for hint) besides the [spec](http://docs.oasis-open.org/camp/camp-spec/v1.1/camp-spec-v1.1.html)?

ahgittin:
There is a Brooklyn camp yaml reference on the website. Most useful thin
would be autocomplete for catalog item types and their configuration keys
both of which are exposed through rest api.
On 26 Aug 2015 19:15, "Alexandru Zbarcea" <notifications@github.com> wrote:

> @ahgittin <https://github.com/ahgittin> - I will take a look to the bug.
> Also keep in mind that is not integrated in all the views, neither the
> *events* aren't propagated.
>
> @tbouron <https://github.com/tbouron>, is there a documentation for the
> CAMP syntax/semantics (for hint) besides the spec
> <http://docs.oasis-open.org/camp/camp-spec/v1.1/camp-spec-v1.1.html>?
>
> —
> <https://github.com/apache/incubator-brooklyn/pull/865#issuecomment-135128315>
> .
>

-- 
 
message after it was sent.

CMoH:
I think it's this page, right?

https://brooklyn.incubator.apache.org/v/latest/yaml/yaml-reference.html

tbouron:
Yes @CMoH, this is the page that @hzbarcea was looking for.

tbouron:
@hzbarcea To be honest, I'm not so sure anymore about the drag and drop editor. It is certainly a good feature for a sales point of view but I'm wondering about the usability. I think that having a good YAML editor with a good autocompletion feature over Brooklyn CAMP spec is a better approach for now.

We could also imagine to have something a bit like swagger, i.e. having a live representation of the YAML as a tree or nodes. That would be great!

ahgittin:
I think a drag-and-drop editor is very important for new users, especially if it generates the YAML which will help them move to become advanced users.

ahgittin:
@azbarcea can you address the issues blocking this being merged?

ahgittin:
ping @azbarcea

azbarcea:
@ahgittin, @CMoH - fixed CodeMirror integration.

Please review. I look forward to your feedback.

azbarcea:
fixed merge ... with latest code base

usage/jsgui/src/main/webapp/assets/css/brooklyn.css
```diff
@@ -25,7 +25,7 @@ BODY {
 textarea {
     white-space: pre;
     word-wrap: normal;
-    overflow-x: scroll;
+/*    overflow-x: scroll; */
```
tbouron: This line can be removed

usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -591,10 +607,14 @@ define([
...
-                    return true;
-                }
+            	if (self.editor && self.editor) {
+            		this.model.yaml = self.editor.getValue();
+	        		if (this.model.yaml) {
+	                    return true;
+	        		}
+            	} else {
+            		console.info("No text in the editor!");
+            	}
```
tbouron: This block has a wrong indentation.

usage/jsgui/src/main/webapp/assets/js/util/brooklyn-utils.js
```diff
@@ -113,6 +113,7 @@ define([
         if ($input.attr("type") === "checkbox") {
             return $input.is(":checked");
         } else {
+        	// TODO: get codemirror.getValue() in case this is a textarea ...
```
tbouron: This line has a wrong indentation.

usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -564,6 +579,7 @@ define([
         },
 
         validate:function () {
+        	log ("validate");
```
tbouron: This line has a wrong indentation.

usage/jsgui/src/main/webapp/assets/js/view/catalog.js
```diff
@@ -134,54 +135,62 @@ define([
...
             this.$("#catalog-add-form").html(template);
         },
         beforeClose: function () {
             if (this.contextView) this.contextView.close();
         },
+        setupCodeEditor: function() {
+        	if (this.editor === null) {
+            	this.editor = CodeMirror.fromTextArea(document.getElementById("new-blueprint"), {
+            		height: "150px",
+                	lineNumbers: true,
```
tbouron: The 2 lines above have a wrong indentation.

usage/jsgui/src/main/webapp/assets/js/view/catalog.js
```diff
@@ -134,54 +135,62 @@ define([
...
+                    extraKeys: {"Ctrl-Space": "autocomplete"},
+                    textWrapping: true,
+                    mode: {name: "yaml", globalVars: true}
+            	});
+
+                this.editor.setValue("# Please add your blueprint here\n");
+        	}
+        },
         showContext: function(event) {
+        	log("showContext");
```
tbouron: This line has a wrong indentation.

azbarcea:
I fixed the indentation ... any thoughts?

tbouron:
LGTM

neykov:
* `bin/brooklyn launch` doesn't work
* RAT is still disabled (`**//src/*`)
* Re-opening the dialog shows the old textarea

usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -180,14 +183,13 @@ define([
...
-            var finishEnabled = finishVisible
+            var finishVisible = (this.currentStep >= 1);
+            var finishEnabled = finishVisible;
             if (!finishEnabled && this.currentStep==0) {
                 if (this.model.mode == "yaml") {
                     // should do better validation than non-empty
                     finishVisible = true;
-                    var yaml_code = this.$("#yaml_code").val()
-                    if (yaml_code) {
+                    if (self.editor && self.editor && self.editor.getValue()) {
```
neykov: Second `self.editor` is not necessary.

usage/jsgui/src/main/license/source-inclusions.yaml
```diff
@@ -38,3 +38,5 @@
 - id: js-uri
 - id: js-yaml.js
 - id: jquery.form.js
+
+- id: codemirror.js
```
neykov: Should add addon & mode files as well?

usage/jsgui/src/main/webapp/assets/js/config.js
```diff
@@ -48,6 +50,15 @@ require.config({
         "zeroclipboard":"libs/ZeroClipboard",
         "js-yaml":"libs/js-yaml",
         
+        "codemirror":"lib/codemirror",
```
neykov: From: https://codemirror.net/doc/manual.html#modloader
`Do not use RequireJS' paths option to configure the path...`

usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -304,6 +306,9 @@ define([
                 this.renderCurrentStep(function callback(view) {
                     // Drop any "None" locations.
                     that.model.spec.pruneLocations();
+
+                    // TODO: see if this preview has conflicts with codemirror. maybe the id="code"
+                    //       has to be switched back to yaml_code
```
neykov: Looks like this needs to be addressed before merging?

usage/jsgui/src/main/webapp/assets/js/util/brooklyn-utils.js
```diff
@@ -113,6 +113,7 @@ define([
         if ($input.attr("type") === "checkbox") {
             return $input.is(":checked");
         } else {
+            // TODO: get codemirror.getValue() in case this is a textarea ...
```
neykov: Either delete or address.

neykov:
Also move `lib/codemirror.js`, `addon/*`, `mode/*` to `libs/codemirror` folder.

usage/jsgui/src/main/webapp/assets/js/config.js
azbarcea: this is true. This is the reason I put ```addon```, ```mode``` in the lib, and not in the codemirror folder. I tried modifying the ```config.js``` as explained in doc but it didn't work. 

Let me know if it works for you ...

azbarcea:
I cannot move ```lib/codemirror.js```,``` addon/*``` to ```libs/codemirror``` folder, see explanation above ...

CMoH:
`bin/brooklyn launch` works - perhaps there was some external reason it didn't on you machine, @neykov

usage/jsgui/src/main/license/source-inclusions.yaml
ahgittin: no need to add the codemirror dependencies here, but we do need a corresponding entry in `usage/dist/licensing/overrides.yaml` and there under `files` list the files pulled in (as done for `jquery.dataTables`) -- it's only used to generate the proper license files

usage/jsgui/src/main/webapp/assets/css/brooklyn.css
```diff
@@ -25,7 +25,6 @@ BODY {
 textarea {
     white-space: pre;
     word-wrap: normal;
-    overflow-x: scroll;
```
ahgittin: i think we rely on textareas scrolling e.g. when we show a blueprint or other formatted code in a textarea.

it perhaps would be better to have a special style for this but if we're removing this style have you made sure that no textareas suffer bad consequences?  if needed either update them or if you need a different `overflow-x` behaviour in the codemirror views define a new css rule.

usage/jsgui/src/main/webapp/index.html
```diff
@@ -26,6 +26,10 @@
 
     <title>Brooklyn JS REST client</title>
 
+	<link rel="stylesheet" href="/assets/css/codemirror.css">
+    <link rel="stylesheet" href="/assets/css/addon/hint/show-hint.css">
+    <link rel="stylesheet" href="/assets/css/complete.css">
```
ahgittin: `complete.css` is not included, gives 404's at runtime

ahgittin:
also note that the catalog yaml syntax is different, as described at https://brooklyn.incubator.apache.org/v/latest/ops/catalog/index.html .  it allows you to paste a simple blueprint or several such.  not a blocker for this but important when completion is enabled; maybe for now a quick win would be to have a sample template with the right syntax, or comments showing the right syntax.

i confirm `bin/brooklyn launch` works nicely.  and switching between tabs in the "add application" wizard causes no problems (but it isn't using the new codemirror any more.)

newly found issues:

* check/fix `overflow-x: scroll` for other textareas

* add license info for codemirror in `overrides.yaml`

* the JS console gives an error trying to load `/assets/css/complete.css`

others still outstanding:

* if possible, put 3rd party deps into `libs/` like the others, but if that's too hard leave as is

* the rat check needs to scan all files or have an explanation.  if it's really really hard to have the codemirror files live under `libs/` then call them out explicitly here with a comment explaining why they have the unusual structure

* update the two `TODO` comments, or expand on who/when they'll get done

if we could add this to the wizard -- or completely rewrite the wizard as part of drag-and-drop -- then that would be good, but not needed in this PR.

i think very soon we should start on a new UI based on angular.

usage/rest-server/src/test/java/org/apache/brooklyn/rest/BrooklynRestApiLauncher.java
```diff
@@ -314,6 +314,7 @@ public static void main(String[] args) throws Exception {
     public static Server startRestResourcesViaFilter() {
         return new BrooklynRestApiLauncher()
                 .mode(StartMode.FILTER)
+                .forceUseOfDefaultCatalogWithJavaClassPath(true)
```
ahgittin: we don't want to do this by default, it slows things down.  can you revert?

ahgittin:
@azbarcea what's the status of this?

the latest changes address the licensing and rat issues, however still outstanding are:

* check/fix `overflow-x: scroll` for other textareas
* error in JS at rungime trying to load `/assets/css/complete.css`
* 3rd party deps into libs/ like the others (if not too hard, else comment)
* provide more info where you have added TODO comments

plus new issue:

* revert auto-scanning of classpath for REST API; it significantly increases startup time and typically we don't want it (you can trigger a scan by POSTing a suitable `catalog.bom` if you need to, or add those explicit items you wish to have); if you think it should be there, please give more details

usage/jsgui/src/main/webapp/index.html
sjcorbett: + these files should not be referenced here. They should be included in [styles.css](https://github.com/apache/incubator-brooklyn/blob/master/usage/jsgui/src/main/webapp/assets/css/styles.css) so that they can be concatenated into one file at build time.

usage/rest-server/src/test/java/org/apache/brooklyn/rest/BrooklynRestApiLauncher.java
azbarcea: Well this is the only way I know to see something in the catalog and test brooklyn.  `$ bin\brooklyn launch` uses `.forceUseOfDefaultCatalogWithJavaClassPath(true)`. Don't you want the test to be identical with the way is used? I think it slows down the build (the tests), but has no other consequence.

Do you have any suggestions?

azbarcea:
I am addressing all the issues above ...

ahgittin:
@azbarcea many of those issues don't look like they've been addressed.  what's the latest?

azbarcea:
Having issues with the current Modal Wizzard and adding the CodeMirror editor, I'm proposing having the Editor as a tab next to Home.

apachebrooklyn-editor-blueprint

apachebrooklyn-editor-detail

The proposal and the updates will continue on this PR.

ahgittin:
@azbarcea +1.  maybe call it "Composer" rather than "Editor", and I think don't show the catalog on this tab (we have the catalog tab).

azbarcea:
@ahgittin I'll rename it to composer! +1

Regarding the Catalog, I think is a good proposal (+1), and I can have propose it as a different PR, based also on the community feedback.

ahgittin:
merge conflicts with recent swagger changes but i've resolved those; now reviewing

ahgittin:
a lot of `log` statements in the JS -- probably okay but looks a little weird if someone opens the JS console.  maybe logging should be disabled by default (ie `log(...)` has no effect) but there would be an easy way to enable logging.

usage/jsgui/src/main/webapp/index.html
```diff
@@ -26,6 +26,11 @@
 
     <title>Brooklyn JS REST client</title>
 
+    <!-- TODO: following CSS to config.js -->
```
ahgittin: what does this mean?

usage/jsgui/src/main/license/files/LICENSE
```diff
@@ -353,6 +353,13 @@ This project includes the software: ZeroClipboard
   Used under the following license: The MIT License (http://opensource.org/licenses/MIT)
   Copyright (c) Jon Rohan, James M. Greene (2014)
 
+This project includes the software: CodeMirror
```
ahgittin: looks like you manually added this?  this file is autogenerated based on `overrides.yaml` which is where this data should be added.  (and it will update other `LICENSE` files as well because we are obligated to include info this not just in the UI license file but in the source and binary dists as well.)

ahgittin:
the following issues are still outstanding:

* check/fix `overflow-x: scroll` for textareas, likely others depend on this style you've removed -- https://github.com/apache/incubator-brooklyn/pull/865/files#diff-2c00b2e9ba84934a94b937199763e780L28
* 3rd party deps into `libs/` like the others, not in `lib/` and `addons/` (or say why not)
* revert auto-scanning of classpath for REST API; it significantly increases startup time and typically we don't want it (you can trigger a scan by POSTing a suitable catalog.bom if you need to, or add those explicit items you wish to have); if you think it should be there, please give more details -- https://github.com/apache/incubator-brooklyn/pull/865/files#diff-b8eba103a5475b653aaba9dd4fe8827cR55

new ones:

* UI does not run:  `:8081/assets/js/router.js:39 Uncaught TypeError: serverStatus.addCallback is not a function` (and we get the same error with JS jasmine tests: `[ERROR] java.lang.RuntimeException: org.openqa.selenium.WebDriverException: com.gargoylesoftware.htmlunit.ScriptException: Wrapped com.gargoylesoftware.htmlunit.ScriptException: Wrapped com.gargoylesoftware.htmlunit.ScriptException: TypeError: Cannot find function addCallback in object [object Object]. (http://localhost:57004/src/js/router.js#39)`)

* license metadata needs to be added to `overrides.yaml` (the `LICENSE` files are autogenerated from this) - https://github.com/apache/incubator-brooklyn/pull/865/files#diff-001a0bdd771ea5cf50f47cd08ce6a0dcR356

ability to test is limited by the above issues!

usage/jsgui/src/main/webapp/assets/js/router.js
```diff
@@ -18,14 +18,15 @@
 */
 define([
     "brooklyn", "underscore", "jquery", "backbone",
+    "codemirror",
```
ahgittin: this is what is breaking the run -- we don't use it locally so we can remove it I think; but if it's added of course there needs to be a corresponding `CodeMirror` argument to the following `function`

ahgittin:
i've fixed the problem with startup.  now looks much better, and tests work.

however:
* i thought "Editor" was going to be called "Composer"
* when using the pop-up wizard there should be a flow by which i can go to the composer populated with the YAML for what i've built (this is what the `preview` button typically does; maybe we should have the preview enabled even on the first step of the wizard if i have something selected)
* on catalog page, add, if i click "yaml" to add a yaml catalog item, nothing happens -- i think it should take me to the blueprint editor, perhaps pre-populated with a sample item a la http://brooklyn.apache.org/v/latest/ops/catalog/index.html (NB catalog has a different syntax than normal deployments; it also accepts the normal deployment syntax), or see below
* do we want a "Help" button on the composer, e.g. with a dropdown that has "insert template blueprint", "insert template catalog addition", "go to online documentation"
* on the editor page what is "Delete" supposed to do?  currently seems to do nothing.  perhaps remove?  or comment out with TODO if you think it might do something in future.  (i'm not sure it does.)
* on the editor page the "Run" button doesn't work; it constructs a POST but it is malformed
* no feedback is given after clicking "Run"; on success it should take you to the "Applications" tab with the app highlighted (you have the app ID in the response code so that should be easy), and on failure it should show a message
* there should be an "Add to Catalog" button on the editor/composer which has same feedback after clicking, taking you to catalog page (same as if you clicked the old catalog add yaml button)

ahgittin:
my fix for the `router.js` failure and merge conflicts is at https://github.com/ahgittin/incubator-brooklyn/tree/ui

@azbarcea I suggest you merge it in here

usage/jsgui/src/main/webapp/index.html
azbarcea: this line means that the CSS from index.html should be minimized as well ... I will address this issue!

usage/jsgui/src/main/license/files/LICENSE
azbarcea: I didn't know that. I will address this issue too ...

azbarcea:
merged with @ahgittin `ui` branch ... addressing the above issues   ...

PR #867, opened 2015-08-25
Sanitizes the classpath in CatalogClasspathDo to prevent issue when launching brooklyn with jsvc

nakomis:


hzbarcea:
Pretty straightforward, reviewed, tested, ok. Merging.
Adding a unit test would have been good too.

PR #868, opened 2015-08-26
Adding timestamp to the template options' name

ygy:
- timeStamp replaces the previously used randId. 
- timeStamp uses the standard unix timestamp represented
  as a 8-char hex string.
- It represents the moment in time when the name is constructed. 
- It gives the possibility to search easily for instances, security
  groups, keypairs, etc based on timestamp without complicated
  enumeration

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/location/cloud/names/BasicCloudMachineNamer.java
+++ b/core/src/main/java/org/apache/brooklyn/core/location/cloud/names/BasicCloudMachineNamer.java
@@ -41,9 +41,15 @@ public class BasicCloudMachineNamer extends AbstractCloudMachineNamer {
         StringShortener shortener = Strings.shortener().separator("-");
         shortener.append("system", "brooklyn");
         
-        // randId often not necessary, as an 8-char hex identifier is added later (in jclouds? can we override?)
-        // however it can be useful to have this early in the string, to prevent collisions in places where it is abbreviated 
-        shortener.append("randId", Identifiers.makeRandomId(4));
+        /* timeStamp replaces the previously used randId. 
+         * 
+         * timeStamp uses the standard unix timestamp represented as a 8-char hex string.
+         * 
+         * It represents the moment in time when the name is constructed. 
+         * It gives the possibility to search easily for instances, security groups, keypairs, etc
+         * based on timestamp without complicated enumeration        
+         */ 
+        shortener.append("timeStamp", Long.toString(System.currentTimeMillis() / 1000L, 16));
         
         String user = System.getProperty("user.name");
         if (!"brooklyn".equals(user))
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/brooklyn/core/location/cloud/names/BasicCloudMachineNamer.java
```diff
@@ -41,9 +41,15 @@ protected String generateNewIdOfLength(ConfigBag setup, int len) {
...
-        shortener.append("randId", Identifiers.makeRandomId(4));
+        /* timeStamp replaces the previously used randId. 
+         * 
+         * timeStamp uses the standard unix timestamp represented as a 8-char hex string.
+         * 
+         * It represents the moment in time when the name is constructed. 
+         * It gives the possibility to search easily for instances, security groups, keypairs, etc
+         * based on timestamp without complicated enumeration        
+         */ 
+        shortener.append("timeStamp", Long.toString(System.currentTimeMillis() / 1000L, 16));
```
bostko: You could use Long.toString(1440680387, Character.MAX_RADIX) or even more explicitly Long.toString(l, 36)
This narrows down the required string length for the timestamp to 6

CMoH:
I might be asking the wrong question here, but what will happen if more machines are created within the same second? Will they get the same name? Would this be a problem or not?

ygy:
It will be highly unlikely to get 2 machines with one and the same name. 
The name is comprised by several other parameters like `appId` and `entityId`. They are random strings as well.
Most importantly jclouds itself append a randomly generated 8-char suffix at the end.

CMoH:
I see - thanks for clarifying this.

core/src/main/java/org/apache/brooklyn/core/location/cloud/names/BasicCloudMachineNamer.java
ygy: Thanks @bostko 
Addressed in commit 7fff655b05df098d9f35c97863da0d9902bf420a

aledsage:
@ygy the build is failing at `CloudMachineNamerTest.testLengthMaxPermittedForMachineName` (see https://builds.apache.org/job/incubator-brooklyn-pull-requests/org.apache.brooklyn$brooklyn-core/1746/testReport/junit/org.apache.brooklyn.core.location.cloud/CloudMachineNamerTest/testLengthMaxPermittedForMachineName/). Can you take a look at fixing that please?

    java.lang.AssertionError: expected [10] but found [9]
    	at org.testng.Assert.fail(Assert.java:94)
    	at org.testng.Assert.failNotEquals(Assert.java:494)
    	at org.testng.Assert.assertEquals(Assert.java:123)
    	at org.testng.Assert.assertEquals(Assert.java:370)
    	at org.testng.Assert.assertEquals(Assert.java:380)
    	at org.apache.brooklyn.core.location.cloud.CloudMachineNamerTest.testLengthMaxPermittedForMachineName(CloudMachineNamerTest.java:118)

aledsage:
Thanks @ygy - merging.

PR #869, opened 2015-08-28
Fix AbstractSoftwareProcessWinRmDriver tags

aledsage:
- See comments / discussion in https://github.com/apache/incubator-brooklyn/pull/866


aledsage:
Merging this one-line fix now.

PR #870, opened 2015-08-28
Don't treat *all* config as a "required open port"

alasdairhodge:
Make SoftwareProcessImpl *much* fussier about which config it implicitly adds to the list of required open ports.


alasdairhodge:
Context: a relatively recent change meant that *all* config keys were searched for any type that can be coerced to a `PortRange`, which was then added to the set of "required open ports" by `SoftwareProcessImpl`. However, since type coercions are defined from integer/long/string, this resulted in spurious port values being added, and in at least one customer scenario, a runtime exception when a value >16 bits was incorrectly subsumed into the port set.

richardcloudsoft:
This is a better idea IMO than just adding anything which can be coerced to an integer to the security group. Although we do need to keep an eye on the list for anyone who's blueprints suddenly stop working because they relied on the old behaviour. Thanks @alasdairhodge; merged.

PR #872, opened 2015-08-30
Handle class-renames in persisted state (Work In Progress)

aledsage:
For handling the class renames that have happened between `0.7.0` and now (e.g. for producing a `0.8.0-M1` or a `0.8.0` GA).

This is work in progress - it will allow early review and also the opportunity for someone else to pick it up while I'm on holiday (@ahgittin @neykov !). It is not yet tested with a real brooklyn instance!

Note that the long list of class renames was (mostly) auto-generated using https://github.com/brooklyncentral/version-differ. That is a very early stage tool for comparing two versions of a repo to find what classes have been moved. It writes out the changes + the ambiguous files, and also writes out the confidence of that change (based on the similarity of the source, ignoring license/imports).

I'm considering extending the above tool to write out details of new/deleted methods, the absence of `@since`, the addition of `@deprecated`, etc. If we do that, then this would be useful for auto-generating more detailed release notes.

I didn't use a combination of `git` and `gawk` because git wasn't aware for quite a few files that it was a move - presumably because the imports had change substantially, there was too much change for it to infer that it was the same file.


aledsage:
Note that we went for handling the class-renames in xstream deserialization. And in catalog-loading where it has fallen back to classpath-based loading.

The alternative would have been to transform the persisted state (e.g. using xslt). However, the classname can appear in many different ways / places within the xml, which makes it hard to reliably transform. One can't just do a regex conversion because the classname might have been used as part of a sensor / config key name (which has not been changed).

ahgittin:
I'm working on this.  A few minor mistakes (the path to the `deserializedClassNames.properties`), but I've fixed them.  Now it's failing because some classes are loaded elsewhere (in `RebindIteration`, calling to `Reflections`) so I'm refactoring to pass the map in there.

ahgittin:
BTW @aledsage with `git diff` you can use `-M` to be more generous when detecting renames; and you'll also need to increase `merge.renameLimit` because normally it defaults to 100 files.

ahgittin:
Incorporated into #873 which fixes some minor but fatal flaws in this (untested) but essentially keeps all the real work which was done here.  This should close automatically when that is merged BTW.

PR #873, opened 2015-08-31
Class renames

ahgittin:
Builds on #872 (which this incorporates):

* adding a few more places to use the map when loading
* doing a lookup based on the segment before `$` rather than an expensive map scan for any `startsWith`
* refactoring so rename is supported in `Reflections` (which was doing many of the `loadClass` calls in any case)

All renames get a debug log message of the form `Mapping <old> to <new>` .

Have tested this with some 070-incubating serialized state, but more testing would be useful.

Also have started a fine-tooth comb review of text/plan mentions of old classes, using a `grep -f <list> .../src` where `<list>` is the `awk { print $1 } | sort` of the `deserializedClassRenames.properties` map with blank lines removed; a bit more to come but wanted to get this out for review.


ahgittin:
As this is primarily a tidy-up of @aledsage 's work, plus doc tweaks, I'm going to merge this and update root docs and snapshot version docs on the site for comments.

PR #874, opened 2015-09-02
Update for 0.8.0 release

ahgittin:
This tidies the release process instructions (which were excellent, @rdowner !).

It also improves a problem with the download links at Apache (on top of #873).

And it adds `r.js` which I think completes addressing the comments on licenses from the 0.7.0-incubating release, /cc @hzbarcea .

For reference the comments there were:

> LICENCE is missing standard appendix

addressed with LICENSE file overhaul, now includes standard appendices and clearer format

> probably no need for the full text of MIT and BSD as the short version is preferred and already used

we still use the full text but this appears to be standard practice in other projects, and safer

> LICENSE is missing:
>  riak (see ./software/nosql/src/main/resources/brooklyn/entity/nosql/riak/vm.args + others)

these files are *our* work based on config files supplied by riak. both theirs and ours are examples which people are intended to modify,  i understand in such cases -- same as cassandra and mysql config files -- these should not have license headers nor be claimed as used under license.  people need to be able to modify these without license restrictions!

> normalize.css (see ./sandbox/mobile-app/src/main/webapp/assets/mobile/libs/mcasimir-mobile-angular-ui-be39011/dist/css/mobile-angular-ui-base.css)
> mobileangular-ui (see ./sandbox/mobile-app/src/main/webapp/assets/mobile/libs/mcasimir-mobile-angular-ui-be39011/dist/js/mobile-angular-ui.js)
> angular (see ./sandbox/mobile-app/src/main/webapp/assets/mobile/libs/angular-1.2.19)

these are in a sandbox project not included in the releases AFAIK, so i think it is correct not to list them?  (if this is incorrect i think we should simply remove sandbox items; they can be maintained outwith apache or in a sub-git-repo.)

> r.js (see ./usage/jsgui/src/build/requirejs-maven-plugin/r.js)

good spot, added this

---

we still need to:
* run the license update to incorporate r.js
* back port this to 0.8.0 prior to releasing it and run the license update there also

i suggest that be done by the person merging this.

### DIFF
--- a/docs/website/developers/committers/release-process/make-release-artifacts.md
+++ b/docs/website/developers/committers/release-process/make-release-artifacts.md
@@ -5,11 +5,11 @@ navgroup: developers
 ---
 
 A release script is provided in `release/make-release-artifacts.sh`. This script will prepare all the release artifacts.
-It is written to account for several Apache requirements, so you are advised to use it rather than "rolling your own".
+It is written to account for several Apache requirements, so you are strongly advised to use it rather than "rolling your own".
 
 The release script will:
 
-- **Create source code and binary distribution artifacts** and place them in a directory on your workstation.
+- **Create source code and binary distribution artifacts** and place them in a temporary staging directory on your workstation, usually `releases/`.
 - **Create Maven artifacts and upload them to a staging repository** located on the Apache Nexus server.
 
 The script has a single required parameter `-r` which is given the release candidate number - so `-r1` will create
@@ -23,6 +23,8 @@ To run the script:
 
 It will show you the release information it has deduced, and ask yes-or-no if it can proceed. Please note that the
 script will thoroughly clean the Git workspace of all uncommited and unadded files.
+
+**You really probably want to run this against a secondary checkout.** It will wipe `.project` files and other IDE metadata, and bad things can happen if an IDE tries to write while the script is running. Also as it takes a long time, this means your workspace is not tied up. One quick and easy way to do this is to `git clone` the local directory of your primary checkout to a secondary location.
 
 A few minutes into the script you will be prompted for the passphrase to your GnuPG private key. You should only be
 asked this question once; the GnuPG agent will cache the password for the remainder of the build.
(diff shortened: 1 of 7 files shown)


docs/website/developers/committers/release-process/make-release-artifacts.md
```diff
@@ -24,6 +24,8 @@ To run the script:
 It will show you the release information it has deduced, and ask yes-or-no if it can proceed. Please note that the
 script will thoroughly clean the Git workspace of all uncommited and unadded files.
 
+**You really probably want to run this against a secondary checkout.** It will wipe `.project` files and other IDE metadata, and bad things can happen if an IDE tries to write while the script is running. Also as it takes a long time, this means your workspace is not tied up. One quick and easy way to do this is to `git clone` the local directory of your primary checkout to a secondary location.
+
```
rdowner: (Minor) There's also a Vagrantfile for setting up a completely clean build environment in a VM. Maybe I'll add docs for that separately, however.

docs/website/developers/committers/release-process/prerequisites.md
```diff
@@ -32,16 +32,24 @@ pre-release software to the global release mirror network!
...
 
-The release manager must have a GPG key to be used to sign the release.
+The release manager must have a GPG key to be used to sign the release. See below to install `gpg2`
+(with a `gpg` alias).  The steps here also assume you have the following set
+(not using `whoami` if that's not appropriate):
+
+{% highlight bash %}
+GPG_KEY=`whoami`@apache.org
+SVN_USERNAME=`whoami`
+{% endhighlight %}
```
rdowner: (Minor) I'd describe these in the "environment variable setup" page earlier.

rdowner:
Ready to merge. A few minor comments but they can be addressed in a follow-up PR.

`closer.cgi` seems to have broken recently but I don't know why. Thanks for finding a solution to that.

docs/website/developers/committers/release-process/prerequisites.md
ahgittin: They're only ever needed in this section, which is why I think they're better here.

docs/website/developers/committers/release-process/make-release-artifacts.md
ahgittin: Good idea

ahgittin:
Agree, very odd about the `closer.cgi`, I was sure that used to work.

PR #877, opened 2015-09-02
Fix for BROOKLYN-169 - CreateUserPolicy always resets the password

ygy:
- Added parameter (BYON_USER_RESET_PASSWORD)
- BYON_USER_RESET_PASSWORD controls when password to be reset
- Whether to reset the password of the BYON login user in case it is
  used to create a new one, e.g. via a custom security policy such as
  org.apache.brooklyn.policy.jclouds.os.CreateUserPolicy"

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/policy/jclouds/os/CreateUserPolicy.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/policy/jclouds/os/CreateUserPolicy.java
@@ -87,6 +87,12 @@ public class CreateUserPolicy extends AbstractPolicy implements SensorEventListe
             "createuser.vm.user.credentials",
             "The \"<user> : <password> @ <hostname>:<port>\"");
 
+    @SetFromFlag("resetLoginUser")
+    public static final ConfigKey<Boolean> RESET_LOGIN_USER = ConfigKeys.newBooleanConfigKey(
+            "createuser.vm.user.resetLoginUser",
+            "Whether to reset the password used for user login",
+            false);
+
     public void setEntity(EntityLocal entity) {
         super.setEntity(entity);
         subscribe(entity, AbstractEntity.LOCATION_ADDED, this);
@@ -110,6 +116,7 @@ public class CreateUserPolicy extends AbstractPolicy implements SensorEventListe
     
     protected void addUser(Entity entity, SshMachineLocation machine) {
         boolean grantSudo = getRequiredConfig(GRANT_SUDO);
+        boolean resetPassword = getRequiredConfig(RESET_LOGIN_USER);
         String user = getRequiredConfig(VM_USERNAME);
         String password = Identifiers.makeRandomId(12);
         String hostname = machine.getAddress().getHostName();
@@ -125,7 +132,7 @@ public class CreateUserPolicy extends AbstractPolicy implements SensorEventListe
                 .adminUsername(user)
                 .adminPassword(password)
                 .grantSudoToAdminUser(false)
-                .resetLoginPassword(true)
+                .resetLoginPassword(resetPassword)
                 .loginPassword(password)
                 .authorizeAdminPublicKey(false)
                 .adminPublicKey("ignored")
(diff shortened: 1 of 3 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/policy/jclouds/os/CreateUserPolicy.java
```diff
@@ -110,6 +110,7 @@ public void run() {
     
     protected void addUser(Entity entity, SshMachineLocation machine) {
         boolean grantSudo = getRequiredConfig(GRANT_SUDO);
+        boolean resetPassword = machine.getConfig(SshMachineLocation.BYON_USER_RESET_PASSWORD);
```
sjcorbett: Is this addition still necessary given the similar change ("resetLoginPassword ") in 42be5897 by @bostko?

sjcorbett:
Repeating my comment hidden by GitHub: is this addition still necessary given the similar change ("resetLoginPassword ") in 42be589 by @bostko?

ygy:
@sjcorbett Thanks for asking
This change is still valid. A customer which heavily relies on such policies needs it.
It has been verified with a set of private tests too.

@bostko already commented on his PR. I will integrate the test prepared in PR #881 in a separate PR.

core/src/main/java/org/apache/brooklyn/location/ssh/SshMachineLocation.java
```diff
@@ -152,6 +152,12 @@
     public static final ConfigKey<Boolean> DETECT_MACHINE_DETAILS = ConfigKeys.newBooleanConfigKey("detectMachineDetails",
             "Attempt to detect machine details automatically. Works with SSH-accessible Linux instances.", true);
 
+    public static final ConfigKey<Boolean> BYON_USER_RESET_LOGIN = ConfigKeys.newBooleanConfigKey(
```
ahgittin: this should be a config on `CreateUserPolicy` as that is the only thing it affects, and it affects everything there

on a different note if that policy is tied to jclouds it will have no effect on BYON deployments.  if it isn't related to jclouds it should probably be moved to the policies project and a different package.  (we should leave the old instance in place as an empty class extending the item in the new location, but deprecated, so that references to the old one will still work.)

ygy:
@ahgittin Thanks for the suggestions.
It was in the `CreateUserPolicy` at the beginning. I've moved it to `SshMachineLocation` so it is backward compatible as much as it can. It guarantees that customers will make no changes at their end.

I will address your comments taking into account that customers will have to change theirs BPs, because of the package renaming

aledsage:
Thank @ygy 

* I agree with Alex: the `BYON_USER_RESET_PASSWORD` config key should be on the policy rather than on `SshMachineLocation`. I don't understand your comment "I've moved it to `SshMachineLocation` so it is backward compatible as much as it can. It guarantees that customers will make no changes at their end." This is a new config key. Also, see below about the default being `false`.
* `BYON_USER_RESET_PASSWORD` should default to `false`. We only got away with it with VMs provisioned with jclouds because the `user` was using an ssh-key rather than a password. We should not touch the existing user's configuration at all.
* The name `byonResetLogin` feels wrong. This happens for byon, but it would also happen for non-byon locations. I'd change it to `RESET_LOGIN_USER` for the constant, and `createuser.vm.user.resetLoginUser` for the string name, with a `@SetFromFlag("resetLoginUser")`.

@ahgittin it's in location-jclouds because it uses the jclouds `AdminAccess.builder` to build the command to be executed. However, it then executes it directly using the `SshMachineLocation` so works against BYON as well.

aledsage:
Also see https://github.com/apache/incubator-brooklyn/pull/881, which had `CreateUserPolicyWithEntityLiveTest` but was then closed. However, how does that test differ from `CreateUserPolicyLiveTest`? Is it different enough to justify adding that extra test? And if we do include it, then please see the assertions in `CreateUserPolicyLiveTest` which are better than just checking that no exceptions were thrown.

ygy:
@aledsage thanks for the recommendations 

This PR has been ammended based on the feedback in the previous comments

locations/jclouds/src/test/java/org/apache/brooklyn/policy/jclouds/os/CreateUserPolicyLiveTest.java
```diff
@@ -73,6 +73,7 @@ public void runTestCreatesUser(MachineProvisioningLocation<SshMachineLocation> l
             app.createAndManageChild(EntitySpec.create(TestEntity.class)
                     .policy(PolicySpec.create(CreateUserPolicy.class)
                             .configure(CreateUserPolicy.GRANT_SUDO, true)
+                            .configure(CreateUserPolicy.RESET_LOGIN_USER, true)
```
aledsage: Why use `RESET_LOGIN_USER=true` here? If anyone were to run `testIntegrationCreatesUser()`, it would reset their user password! We should either delete `testIntegrationCreatesUser()`, or remove this (to rely on the default of false).

aledsage:
Thanks @ygy - one last comment, then it's good to merge.

aledsage:
Thanks @ygy - merging now.

PR #878, opened 2015-09-02
changed css for app home page

morganbrooke:
Just some aesthetic changes to the landing page that had been recommended to me. 
No major functionality alterations. 
screen shot 2015-09-02 at 3 11 27 pm

### DIFF
--- a/usage/jsgui/src/main/webapp/assets/css/base.css
+++ b/usage/jsgui/src/main/webapp/assets/css/base.css
@@ -147,7 +147,7 @@ div#catalog-applications-empty {
 }
 
 .navbar .nav>li>a:hover {
-    background-color: #A8B8B0;
+    background-color: #6BB445;
     foreground-color: #261;
     color: #261;
     text-decoration: none;
@@ -166,7 +166,7 @@ div#catalog-applications-empty {
     text-shadow: 0 0px 0;
 }
 .navbar .nav > li > a.active {
-    background-color: #492;
+    background-color: #6BB445;
 }
 
 ul.dropdown-menu {
@@ -324,21 +324,19 @@ input[type="color"]:focus,.uneditable-input:focus {
 
 .roundedSummary {
     float: left;
-    border: 1px solid #d4d4d4;
+    border: 1px solid #6BB445;
     -webkit-border-radius: 15px;
     -moz-border-radius: 15px;
     border-radius: 15px;
-    background-color: #f7f6e8;
+    background-color: #DDDDDD !important;
     margin: 10px 15px 0px 0px;
-    padding: 20px 20px;
     width: 264px;
-    height: 160px;
+    height: 60px;
     line-height: 1.2;
-    font-size: 140%;
-    display: inline-block;
-    text-align: left;
-    background: #f9f9f9 url(../images/roundedSummary-background.png) top
-        repeat-x !important;
+    font-size: 120%;
+    display: inline;
+    text-align: center;
+
 }
 
 .roundedSummary:last-child {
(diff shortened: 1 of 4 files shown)


usage/jsgui/src/main/webapp/assets/css/base.css
```diff
@@ -362,24 +360,31 @@ input[type="color"]:focus,.uneditable-input:focus {
...
+  background-image: -webkit-gradient(linear, 0 0, 0 100%, from(#90C858), to(#609040));
+  background-image: -webkit-linear-gradient(top, #90C858, #609040);
+  background-image: -o-linear-gradient(top, #90C858, #609040);
+  background-image: -moz-linear-gradient(top, #90C858, #609040);
+  background-image: linear-gradient(top, #90C858, #609040);
+  background-repeat: repeat-x;
+  border-color: #609040 #609040 #609040;
+  border-color: rgba(0, 0, 0, 0.1) rgba(0, 0, 0, 0.1) rgba(0, 0, 0, 0.25);
+  filter: progid:dximagetransform.microsoft.gradient(startColorstr='#90C858', endColorstr='#609040', GradientType=0);
+  filter: progid:dximagetransform.microsoft.gradient(enabled=false);
```
sjcorbett: I'm not too familiar with this function but a bit of research suggests it's deprecated from IE9 onwards: https://msdn.microsoft.com/en-us/library/ms532997(v=vs.85).aspx. Is it something we want to include?

usage/jsgui/src/main/webapp/assets/css/base.css
```diff
@@ -362,24 +360,31 @@ input[type="color"]:focus,.uneditable-input:focus {
...
+  *background-color: #609040;
+  background-image: -ms-linear-gradient(top, #90C858, #609040);
+  background-image: -webkit-gradient(linear, 0 0, 0 100%, from(#90C858), to(#609040));
+  background-image: -webkit-linear-gradient(top, #90C858, #609040);
+  background-image: -o-linear-gradient(top, #90C858, #609040);
+  background-image: -moz-linear-gradient(top, #90C858, #609040);
+  background-image: linear-gradient(top, #90C858, #609040);
+  background-repeat: repeat-x;
+  border-color: #609040 #609040 #609040;
+  border-color: rgba(0, 0, 0, 0.1) rgba(0, 0, 0, 0.1) rgba(0, 0, 0, 0.25);
```
sjcorbett: Why have two definitions of border-color?

morganbrooke: It gives it a gradient effect(lighter to darker).

morganbrooke: Which is already present on the other buttons on the page. 
Just creating a standard for all of the buttons.

sjcorbett: I had to read up a little - am I right that if a browser understands the `rgba` format it will use the second definition of `border-color` and otherwise it will use the first (#609040)?

morganbrooke: Exactly, it adds a bit of transparency to the the color allowing it to fill a bit more gracefully. And if not, like you said, it will use the other color.

usage/jsgui/src/main/webapp/assets/css/base.css
morganbrooke: This I'm looking into, I wasn't familiar with this issue. But I'm looking at what I can do to correct it. Thanks for pointing it out.

ahgittin:
Nice.  Two things I notice:

* The height of the "Add Application" lozenge is different (a few pixels shorter) to the two status lozenges
* Applications listed should not have the underline except on hover, I think, to match behaviour elsewhere

Re the IE issue, the style directives should be ignored where deprecated or not supported, right, so maybe it doesn't matter?

morganbrooke:
I added the underline because it was not at all obvious to me that the applications listed were possible to be clicked on. Many of the team members agreed on this issue. 

The height difference I was having issues with because I believe (still sifting through things) there are other div elements in it that are messing with the DIV heights of the page and when the  "add application" lozenge was set to the same height, the 3 box's were not all at the same level.

ahgittin:
Yes agree with the need to style apps so they look like a link, but we should also try to keep a consistent style.  Maybe just making it green is enough (no underline) ?

Re heights yes it's annoying -- feel free to totally change that entire row if that makes things easier!

usage/jsgui/src/main/webapp/assets/tpl/home/applications.html
```diff
@@ -40,8 +40,8 @@
     <div id="applications">
         <table class="table table-striped table-bordered table-condensed">
             <thead>
-            <th>Application</th>
-            <th>Status</th>
+            <th><h3>Application</h3></th>
```
robertgmoss: The HTML standard forbids the use of Heading Content in a `th` element.  Suggest using css to style them instead.

ahgittin:
@morganbrooke feel free to keep this open - you can push add'l items to it.  (or maybe you are already planning to reopen once you've updated it?)

PR #884, opened 2015-09-08
Security provider sooner and REST docs

ahgittin:
Fix some irritants and long-running wishlist items during review of 080 (but which don't need to block a release)


ahgittin:
FYI this fixes the problem when running on a server that server-generated credentials for external access aren't logged until *after* someone logs in.  Now they are logged immediately on startup.

ahgittin:
Spurious failure -- `org.apache.brooklyn.entity.nosql.cassandra.CassandraDatacenterTest.testPopulatesInitialSeeds` :(

PR #885, opened 2015-09-09
Adds CUSTOMIZE_LATCH to redis slaves, allowing async start

nakomis:


aledsage:
@nakomis looks good - I presume `RedisClusterIntegrationTest` already covers the tests sufficiently, so no need for anything new?

Merging now.

PR #887, opened 2015-09-10
Misc fixes working with plans

ahgittin:
Change CAMP parsing so it can load items which use other transformers.  Previously it assumed referenced types were either java or yaml; now it uses the plan transformer extensions.  Messier than I had thought so please review commits.

Misc other tidies including errors and generics, as described in commits.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
@@ -26,7 +26,6 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
-import java.util.Set;
 
 import javax.annotation.Nullable;
 
@@ -34,6 +33,7 @@ import org.apache.brooklyn.api.catalog.BrooklynCatalog;
 import org.apache.brooklyn.api.catalog.CatalogItem;
 import org.apache.brooklyn.api.catalog.CatalogItem.CatalogBundle;
 import org.apache.brooklyn.api.catalog.CatalogItem.CatalogItemType;
+import org.apache.brooklyn.api.internal.AbstractBrooklynObjectSpec;
 import org.apache.brooklyn.api.location.Location;
 import org.apache.brooklyn.api.location.LocationSpec;
 import org.apache.brooklyn.api.mgmt.ManagementContext;
@@ -312,7 +312,7 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
     }
 
     @Override
-    public <T, SpecT> SpecT createSpec(CatalogItem<T, SpecT> item) {
+    public <T, SpecT extends AbstractBrooklynObjectSpec<? extends T, SpecT>> SpecT createSpec(CatalogItem<T, SpecT> item) {
         if (item == null) return null;
         @SuppressWarnings("unchecked")
         CatalogItemDo<T,SpecT> loadedItem = (CatalogItemDo<T, SpecT>) getCatalogItemDo(item.getSymbolicName(), item.getVersion());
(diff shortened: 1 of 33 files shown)


core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -800,7 +809,11 @@ private boolean attemptType(String key, CatalogItemType candidateCiType) {
                     // (when we're not given a key, the previous block should apply)
                     errors.add(e);
                 } else {
-                    // all other cases, the error is probably due to us not getting the type right, ignore it
+                    // all other cases, the error is probably due to us not getting the type right, probably ignore it
+                    // but cache it if we've checked entity, we'll use that as fallback errors
```
aledsage: Not sure I follow - why would we get the type wrong for non-entities? Have we guessed that it was an entity, without being able to tell from the context if its instead a policy/enricher/etc?

core/src/main/java/org/apache/brooklyn/core/plan/PlanToSpecFactory.java
```diff
@@ -95,13 +96,18 @@ public static PlanToSpecTransformer forPlanType(ManagementContext mgmt, String p
...
+                if (result==null) {
+                    transformersWhoDontSupport.add(t.getShortDescription() + " (returned null)");
+                    continue;
+                }
+                return Maybe.of(result);
             } catch (PlanNotRecognizedException e) {
                 transformersWhoDontSupport.add(t.getShortDescription() +
                     (Strings.isNonBlank(e.getMessage()) ? " ("+e.getMessage()+")" : ""));
-            } catch (Exception e) {
+            } catch (Throwable e) {
```
aledsage: Out of interest, why catch `Throwable`? We immediately do a `Exceptions.propagateIfFatal(e)` so if it was an `Error` then we'd rethrow immediately (as though we hadn't caught). Are there other throwable types that we're hitting?

(I'm fine with the change; just want to know what throwables we should expect to encounter).

utils/common/src/test/java/org/apache/brooklyn/util/time/TimeTest.java
```diff
@@ -152,7 +152,13 @@ public void testElapsedSince() {
         Assert.assertFalse(Time.hasElapsedSince(aFewSecondsAgo, Duration.TEN_SECONDS));
         Assert.assertTrue(Time.hasElapsedSince(-1, Duration.TEN_SECONDS));
     }
-    
+
+    @Test
+    public void testParseTime() {
+        Assert.assertEquals(Time.parseElapsedTime("1s"), 1000);
```
aledsage: There's a lot of logic in `parseElapsedTimeAsDouble`. Would be good to test more scenarios, e.g. the inverse of of the tests in `testMakeStringExact_secondsAndMillis()`.

aledsage:
@ahgittin looks good - only very minor comments/questions. Happy for this to be merged.

utils/common/src/test/java/org/apache/brooklyn/util/time/TimeTest.java
ahgittin: agree

core/src/main/java/org/apache/brooklyn/core/plan/PlanToSpecFactory.java
ahgittin: i think i was hoping to catch some weird errors, didn't realize propagateIfFatal would grab all `Error`.  maybe for some things like class linking etc we don't treat them as fatal?

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
ahgittin: we're parsing CAMP, then CAMP refers to something in catalog.  the code before this change would use camp-specific logic to handle the catalog entry making use of the parent entity's class loading context if needed.  but that breaks if it refers to a non-camp non-java catalog item (eg tosca).  so this now also looks it up using the catalog mechanism (although in that case it does not have any default class loading context.)

the CLC should only matter for java types.  it's working now, but would like to clean this whole area up massively!

PR #888, opened 2015-09-10
Changes propagator getDestinationSensor to lookup on sensor name, rather than sensor equality

nakomis:
This resolves #BROOKLYN-171

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/enricher/stock/Propagator.java
+++ b/core/src/main/java/org/apache/brooklyn/enricher/stock/Propagator.java
@@ -39,6 +39,7 @@ import org.apache.brooklyn.util.core.task.ValueResolver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
@@ -162,7 +163,7 @@ public class Propagator extends AbstractEnricher implements SensorEventListener<
         // propagate upwards
         Sensor<?> sourceSensor = event.getSensor();
         Sensor<?> destinationSensor = getDestinationSensor(sourceSensor);
-        
+
         if (!sensorFilter.apply(sourceSensor)) {
             return; // ignoring excluded sensor
         }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/brooklyn/enricher/stock/Propagator.java
```diff
@@ -195,7 +196,15 @@ public void emitAllAttributes(boolean includeNullValues) {
         }
     }
 
-    private Sensor<?> getDestinationSensor(Sensor<?> sourceSensor) {
-        return sensorMapping.containsKey(sourceSensor) ? sensorMapping.get(sourceSensor): sourceSensor;
+    private Sensor<?> getDestinationSensor(final Sensor<?> sourceSensor) {
+        Optional<? extends Sensor<?>> mappingSensor = Iterables.tryFind(sensorMapping.keySet(), new Predicate<Sensor<?>>() {
```
aledsage: Worth a comment saying why we're doing this, rather than the old simpler `sensorMapping.get(sourceSensor)`. e.g.:

    // sensor equality includes the type; we want just name-equality so will use predicate.

core/src/test/java/org/apache/brooklyn/enricher/stock/SensorPropagatingEnricherTest.java
```diff
@@ -215,4 +213,43 @@ public void testEnricherSpecThrowsOnPropagatesAndPropagatesAllSet() throws Excep
             if (ise == null) throw e;
         }
     }
+
+    @Test
+    public void testPropagateToDynamicSensor() {
```
aledsage: Would be good to have a specific test for the actual fix, rather than just this more complicated example. Maybe something like:

    @Test
    public void testSensorPropagatedWhenMappingUsedSameNameButDifferentType() throws Exception {
        AttributeSensor<String> origSensor = Sensors.newSensor(String.class, "origSensor");
        AttributeSensor<Object> sourceSensorFromYaml = Sensors.newSensor(Object.class, "origSensor");
        AttributeSensor<Object> targetSensor = Sensors.newSensor(Object.class, "newSensor");
        app.addEnricher(Enrichers.builder()
                .propagating(ImmutableMap.of(sourceSensorFromYaml, targetSensor))
                .from(entity)
                .build());
        entity.sensors().set(origSensor, "myval");
        EntityTestUtils.assertAttributeEqualsEventually(app, targetSensor, "myval");
    }

aledsage:
@nakomis looks good - a couple of minor comments. Do you want to address them now? Or shall we merge and then do it? Ping me when it's time to merge.

aledsage:
@nakomis - merging now, and then I'll make the changes (including the new unit test) in a new PR.

PR #890, opened 2015-09-11
Config constraints

sjcorbett:
Allows constraints to be specified on config keys. 

Constraints are eagerly enforced when entities are first managed and when policies and enrichers are created. At other times they are not enforced. Constraints can only be checked on ConfigKeys that are declared on the object's type hierarchy.

A constraint is a `Predicate<T>`, where `T` is the type of the `ConfigKey` in question. Guava's predicates provide almost all of the obvious use cases. I've written `ResourcePredicates.urlExists()` to use with config keys that represent things like files Brooklyn will load and run through Freemarker. I've verified that this works with resources in bundles. Adding this required a couple of adjustments to `NonDeploymentManagementContext`.

See the tests in `ConfigKeyConstraintTest` for most scenarios considered.

Here's what you see if you give an invalid value for a config key using the `urlExists` predicate:
hello

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/config/ConfigConstraints.java
+++ b/core/src/main/java/org/apache/brooklyn/core/config/ConfigConstraints.java
@@ -19,13 +19,16 @@
 
 package org.apache.brooklyn.core.config;
 
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 
 import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.location.Location;
 import org.apache.brooklyn.api.objs.BrooklynObject;
 import org.apache.brooklyn.api.objs.EntityAdjunct;
 import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.location.AbstractLocation;
 import org.apache.brooklyn.core.objs.AbstractEntityAdjunct;
 import org.apache.brooklyn.core.objs.BrooklynObjectPredicate;
 import org.apache.brooklyn.core.objs.BrooklynObjectInternal;
@@ -74,6 +77,18 @@ public abstract class ConfigConstraints<T extends BrooklynObject> {
         }
     }
 
+    public static <T> void assertValid(Entity entity, ConfigKey<T> key, T value) {
+        if (!new EntityConfigConstraints(entity).isValueValid(key, value)) {
+            throw new ConstraintViolationException("Invalid value for " + key + " on " + entity + ": " + value);
+        }
+    }
+
+    public static <T> void assertValid(Location location, ConfigKey<T> key, T value) {
+        if (!new LocationConfigConstraints(location).isValueValid(key, value)) {
+            throw new ConstraintViolationException("Invalid value for " + key + " on " + location + ": " + value);
+        }
+    }
+
     private static String errorMessage(BrooklynObject object, Iterable<ConfigKey<?>> violations) {
         StringBuilder message = new StringBuilder("Error configuring ")
                 .append(object.getDisplayName())
@@ -97,10 +112,6 @@ public abstract class ConfigConstraints<T extends BrooklynObject> {
 
     abstract Iterable<ConfigKey<?>> getBrooklynObjectTypeConfigKeys();
 
-    public boolean isValid() {
-        return Iterables.isEmpty(getViolations());
-    }
-
     public Iterable<ConfigKey<?>> getViolations() {
         return validateAll();
     }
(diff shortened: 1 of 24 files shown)


sjcorbett:
An annoyance of testing constraints pre-management is that an entity's `init` method has been run. Many entities create feeds and enrichers in this method. By throwing when configuration is invalid the creation of the entity is stopped but its adjuncts are left behind. When persistence is switched on this leaves orphaned mementos behind. Suggestions?

core/src/main/java/org/apache/brooklyn/core/config/ConfigConstraints.java
```diff
@@ -0,0 +1,165 @@
...
+        return Iterables.isEmpty(getViolations());
+    }
+
+    public Iterable<ConfigKey<?>> getViolations() {
+        return validateAll();
+    }
+
+    @SuppressWarnings("unchecked")
+    private Iterable<ConfigKey<?>> validateAll() {
+        List<ConfigKey<?>> violating = Lists.newArrayList();
```
mikezaccardo: Trivial, but I'd use a `LinkedList` here since it's of an unknown size and just going to be iterated on.

core/src/main/java/org/apache/brooklyn/core/config/ConfigConstraints.java
```diff
@@ -0,0 +1,165 @@
...
+        BrooklynObjectInternal.ConfigurationSupportInternal configInternal = getConfigurationSupportInternal();
+
+        Iterable<ConfigKey<?>> configKeys = getBrooklynObjectTypeConfigKeys();
+        LOG.trace("Checking config keys on {}: {}", getBrooklynObject(), configKeys);
+        for (ConfigKey<?> configKey : configKeys) {
+            // getRaw returns null if explicitly set and absent if config key was unset.
+            Object value = configInternal.getRaw(configKey).or(configKey.getDefaultValue());
```
mikezaccardo: If `getRaw` can return `null` if explicitly set then won't the `.or(...)` NPE?

core/src/main/java/org/apache/brooklyn/core/config/ConfigConstraints.java
```diff
@@ -0,0 +1,165 @@
...
+
+            if (value == null || value.getClass().isAssignableFrom(configKey.getType())) {
```
mikezaccardo: Should anything be done if this condition is not met?

mikezaccardo:
LGTM -- tests pass and only minor comments

core/src/main/java/org/apache/brooklyn/core/config/ConfigConstraints.java
ahgittin: @mikezaccardo what the comment means is that `getRaw()` can return a wrapped null (`Maybe.of(null)`) -- it never returns `null`

but @sjcorbett worth a comment why it's doing `getRaw` + `getDefault` in the first place.  presumably this is to prevent blocking etc.  but maybe that should instead live in a `getNonBlocking` method on `ConfigurationSupportInternal`?  (also it might be okay to allow a small time window in order to evaluate dependent tasks -- cf `EntityConfigResource.get`?)

ahgittin:
+1 -- excellent addition.  i see it evolving where and how validation happens but this is a great start.  merge when you're happy with the comments.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
```diff
@@ -268,9 +270,26 @@ public void manage(Entity e) {
                     new Exception("source of duplicate management of "+e));
             return;
         }
+        // Validate all config before attempting to manage any entity. Do this here rather
+        // than in manageRecursive so that rebind is unaffected.
+        validateDescendantConfig(e);
```
neykov: What's the reason to do it here rather than in `InternalEntityFactory` (analogously to the policy path)?

neykov:
Looks good. Are you leaving out validation at `confgi().set()` time on purpose or it's just out of scope for now? Could be harder to get it in at a later point due to existing invalid values. +1 to merge.

core/src/main/java/org/apache/brooklyn/core/config/ConfigConstraints.java
neykov: Coerce instead of checking for type compatibility (for the case when the value is not a future).

neykov: which is actually what @ahgittin suggests above - using a `getNonBlocking()` method, which will coerce internally

core/src/main/java/org/apache/brooklyn/core/config/ConstraintViolationException.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * A {@link ConstraintViolationException} indicates one or more problems applying
+ * values for {@link org.apache.brooklyn.config.ConfigKey ConfigKeys} when creating
+ * a {@link org.apache.brooklyn.api.objs.BrooklynObject}.
+ */
+public class ConstraintViolationException extends RuntimeException {
+    private static final long serialVersionUID = -6719912119648996815L;
+
+    public ConstraintViolationException(String message) {
```
aledsage: I'm a big fan of also declaring `public ConstraintViolationException(String message, Throwable cause)` for pretty much all exception classes - e.g. see that constructor signature being added to `IllegalStateException` only in Java 1.5.

utils/common/src/main/java/org/apache/brooklyn/util/guava/PredicateWithContext.java
```diff
@@ -0,0 +1,33 @@
...
+import com.google.common.base.Predicate;
+
+/**
+ * A marker interface for predicates that may take a context object into account.
+ * <p>
+ * Such predicates should handle absent contexts sensibly.
+ */
+public interface PredicateWithContext<P, T> extends Predicate<P> {
+
+    boolean apply(P value, T context);
```
aledsage: Bike shed: can we use "input" instead of "value", as that is used throughout guava.

core/src/test/java/org/apache/brooklyn/core/config/ConfigKeyConstraintTest.java
```diff
@@ -0,0 +1,285 @@
...
+        try {
+            Entities.manage(testEntity);
+            fail("Expected exception when managing child with invalid config");
+        } catch (Exception e) {
+            Throwable t = Exceptions.getFirstThrowableOfType(e, ConstraintViolationException.class);
+            assertNotNull(t);
+        }
+    }
+
+    // Test fails because config keys that are not on an object's interfaces cannot be checked automatically.
```
aledsage: I'd include a `TODO` at the start of the comment.

core/src/test/java/org/apache/brooklyn/core/config/ConfigKeyConstraintTest.java
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    @Test
+    public void testExceptionWhenSubclassSetsInvalidDefaultValue() {
+        try {
+            app.createAndManageChild(EntitySpec.create(EntityProvidingDefaultValueForConfigKeyInRange.class));
+            fail("Expected exception when managing entity setting invalid default value");
+        } catch (Exception e) {
+            Throwable t = Exceptions.getFirstThrowableOfType(e, ConstraintViolationException.class);
+            assertNotNull(t);
```
aledsage: personal preference for something like:

    if (Exceptions.getFirstThrowableOfType(e, ConstraintViolationException.class) == null) {
        throw e;
    }

Otherwise we lose the underlying exception - which is normally fine, but is a real pain when a test fails non-deterministically in the asfbot!

aledsage:
@sjcorbett looks great - can you rebase against master please (it currently says "this branch has conflicts that must be resolved").

My comments are all really minor - they don't need addressed before merging, but up to you.

Any thoughts on the comments from @neykov ?

core/src/main/java/org/apache/brooklyn/core/config/ConfigConstraints.java
sjcorbett: @neykov Is that what @ahgittin suggests? None of the methods in `ConfigurationSupportInternal` do any coercion.

neykov: Turns out it's insanely deep down the call stack, in https://github.com/apache/incubator-brooklyn/blob/master/core/src/main/java/org/apache/brooklyn/util/core/task/ValueResolver.java#L393, through https://github.com/apache/incubator-brooklyn/blob/master/core/src/main/java/org/apache/brooklyn/core/config/BasicConfigKey.java#L233.

neykov: @sjcorbett Reading the comment again, I agree that the two things are not the same - @ahgittin's focus is on blocking, while mine is on coercion. My thinking was that both would be addressed by `getNonBlocking` if it follows the existing behaviour of `get()` (blocks and then coerces), but with a very short timeout on the block.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
sjcorbett: Only that this came first and the idea was to do validate constraints pre-management. You're right that it's better in the entity factory. As an added bonus, this makes an execution context available for the `getNonBlocking` suggestion.

sjcorbett:
Thanks for the feedback everyone. I think I've incorporated all of the suggestions. In doing so I've added new commits 5423137, aa5c00f and 9ba5391. I'm happy for this to be re-reviewed or merged.

sjcorbett:
Since there are no more comments I will merge this.

PR #891, opened 2015-09-14
Fix FileBasedObjectStore#moveFile

alexmuntean:
- change method to use java.nio.file.Files
- add condition to delete existing destFile

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/FileBasedObjectStore.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/FileBasedObjectStore.java
@@ -24,6 +24,10 @@ import static java.lang.String.format;
 import java.io.File;
 import java.io.FileFilter;
 import java.io.IOException;
+import java.nio.file.AtomicMoveNotSupportedException;
+import java.nio.file.DirectoryNotEmptyException;
+import java.nio.file.Files;
+import java.nio.file.StandardCopyOption;
 import java.text.SimpleDateFormat;
 import java.util.Arrays;
 import java.util.Date;
@@ -37,9 +41,6 @@ import javax.annotation.Nullable;
 import org.apache.brooklyn.api.mgmt.ManagementContext;
 import org.apache.brooklyn.api.mgmt.ha.HighAvailabilityMode;
 import org.apache.brooklyn.core.server.BrooklynServerConfig;
-import org.apache.brooklyn.util.collections.MutableList;
-import org.apache.brooklyn.util.collections.MutableMap;
-import org.apache.brooklyn.util.core.internal.ssh.process.ProcessTool;
 import org.apache.brooklyn.util.exceptions.Exceptions;
 import org.apache.brooklyn.util.exceptions.FatalConfigurationRuntimeException;
 import org.apache.brooklyn.util.io.FileUtil;
@@ -53,7 +54,6 @@ import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.FluentIterable;
 import com.google.common.collect.ImmutableList;
-import com.google.common.io.Files;
 import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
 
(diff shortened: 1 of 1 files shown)


ahgittin:
+1 to switching to `nio` for this.

However we should keep the warning if we can't do this atomically.

Suggest specifying `StandardCopyOption#ATOMIC_MOVE ATOMIC_MOVE` and falling back to warning + strategy employed here.

alexmuntean:
@ahgittin commit replaced. what warning and why is that needed?

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/FileBasedObjectStore.java
```diff
@@ -346,51 +345,17 @@ protected File backupDirByMoving(File dir) throws InterruptedException, IOExcept
...
-            String cmd = "mv '"+srcFile.getAbsolutePath()+"' '"+destFile.getAbsolutePath()+"'";
-            
-            int exitStatus = new ProcessTool().execCommands(MutableMap.<String,String>of(), MutableList.of(cmd), null);
-            // prefer the above to the below because it wraps it in the appropriate bash
-//            Process proc = Runtime.getRuntime().exec(cmd);
-//            result = proc.waitFor();
-            
-            if (log.isTraceEnabled()) log.trace("FS move of {} to {} completed, code {}", new Object[] { srcFile, destFile, exitStatus });
-            if (exitStatus == 0) return;
+        if (destFile.exists()) {
```
aledsage: Instead of doing the delete followed by the move, can we first try (this will work on some filesystems, replacing the existing file atomically):

    Files.move(srcFile.toPath(), destFile.toPath(), StandardCopyOption.ATOMIC_MOVE);

If that fails (with an `AtomicMoveNotSupportedException`), then we can fallback to other options. We'll want to include the log.warn in this situation. Not sure how complicated we want to make it! I'm not sure how much the exception would destroy our performance if it's happening one or two times on every attempted file move.

Could try deleting and then atomically moving:

    if (destFile.exists()) {
        destFile.delete();
    }
    Files.move(srcFile.toPath(), destFile.toPath(), StandardCopyOption.ATOMIC_MOVE);

And could try a non-atomic move:

    Files.move(srcFile.toPath(), destFile.toPath(), StandardCopyOption.REPLACE_EXISTING);

See discussion in http://stackoverflow.com/questions/595631/how-to-atomically-rename-a-file-in-java-even-if-the-dest-file-already-exists and description in http://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#move(java.nio.file.Path,%20java.nio.file.Path,%20java.nio.file.CopyOption...)

aledsage:
@alexmuntean I've added a comment about how we could get "more atomic" - deleting the file and then moving leaves a period of time where the file does not exist, so it's not as atomic as it could be.

alexmuntean:
@aledsage changed. also added a dir check.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/FileBasedObjectStore.java
alexmuntean: I understand. about the exception being thrown... if we have any performance issues because of that we can change it then.. but I don't think so..

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/FileBasedObjectStore.java
```diff
@@ -346,53 +346,27 @@ protected File backupDirByMoving(File dir) throws InterruptedException, IOExcept
...
-            int exitStatus = new ProcessTool().execCommands(MutableMap.<String,String>of(), MutableList.of(cmd), null);
-            // prefer the above to the below because it wraps it in the appropriate bash
-//            Process proc = Runtime.getRuntime().exec(cmd);
-//            result = proc.waitFor();
-            
-            if (log.isTraceEnabled()) log.trace("FS move of {} to {} completed, code {}", new Object[] { srcFile, destFile, exitStatus });
-            if (exitStatus == 0) return;
+
+        try {
+            Files.move(srcFile.toPath(), destFile.toPath(), StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING);
```
aledsage: Note the `REPLACE_EXISTING` is ignored when ATOMIC_MOVE is used - see http://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#move(java.nio.file.Path,%20java.nio.file.Path,%20java.nio.file.CopyOption...), which says "all other options are ignored".

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/FileBasedObjectStore.java
```diff
@@ -346,53 +346,27 @@ protected File backupDirByMoving(File dir) throws InterruptedException, IOExcept
...
+        } catch (AtomicMoveNotSupportedException e) {
+            Files.move(srcFile.toPath(), destFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
```
aledsage: Personally I'd include the one-off log.warn to say that the move is not atomic. Otherwise the user gets no indication (and our log won't tell us that the filesystem didn't support atomic moves, which might be vital info for debugging error reports later).

aledsage:
Merging now; we can add the log.warn in a separate PR - thanks @alexmuntean

PR #894, opened 2015-09-14
Mongodb authentication

nakomis:


parent/pom.xml
```diff
@@ -107,7 +107,7 @@
         <jsonPath.version>2.0.0</jsonPath.version>
         <commons-compress.version>1.4</commons-compress.version>
         <qpid.version>0.20</qpid.version>
-        <mongodb.version>2.11.4</mongodb.version>
+        <mongodb.version>3.0.3</mongodb.version>
```
aledsage: The `AbstractMongoDBServer` still uses version 2.6.5. Should this really be version 3.x of the client?

robertgmoss:
Note: I paired with @nakomis on this and reviewed each change during development, too.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/AbstractMongoDBSshDriver.java
```diff
@@ -73,14 +75,40 @@ public void install() {
...
+        List<String> commands = new LinkedList<>();
+        commands.add(String.format("mkdir -p %s", getDataDirectory()));
+
+        if (MongoDBAuthenticationUtils.usesAuthentication(entity)) {
+            String destinationLocation = Os.mergePaths(getRunDir(), "mongodb-keyfile");
+            entity.sensors().set(AbstractMongoDBServer.MONGODB_KEYFILE_DESTINATION, destinationLocation);
+            String keyfileContents = entity.config().get(AbstractMongoDBServer.MONGODB_KEYFILE_CONTENTS);
+            if (keyfileContents == null) {
+                String keyfileUrl = entity.config().get(AbstractMongoDBServer.MONGODB_KEYFILE_URL);
+                copyResource(keyfileUrl, destinationLocation);
```
sjcorbett: I think it's worth checking that `keyfileUrl` isn't null too just in case someone changes the behaviour of `usesAuthentication` later on without realising that this block implicitly relies on `keyfileUrl` being non-null.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/AbstractMongoDBSshDriver.java
```diff
@@ -151,24 +186,38 @@ protected String getConfFile() {
...
+        builder.add("--fork");
+        return builder;
     }
     
     protected void launch(ImmutableList.Builder<String> argsBuilder) {
         String args = Joiner.on(" ").join(argsBuilder.build());
-        String command = String.format("%s/bin/mongod %s > out.log 2> err.log < /dev/null", getExpandedInstallDir(), args);
-        LOG.info(command);
+        List<String> commands = new LinkedList<String>();
+        commands.add(String.format("%s/bin/mongod %s > out.log 2> err.log < /dev/null", getExpandedInstallDir(), args));
```
sjcorbett: It's worth changing the stdout/err redirections to append to the files rather than overwrite them so we can keep the logs from the create-user script.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/AbstractMongoDBSshDriver.java
```diff
@@ -151,24 +186,38 @@ protected String getConfFile() {
...
+        builder.add("--port", port.toString());
+        builder.add("--fork");
+        return builder;
     }
     
     protected void launch(ImmutableList.Builder<String> argsBuilder) {
         String args = Joiner.on(" ").join(argsBuilder.build());
-        String command = String.format("%s/bin/mongod %s > out.log 2> err.log < /dev/null", getExpandedInstallDir(), args);
-        LOG.info(command);
+        List<String> commands = new LinkedList<String>();
```
sjcorbett: Any need for the list?

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/MongoDBAuthenticationMixins.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.brooklyn.util.core.flags.SetFromFlag;
+
+public interface MongoDBAuthenticationMixins {
+    @SetFromFlag("mongodbKeyfileContents")
+    ConfigKey<String> MONGODB_KEYFILE_CONTENTS = ConfigKeys.newStringConfigKey(
+            "mongodb.keyfile.contents", "Contents of the keyfile used for authentication");
+
+    @SetFromFlag("mongodbKeyfileUrl")
+    ConfigKey<String> MONGODB_KEYFILE_URL = ConfigKeys.newStringConfigKey(
+            "mongodb.keyfile.url", "Location of the keyfile used for authentication");
```
sjcorbett: Worth noting the relationship between the keyfile url and contents config keys in both of their descriptions.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/MongoDBAuthenticationMixins.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.brooklyn.api.sensor.AttributeSensor;
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.sensor.BasicAttributeSensorAndConfigKey;
+import org.apache.brooklyn.core.sensor.Sensors;
+import org.apache.brooklyn.util.core.flags.SetFromFlag;
+
+public interface MongoDBAuthenticationMixins {
+    @SetFromFlag("mongodbKeyfileContents")
+    ConfigKey<String> MONGODB_KEYFILE_CONTENTS = ConfigKeys.newStringConfigKey(
```
sjcorbett: Thinking about the two config options more, do we even need keyfileContents? It complicates the logic and is less secure than giving a URL and isn't used in any of the tests in this PR. What arguments for its inclusion are there?

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/MongoDBAuthenticationUtils.java
```diff
@@ -0,0 +1,79 @@
...
+    private MongoDBAuthenticationUtils(){}
+
+    /**
+     * @return true if either the keyfile contents or keyfile url is set, false otherwise. If both are set, an IllegalStateException is thrown
+     */
+    public static boolean usesAuthentication(Entity entity) {
+        String keyfileContents = entity.config().get(MongoDBAuthenticationMixins.MONGODB_KEYFILE_CONTENTS);
+        String keyfileUrl = entity.config().get(MongoDBAuthenticationMixins.MONGODB_KEYFILE_URL);
+        if (!(Strings.isEmpty(keyfileContents) || Strings.isEmpty(keyfileUrl))) {
+            throw new IllegalStateException("keyfile contents and keyfile location cannot both be set");
```
sjcorbett: I've seen lots of entities use a this-config-key-is-preferred strategy rather than failing outright. Any reason to avoid this? Given the comment on keyfileContents this may not matter.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/MongoDBAuthenticationUtils.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+        if (Strings.isEmpty(keyfileContents) && Strings.isEmpty(keyfileUrl)) {
```
sjcorbett: Just return the negation of the if condition. Or `return Strings.isNonBlank(contents) || Strings.isNonBlank(url)`.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/MongoDBAuthenticationUtils.java
```diff
@@ -0,0 +1,79 @@
...
+            return false;
+        }
+        return true;
+    }
+
+    public static String getRootPassword(Entity entity) {
+        String password = entity.config().get(MongoDBAuthenticationMixins.ROOT_PASSWORD);
+        if (Strings.isEmpty(password)) {
+            LOG.debug(entity + " has no password specified for " + MongoDBAuthenticationMixins.ROOT_PASSWORD.getName() + "; using a random string");
+            password = Strings.makeRandomId(8);
```
sjcorbett: Why only eight characters? May as well make it long.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/MongoDBAuthenticationUtils.java
```diff
@@ -0,0 +1,79 @@
...
+            entity.sensors().set(MongoDBAuthenticationMixins.ROOT_PASSWORD, password);
+            entity.config().set(MongoDBAuthenticationMixins.ROOT_PASSWORD, password);
+        }
+        return password;
+    }
+
+    public static void setAuthenticationConfig(EntitySpec<?> spec, Entity source) {
```
sjcorbett: The order of the arguments to these two methods surprised me. Worth documenting the effect of both or giving the arguments meaningful names.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/MongoDBClientSupport.java
```diff
@@ -45,24 +47,36 @@
...
     // Set client to automatically reconnect to servers.
     private static final MongoClientOptions connectionOptions = MongoClientOptions.builder()
-            .autoConnectRetry(true)
             .socketKeepAlive(true)
             .build();
 
     private static final BasicBSONObject EMPTY_RESPONSE = new BasicBSONObject();
 
-    public MongoDBClientSupport(ServerAddress standalone) {
+    public MongoDBClientSupport(ServerAddress standalone, boolean usesAuthentication, String username, String password, String authenticationDatabase) {
```
sjcorbett: I think a nicer way to express this is to use two constructors, one with a single `ServerAddress` argument and one as the line above, without `usesAuthentication`. The constructors can set `usesAuthentication` themselves. The `forServer` method can decide which to call by checking `MongoDBAuthenticationUtils.usesAuthentication`.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/MongoDBClientSupport.java
```diff
@@ -96,12 +111,18 @@ private HostAndPort getServerHostAndPort() {
...
             DB db = client.getDB(database);
-            CommandResult status;
-            try {
-                status = db.command(command);
-            } catch (MongoException e) {
-                LOG.warn("Command " + command + " on " + getServerAddress() + " failed", e);
-                return Optional.absent();
+            CommandResult status = null;
+            // The mongoDB client can occasionally fail to connect. Try up to 5 times to run the command
+            for (int i = 0; i < 5; i++) {
```
sjcorbett: Would a `Repeater` be of any use? It allows you to give backoffs, etc. It's a cleaner way of expressing this block.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/MongoDBReplicaSetImpl.java
```diff
@@ -141,7 +142,12 @@ public Entity apply(@Nullable Collection<Entity> entities) {
 
     @Override
     protected EntitySpec<?> getMemberSpec() {
-        return getConfig(MEMBER_SPEC, EntitySpec.create(MongoDBServer.class));
+        EntitySpec<?> spec = config().get(MEMBER_SPEC);
+        if (spec == null) {
+            spec = EntitySpec.create(MongoDBServer.class);
```
sjcorbett: setConfig too?

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/MongoDBReplicaSetImpl.java
```diff
@@ -394,10 +401,18 @@ public void onManagementStopped() {
             // Ignored
         }
         @Override protected void onEntityAdded(Entity member) {
-            ((MongoDBReplicaSetImpl)entity).serverAdded((MongoDBServer) member);
+            try {
+                ((MongoDBReplicaSetImpl) entity).serverAdded((MongoDBServer) member);
+            } catch (Exception e) {
+                ServiceStateLogic.ServiceNotUpLogic.updateNotUpIndicator((EntityLocal)member, "Failed to update replicaset", e);
```
sjcorbett: Would prefer this to happen in `serverAdded` rather than the event listener.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/MongoDBReplicaSetImpl.java
```diff
@@ -394,10 +401,18 @@ public void onManagementStopped() {
...
+            }
         }
         @Override protected void onEntityRemoved(Entity member) {
-            ((MongoDBReplicaSetImpl)entity).serverRemoved((MongoDBServer) member);
+            try {
+                ((MongoDBReplicaSetImpl) entity).serverRemoved((MongoDBServer) member);
+            } catch (Exception e) {
+                ServiceStateLogic.ServiceNotUpLogic.updateNotUpIndicator((EntityLocal)member, "Failed to update replicaset", e);
```
sjcorbett: Likewise.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/ReplicaSetConfig.java
```diff
@@ -148,6 +148,10 @@ public ReplicaSetConfig member(String hostname, Integer port, Integer id) {
         BasicBSONObject member = new BasicBSONObject();
         member.put("_id", id);
         member.put("host", String.format("%s:%s", hostname, port));
+        if (members.contains(member)) {
+            LOG.warn("Ignoring attempt to add entity to a set of which it is already a member");
```
sjcorbett: When does this case happen? Suggests calling code needs to be fixed.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/sharding/CoLocatedMongoDBRouterImpl.java
```diff
@@ -19,32 +19,47 @@
...
 import org.apache.brooklyn.entity.software.base.SameServerEntityImpl;
+import org.apache.brooklyn.util.exceptions.Exceptions;
 
 import com.google.common.base.Predicates;
 import com.google.common.collect.Iterables;
 
-public class CoLocatedMongoDBRouterImpl extends SameServerEntityImpl implements CoLocatedMongoDBRouter {
+public class CoLocatedMongoDBRouterImpl extends SameServerEntityImpl implements CoLocatedMongoDBRouter, MongoDBAuthenticationMixins {
+
+    private MongoDBRouter router;
```
sjcorbett: This won't persist well. Suggest a `getRouter()` method which just looks for `Entities.descendants(this, MongoDBRouter.class)`.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/sharding/CoLocatedMongoDBRouterImpl.java
```diff
@@ -19,32 +19,47 @@
...
-        
+        router = addChild(EntitySpec.create(MongoDBRouter.class)
+                .configure(MongoDBRouter.CONFIG_SERVERS,
+                        DependentConfiguration.attributeWhenReady(
+                                getConfig(CoLocatedMongoDBRouter.SHARDED_DEPLOYMENT),
+                                MongoDBConfigServerCluster.CONFIG_SERVER_ADDRESSES)));
         for (EntitySpec<?> siblingSpec : getConfig(CoLocatedMongoDBRouter.SIBLING_SPECS)) {
             addChild(siblingSpec);
         }
+        sensors().set(ROUTER, (MongoDBRouter) Iterables.tryFind(getChildren(), Predicates.instanceOf(MongoDBRouter.class)).get());
```
sjcorbett: You already have a reference to `router`!

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/sharding/CoLocatedMongoDBRouterImpl.java
```diff
@@ -54,17 +69,24 @@ protected void doStart(Collection<? extends Location> locations) {
...
-                                getConfig(CoLocatedMongoDBRouter.SHARDED_DEPLOYMENT), 
-                                MongoDBConfigServerCluster.CONFIG_SERVER_ADDRESSES)));
+
+        try {
+            // Do not attempt to read the configuration until after the router has been added to the cluster
+            // as it is at this point that the authentication configuration is set
+            Task<?> clusterTask = DependentConfiguration.attributeWhenReady(router, AbstractGroup.FIRST);
+            Entities.submit(this, clusterTask);
+            clusterTask.get();
+        } catch (Exception e) {
```
sjcorbett: What exceptions are being swallowed here? If it's only the task interrupted and execution exceptions you may as well just call `getUnchecked`.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/sharding/CoLocatedMongoDBRouterImpl.java
```diff
@@ -54,17 +69,24 @@ protected void doStart(Collection<? extends Location> locations) {
...
+            Exceptions.propagateIfFatal(e);
+        }
+        MongoDBAuthenticationUtils.setAuthenticationConfig(router, this);
+        router.sensors().set(MongoDBAuthenticationMixins.ROOT_PASSWORD, router.config().get(MongoDBAuthenticationMixins.ROOT_PASSWORD));
+        router.sensors().set(MongoDBAuthenticationMixins.ROOT_USERNAME, router.config().get(MongoDBAuthenticationMixins.ROOT_USERNAME));
+        router.sensors().set(MongoDBAuthenticationMixins.AUTHENTICATION_DATABASE, router.config().get(MongoDBAuthenticationMixins.AUTHENTICATION_DATABASE));
         Entities.manage(router);
```
sjcorbett: Appreciate this is unchanged in this pull request. Do you recall why the router is only managed here? Shouldn't it happen in `init`?

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/sharding/MongoDBRouterSshDriver.java
```diff
@@ -48,4 +48,8 @@ public void launch() {
                 .body.append(command).execute();
     }
 
+    @Override
+    public boolean isRunning() {
+        return super.isRunning();
```
sjcorbett: Useless override.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/sharding/MongoDBShardedDeploymentImpl.java
```diff
@@ -53,22 +54,36 @@
...
         addEnricher(Enrichers.builder()
                 .propagating(MongoDBConfigServerCluster.CONFIG_SERVER_ADDRESSES)
                 .from(getAttribute(CONFIG_SERVER_CLUSTER))
                 .build());
-        
+
+        // Advertise even if default are used (root password is set in MongoDBAuthenticationUtils)
+        sensors().set(MongoDBAuthenticationMixins.AUTHENTICATION_DATABASE, config().get(MongoDBAuthenticationMixins.AUTHENTICATION_DATABASE));
+        sensors().set(MongoDBAuthenticationMixins.ROOT_USERNAME, config().get(MongoDBAuthenticationMixins.ROOT_USERNAME));
+        sensors().set(MongoDBAuthenticationMixins.MONGODB_KEYFILE_DESTINATION, config().get(MongoDBAuthenticationMixins.MONGODB_KEYFILE_CONTENTS));
```
sjcorbett: I think you've used `MONGODB_KEYFILE_CONTENTS` incorrectly.

utils/common/src/main/java/org/apache/brooklyn/util/ssh/BashCommands.java
```diff
@@ -625,6 +625,13 @@ public static String pipeTextTo(String text, String command) {
                 +"\n"+"EOF_"+id+"\n";
     }
 
+    public static String pipeTextToFile(String text, String filepath) {
+        String id = Identifiers.makeRandomId(8);
+        return "cat > " + filepath + " << EOL_" + id + "\n"
```
sjcorbett: `cat > \"" + filepath + "\" ...` would be preferable to handle paths with spaces.

parent/pom.xml
nakomis: The 3.x client is compatible with 2.x[1], and addresses intermittent connection issues that we were seeing with the 2.11.4 client

[1]: http://docs.mongodb.org/ecosystem/drivers/driver-compatibility-reference/#java-driver-compatibility

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/MongoDBAuthenticationMixins.java
nakomis: The (anticipated) use-case of this is where the person deploying mongoDB is not the person who configured the Brooklyn server, and may not necessarily have access to put the keyfile on Brooklyn's classpath. Providing the keyfile contents is easier (and more secure) than putting it on S3 (e.g.) and providing a URL

sjcorbett: I'm hesitant to support use cases we don't know we need yet. I'm not aware of any other entities that make allowance for this scenario - all the examples I can think of just use a URL property for config files and so forth. I'm not insisting this be changed, only that we only include it if we think it's absolutely necessary.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/sharding/CoLocatedMongoDBRouterImpl.java
nakomis: I don't recall

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/sharding/MongoDBRouterSshDriver.java
nakomis: It was a useful place to put a breakpoint :-) Completely forgot to remove it...

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/sharding/CoLocatedMongoDBRouterImpl.java
sjcorbett: I was being slow - it happened here because the router was added here. Suggest moving the manage call to `init` to be alongside `addChild`.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/sharding/CoLocatedMongoDBRouterImpl.java
```diff
@@ -19,32 +19,47 @@
...
         for (EntitySpec<?> siblingSpec : getConfig(CoLocatedMongoDBRouter.SIBLING_SPECS)) {
             addChild(siblingSpec);
         }
+        sensors().set(ROUTER, (MongoDBRouter) Iterables.tryFind(getChildren(), Predicates.instanceOf(MongoDBRouter.class)).get());
     }
 
     @Override
     protected void doStart(Collection<? extends Location> locations) {
+
         // TODO Changed to create the router child after init as a workaround.
```
sjcorbett: Is this still valid now you've moved the creation of the router back to init?

utils/common/src/main/java/org/apache/brooklyn/util/ssh/BashCommands.java
```diff
@@ -625,6 +625,13 @@ public static String pipeTextTo(String text, String command) {
                 +"\n"+"EOF_"+id+"\n";
     }
 
+    public static String pipeTextToFile(String text, String filepath) {
+        String id = Identifiers.makeRandomId(8);
```
sjcorbett: Do you need a random string? You could just hard code a long string of garbage.

nakomis:
PR Comments addressed

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/sharding/MongoDBRouterSshDriver.java
sjcorbett: Still forgotten?

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/MongoDBAuthenticationUtils.java
sjcorbett: Could you comment on your preference for throwing an exception? If you're going to keep it then it should be thrown before any machines are provisioned.

nakomis:
Updated as requested

sjcorbett:
@nakomis Thanks. Will run integration/live tests and (all being well!) merge.

PR #896, opened 2015-09-15
Move to java7 setPosixFilePermissions

CMoH:
On my machine this improved performance from 34903 accessor put() calls
per second to 61553 using the new implementation.

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/util/io/FilePermissions.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/io/FilePermissions.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2015 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.brooklyn.util.io;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.attribute.PosixFileAttributeView;
+import java.nio.file.attribute.PosixFilePermission;
+import static java.nio.file.attribute.PosixFilePermission.GROUP_EXECUTE;
+import static java.nio.file.attribute.PosixFilePermission.GROUP_READ;
+import static java.nio.file.attribute.PosixFilePermission.GROUP_WRITE;
+import static java.nio.file.attribute.PosixFilePermission.OTHERS_READ;
+import static java.nio.file.attribute.PosixFilePermission.OTHERS_WRITE;
+import static java.nio.file.attribute.PosixFilePermission.OTHERS_EXECUTE;
+import static java.nio.file.attribute.PosixFilePermission.OWNER_READ;
+import static java.nio.file.attribute.PosixFilePermission.OWNER_WRITE;
+import static java.nio.file.attribute.PosixFilePermission.OWNER_EXECUTE;
+import java.util.EnumSet;
+import java.util.Set;
+import org.apache.brooklyn.util.os.Os;
+
+/**
+ * Portable file permissions.
+ *
+ * @author Ciprian Ciubotariu <cheepeero@gmx.net>
+ */
+public class FilePermissions {
+
+    private final int posixMode;
+    private final Set<PosixFilePermission> posixPermissions = EnumSet.noneOf(PosixFilePermission.class);
+
+    public FilePermissions(int posixMode) {
+        this.posixMode = posixMode;
+        if ((posixMode & 0400) != 0) posixPermissions.add(OWNER_READ);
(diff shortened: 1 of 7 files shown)


utils/common/src/main/java/org/apache/brooklyn/util/io/FileUtil.java
```diff
@@ -73,36 +77,11 @@ public static void setFilePermissionsTo700(File file) throws IOException {
...
-        boolean setExec = file.setExecutable(false, false);
-        
-        if (setRead && setWrite && setExec) {
-            if (LOG.isTraceEnabled()) LOG.trace("Set permissions to 600 for file {}", file.getAbsolutePath());
-        } else {
-            if (loggedSetFilePermissionsWarning) {
-                if (LOG.isTraceEnabled()) LOG.trace("Failed to set permissions to 600 for file {}: setRead={}, setWrite={}, setExecutable={}",
-                        new Object[] {file.getAbsolutePath(), setRead, setWrite, setExec});
-            } else {
-                if (Os.isMicrosoftWindows()) {
```
aledsage: On windows, will we get an `UnsupportedOperationException` from `Files.setPosixFilePermissions`? If so, it would be good to preserve the old behaviour, rather than fail: to log about it (but to guard that log with the `loggedSetFilePermissionsWarning` check so that we only log once).

utils/common/src/main/java/org/apache/brooklyn/util/io/FileUtil.java
```diff
@@ -73,36 +77,11 @@ public static void setFilePermissionsTo700(File file) throws IOException {
...
-        file.setWritable(false, false);
-        file.setReadable(true, true);
-        file.setWritable(true, true);
-        
-        boolean setRead = file.setReadable(false, false) & file.setReadable(true, true);
-        boolean setWrite = file.setWritable(false, false) & file.setWritable(true, true);
-        boolean setExec = file.setExecutable(false, false);
-        
-        if (setRead && setWrite && setExec) {
-            if (LOG.isTraceEnabled()) LOG.trace("Set permissions to 600 for file {}", file.getAbsolutePath());
```
aledsage: Would be good to preserve the `LOG.trace` when we change the file permission.

utils/common/src/main/java/org/apache/brooklyn/util/io/FileUtil.java
```diff
@@ -73,36 +77,11 @@ public static void setFilePermissionsTo700(File file) throws IOException {
...
-        } else {
```
aledsage: The previous code relied on `file.setReadable` returning false, so as to do a single `LOG.warn` if the file was not accessible. I presume that `Files.setPosixFilePermissions` will throw an `IOException`.

I worry about us propagating the exception, without changing the callers. For example, see `FileBasedStoreObjectAccessor.put`. This will now fail, rather than just writing to the file without changing the file permission. We'd definitely want the warn (done just once, rather than many times a second!) in such a situation.

aledsage:
Thanks @CMoH - some concerns about the change in behaviour for exception handling; and would be good to keep more of the old logging.

CMoH:
I'd very much appreciate a review and testing on Windows. I'm not very satisfied with the solution, so if you have any ideas, please pitch in.

utils/common/src/main/java/org/apache/brooklyn/util/io/FilePermissions.java
```diff
@@ -0,0 +1,93 @@
...
+            }
+            done = true;
+        } catch (UnsupportedOperationException ex) {}
+        
+        if (!done) {
+            // ~42000 calls per sec
+            // TODO: what happens to group permissions ?
+            boolean setRead = file.setReadable(posixPermissions.contains(OTHERS_READ), false) & file.setReadable(posixPermissions.contains(OWNER_READ), true);
+            boolean setWrite = file.setWritable(posixPermissions.contains(OTHERS_WRITE), false) & file.setWritable(posixPermissions.contains(OWNER_WRITE), true);
+            boolean setExec = file.setExecutable(posixPermissions.contains(OTHERS_EXECUTE), false) & file.setExecutable(posixPermissions.contains(OWNER_EXECUTE), true);
```
aledsage: This ignores if it was 700 or 600.

utils/common/src/main/java/org/apache/brooklyn/util/io/FilePermissions.java
```diff
@@ -0,0 +1,93 @@
...
+        Path filePath = file.toPath();
+
+        // the appropriate condition is actually Files.getFileStore(filePath).supportsFileAttributeView(PosixFileAttributeView.class)
+        // but that downs performance to ~9000 calls per second
+
+        boolean done = false;
+        try {
+            // ~59000 calls per sec
+            if (!Os.isMicrosoftWindows()) {
+                Files.setPosixFilePermissions(filePath, posixPermissions);
```
aledsage: We could log (once) to say that file permissions will be ignored because we are running on windows. Is that worth it?

utils/common/src/main/java/org/apache/brooklyn/util/io/FilePermissions.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.EnumSet;
+import java.util.Set;
+import org.apache.brooklyn.util.os.Os;
+
+/**
+ * Portable file permissions.
+ *
+ * @author Ciprian Ciubotariu <cheepeero@gmx.net>
+ */
+public class FilePermissions {
```
aledsage: We should mark this as `@Beta` if not certain about the approach.

aledsage:
Thanks @CMoH - let's merge this now; we can make more improvements in other PRs.

PR #898, opened 2015-09-15
Uses BrooklynAccessUtils.getBrooklynAccessibleAddress for DATASTORE_U…

nakomis:
…RL, replaces get/set Config/Attribute with config() and sensors()

### DIFF
--- a/software/database/src/main/java/org/apache/brooklyn/entity/database/crate/CrateNodeImpl.java
+++ b/software/database/src/main/java/org/apache/brooklyn/entity/database/crate/CrateNodeImpl.java
@@ -20,6 +20,7 @@ package org.apache.brooklyn.entity.database.crate;
 
 import org.apache.brooklyn.core.config.render.RendererHints;
 import org.apache.brooklyn.core.entity.Attributes;
+import org.apache.brooklyn.core.location.access.BrooklynAccessUtils;
 import org.apache.brooklyn.enricher.stock.Enrichers;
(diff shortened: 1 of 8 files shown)


sjcorbett:
I'm a bit nervous of changing the meaning of `DATASTORE_URL` like this. We have no visibility into what existing applications are out there that rely on it being sensible to use for within-application communication. My preference would be to either manually set a `BROOKLYN_ACCESSIBLE_DATASTORE_URL` or to use an enricher (if possible) to do the transformation automatically.

@aledsage Any thoughts?

aledsage:
@nakomis I agree with @sjcorbett - if we change the `DATASTORE_URL` meaning to be a potentially NAT'ed ip:port, then that can mean it's not accessible from other VMs within the same private network / zone.

We should be consistent with other types of entity, such as app servers. There the `webapp.url` and `main.uri` are constructed using the `getSubnetHostname()` and unmapped port.

If we want to change the semantics of `DATASTORE_URL` (and all other external endpoints sensors), then we should not do that lightly - it would need to be discussed on the mailing list. Personally, I'd be in favour of leaving this as the VM's url, and then having a mapped version of the sensor where required for external use.

aledsage:
@nakomis can you keep the trivial changes (e.g. changing use of deprecated code, and removing types from `<>` in generics) in a different commit for future PRs please? Where that's hard (e.g. because it would be very fiddly to separate where you accidentally changes things at the same time), then that's fine - but it would be good if that is the exceptional situation.

PR #900, opened 2015-09-15
Replicaset endpoint url

robertgmoss:
Depends on #894

Forms a connection string for replica sets so that clients can connect directly.


sjcorbett:
@robertgmoss Thanks. A few comments to address.

robertgmoss:
@sjcorbett comments addressed

sjcorbett:
@robertgmoss Thanks.

PR #901, opened 2015-09-16
Fixing JcloudsLocationSecurityGroupCustomizer on OpenStack

grkvlt:
Add synchronization and checks for duplicate rules in addPermission for the `JcloudsLocationSecurityGroupCustomizer` as used by Clocker. When deploying to OpenStack, duplicate rules will cause the `addPermission` method to fail, if the customizer is called in parallel on the same group. So, we serialize the calls with a `synchronized` block, and inside the critical section we make sure to only add rules that are not already present.


sjcorbett:
@grkvlt Thanks. Will merge.

PR #902, opened 2015-09-16
Make Repeater a Callable

grkvlt:
Easy to do, so why not? Allows it to be passed directly to the `body()` method of a `TaskBuilder` for instance.


sjcorbett:
Will merge.

PR #903, opened 2015-09-16
Fix MySqlCluster script init

neykov:
`CREATION_SCRIPT_{URL, CONTENT} `was applied to the master node pre-replication-init so there was no way to initialize the slave nodes. Improved by:
  * Applying `CREATION_SCRIPT_{URL, CONTENT} ` on all nodes, including slaves pre-replication-init
  * Introducing a new `MASTER_CREATION_SCRIPT_{URL, CONTENT} ` config applied on master only post-replication-init

### DIFF
--- a/software/database/src/test/java/org/apache/brooklyn/entity/database/mysql/MySqlClusterIntegrationTest.java
+++ b/software/database/src/test/java/org/apache/brooklyn/entity/database/mysql/MySqlClusterIntegrationTest.java
@@ -19,18 +19,20 @@
 package org.apache.brooklyn.entity.database.mysql;
 
 import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.location.Location;
 import org.apache.brooklyn.core.test.BrooklynAppLiveTestSupport;
 import org.apache.brooklyn.util.os.Os;
 import org.testng.annotations.Test;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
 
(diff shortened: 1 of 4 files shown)


neykov:
Build failure not related, forced a new one.

software/database/src/test/java/org/apache/brooklyn/entity/database/mysql/MySqlClusterIntegrationTest.java
```diff
@@ -40,4 +42,22 @@ public void test_localhost() throws Exception {
...
+                String runDir = member.getAttribute(MySqlNode.RUN_DIR);
+                if (runDir != null) {
+                    Os.deleteRecursively(runDir);
+                }
+            }
+        }
+    }
+
+    private Location getLocation() {
+        return mgmt.getLocationRegistry().resolve("vagrant", ImmutableMap.of("minRam", "2G"));
```
aledsage: Needs comment to say what config is required in brooklyn.properties for this to work. Presumably it won't work for other people, or for jenkins?

aledsage:
Looks good; just one comment about vagrant location.

software/database/src/test/java/org/apache/brooklyn/entity/database/mysql/MySqlClusterIntegrationTest.java
neykov: This wasn't supposed to get committed. Replaced with `localhost`.

sjcorbett:
@neykov Thanks. Will merge.

PR #906, opened 2015-09-18
Capture exit code in AbstractSoftwareProcessWinrmDriver

bostko:
- Refactor AbstractSoftwareProcessWinrmDriver implementation like AbstractSoftwareProcessSshDriver
- ScriptHelperUnitTest exit code test

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessWinRmDriver.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessWinRmDriver.java
@@ -19,14 +19,17 @@
 package org.apache.brooklyn.entity.software.base;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Maps;
 import io.cloudsoft.winrm4j.winrm.WinRmToolResponse;
 import org.apache.brooklyn.api.entity.EntityLocal;
 import org.apache.brooklyn.api.mgmt.Task;
 import org.apache.brooklyn.api.sensor.AttributeSensor;
-import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.entity.BrooklynConfigKeys;
+import org.apache.brooklyn.core.entity.Entities;
 import org.apache.brooklyn.core.mgmt.BrooklynTaskTags;
 import org.apache.brooklyn.core.sensor.Sensors;
+import org.apache.brooklyn.entity.software.base.lifecycle.NativeWindowsScriptRunner;
+import org.apache.brooklyn.entity.software.base.lifecycle.WinRmExecuteHelper;
 import org.apache.brooklyn.location.winrm.WinRmMachineLocation;
 import org.apache.brooklyn.util.core.task.Tasks;
 import org.apache.brooklyn.util.exceptions.ReferenceWithError;
@@ -46,7 +49,9 @@ import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 
-public abstract class AbstractSoftwareProcessWinRmDriver extends AbstractSoftwareProcessDriver {
+import static org.apache.brooklyn.util.JavaGroovyEquivalents.elvis;
+
+public abstract class AbstractSoftwareProcessWinRmDriver extends AbstractSoftwareProcessDriver implements NativeWindowsScriptRunner {
     private static final Logger LOG = LoggerFactory.getLogger(AbstractSoftwareProcessWinRmDriver.class);
 
     AttributeSensor<String> WINDOWS_USERNAME = Sensors.newStringSensor("windows.username",
@@ -60,10 +65,26 @@ public abstract class AbstractSoftwareProcessWinRmDriver extends AbstractSoftwar
         entity.setAttribute(WINDOWS_PASSWORD, location.config().get(WinRmMachineLocation.PASSWORD));
     }
 
+    /** @see #newScript(Map, String) */
+    protected WinRmExecuteHelper newScript(String phase) {
+        return newScript(Maps.<String, Object>newLinkedHashMap(), phase);
+    }
+
+    protected WinRmExecuteHelper newScript(Map<String, ?> flags, String phase) {
+        if (!Entities.isManaged(getEntity()))
+            throw new IllegalStateException(getEntity() + " is no longer managed; cannot create script to run here (" + phase + ")");
+
+        WinRmExecuteHelper s = new WinRmExecuteHelper(this, phase + " " + elvis(entity, this));
+        return s;
+    }
+
     @Override
     public void runPreInstallCommand() {
         if (Strings.isNonBlank(getEntity().getConfig(VanillaWindowsProcess.PRE_INSTALL_COMMAND)) || Strings.isNonBlank(getEntity().getConfig(VanillaWindowsProcess.PRE_INSTALL_POWERSHELL_COMMAND))) {
-            executeCommand(VanillaWindowsProcess.PRE_INSTALL_COMMAND, VanillaWindowsProcess.PRE_INSTALL_POWERSHELL_COMMAND, true);
+            executeCommandInTask(
+                    getEntity().getConfig(VanillaWindowsProcess.PRE_INSTALL_COMMAND),
+                    getEntity().getConfig(VanillaWindowsProcess.PRE_INSTALL_POWERSHELL_COMMAND),
+                    "pre-install-command");
         }
         if (entity.getConfig(VanillaWindowsProcess.PRE_INSTALL_REBOOT_REQUIRED)) {
             rebootAndWait();
@@ -78,27 +99,59 @@ public abstract class AbstractSoftwareProcessWinRmDriver extends AbstractSoftwar
     @Override
     public void runPostInstallCommand() {
         if (Strings.isNonBlank(entity.getConfig(BrooklynConfigKeys.POST_INSTALL_COMMAND)) || Strings.isNonBlank(getEntity().getConfig(VanillaWindowsProcess.POST_INSTALL_POWERSHELL_COMMAND))) {
-            executeCommand(BrooklynConfigKeys.POST_INSTALL_COMMAND, VanillaWindowsProcess.POST_INSTALL_POWERSHELL_COMMAND, true);
+            executeCommandInTask(
+                    getEntity().getConfig(VanillaWindowsProcess.POST_INSTALL_COMMAND),
+                    getEntity().getConfig(VanillaWindowsProcess.POST_INSTALL_POWERSHELL_COMMAND),
+                    "post-install-command");
         }
     }
 
     @Override
     public void runPreLaunchCommand() {
         if (Strings.isNonBlank(entity.getConfig(BrooklynConfigKeys.PRE_LAUNCH_COMMAND)) || Strings.isNonBlank(entity.getConfig(VanillaWindowsProcess.PRE_LAUNCH_POWERSHELL_COMMAND))) {
-            executeCommand(BrooklynConfigKeys.PRE_LAUNCH_COMMAND, VanillaWindowsProcess.PRE_LAUNCH_POWERSHELL_COMMAND, true);
+            executeCommandInTask(
+                    getEntity().getConfig(VanillaWindowsProcess.PRE_LAUNCH_COMMAND),
+                    getEntity().getConfig(VanillaWindowsProcess.PRE_LAUNCH_POWERSHELL_COMMAND),
+                    "pre-launch-command");
         }
     }
 
     @Override
     public void runPostLaunchCommand() {
         if (Strings.isNonBlank(entity.getConfig(BrooklynConfigKeys.POST_LAUNCH_COMMAND)) || Strings.isNonBlank(entity.getConfig(VanillaWindowsProcess.POST_LAUNCH_POWERSHELL_COMMAND))) {
-            executeCommand(BrooklynConfigKeys.POST_LAUNCH_COMMAND, VanillaWindowsProcess.POST_LAUNCH_POWERSHELL_COMMAND, true);
+            executeCommandInTask(
+                    getEntity().getConfig(VanillaWindowsProcess.POST_LAUNCH_COMMAND),
+                    getEntity().getConfig(VanillaWindowsProcess.POST_LAUNCH_POWERSHELL_COMMAND),
+                    "post-launch-command");
         }
     }
 
     @Override
     public WinRmMachineLocation getLocation() {
         return (WinRmMachineLocation)super.getLocation();
+    }
+
+    public WinRmMachineLocation getMachine() {
+        return getLocation();
+    }
+
+    protected int executeCommandInTask(String command, String psCommand, String phase) {
+        return newScript(phase)
+                .setCommand(command)
+                .setPsCommand(psCommand)
+                .failOnNonZeroResultCode()
+                .gatherOutput()
+                .execute();
+    }
+
+    @Override
+    public int executeNativeCommand(Map flags, String command, String phase) {
+        return executeNativeOrPsCommand(flags, command, null, phase, true);
+    }
+
+    @Override
+    public int executePsCommand(Map flags, String command, String phase) {
+        return executeNativeOrPsCommand(flags, null, command, phase, true);
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
```diff
@@ -18,32 +18,6 @@
  */
 package org.apache.brooklyn.location.winrm;
 
-import static com.google.common.base.Preconditions.checkNotNull;
```
aledsage: I prefer not having changes that just re-order imports.

Also, the vast majority of our code uses the Eclipse convention for ordering imports (this looks like the Intellij default). I believe that (some) Intellij users change their settings to use the Eclipse ordering, for Brooklyn.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessWinRmDriver.java
```diff
@@ -60,10 +66,27 @@ public AbstractSoftwareProcessWinRmDriver(EntityLocal entity, WinRmMachineLocati
...
+
+        WinRmExecuteHelper s = new WinRmExecuteHelper(this, phase + " " + elvis(entity, this));
+        return s;
+    }
+
     @Override
     public void runPreInstallCommand() {
         if (Strings.isNonBlank(getEntity().getConfig(VanillaWindowsProcess.PRE_INSTALL_COMMAND)) || Strings.isNonBlank(getEntity().getConfig(VanillaWindowsProcess.PRE_INSTALL_POWERSHELL_COMMAND))) {
-            executeCommand(VanillaWindowsProcess.PRE_INSTALL_COMMAND, VanillaWindowsProcess.PRE_INSTALL_POWERSHELL_COMMAND, true);
+            executeCommand(
```
aledsage: Why is this method using `executeCommand`, but the other methods use `executeNativeOrPsCommand`?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessWinRmDriver.java
```diff
@@ -154,11 +214,11 @@ protected WinRmToolResponse executeCommand(ConfigKey<String> regularCommandKey,
         Task<?> currentTask = Tasks.current();
         if (currentTask != null) {
             writeToStream(stdIn, Strings.isBlank(regularCommand) ? powershellCommand : regularCommand);
-            Tasks.addTagDynamically(BrooklynTaskTags.tagForStreamSoft(BrooklynTaskTags.STREAM_STDIN, stdIn));
+            Tasks.addTagDynamically(BrooklynTaskTags.tagForStreamSoft(BrooklynTaskTags.STREAM_STDIN, stdIn)); flags.put("stdin", stdIn);
 
             if (BrooklynTaskTags.stream(currentTask, BrooklynTaskTags.STREAM_STDOUT)==null) {
-                Tasks.addTagDynamically(BrooklynTaskTags.tagForStreamSoft(BrooklynTaskTags.STREAM_STDOUT, stdOut));
-                Tasks.addTagDynamically(BrooklynTaskTags.tagForStreamSoft(BrooklynTaskTags.STREAM_STDERR, stdErr));
+                Tasks.addTagDynamically(BrooklynTaskTags.tagForStreamSoft(BrooklynTaskTags.STREAM_STDOUT, stdOut)); flags.put("stdout", stdOut);
```
aledsage: Can you separate these commands onto multiple lines - it risks being missed by someone reading the code when at the end of a long line like this. Same for the code above.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessWinRmDriver.java
```diff
@@ -134,18 +189,23 @@ protected void createDirectory(String directoryName, String summaryForLogging) {
...
+                return new WinRmToolResponse("", "", 0).getStatusCode();
             } else {
-                throw new IllegalStateException(String.format("Exactly one of %s or %s must be set", regularCommandKey.getName(), powershellCommandKey.getName()));
+                throw new IllegalStateException(String.format("Exactly one of %s or %s must be set", regularCommand, powershellCommand));
             }
         } else if (!Strings.isBlank(regularCommand) && !Strings.isBlank(powershellCommand)) {
-            throw new IllegalStateException(String.format("%s and %s cannot both be set", regularCommandKey.getName(), powershellCommandKey.getName()));
+            throw new IllegalStateException(String.format("%s and %s cannot both be set", regularCommand, powershellCommand));
         }
+        Map flags = Maps.newLinkedHashMap();
```
aledsage: I don't understand how you are using `flags` and `flags2`. You put things into `flags2`, but it looks like the map is never used?

Now that you don't return `WinRmToolResponse`, we need another way for someone to get the stdout/stderr. The way ssh does it is to allow someone to pass in an `OutputStream` in the `flags2` parameter, using the keys `out` and `err`. If they are passed in, then they are populated with the stdout/stderr. That approach is more fiddly code than is really required for this use-case (where the stdout/stderr is not available incrementally anyway) but we should make sure that the stdout/stderr is accessible.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/WinRmExecuteHelper.java
```diff
@@ -0,0 +1,217 @@
...
+        Map<?,?> env = (Map<?,?>) flags.get("env");
+        if (env!=null) {
+            // if not explicitly set, env will come from getShellEnv in AbstractSoftwareProcessSshDriver.execute,
+            // which will also update this tag appropriately
+            tb.tag(BrooklynTaskTags.tagForEnvStream(BrooklynTaskTags.STREAM_ENV, env));
+        }
+
+        if (gatherOutput) {
+            stdout = new ByteArrayOutputStream();
+            tb.tag(BrooklynTaskTags.tagForStreamSoft(BrooklynTaskTags.STREAM_STDOUT, stdout));
```
aledsage: Why do we now have this code here and in `AbstractSoftwareProcessWinRmDriver.executeCommand`? Is there a way to just have it in one place? e.g. could `AbstractSoftwareProcessWinRmDriver.executeCommand` use the `WinRmExecuteHelper` somehow?

software/base/src/test/java/org/apache/brooklyn/entity/software/base/lifecycle/ScriptHelperUnitTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+public class ScriptHelperUnitTest {
+    private static final String NON_ZERO_CODE_COMMAND = "false";
+
+    @Test
+    public void testZeroExitCode() {
+        DoNothingSoftwareProcessDriver sshRunner = mock(DoNothingSoftwareProcessDriver.class);
+
+        ScriptHelper scriptHelper = new ScriptHelper(sshRunner, "test-zero-code-task");
+        Asserts.assertTrue(scriptHelper.executeInternal() == 0, "ScriptHelper doesn't return zero code");
```
aledsage: Better to use `Asserts.assetEquals(scriptHelper.executeInternal(), 0)`. With your code, the error won't tell us what the actual exit code is, which hides information.

software/base/src/test/java/org/apache/brooklyn/entity/software/base/lifecycle/ScriptHelperUnitTest.java
```diff
@@ -0,0 +1,63 @@
...
+        DoNothingSoftwareProcessDriver sshRunner = mock(DoNothingSoftwareProcessDriver.class);
+        when(sshRunner.execute(any(Map.class), any(List.class), any(String.class))).thenReturn(1);
+
+        ScriptHelper scriptHelper = new ScriptHelper(sshRunner, "test-zero-code-task")
+                .body.append(NON_ZERO_CODE_COMMAND);
+        Asserts.assertTrue(scriptHelper.executeInternal() != 0, "ScriptHelper return zero code for non-zero code task");
+    }
+
+    @Test(expectedExceptions = IllegalStateException.class)
+    public void testNonZeroExitCodeException() {
```
aledsage: This method looks the same as `testNonZeroExitCode`. Should it be asserting about an exception being thrown by the mock, rather than the mock returning 1?

software/base/src/test/java/org/apache/brooklyn/entity/software/base/lifecycle/WinRmExecuteHelperUnitTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+public class WinRmExecuteHelperUnitTest {
+    private static final String NON_ZERO_CODE_COMMAND = "false";
+
+    @Test
+    public void testZeroExitCode() {
+        DoNothingWinRmSoftwareProcessDriver nativeWindowsScriptRunner = mock(DoNothingWinRmSoftwareProcessDriver.class);
+
+        WinRmExecuteHelper scriptHelper = new WinRmExecuteHelper(nativeWindowsScriptRunner, "test-zero-code-task");
+        Asserts.assertTrue(scriptHelper.executeInternal() == 0, "WinRmExecuteHelper doesn't return zero code");
```
aledsage: As commented above, don't throw away what the actual value was; I personally wouldn't worry about the custom error message as it will be obvious from the line of code and from the test name.

software/base/src/test/java/org/apache/brooklyn/entity/software/base/lifecycle/ScriptHelperUnitTest.java
aledsage: This method's body looks the same as `testNonZeroExitCode`. Given you are actually asserting with `expectedExceptions` above, then strip out the extra code (e.g. the `assertTrue` at the end of this method).

software/base/src/test/java/org/apache/brooklyn/entity/software/base/lifecycle/WinRmExecuteHelperUnitTest.java
```diff
@@ -0,0 +1,62 @@
...
+        DoNothingWinRmSoftwareProcessDriver nativeWindowsScriptRunner = mock(DoNothingWinRmSoftwareProcessDriver.class);
+        when(nativeWindowsScriptRunner.executeCommand(any(Map.class), any(String.class), any(String.class), any(String.class), any(Boolean.class))).thenReturn(1);
+
+        WinRmExecuteHelper scriptHelper = new WinRmExecuteHelper(nativeWindowsScriptRunner, "test-zero-code-task")
+                .setCommand(NON_ZERO_CODE_COMMAND);
+        Asserts.assertTrue(scriptHelper.executeInternal() != 0, "WinRmExecuteHelper return zero code for non-zero code task");
+    }
+
+    @Test(expectedExceptions = IllegalStateException.class)
+    public void testNonZeroExitCodeException() {
```
aledsage: This method's body looks the same as `testNonZeroExitCode`. Given you are actually asserting with `expectedExceptions` above, then strip out the extra code (e.g. the `assertTrue` at the end of this method).

aledsage:
We need live tests that the different exit codes are actually respected, when thrown by WinRM powershell or native commands - those are vital for having confidence that the WinRM stuff is working.

aledsage:
I'm concerned that we've made the code more complicated by adding the `WinRmExecuteHelper.java`, but without it having the same functionality as the ssh version (e.g. it's impossible to get the stdout/stderr back when you use it).

We don't actually need the `WinRmExecuteHelper` currently (because things are called from yaml) so we can defer that, but if we're deferring it then it feels like we could have deferred adding this class just now.

With this change, we are much of the way there to having the more powerful execution helper, but we have lost functionality compared to where we were before (i.e. you can't get the stdout/stderr if writing Java).

software/base/src/test/java/org/apache/brooklyn/entity/software/base/lifecycle/ScriptHelperUnitTest.java
bostko: We should think about extending `org.apache.brooklyn.test.Asserts` from `org.testng.Assert`

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessWinRmDriver.java
```diff
@@ -134,17 +189,20 @@ protected void createDirectory(String directoryName, String summaryForLogging) {
         getLocation().executePsScript("New-Item -path \"" + directoryName + "\" -type directory -ErrorAction SilentlyContinue");
     }
 
-    protected WinRmToolResponse executeCommand(ConfigKey<String> regularCommandKey, ConfigKey<String> powershellCommandKey, boolean allowNoOp) {
-        String regularCommand = getEntity().getConfig(regularCommandKey);
-        String powershellCommand = getEntity().getConfig(powershellCommandKey);
-        if (Strings.isBlank(regularCommand) && Strings.isBlank(powershellCommand)) {
+    public int executeCommandInTask(String regularCommand, String powershellCommand, String phase, Boolean allowNoOp) {
```
aledsage: It is surprising that the presence of the `Boolean allowNoOp` in the overloaded method causes it to go down a completely different code path, not calling `newScript(...)...` in the same was as the other `executeCommandInTask` method.

Do you really mean for this `executeCommandInTask` method that it is *not* executed in a task? Or should it do the no-op guard, and then delegate to the other `executeCommandInTask` method?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessWinRmDriver.java
```diff
@@ -134,37 +187,40 @@ protected void createDirectory(String directoryName, String summaryForLogging) {
...
-        } else if (!Strings.isBlank(regularCommand) && !Strings.isBlank(powershellCommand)) {
-            throw new IllegalStateException(String.format("%s and %s cannot both be set", regularCommandKey.getName(), powershellCommandKey.getName()));
+        } else if (!Strings.isBlank(regularCommand) && !Strings.isBlank(powerShellCommand)) {
+            throw new IllegalStateException(String.format("%s and %s cannot both be set", regularCommand, powerShellCommand));
         }
 
         ByteArrayOutputStream stdIn = new ByteArrayOutputStream();
-        ByteArrayOutputStream stdOut = new ByteArrayOutputStream();
-        ByteArrayOutputStream stdErr = new ByteArrayOutputStream();
+        ByteArrayOutputStream stdOut = flags.get("out") != null ? (ByteArrayOutputStream)flags.get("out") : new ByteArrayOutputStream();
```
aledsage: We should not assume that the caller has passed in a `ByteArrayOutputStream` when they have populated the `out` or `err` key in `flags`. 

Let's not worry about it in this PR though. I see that `AbstractSoftwareProcessSshDriver` is bad as well, but in a different way. It overwrites any stdout and stderr keys (if any) that were passed in by the user.

software/base/src/test/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessStreamsTest.java
```diff
@@ -105,8 +105,8 @@ protected String getStreamOrFail(Task<?> task, String streamType) {
             String stdin = getStreamOrFail(subTask, BrooklynTaskTags.STREAM_STDIN);
             String stdout = getStreamOrFail(subTask, BrooklynTaskTags.STREAM_STDOUT);
             String stderr = getStreamOrFail(subTask, BrooklynTaskTags.STREAM_STDERR);
-            String env = getStreamOrFail(subTask, BrooklynTaskTags.STREAM_ENV);
-            String msg = "stdin="+stdin+"; stdout="+stdout+"; stderr="+stderr+"; env="+env;
+//            String env = getStreamOrFail(subTask, BrooklynTaskTags.STREAM_ENV);
```
aledsage: I strongly dislike commented out that has no comment to say why it's commented out and why it might be useful in the future (versus it being deleted).

aledsage:
Thanks @bostko - nerging now. We can iterate on it in additional PRs as required.

nakomis:
This [1] requires java 8. Additionally, `ScriptHelperUnitTest.testTaskNoGatherOutput` fails!

PR #908, opened 2015-09-19
Fix jboss service not up

aledsage:
Previously, serviceUp was never set to true, for two reasons:
    
    1. because connectServiceUpIsRunning was not called, nothing was
       unsetting the service-not-up-indicator for SERVICE_PROCESS_IS_RUNNING
    
    2. because the service-not-up-indicator for MANAGEMENT_URL_UP said
       suppressDuplicates=true, then it checked if the map was equal to
       the “old” map before publishing. But the map was the same object,
       and that object had been modified in-place. So the object was
       always equal to itself. It therefore never did an emit(), so the
       enricher that listened to it to set serviceUp never fired.

Also fix UpdatingMap for if suppressDuplicates=true:
    
- log.warn to say it should not be set.
- reset config to suppressDuplicates=false

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/enricher/stock/UpdatingMap.java
+++ b/core/src/main/java/org/apache/brooklyn/enricher/stock/UpdatingMap.java
@@ -87,11 +87,30 @@ public class UpdatingMap<S,TKey,TVal> extends AbstractEnricher implements Sensor
 
     public UpdatingMap(Map<Object, Object> flags) {
         super(flags);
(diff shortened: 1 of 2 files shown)


aledsage:
This bug means that Jboss7 is broken in 0.8.0-incubating. Once reviewed+merged into master, we should cherry-pick this into the 0.8.x branch if we are going to do a 0.8.1 release.

aledsage:
asfbot failure caused by a test that was failing in master. I fixed that in PR #913 and have rebased this branch again.

richardcloudsoft:
Code looks good. Good to merge once asfbot says it passes.

PR #909, opened 2015-09-20
Usage goodies

ahgittin:
* CLI accepts `-Da=b -Dc=d ...`
* JS GUI uses a custom cookie ID so as not to conflict with multiple brooklyn sessions or other services (e.g. alien4cloud) coming from the same server

### DIFF
--- a/usage/cli/src/main/java/org/apache/brooklyn/cli/AbstractMain.java
+++ b/usage/cli/src/main/java/org/apache/brooklyn/cli/AbstractMain.java
@@ -33,19 +33,22 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Callable;
 
 import javax.inject.Inject;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.apache.brooklyn.api.catalog.BrooklynCatalog;
 import org.apache.brooklyn.cli.Main.LaunchCommand;
 import org.apache.brooklyn.core.BrooklynVersion;
+import org.apache.brooklyn.util.collections.MutableList;
 import org.apache.brooklyn.util.exceptions.FatalConfigurationRuntimeException;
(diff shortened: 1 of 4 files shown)


usage/cli/src/main/java/org/apache/brooklyn/cli/AbstractMain.java
```diff
@@ -150,6 +153,34 @@ public ToStringHelper string() {
...
+     */
+    public static abstract class BrooklynCommandWithSystemDefines extends BrooklynCommandCollectingArgs {
+        @Option(type = OptionType.GLOBAL, name = { "-D" }, description = "Set java system property")
+        public List<String> defines1;
+
+        @Option(name = { "-D" }, description = "Set java system property")
+        public List<String> defines2;
+
+        public List<String> getDefines() { return MutableList.copyOf(defines1).appendAll(defines2); }
+        public Map<String,String> getDefinesAsMap() { return KeyValueParser.parseMap(Strings.join(getDefines(),",")); }
```
aledsage: Feels like it would be good to have a `KeyValueParser.parseMapEntries(List<String>)`, rather than us having to join them so they can be split apart again (which might lead to problems if a value contained a comma, and was not being quoted here).

But this is good enough for now.

aledsage:
Looks good; one minor comment that could be addressed another time. Merging.

rdowner:
This has a failing unit test - master is red in Jenkins - https://builds.apache.org/job/incubator-brooklyn-master-build/561/ @ahgittin @aledsage 

```
2015-09-21 12:30:22,705 INFO  TESTNG FAILED: "Surefire test" - org.apache.brooklyn.cli.CliTest.testDefaultInfoCommand() finished in 22 ms
io.airlift.command.ParseArgumentsUnexpectedException: Found unexpected parameters: []
	at io.airlift.command.Cli.validate(Cli.java:148)
	at io.airlift.command.Cli.parse(Cli.java:116)
	at io.airlift.command.Cli.parse(Cli.java:97)
	at org.apache.brooklyn.cli.CliTest.testDefaultInfoCommand(CliTest.java:314)
```

PR #910, opened 2015-09-21
Regex replacement dsl

nakomis:
Adds ability to perform a regex replacement in YAML. Either directly as a string, or returning a `Function<String, String>` for use in (e.g.) `Transformer.TRANSFORMATION_FROM_VALUE`


aledsage:
@nakomis There aren't any YAML tests - can you add those please?

core/src/main/java/org/apache/brooklyn/core/sensor/DependentConfiguration.java
```diff
@@ -483,6 +483,108 @@ public T apply(@Nullable U input) {
...
+
+        @Override
+        public Function<String, String> apply(List<Object> input) {
+            Iterator<?> taskArgsIterator = input.iterator();
+            return new RegexReplacer(resolveArgument(pattern, taskArgsIterator), resolveArgument(replacement, taskArgsIterator));
+        }
+
+    }
+
+    private static String resolveArgument(Object argument, Iterator<?> taskArgsIterator) {
```
aledsage: Worth having some javadoc (even though the method is small, it's behaviour is non-obvious). It wasn't clear until looking carefully why would increment the iterator sometimes (i.e. call `next()`) and not other times.

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/BrooklynDslInterpreter.java
```diff
@@ -168,6 +168,17 @@ public Object evaluateOn(Object o, FunctionWithArgs f, boolean deepEvaluation) {
         
         String fn = f.getFunction();
         fn = Strings.removeFromStart(fn, "$brooklyn:");
+        if (fn.startsWith("function.")) {
+            for (Class<?> c : clazz.getClasses()) {
+                if (c.equals(BrooklynDslCommon.Functions.class)) {
+                    if (o instanceof Class) {
```
aledsage: It's not obvious what you're doing here or why. What do you expect to be passed in (i.e. what is `o`)? Why will `clazz` be a sub-type of (or the concrete type) `BrooklynDslCommon.Functions`?

nakomis:
PR Comments addressed

nakomis:
Was based on broken master, hence build failures. Rebased

core/src/main/java/org/apache/brooklyn/enricher/stock/AbstractTransformer.java
```diff
@@ -57,14 +57,16 @@ public AbstractTransformer() {
...
-        Function<SensorEvent<T>, U> transformation = getTransformation();
         this.producer = getConfig(PRODUCER) == null ? entity: getConfig(PRODUCER);
         this.sourceSensor = (Sensor<T>) getRequiredConfig(SOURCE_SENSOR);
         Sensor<?> targetSensorSpecified = getConfig(TARGET_SENSOR);
         this.targetSensor = targetSensorSpecified!=null ? (Sensor<U>) targetSensorSpecified : (Sensor<U>) this.sourceSensor;
         if (producer.equals(entity) && targetSensorSpecified==null) {
+            // We cannot call getTransformation() here to log the tranformation, as it will attempt
+            // to resolve the transformation, which will cause the entity initialization thread to
+            // block
             LOG.error("Refusing to add an enricher which reads and publishes on the same sensor: "+
```
ahgittin: why not log details of the transformer ?  that would be useful i think to debug

core/src/main/java/org/apache/brooklyn/enricher/stock/Transformer.java
```diff
@@ -44,9 +44,6 @@
     public static ConfigKey<Function<?, ?>> TRANSFORMATION_FROM_VALUE = ConfigKeys.newConfigKey(new TypeToken<Function<?, ?>>() {}, "enricher.transformation");
     public static ConfigKey<Function<?, ?>> TRANSFORMATION_FROM_EVENT = ConfigKeys.newConfigKey(new TypeToken<Function<?, ?>>() {}, "enricher.transformation.fromevent");
     
-    public Transformer() {
```
ahgittin: i like to include the explicit no-args constructor if it's intended; no harm in removing it but no point either....

core/src/main/java/org/apache/brooklyn/core/sensor/DependentConfiguration.java
```diff
@@ -483,6 +483,114 @@ public T apply(@Nullable U input) {
...
+            resolvedArgument = taskArgsIterator.next();
+        } else if (argument instanceof DeferredSupplier) {
+            resolvedArgument = ((DeferredSupplier<?>) argument).get();
+        } else {
+            resolvedArgument = argument;
+        }
+        return String.valueOf(resolvedArgument);
+    }
+
+    public static class RegexReplacer implements Function<String, String> {
```
ahgittin: move to `StringFunctions` ?

core/src/main/java/org/apache/brooklyn/core/sensor/DependentConfiguration.java
```diff
@@ -483,6 +483,114 @@ public T apply(@Nullable U input) {
...
+        public String apply(@Nullable List<Object> input) {
+            Iterator<?> taskArgsIterator = input.iterator();
+            String resolvedSource = resolveArgument(source, taskArgsIterator);
+            String resolvedPattern = resolveArgument(pattern, taskArgsIterator);
+            String resolvedReplacement = resolveArgument(replacement, taskArgsIterator);
+            return new RegexReplacer(resolvedPattern, resolvedReplacement).apply(resolvedSource);
+        }
+    }
+
+    public static class RegexTransformerFunction implements Function<List<Object>, Function<String, String>> {
```
ahgittin: feels a bit convoluted but okay with me if you think this is the right way.  maybe mark `@Beta` ?

core/src/main/java/org/apache/brooklyn/enricher/stock/Transformer.java
hzbarcea: I went back and forth for year on the implicit vs explicit argument. An argument would be that if it's there with no value, one may infer that there was an intention to do something but then the developer forgot. One convention made in other projects is to add a comment in the empty block that says something like "intentionally left blank". This applies even more to empty catch() blocks. My $0.02.

usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/DslAndRebindYamlTest.java
```diff
@@ -241,4 +243,72 @@ private Entity entityWithFormatString() throws Exception {
             "    test.confName: $brooklyn:formatString(\"hello %s\", \"world\")");
     }
 
+
+    /*
+           - type: org.apache.brooklyn.enricher.stock.Transformer
```
aledsage: strange indentation - is this a tab?

core/src/main/java/org/apache/brooklyn/enricher/stock/Transformer.java
aledsage: No particularly strong feelings from me, but I'd lean towards including the no-arg constructor for two reasons:
1. It's consistent with the other enrichers (we could change them all, but having some of each pattern seems more confusing).
2. The contract for enrichers/policies/entities/locations to be instantiated through the `EnricherSpec` etc is that the class must have a no-arg constructor. We don't expect people to call this constructor directly.

The second point means we probably should include the constructor with a comment. We could maybe even change `InternalPolicyFactory` etc so that it can handle calling protected constructors, which would enforce that more.

Anyway, I'm happy to ignore it in this PR.

core/src/main/java/org/apache/brooklyn/core/sensor/DependentConfiguration.java
aledsage: I agree it feels convoluted.

When the function stuff was going to be added, I initially expected it to return a normal(ish) function. The pattern/replacement could come from attributeWhenReady etc, but the `apply(input)` would expect normal input. For example, this could be used in a `Transformer.TRANSFORMATION_FROM_VALUE`.

This use of function seems to be by-passing the use of a transformer in order to get the input via another attributeWhenReady, rather than getting it from the enricher's subscription.

core/src/main/java/org/apache/brooklyn/enricher/stock/AbstractTransformer.java
nakomis: As stated in the comment, we can't call getTransformation() here as the entity is being initialized, so the call will block

sjcorbett:
@nakomis Could you rebase against master?

nakomis:
@sjcorbett rebased

core/src/main/java/org/apache/brooklyn/enricher/stock/AbstractTransformer.java
ahgittin: of course -- sorry for daft comment.  would it be useful to log `JavaClassNames.simpleClassName(this)` ?

ahgittin:
nice bit of work.  good to merge.  @nakomis shall i make that one change?

PR #911, opened 2015-09-21
Refactor: deprecate EntityLocal methods, and group methods on Entity

aledsage:


api/src/main/java/org/apache/brooklyn/api/entity/Entity.java
```diff
@@ -314,4 +361,80 @@
          */
         <T> void emit(Sensor<T> sensor, T value);
     }
+    
+    @Beta
+    public interface PolicySupport {
+        /**
+         * @return an immutable thread-safe view of the policies.
+         */
+        Collection<Policy> getPolicies();
```
robertgmoss: why not make this interface extend Iterable instead?

api/src/main/java/org/apache/brooklyn/api/entity/Entity.java
```diff
@@ -314,4 +361,80 @@
...
+        
+        /**
+         * Removes the given enricher from this entity. 
+         * @return True if the policy enricher at this entity; false otherwise
+         */
+        boolean removeEnricher(Enricher enricher);
+    }
+    
+    @Beta
+    public interface GroupSupport {
```
robertgmoss: Feels like these 3 interfaces could extend a generified super interface

robertgmoss:
Really nice @aledsage - only a couple of minor suggestions from me.

api/src/main/java/org/apache/brooklyn/api/entity/Entity.java
aledsage: Do you mean `public interface PolicySupport extends Iterable<Policy>`? I'd prefer not for a couple of reasons. First if you wrote `entity.policy().` the auto-complete would mix the addPolicy/removePolicy with the other methods on Iterable. Second, the intent is consistency for things on the Entity api - for `config()`, `sensor()` and `subscriptions()` they don't return Iterable so I'm inclined to not have policies/enrichers return an iterable either.

api/src/main/java/org/apache/brooklyn/api/entity/Entity.java
aledsage: What are you thinking of on the generified super interface?

api/src/main/java/org/apache/brooklyn/api/entity/Entity.java
robertgmoss: There is only one method on `Iterable`: `Iterator<T> iterator()`.  And I think it would be nicer if we didn't have `policies().getPolicies()`.  I think the same should be true of `config()` and `sensor()` etc. they would benefit from being an `Iterable`.

api/src/main/java/org/apache/brooklyn/api/entity/Entity.java
robertgmoss: Something like `public interface AdjunctSupport<T extends EntityAdjunct> extends Iterable<T> ` with
`iterator()`, `add()`, and `remove()` with  `public interface EnricherSupport extends AdjunctSupport<Enricher>` perhaps?

api/src/main/java/org/apache/brooklyn/api/entity/Entity.java
robertgmoss: The other benefit is that it can be used in the enhanced for loop, e.g. `for(Policy p : policies())`.  The implementation could just return the `Iterator` from the `Collection` that is currently constructed for `getPolicies()`

aledsage:
@robertgmoss thanks, you've convinced me on both counts (implementing `Iterable` and having the common super interface). I've made those changes, and rebased against master.

Waiting for asfbot and will then merge.

PR #912, opened 2015-09-21
Fix service-not-up-diagnostics, add subscription.notifyOfInitialValue, etc

aledsage:
Various fixes for things encountered (or speculatively encountered for "notifyOfInitialValue"!) when testing upgrade from an old Brooklyn to a new brooklyn.


rdowner:
Code looks OK. Jenkins PR builder reports failure, but this appears to be due to PR #909 breaking master.

aledsage:
I've rebased against master (now that test is fixed in #913 ); can see if asfbot is happy now.

aledsage:
Thanks @rdowner - jenkins has now confirmed that the tests pass. Merging.

PR #913, opened 2015-09-21
Fix CliTest.testDefaultInfoCommand

aledsage:


aledsage:
Fixes failing test introduced in https://github.com/apache/incubator-brooklyn/pull/909, which I didn't notice until after I'd merged it!

Merging this PR now so that master is not left broken.

ahgittin:
looks good -- sorry i missed this + thanks

PR #915, opened 2015-09-21
Resolve DSL functions for a configuration key within any collection

tbouron:
This will look inside a collection (Set, Map or List) and will resolve the value that uses $brooklyn:*


ahgittin:
LGTM

ahgittin:
failures unrelated /cc @aledsage :

* org.apache.brooklyn.policy.loadbalancing.LoadBalancingPolicySoakTest.testLoadBalancingQuickTest
* org.apache.brooklyn.policy.ha.ConnectionFailureDetectorTest.setUp

PR #918, opened 2015-09-22
Adding CloudFoundry support.

kiuby88:
For adding PaaS support, a CloudFoundryLocation is added which is based on PaasLocation.
This new location contains a API client for CloudFoundry Paas Services, (CloudFoundry Java Client), that allows
to use the services of this provider.

Extracting ssh behaviour from SoftwareProcessImpl: provisioning flags

For adding paas support, SoftwareProcess entities should to be agnostics in order to allows to be deployed
on SSH and PaaS locations. So, the machine behaviour and management should be decoupled of the entities
and it should be moved to an isolate class.

SoftwareProcessImpl contained a pair of method for obtaining the needed flags to create and configure a VM.
However as it was mentioned previously, this entity shouldn't contain any ssh (machine) behaviour. Then, this
code is moved to an external class, MachineProvisioningLocationFlagsSupplier that is managed in
SoftwareProcessImpl as a LocationFlagSupplier interface. Moreover, an equivalent class is added to represent
this behaviour (obtain flags location) for PaaS.

SoftwareProcessImpl will select a behaviour for obtaining the flags depending on the selected location for deploying
the entity.

Adding BehaviourFactory for selecting PaaS or SSH behaviour in run time.

Following the previous commits, SoftwareProcessImpl entities should select PaaS or SSH behaviour in run time,
depending on the target location. An entity know the target location when it is started (calling to start effector), then just
when this effector is called the entity will able to know the selected location, so in this moment the entity will able to find
the required behaviour, SSH or PaaS.

The whole behaviour is composed by several classes, then SoftwareProcessImpl uses a factory (BehaviourFactory) which is selected according
to the target location. This factory allows to select in run time all behaviour focus in a location.

Adding PaaS support to TomcatServer entity.

The drivers for CloudFoundry are added to SoftwareProcess entity and JavaWebAppSoftwareProcess implementation entity.
The final PaaS support is added to TomcatServer entity using a new driver which is focused on CloudFoundryPaaSLocation.

A new hierarchy of drivers is added to Brooklyn, these new drivers is oriented to PaaS but they follow the hierarchy
defined for SSH. For example, AbstractSoftwareProcessDriver, AbstractSoftwareProcessSshDriver allow to SoftwareProcess entities
use drivers for location based on machines. Then, to enable the PaaS support for SoftwareProcess entities a set of drivers is added such as
AbstractSoftwareProcessCloudFoundryDriver and AbstractApplicationCloudFoundryDriver.
Fixing dependency convergence.

TODO: 
* It should be necessary defined the start effector body in SoftwareProcessImpl because currently the first target location fixes the behavior of the entity.
* The restart effector should be defined.
* Some methods could be renamed.


robertgmoss:
@kiuby88 Thanks for this, would you mind rebasing and I'll dive into this shortly?

kiuby88:
Thanks for your reply.
@robertgmoss, I will rebase asap.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
```diff
@@ -599,15 +561,67 @@ public void waitForEntityStart() {
      */
     @Override
     public final void start(final Collection<? extends Location> locations) {
+
```
kiuby88: This method fix the type location for this entity, for example ssh or PaaS. So, it is not possible to start (migrate #924) the entity to a new kind of location.

PR #919, opened 2015-09-23
Allow sensors to be aggregated using a Reducer

robertgmoss:
Combines sensors using a function and publishes into a target sensor, e.g

```
location: localhost
services:
- type: org.apache.brooklyn.entity.stock.BasicApplication

  brooklyn.enrichers:
  - type: org.apache.brooklyn.enricher.stock.reducer.Reducer
    brooklyn.config:
      enricher.sourceSensors:
      - $brooklyn:sensor("first.name")
      - $brooklyn:sensor("last.name")
      enricher.targetSensor:
        $brooklyn:sensor("full.name")
      enricher.reducerFunction.transformation: formatString
      enricher.reducerFunction.parameters:
        format: "%s %s"

  brooklyn.initializers:
  - type: org.apache.brooklyn.core.sensor.StaticSensor
    brooklyn.config:
      name: first.name
      targetType: String
      static.value: Foo
  - type: org.apache.brooklyn.core.sensor.StaticSensor
    brooklyn.config:
      name: last.name
      targetType: String
      static.value: Bar
```


core/src/main/java/org/apache/brooklyn/enricher/stock/Enrichers.java
```diff
@@ -676,6 +681,62 @@ public String toString() {
...
+
+    protected abstract static class AbstractReducerBuilder<S, T, B extends AbstractReducerBuilder<S, T, B>> extends AbstractEnricherBuilder<B> {
+        protected AttributeSensor<T> publishing;
+        protected Entity fromEntity;
+        protected List<AttributeSensor<S>> reducing;
+        protected Function<List<S>, T> computing;
+        protected String functionName;
+        private Map<String, Object> parameters;
+
+        public AbstractReducerBuilder(Class<? extends Reducer<S, T>> clazz, List<AttributeSensor<S>> val) {
```
mikezaccardo: Add `@Nonnull` to this and the following args with `null` checks?

core/src/main/java/org/apache/brooklyn/enricher/stock/Transformer.java
```diff
@@ -23,6 +23,8 @@
 import org.apache.brooklyn.api.sensor.SensorEvent;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.entity.Attributes;
```
mikezaccardo: These imports are never used

core/src/main/java/org/apache/brooklyn/enricher/stock/reducer/GenericStringReducer.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.google.common.base.Function;
+import com.google.common.base.Preconditions;
+
+public abstract class GenericStringReducer<T> extends Reducer<T, String>{
+
+    @Override
+    protected Function<List<T>, String> createReducerFunction(
+            String reducerName, Map<String, ?> parameters) {
+        if (reducerName.equals("formatString")){
```
mikezaccardo: Would be good to use `Objects.equal(reducerName, "formatString")` to avoid NPE if `reducerName` is `null`.

core/src/main/java/org/apache/brooklyn/enricher/stock/reducer/GenericStringReducer.java
```diff
@@ -0,0 +1,39 @@
...
+            String format = Preconditions.checkNotNull((String)parameters.get("format"), "format");
```
mikezaccardo: The second argument should be an error message rather than `"format"`, right?

core/src/main/java/org/apache/brooklyn/enricher/stock/reducer/Reducer.java
```diff
@@ -0,0 +1,181 @@
...
+    protected Entity producer;
+    protected List<AttributeSensor<S>> subscribedSensors;
+    protected Sensor<T> targetSensor;
+    protected Function<List<S>, T> reducerFunction;
+
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    @Override
+    public void setEntity(EntityLocal entity) {
+        super.setEntity(entity);
+        Preconditions.checkNotNull(getConfig(SOURCE_SENSORS), "source sensors");
```
mikezaccardo: Same comment about error message

core/src/main/java/org/apache/brooklyn/enricher/stock/reducer/Reducer.java
```diff
@@ -0,0 +1,181 @@
...
+
+        this.producer = getConfig(PRODUCER) == null ? entity : getConfig(PRODUCER);
+        List<AttributeSensor<S>> sensorListTemp = Lists.newArrayList();
+
+        for (Object sensorO : getConfig(SOURCE_SENSORS)) {
+            AttributeSensor<S> sensor = Tasks.resolving(sensorO).as(AttributeSensor.class).timeout(ValueResolver.REAL_QUICK_WAIT).context(producer).get();
+            if(!sensorListTemp.contains(sensor)) {
```
mikezaccardo: Sanity check: neither `AttributeSensor` nor any of its children define an `equals()` method -- is it possible for there to be two sensors here which are logically equal but not actually the same (different memory locations)? If yes, then this check won't work.

core/src/main/java/org/apache/brooklyn/enricher/stock/reducer/StringStringReducer.java
```diff
@@ -0,0 +1,41 @@
...
+    
+    public StringStringReducer() {}
+
+    @Override
+    protected Function<List<String>, String> createReducerFunction(
+            String reducerName, Map<String, ?> parameters) {
+        Function<List<String>, String> function = super.createReducerFunction(reducerName, parameters);
+        if(function != null) return function;
+        
+        if(reducerName.equals("joiner")){
```
mikezaccardo: Same comment about `Objects.equal()`

core/src/main/java/org/apache/brooklyn/enricher/stock/reducer/Reducer.java
```diff
@@ -0,0 +1,181 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(Reducer.class);
+
+    @SetFromFlag("producer")
+    public static ConfigKey<Entity> PRODUCER = ConfigKeys.newConfigKey(Entity.class, "enricher.producer");
+    public static ConfigKey<Sensor<?>> TARGET_SENSOR = ConfigKeys.newConfigKey(new TypeToken<Sensor<?>>() {}, "enricher.targetSensor");
+    public static ConfigKey<List<? extends AttributeSensor<?>>> SOURCE_SENSORS = ConfigKeys.newConfigKey(new TypeToken<List<? extends AttributeSensor<?>>>() {}, "enricher.sourceSensors");
+    public static ConfigKey<Function<List<?>,?>> REDUCER_FUNCTION = ConfigKeys.newConfigKey(new TypeToken<Function<List<?>, ?>>() {}, "enricher.reducerFunction");
+    @SetFromFlag("transformation")
+    public static final ConfigKey<String> REDUCER_FUNCTION_UNTYPED = ConfigKeys.newStringConfigKey("enricher.reducerFunction.untyped",
+        "A string matching a pre-defined named reducer function, such as join");
```
mikezaccardo: Should be "joiner" as the example

core/src/main/java/org/apache/brooklyn/enricher/stock/reducer/Reducer.java
```diff
@@ -0,0 +1,181 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(Reducer.class);
+
+    @SetFromFlag("producer")
+    public static ConfigKey<Entity> PRODUCER = ConfigKeys.newConfigKey(Entity.class, "enricher.producer");
+    public static ConfigKey<Sensor<?>> TARGET_SENSOR = ConfigKeys.newConfigKey(new TypeToken<Sensor<?>>() {}, "enricher.targetSensor");
+    public static ConfigKey<List<? extends AttributeSensor<?>>> SOURCE_SENSORS = ConfigKeys.newConfigKey(new TypeToken<List<? extends AttributeSensor<?>>>() {}, "enricher.sourceSensors");
+    public static ConfigKey<Function<List<?>,?>> REDUCER_FUNCTION = ConfigKeys.newConfigKey(new TypeToken<Function<List<?>, ?>>() {}, "enricher.reducerFunction");
+    @SetFromFlag("transformation")
+    public static final ConfigKey<String> REDUCER_FUNCTION_UNTYPED = ConfigKeys.newStringConfigKey("enricher.reducerFunction.untyped",
```
mikezaccardo: I understand why it's called "untyped" in the context of the code, but is there possibly a more user-friendly name for this?

mikezaccardo:
LGTM -- tests pass and only minor comments

core/src/main/java/org/apache/brooklyn/enricher/stock/Enrichers.java
robertgmoss: This would be inconsistent with the other uses throughout, and could cause confusion.  I'm inclined to say that this should be done in another PR.  Moreover, this is a protected class and not public.  Perhaps a discussion on the mailing list about whether this should be adopted widely.

robertgmoss:
Review comments addressed

core/src/test/java/org/apache/brooklyn/enricher/stock/reducer/ReducerTest.java
```diff
@@ -0,0 +1,255 @@
...
+            return result.toString();
+        }
+    }
+    
+    public static class StringIntegerReducer extends Reducer<String, Integer> {
+        
+        public StringIntegerReducer() {}
+
+        @Override
+        protected Function<List<String>, Integer> createReducerFunction(
```
neykov: Have this as the default implementation in `Reducer`?

core/src/main/java/org/apache/brooklyn/enricher/stock/Enrichers.java
```diff
@@ -171,6 +173,9 @@ public PropagatorBuilder propagatingAllBut(Iterable<? extends Sensor<?>> vals) {
         public JoinerBuilder joining(AttributeSensor<?> source) {
             return new JoinerBuilder(source);
         }
+        public <S, T> ReducerBuilder<S, T> reducing(Class<? extends Reducer<S, T>> clazz, List<AttributeSensor<S>> sourceSensors) {
```
neykov: Would be nice to have an additional method factory which just uses the generic `Reducer` implementation, working with `Function<List<?>, ?>` params - which I guess would be the main way it's used in java land.

neykov:
All the comments I have are in bike-shedding land, so feel free to ignore. Here are some more:
  * From a user perspective I would prefer to use a single `Reducer` which knows how to load any transformation alias (as opposed to having to look up which implementation does which alias).
  * When working with sensors (especially in yaml land) you never know what the type of the input sensor(s) will be even if the interface is typed in Java. So as a rule of thumb it's nice to `coerce` to the required type instead of forcing the sensor to be of the required type. This would also alleviate the need for all the XXXStringReducer alternatives.


Looks good and can be merged as is.

core/src/main/java/org/apache/brooklyn/enricher/stock/reducer/Reducer.java
```diff
@@ -0,0 +1,187 @@
...
+    @Override
+    public void onEvent(SensorEvent<Object> event) {
+        Sensor<T> destinationSensor = (Sensor<T>) getConfig(TARGET_SENSOR);
+
+        List<S> values = Lists.newArrayList();
+
+        for (AttributeSensor<S> sourceSensor : subscribedSensors) {
+            S resolvedSensorValue = entity.sensors().get(sourceSensor);
+            if (resolvedSensorValue == null) {
+                // only apply function if all values are resolved
```
aledsage: Is this always the behaviour we want - to not call the function if any of the values are null? My gut feel is to leave that up to the function (it can filter nulls if it wants).

Doing it this way could lead to strange behaviour - e.g. sensor X and Y get set to "a" and "b", then X gets set to `null`. The function will not be notified about the change of X.

I don't have any particular use-cases in mind where we'd want to know about the null - it's just a gut feel that leaving it up to the function gives more flexibility.

core/src/main/java/org/apache/brooklyn/enricher/stock/reducer/StringStringReducer.java
```diff
@@ -0,0 +1,42 @@
...
+    
+    public StringStringReducer() {}
+
+    @Override
+    protected Function<List<String>, String> createReducerFunction(
+            String reducerName, Map<String, ?> parameters) {
+        Function<List<String>, String> function = super.createReducerFunction(reducerName, parameters);
+        if(function != null) return function;
+        
+        if(Objects.equals(reducerName, "joiner")){
```
aledsage: This feels like a strange place to define the names of functions that we'll accept.

I presume these conveniences are for yaml (as in Java one would just use `StringFunctions.joiner()` directly)?

Should we be adding support for different functions to https://github.com/nakomis/incubator-brooklyn/blob/master/usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java ? As is done in https://github.com/apache/incubator-brooklyn/pull/910/files#diff-d108c16ec590974e317590c54f71e9da

core/src/main/java/org/apache/brooklyn/enricher/stock/reducer/Reducer.java
```diff
@@ -0,0 +1,187 @@
...
+
+        Object result = reducerFunction.apply(values);
+
+        if (LOG.isTraceEnabled()) LOG.trace("enricher {} got {}, propagating via {} as {}",
+                new Object[] {this, event, entity, reducerFunction, destinationSensor});
+
+        emit((Sensor<T>)destinationSensor, result);
+    }
+   
+    public static class JoinerReducerFunction<A> implements Function<List<A>, String> {
```
aledsage: Why not use `StringFunctions.joiner(separator)` (which delegates to `Strings.join` for its `apply` method)?

core/src/main/java/org/apache/brooklyn/enricher/stock/reducer/Reducer.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    public static class JoinerFunction extends JoinerReducerFunction<String>{
+        
+        public JoinerFunction(Object separator) {
+            super(separator);
+        }
+    }
+
+    public static class ToStringReducerFunction<A> implements Function<List<A>, String> {
```
aledsage: See guava's `Functions#toStringFunction()` (which we delegate to in Brooklyn's `StringFunctions.toStringFunction`).

core/src/main/java/org/apache/brooklyn/enricher/stock/reducer/Reducer.java
```diff
@@ -0,0 +1,187 @@
...
+
+        @Override
+        public String apply(List<A> input) {
+            return input.toString();
+        }
+        
+    }
+    
+    public static class FormatStringReducerFunction<T> implements Function<List<T>, String> {
```
aledsage: See `StringFunctions.formatterForArray` (might need a new variant for `Iterable<T>`).

core/src/main/java/org/apache/brooklyn/enricher/stock/reducer/StringStringReducer.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package org.apache.brooklyn.enricher.stock.reducer;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+import com.google.common.base.Function;
+
+public class StringStringReducer extends GenericStringReducer<String> {
```
aledsage: Agree with Svet's comment - the `StringStringReducer` is a bit worrying. It makes me think we're only handling a sub-set of the possible use-cases, and opening ourselves to having many variants longer term. Is there anything smart we can do with generics instead, and/or coercion?

core/src/main/java/org/apache/brooklyn/util/core/sensor/SensorPredicates.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.google.common.base.Predicate;
+
+public class SensorPredicates {
+
+    private SensorPredicates() {
+        // not instantiable
+    }
+    
+    public static Predicate<Sensor<?>> sensorNameEqualTo(String sensorName) {
```
aledsage: I'd personally go for `nameEqualTo` (it's implicit that it's a sensor, based on the class name). That would be more in-keeping with naming in our other `*Predicates` classes.

core/src/main/java/org/apache/brooklyn/util/core/sensor/SensorPredicates.java
```diff
@@ -0,0 +1,49 @@
...
+        return new SensorNameEquals(sensorName);
```
aledsage: I'd do `checkNotNull(sensorName, "sensorName)`, rather than risk it throwing an NPE in apply (at which point the stacktrace will tell us nothing about where the null was passed in).

(If it were ever possible to have a null sensor name, then I'd let us pass in null to this method).

core/src/main/java/org/apache/brooklyn/util/core/sensor/SensorPredicates.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    private static class SensorNameEquals implements Predicate<Sensor<?>> {
+        
+        private String sensor;
```
aledsage: use `final` wherever possible.

core/src/test/java/org/apache/brooklyn/enricher/stock/reducer/ReducerTest.java
```diff
@@ -0,0 +1,255 @@
...
+                    return null;
+                } else {
+                    result.append(value);
+                }
+            }
+            return result.toString();
+        }
+    }
+    
+    public static class StringIntegerReducer extends Reducer<String, Integer> {
```
aledsage: It's worrying that you're defining this in the test - that presumably means it's impossible to do it in yaml, because this class doesn't exist and there is no other way to do it other than writing the code yourself?

ahgittin:
@robertgmoss looks really useful.  as soon as you can respond to the copious but good comments from @aledsage and @neykov let's merge!

robertgmoss:
@aledsage Comments addressed

core/src/test/java/org/apache/brooklyn/enricher/stock/reducer/ReducerTest.java
```diff
@@ -0,0 +1,240 @@
...
+            entity.enrichers().add(Enrichers.builder()
+                .reducing(Reducer.class, ImmutableList.<AttributeSensor<?>>of(STR1, STR2))
+                .from(entity)
+                .computing("unknown function name", ImmutableMap.<String, Object>of("separator", "-"))
+                .publishing(STR3)
+                .build()
+            );
+            Asserts.fail("Expected exception when adding reducing enricher with unknown named function");
+        } catch (Exception e) {
+            Throwable t = Exceptions.getFirstThrowableOfType(e, IllegalStateException.class);
```
aledsage: I much prefer:

    if (Exceptions.getFirstThrowableOfType(e, IllegalStateException.class) == null) {
        throw e;
    }

Otherwise we lose the actual exception. This is normally ok, but it's a real pain if a test fails non-deterministically in the asfbot with no information about what went wrong.

utils/common/src/main/java/org/apache/brooklyn/util/text/StringFunctions.java
```diff
@@ -67,6 +70,15 @@ public String apply(@Nullable Object[] input) {
             }
         };
     }
+    
+    /** given e.g. "hello %s %s" returns a function which will insert an Iterable of two strings into that pattern */
+     public static Function<Iterable<?>, String> formatterForIterable(final String pattern) {
+        return new Function<Iterable<?>, String>() {
+            public String apply(@Nullable Iterable<?> input) {
+                return String.format(pattern, Iterables.toArray(input, Object.class));
```
aledsage: This will NPE if `input` is null, yet the parameter is marked `@Nullable`. Should we remove the annotation, or explicitly check for null (e.g. to return null if input is null)?

utils/common/src/main/java/org/apache/brooklyn/util/text/StringFunctions.java
```diff
@@ -154,4 +166,8 @@ public String apply(String input) {
         };
     }
 
+
```
aledsage: Pedantic: remove blank lines.

aledsage:
Thanks @robertgmoss - looks really good. Only some extremely minor comments. Happy to either merge now or you can take a look at those and then I'll merge.

aledsage:
Excellent, thanks @robertgmoss - merging.

PR #920, opened 2015-09-23
Pass EntitlementContext to effectors

bostko:


bostko:
@ygy can you review/test this

usage/rest-server/src/main/java/org/apache/brooklyn/rest/util/BrooklynRestResourceUtils.java
```diff
@@ -444,9 +447,13 @@ public void run() {
         if (mgmt.getEntitlementManager().isEntitled(Entitlements.getEntitlementContext(),
                 Entitlements.INVOKE_EFFECTOR, Entitlements.EntityAndItem.of(entity, 
                     StringAndArgument.of("expunge", MutableMap.of("release", release))))) {
+            Map<String, Object> flags = MutableMap.<String, Object>of("displayName", "expunging " + entity, "description", "REST call to expunge entity "
+                    + entity.getDisplayName() + " (" + entity + ")");
+            if (Entitlements.getEntitlementContext() != null) {
+                flags.put("tags", MutableSet.of(BrooklynTaskTags.tagForEntitlement(Entitlements.getEntitlementContext())));
```
aledsage: It feels like we shouldn't have to treat `expunge` differently from the other REST api operations. 

It looks like the reason we aren't getting the entitlementContext set automatically is that it's using `getExecutionManager().submit(...)`. For normal effectors, it goes through the `BasicExecutionContext.submitInternal` which adds the `EntitlementContext` tag.

I see next to this method `destroy(Application)` which would have the same problem (it is called by the rest `ApplicationApi.delete` operation).

Therefore it feels like a more holistic fix would be good.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -618,7 +620,16 @@ public void restart(ConfigBag parameters) {
         if (isRestartMachine==RestartMachineMode.FALSE) {
             DynamicTasks.queue("stopping (process)", new StopProcessesAtMachineTask());
         } else {
-            DynamicTasks.queue("stopping (machine)", new StopMachineTask());
+            Map<String, Object> stopMachineFlags =  MutableMap.of();
+            if (Entitlements.getEntitlementContext() != null) {
+                stopMachineFlags.put("tags", MutableSet.of(BrooklynTaskTags.tagForEntitlement(Entitlements.getEntitlementContext())));
```
aledsage: Again this feels like a fix for just one situation. Presumably there are lots of places that have an effector/task that queues sub-tasks. Those other places may not inherit the EntitlementContext, if we only fix it here.

But saying that, if the DynamicTasks.queue ends up submitting it via `BasicExecutionContext.submitInternal` then it will happen automatically (I think).

aledsage:
Thanks @bostko 

It would be good to have a test that the EntitlementContext has been passed through correctly (otherwise we risk a regression where we break this again in the future).

I worry about what the general solution should be - so that we are not just fixing it in one place. But we can think about that and tackle it in a separate PR.

Merging this now, and I'll take a look at how we should test it.

PR #921, opened 2015-09-25
winrm docs: more tips/limitations

aledsage:


aledsage:
@bostko thanks for reviewing (via separate chat). I've incorporated your comments. Merging now.

PR #923, opened 2015-09-28
Improve LICENSE notices

ahgittin:
for BSD2 code embedded in `r.js` and others not explicitly attributed

full `mvn clean install` is fine and have tested GUI changes incl sensors


hzbarcea:
Looks good. Tested. Merging.

PR #924, opened 2015-09-28
Added basic migration capability to org.apache.brooklyn.entity.webapp.*

adriannieto:
This PR adds an Effector migrate(locationSpec) to Entities that override JavaWebAppSoftwareProcess. This Effector moves the Entity to a new location. Currently it doesn't support nested entities to be migrated. 

This is a work in progress feature, my plan is to extend this support to non-clustered databases in the future.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/initializer/AddMigrateEffectorEntityInitializer.java
+++ b/core/src/main/java/org/apache/brooklyn/core/initializer/AddMigrateEffectorEntityInitializer.java
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.core.initializer;
+
+import com.google.common.annotations.Beta;
+import com.google.common.collect.Lists;
+import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.entity.EntityInitializer;
+import org.apache.brooklyn.api.entity.EntityLocal;
+import org.apache.brooklyn.api.location.Location;
+import org.apache.brooklyn.core.effector.MigrateEffector;
+import org.apache.brooklyn.core.effector.EffectorBody;
+import org.apache.brooklyn.core.entity.Attributes;
+import org.apache.brooklyn.core.entity.Entities;
+import org.apache.brooklyn.core.entity.EntityInternal;
+import org.apache.brooklyn.core.entity.lifecycle.Lifecycle;
+import org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogic;
+import org.apache.brooklyn.core.entity.trait.Startable;
+import org.apache.brooklyn.util.collections.MutableMap;
+import org.apache.brooklyn.util.core.config.ConfigBag;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Beta
+public class AddMigrateEffectorEntityInitializer implements EntityInitializer {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AddMigrateEffectorEntityInitializer.class);
+
+    @Override
+    public void apply(final EntityLocal entity) {
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("Adding Migrate effector to {}", entity);
+        }
+
+        ((EntityInternal) entity).getMutableEntityType().addEffector(MigrateEffector.MIGRATE, new EffectorBody<Void>() {
+            @Override
+            public Void call(ConfigBag parameters) {
+                migrate((EntityInternal) entity, (String) parameters.getStringKey(MigrateEffector.MIGRATE_LOCATION_SPEC));
+                return null;
+            }
+        });
+    }
+
+    private static void migrate(EntityInternal entity, String locationSpec) {
+        if (entity.sensors().get(Attributes.SERVICE_STATE_ACTUAL) != Lifecycle.RUNNING) {
+            // Is it necessary to check if the whole application is healthy?
+            throw new RuntimeException("The entity needs to be healthy before the migration starts");
+        }
+
+        if (entity.getParent() != null && !entity.getParent().equals(entity.getApplication())) {
+            /*
+             * TODO: Allow nested entites to be migrated
+             * If the entity has a parent different to the application root the migration cannot be done right now,
+             * as it could lead into problems to deal with hierarchies like SameServerEntity -> Entity
+             */
+
+            throw new RuntimeException("Nested entities cannot be migrated right now");
+        }
+
+        // Retrieving the location from the catalog.
+        Location newLocation = Entities.getManagementContext(entity).getLocationRegistry().resolve(locationSpec);
+
+        // TODO: Find a better way to check if you're migrating an entity to the exactly same VM. This not always works.
+        for (Location oldLocation : entity.getLocations()) {
+            if (oldLocation.containsLocation(newLocation)) {
+                LOG.warn("You cannot migrate an entity to the same location, the migration process will stop right now");
+                return;
+            }
+        }
+
+        LOG.info("Migration process of " + entity.getId() + " started.");
+
+        // When we have the new location, we free the resources of the current instance
+        entity.invoke(Startable.STOP, MutableMap.<String, Object>of()).blockUntilEnded();
+
+        // Clearing old locations to remove the relationship with the previous instance
+        entity.clearLocations();
+        entity.addLocations(Lists.newArrayList(newLocation));
+
+        // Starting the new instance
+        entity.invoke(Startable.START, MutableMap.<String, Object>of()).blockUntilEnded();
+
+        // Refresh all the dependent entities
+
+        for (Entity applicationChild : entity.getApplication().getChildren()) {
+            // TODO: Find a better way to refresh the application configuration.
+            // TODO: Refresh nested entities or find a way to propagate the restart properly.
+
+            // Restart any entity but the migrated one.
+            if (applicationChild instanceof Startable) {
+                if (entity.equals(applicationChild)) {
+                    // The entity is sensors should rewired automatically on stop() + restart()
+                } else {
+                    // Restart the entity to fetch again all the dependencies (ie. attributeWhenReady ones)
+                    ((Startable) applicationChild).restart();
+                }
+            }
+        }
+
+        LOG.info("Migration process of " + entity.getId() + " finished.");
+    }
+}
--- a/software/webapp/src/test/java/org/apache/brooklyn/entity/webapp/WebAppMigrationLiveTest.java
+++ b/software/webapp/src/test/java/org/apache/brooklyn/entity/webapp/WebAppMigrationLiveTest.java
@@ -1,0 +1,141 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.entity.webapp;
+
+import com.google.common.collect.ImmutableList;
+import org.apache.brooklyn.api.entity.EntitySpec;
+import org.apache.brooklyn.api.location.ProvisioningLocation;
+import org.apache.brooklyn.core.effector.Effectors;
+import org.apache.brooklyn.core.entity.Attributes;
+import org.apache.brooklyn.core.initializer.AddMigrateEffectorEntityInitializer;
+import org.apache.brooklyn.core.internal.BrooklynProperties;
+import org.apache.brooklyn.core.test.BrooklynAppLiveTestSupport;
+import org.apache.brooklyn.core.test.entity.LocalManagementContextForTests;
+import org.apache.brooklyn.core.effector.MigrateEffector;
+import org.apache.brooklyn.entity.webapp.jboss.JBoss6Server;
+import org.apache.brooklyn.entity.webapp.jboss.JBoss7Server;
+import org.apache.brooklyn.entity.webapp.jetty.Jetty6Server;
+import org.apache.brooklyn.entity.webapp.tomcat.Tomcat8Server;
+import org.apache.brooklyn.entity.webapp.tomcat.TomcatServer;
+import org.apache.brooklyn.test.EntityTestUtils;
+import org.apache.brooklyn.test.HttpTestUtils;
+import org.apache.brooklyn.test.support.TestResourceUnavailableException;
+import org.apache.brooklyn.util.collections.MutableMap;
+import org.apache.brooklyn.util.time.Duration;
+import org.reflections.Reflections;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+// This test could extend AbstractEc2LiveTest but then it will not be able to be parametrized properly.
+public class WebAppMigrationLiveTest extends BrooklynAppLiveTestSupport {
+    private static final String PROVIDER = "aws-ec2";
+    private static final String REGION_NAME = "us-west-2";
+    private static final String LOCATION_SPEC = PROVIDER + (REGION_NAME == null ? "" : ":" + REGION_NAME);
+
+    private BrooklynProperties brooklynProperties;
+    private ProvisioningLocation amazonProvisioningLocation;
+
+    @Override
+    @BeforeMethod(alwaysRun = true)
+    public void setUp() throws Exception {
+        brooklynProperties = BrooklynProperties.Factory.newDefault();
+
+        // EC2 setup obtained from AbstractEc2LiveTest
+        brooklynProperties.remove("brooklyn.jclouds." + PROVIDER + ".image-description-regex");
+        brooklynProperties.remove("brooklyn.jclouds." + PROVIDER + ".image-name-regex");
+        brooklynProperties.remove("brooklyn.jclouds." + PROVIDER + ".image-id");
+        brooklynProperties.remove("brooklyn.jclouds." + PROVIDER + ".inboundPorts");
+        brooklynProperties.remove("brooklyn.jclouds." + PROVIDER + ".hardware-id");
+
+        // Also removes scriptHeader (e.g. if doing `. ~/.bashrc` and `. ~/.profile`, then that can cause "stdin: is not a tty")
+        brooklynProperties.remove("brooklyn.ssh.config.scriptHeader");
+        mgmt = new LocalManagementContextForTests(brooklynProperties);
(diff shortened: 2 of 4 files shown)


software/webapp/src/test/java/org/apache/brooklyn/entity/webapp/WebAppMigrationLiveTest.java
```diff
@@ -0,0 +1,143 @@
...
+        mgmt = new LocalManagementContextForTests(brooklynProperties);
+        amazonProvisioningLocation = (ProvisioningLocation) mgmt.getLocationRegistry().resolve(LOCATION_SPEC);
+        super.setUp();
+    }
+
+
+    @Override
+    @AfterMethod(alwaysRun = true)
+    public void tearDown() throws Exception {
+        app.stop();
```
sjcorbett: The call to `Entities.destroyAll` in the superclass method will take care of this for you.

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppSoftwareProcessImpl.java
```diff
@@ -202,4 +208,75 @@ public String getHttpsSslKeystorePassword() {
...
+        refreshDependantEntities();
+
+        LOG.info("Migration process of " + this.getId() + " finished.");
+
+    }
+
+    /*
+     * TODO: Find a better way to refresh the application configuration.
+     */
+    private void refreshDependantEntities() {
```
sjcorbett: `Dependent`

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppSoftwareProcessImpl.java
```diff
@@ -202,4 +208,75 @@ public String getHttpsSslKeystorePassword() {
...
+        // TODO: Refresh nested entities or find a way to propagate the restart properly.
+        for (Entity entity : getApplication().getChildren()) {
+
+            // Restart any entity but the migrated one.
+            if (entity instanceof Startable) {
+                if(this.equals(entity)){
+                    //The entity is sensors should rewired automatically on stop() + restart()
+                }else{
+                    // Restart the entity to fetch again all the dependencies (ie. attributeWhenReady ones)
+                    ((Startable) entity).restart();
```
sjcorbett: We don't want to restart every startable entity in the application. What is the goal of this loop?

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppSoftwareProcessImpl.java
```diff
@@ -202,4 +208,75 @@ public String getHttpsSslKeystorePassword() {
...
+            throw new RuntimeException("Nested entities cannot be migrated right now");
+        }
+
+        // Retrieving the location from the catalog.
+        Location newLocation = getManagementContext().getLocationRegistry().resolve(locationSpec);
+
+        // TODO: Find a better way to check if you're migrating an entity to the exactly same VM. This not always works.
+        for(Location oldLocation : getLocations()){
+            if(oldLocation.containsLocation(newLocation))
+                throw new RuntimeException("You cannot migrate an entity to the same location");
```
sjcorbett: Simpler than throwing an exception would be to log and return immediately.

sjcorbett:
@adriannieto thanks for your contribution.

I've made a few comments on and asked a few questions about the implementation as it is. 

It's interesting to see that nothing in the implementation of `JavaWebAppSoftwareProcessImpl.migrate` is specific to the webapp entity. A cleaner approach would be to create an [entity initializer](https://github.com/apache/incubator-brooklyn/blob/0c3e7ea5381d91263c6e492063fe3f42e4aef8c4/api/src/main/java/org/apache/brooklyn/api/entity/EntityInitializer.java) that adds the `migrate` effector and implementation to an entity dynamically. This way entity writers can choose when and where it is available. In fact you might not even need to write the initialiser - `org.apache.brooklyn.core.effector.AddEffector` looks like a useful starting point, though you may need to extend it to allow the effector body to be specified.

You can refer to [vanilla-bash-netcat-w-client.yaml](https://github.com/apache/incubator-brooklyn/blob/0c3e7ea5381d91263c6e492063fe3f42e4aef8c4/docs/guide/yaml/example_yaml/vanilla-bash-netcat-w-client.yaml#L29-L40) for an example of how initialisers are used from Yaml.

software/webapp/src/test/java/org/apache/brooklyn/entity/webapp/WebAppMigrationLiveTest.java
adriannieto: Thank you, I will remove it then.

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppSoftwareProcessImpl.java
adriannieto: Lets figure that there are some dependencies  (for example, some attributeWhenReady value) between the migrated entity and another one. 

As Brooklyn doesn't have a dependency graph in the internal model, I have to just stop and start every Entity to refresh the sensors.

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppSoftwareProcessImpl.java
adriannieto: Won't it mean than the migration ended successfully when it doesn't? Anyway if this is the normal Brooklyn behaviour I have no problem with this.

adriannieto:
@sjcorbett  thank you very much for the review, I was reading further about how to initialise the effectors and I will modify everything according to your suggestions.

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppSoftwareProcessImpl.java
sjcorbett: Arguably a migration to the exact same location is always successful without doing anything. I suppose it depends on the semantics of the effector.

adriannieto:
@sjcorbett Hello, Jenkins  is complaining because it can't merge automatically, but this branch builds on my machine. Should I rebase to the master before continuing the review? I wasn't sure that a push force to my branch were a good idea.

sjcorbett:
Yes please, rebase on master. Force pushing your branch is absolutely fine.

sjcorbett:
@adriannieto My apologies for leaving this for an entire month. I've had a quick scan of the updated changes and things look good. I have a few more comments to make that I'll leave shortly.

Your branch conflicts with master but it only involves a couple of imports in `JavaWebAppSshDriver` so will be easy to resolve.

adriannieto:
@sjcorbett thank you, I will rebase it ASAP. Just one question, are the package locations fine? For example I added AllowsMigration definition to base package, do you want any modification on this?

software/webapp/src/test/java/org/apache/brooklyn/entity/webapp/WebAppMigrationLiveTest.java
```diff
@@ -0,0 +1,136 @@
...
+        mgmt = new LocalManagementContextForTests(brooklynProperties);
+        amazonProvisioningLocation = (ProvisioningLocation) mgmt.getLocationRegistry().resolve(LOCATION_SPEC);
+        super.setUp();
+    }
+
+    @DataProvider(name = "webAppEntities")
+    public Object[][] webAppEntities() {
+        Object[][] dataProviderResult;
+
+        Reflections reflections = new Reflections("org.apache.brooklyn.entity.webapp");
```
sjcorbett: This and `getEntitiesFromClasses` is quite cute but it would be much (much!) clearer to just return a reference to the classes explicitly. i.e. `new Object[][]{TomcatServer.class}, {Jboss7Server.class}, ...}`

sjcorbett: My mistake, should return an array of the specs. But I think you get my point.

software/webapp/src/test/java/org/apache/brooklyn/entity/webapp/WebAppMigrationLiveTest.java
```diff
@@ -0,0 +1,136 @@
...
+        // Run the webserver in the premigration location
+        app.start(ImmutableList.of(amazonProvisioningLocation));
+        EntityTestUtils.assertAttributeEqualsEventually(server, Attributes.SERVICE_UP, Boolean.TRUE);
+
+        // Check if the WebServer is reachable in the premigration location
+        String preMigrationUrl = server.getAttribute(JavaWebAppSoftwareProcess.ROOT_URL);
+        HttpTestUtils.assertUrlReachable(preMigrationUrl);
+
+        // Start the migration process
+        Effectors.invocation(AllowsMigration.MIGRATE, MutableMap.of("locationSpec", LOCATION_SPEC), server).asTask().blockUntilEnded();
```
sjcorbett: Give a `Duration` timeout argument to `blockUntilEnded` just in case something goes wrong.

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppMigrateEffector.java
```diff
@@ -0,0 +1,115 @@
...
+            @Override
+            public Void call(ConfigBag parameters) {
+                migrate((EntityInternal) entity, (String) parameters.getStringKey(AllowsMigration.MIGRATE_LOCATION_SPEC));
+                return null;
+            }
+        });
+    }
+
+    private static void migrate(EntityInternal entity, String locationSpec) {
+        if (ServiceStateLogic.getExpectedState(entity) != Lifecycle.RUNNING) {
```
sjcorbett: Should use `entity.sensors().get(Attributes.SERVICE_STATE_ACTUAL)` instead.

sjcorbett: (rather than `ServiceStateLogic.getExpectedState(entity)`)

software/webapp/src/test/java/org/apache/brooklyn/entity/webapp/WebAppMigrationLiveTest.java
adriannieto: Ok, I was considering this just to be "future-proof" just in case you add a new webapp entity. But as you're moving to YAML based entities maybe this test will not be necessary anymore.

sjcorbett: Yeah, I appreciate the sentiment. In truth things change rarely enough that it is fine to list items.

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppMigrateEffector.java
```diff
@@ -0,0 +1,115 @@
...
+import org.apache.brooklyn.core.entity.lifecycle.Lifecycle;
+import org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogic;
+import org.apache.brooklyn.core.entity.trait.Startable;
+import org.apache.brooklyn.entity.software.base.AllowsMigration;
+import org.apache.brooklyn.util.collections.MutableMap;
+import org.apache.brooklyn.util.core.config.ConfigBag;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class JavaWebAppMigrateEffector implements EntityInitializer {
```
sjcorbett: Nothing about this class is particular to Java web apps. How about calling it `MigrateEntityEffector` and moving it to `org.apache.brooklyn.core.initializer` (in the core module, you'll have to create it).

sjcorbett: Also suggest annotating this `@Beta`.

sjcorbett: Actually, call it `AddMigrateEffectorEntityInitializer`.

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppMigrateEffector.java
```diff
@@ -0,0 +1,115 @@
...
+            // TODO: Find a better way to refresh the application configuration.
+            // TODO: Refresh nested entities or find a way to propagate the restart properly.
+
+            // Restart any entity but the migrated one.
+            if (applicationChild instanceof Startable) {
+                if (entity.equals(applicationChild)) {
+                    // The entity is sensors should rewired automatically on stop() + restart()
+                } else {
+                    // Restart the entity to fetch again all the dependencies (ie. attributeWhenReady ones)
+                    ((Startable) applicationChild).restart();
```
sjcorbett: The onus on refreshing dependencies is not on this effector but on the other entities. Restarting everything else is not acceptable. We have to rely on other entities subscribing to events on the appropriate sensors and taking the appropriate action. For example, an Nginx entity would have a subscription to a TomcatServer's hostname and port. These would automatically be republished when `entity` was restarted.

sjcorbett: @aledsage Do you have any thoughts on this interesting effector?

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppMigrateEffector.java
adriannieto: Nice idea, I also moved org.apache.brooklyn.entity.software.base.AllowsMigration to org.apache.brooklyn.core.effector.MigrateEffector. WDYT?

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppMigrateEffector.java
adriannieto: @sjcorbett as we don't have a way to define relationships in Brooklyn, how we are going to know which entities depend between them?

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppMigrateEffector.java
adriannieto: Done

software/webapp/src/test/java/org/apache/brooklyn/entity/webapp/WebAppMigrationLiveTest.java
adriannieto: Done.

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppMigrateEffector.java
sjcorbett: We're not going to know which entities depend on each other. As I tried to describe in my first comment we must configure the other entities to behave sensibly when one is migrated.

The guiding principle to follow is that entities should act on information they receive (i.e. sensor events), not be unilaterally acted upon as is being done here.

Incidentally, support for more general relationships is making its way into Brooklyn. See @ahgittin's recent work in #987 for the foundations.

core/src/main/java/org/apache/brooklyn/core/initializer/AddMigrateEffectorEntityInitializer.java
```diff
@@ -0,0 +1,118 @@
...
+            // TODO: Find a better way to refresh the application configuration.
+            // TODO: Refresh nested entities or find a way to propagate the restart properly.
+
+            // Restart any entity but the migrated one.
+            if (applicationChild instanceof Startable) {
+                if (entity.equals(applicationChild)) {
+                    // The entity is sensors should rewired automatically on stop() + restart()
+                } else {
+                    // Restart the entity to fetch again all the dependencies (ie. attributeWhenReady ones)
+                    ((Startable) applicationChild).restart();
```
sjcorbett: Note to self/other reviewers since GitHub has decoupled conversation from code - see https://github.com/apache/incubator-brooklyn/pull/924#discussion_r45379381 for discussion on this loop before merging.

software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/JavaWebAppMigrateEffector.java
adriannieto: Thank you for your explanation, regarding to the "notification" thing, I know that we already have the emit() method but probably is not ready yet to emit that kind of events.

What do you suggest? Maybe extending this system + relationships on the event will be enough to recognize which entities should be refreshed.

sjcorbett: I suggest simply deleting this whole loop. When the effector restarts the target entity that entity's sensors are going to be republished. I think that should be sufficient information for other entities that need to be updated. For example..

* migrated entity = Tomcat
* entity that needs to be updated = Nginx
* When first creating the application the Nginx entity is configured to subscribe to changes to the Tomcat server's hostname and port. When it is notified of these changes it refreshes its configuration.
* When the Tomcat entity is migrated it is restarted. Its hostname and port change _automatically_. 
* The Nginx entity is notified of the new values because of its subscription. It reloads its configuration and uses the new hostname and port for Tomcat.

Does this make sense? The migrate effector needs only to act on the target entity.

adriannieto: It totally makes sense for me, the problem is I'm not sure if when the sensors are republished the entity automatically refreshes himself. I think not. This is why I'm restarting the entities. 

Is there any way to process this "sensor republished" event?

sjcorbett: You're right that entities do not refresh themselves when sensors are republished. I didn't phrase it very well previously but I was trying to say that it is the job of the user of MigrateEffector to configure this. It's definitely possible if you create your application with Java. e.g. - 

```
tomcat = EntitySpec.create(TomcatServer.class)
nginx  = EntitySpec.create(NginxServer.class)
nginx.subsribe(tomcat, sensor, sensorEventListenerThatUpdatesNginx'sConfig)
```
I'm not sure if it's so straightforward from YAML. I'll take a look too.

sjcorbett: Here's a really simple initialiser that does the basic job. It should do more error checking around the state of the load balancer and the entities whose hostnames have changed. YAML blueprints would include it on whatever needed to respond to a migrate effector:

```
import org.apache.brooklyn.api.entity.EntityInitializer;
import org.apache.brooklyn.api.entity.EntityLocal;
import org.apache.brooklyn.api.sensor.SensorEvent;
import org.apache.brooklyn.api.sensor.SensorEventListener;
import org.apache.brooklyn.core.entity.Attributes;
import org.apache.brooklyn.core.entity.lifecycle.Lifecycle;
import org.apache.brooklyn.entity.proxy.LoadBalancer;

public class LoadBalancerReconfigurer implements EntityInitializer {

    @Override
    public void apply(EntityLocal entity) {
        if (entity instanceof LoadBalancer) {
            LoadBalancer lb = (LoadBalancer) entity;
            entity.subscribeToMembers(
                    entity.config().get(LoadBalancer.SERVER_POOL),
                    Attributes.HOSTNAME,
                    new RefreshConfigSensorEventListener(lb));
        } else {
            // log message that this initialiser only works with load balancers.
        }
    }

    private static class RefreshConfigSensorEventListener implements SensorEventListener<String> {
        final LoadBalancer lb;

        private RefreshConfigSensorEventListener(LoadBalancer lb) {
            this.lb = lb;
        }

        @Override
        public void onEvent(SensorEvent<String> event) {
            Lifecycle lbState = lb.sensors().get(Attributes.SERVICE_STATE_ACTUAL);
            if (Lifecycle.RUNNING.equals(lbState)) {
                lb.update();
            } else {
                // log not updating when in state
            }
        }
    }
}
```

adriannieto: Nice, let me take a look to this implementation and I will answer you ASAP.

mogul:
Migration in Brooklyn has been a great mystery to me, and this PR is the only relevant one I saw mentioned in the mailing list when I went searching. I'm excited to see it as it answers some questions for me, but I'm having trouble following what happened/should have happened with PRs like this as of Brooklyn's graduation from incubation... Did this PR and discussion continue elsewhere, or did it just happen to land at a natural stopping point just prior to the graduation?

bostko:
Hi @mogul Glad to hear that you are excited about  Apache Brooklyn!
Indeed the project is moved to another repo because the project graduated from Incubator. The github repo is now located at http://github.com/apache/brooklyn . Please check http://brooklyn.apache.org/community/mailing-lists.html for mailing lists information.
@adriannieto can you please open Pull Request with the changes against [brooklyn-server](https://github.com/apache/brooklyn-server) and [brooklyn-library](https://github.com/apache/brooklyn-library).

adriannieto:
Yes, no problem but I'm not sure about if I will be able to work on this
topic right now.

However you could use it as an starting point.

sjcorbett:
@mogul This PR reached a natural stopping point. I don't think it would take much more work to get it merged. We'd be happy to answer any questions you have on the mailing list.

PR #925, opened 2015-09-28
MS SQL blueprint - improve installation script

bostko:
pass the exit code from the install command to the install script

@nakomis could you review that please.


nakomis:
Looks like a sensible change. Does the `MssqlBlueprintLiveTest` still work, and still correctly install SQL?

bostko:
Yes it works.
I am testing now other mssql scripts whether they return exit code from the commands.

aledsage:
Thanks @bostko (and thanks @nakomis for reviewing). Merging now.

PR #927, opened 2015-09-29
Add VanillaWindowsProcess to catalog items

bostko:


software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcess.java
```diff
@@ -29,6 +30,7 @@
 
 import com.google.common.collect.ImmutableSet;
 
+@Catalog(name="Vanilla Windows Process", description="A basic windows entity configured with scripts, e.g. for launch, check-running and stop")
```
rdowner: A really trivial request but could you make it `Windows` instead of `windows` in the description? As it's in a prominent location - and a trademarked description - we should be accurate. Thanks!

neykov:
Also would be nice to add `@CatalogConfig` to the required config keys so they are displayed in the UI.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcess.java
```diff
@@ -49,14 +52,34 @@
             "indicates that a reboot should be performed after the install command is run", false);
     ConfigKey<Boolean> CUSTOMIZE_REBOOT_REQUIRED = ConfigKeys.newBooleanConfigKey("customize.reboot.required",
```
aledsage: I'd include `priority=...` on each, to force a sensible order in their listing within the wizard. Not sure what guarantees we get here (probably it's just in the order returned by http://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#getFields(), which says "not sorted and are not in any particular order").

I personally would have also added `@CatalogConfig` to the other powershell commands such as `PRE_INSTALL_POWERSHELL_COMMAND` etc. It's easy enough for someone to scroll down, filling in only what they want. But if they are not annotated then that config option is very hard to use from the wizard.

aledsage:
Thanks @bostko - merging this now. I'll take a look at the priority and other `@CatalogConfig` in a separate PR.

PR #928, opened 2015-09-29
Hello world page- revised

morganbrooke:
screen shot 2015-09-29 at 1 16 34 pm
Changed the Hello, World sample application landing page.


usage/dist/src/main/dist/pid_java
```diff
@@ -0,0 +1 @@
+83209
```
ahgittin: this file should be removed.  i think it's the one causing the RAT error.

morganbrooke: Great, will remove then update.

Thank you!

On 1 October 2015 at 06:16, Alex Heneveld <notifications@github.com> wrote:

> In usage/dist/src/main/dist/pid_java
> <https://github.com/apache/incubator-brooklyn/pull/928#discussion_r40911191>
> :
>
> > @@ -0,0 +1 @@
> > +83209
>
> this file should be removed. i think it's the one causing the RAT error.
>
> —
> <https://github.com/apache/incubator-brooklyn/pull/928/files#r40911191>.
>

-- 
 
message after it was sent.

ahgittin:
probably also worth renaming the images to not have spaces, those still sometimes cause problems.  i suggest `bridge-large.png` and `bridge-small.png` or similar (`bridge-400px.png` is even better if you know the width.)

i also suggest squashing all the commits:

    git rebase master
    git reset master
    # remove/rename files and update references
    git commit -A
    git push -f

docs/guide/yaml/winrm/index.md
```diff
@@ -355,6 +355,18 @@ refer to [How and Why to re-authenticate withing a powershell script](#how-and-w
 If using the imageId of a Windows community AMI, you may find that the AMI is deleted after a few weeks.
 See [Windows AMIs on AWS](#windows-amis-on-aws) above.
 
+### VM Provisioning Times Out
```
aledsage: I'm not sure what has happened with the rebase. These changes are already in master, rather than edits in your PR. Can you try doing a `git rebase master` again please?

If they are left in, it will probably do no harm (but we should double-check after merge that it doesn't now have the text twice or anything like that)..

aledsage:
Looks good merging (I will presume that git does the right thing for the mods in winrm docs shown in the diff view!)

I did find it a little difficult to read the text of "See the available illustrations here." against the green bridge background. But not too difficult.

PR #929, opened 2015-10-01
Better group member display

alasdairhodge:
* Display group members alongside children.
* Each entity still has one "direct" position in the hierarchy, but may also have several "indirect" references.
* Indirect nodes can be interacted with in-place as normal:
  * show details, invoke effectors, expand children, etc.
* Indirect subordinates have a distinct visual style to avoid confusion.

@grkvlt: Would love to know if this plays well with Clocker, which I know makes heavy use of groups.

screen shot 2015-10-01 at 15 31 41

### DIFF
--- a/usage/jsgui/src/main/webapp/assets/css/base.css
+++ b/usage/jsgui/src/main/webapp/assets/css/base.css
@@ -580,6 +580,15 @@ ol.tree {
 .entity_tree_node_wrapper.active .entity_tree_node {
     font-weight: bold;
 }
+.entity_tree_node_wrapper .indirection-icon {
+    opacity: 0.7;
+    margin-left: -5px;
+}
+.tree-box.indirect > .entity_tree_node_wrapper a {
+    font-style: italic !important;
+    color: #666;
+}
+
 #tree label {
 /* remove the folder, and align with + - icons */
 background: none;
--- a/usage/jsgui/src/main/webapp/assets/js/model/app-tree.js
+++ b/usage/jsgui/src/main/webapp/assets/js/model/app-tree.js
@@ -20,27 +20,31 @@ define([
     "underscore", "backbone"
 ], function (_, Backbone) {
 
-    var AppTree = {}
+    var AppTree = {};
 
     AppTree.Model = Backbone.Model.extend({
-        defaults:function () {
+        defaults: function() {
             return {
-                id:"",
-                name:"",
-                type:"",
-                iconUrl:"",
-                serviceUp:"",
-                serviceState:"",
-                applicationId:"",
-                parentId:"",
-                children:[]
-            }
+                id: "",
+                name: "",
+                type: "",
+                iconUrl: "",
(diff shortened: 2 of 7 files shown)


aledsage:
Thanks @alasdairhodge 

I like the curved arrow icon and greyer text - a nice way to represent that they are children.

My concern is what happens when someone expands one of these indirect nodes to see its children and members. I presume everything below it keeps that same look (the icon + grey text). So how does someone tell the difference between a child and a member of an indirect node?

Throwing around some other ideas... we could have sub-headings for "children" and "members". That would take up a lot of space unfortunately, but would also be extensible for other relationships (e.g. showing the locations of an entity, or other dependencies).

@morganbrooke any comments on how best to represent this?

ahgittin:
feels great for now.

longer term i'd like to have a graphical view (not a tree view) which allows for multiple types of relationships, including "managed child" (which is what we'd inject the current hierarchy into), rather than try to make this tree all-singing-all-dancing.

alasdairhodge:
Thanks for reviewing, @aledsage and @ahgittin. I deliberately only distinguished between "direct" and "indirect" subordinates, anticipating several "indirect" relationship types in the future. Will keep an eye out for cases where this new approach is confusing and refine accordingly.

alasdairhodge:
We should now (arguably) remove `DelegateEntity` and `AbstractGroup.MEMBER_DELEGATE_*` (following a deprecation cycle as none are `@Beta`). Might also be a good time to remove the ill-conceived `FIRST_MEMBER` from `AbstractGroup`.

ahgittin:
+1

On 08/10/2015 14:17, Alasdair Hodge wrote:
>
> We should now (arguably) remove |DelegateEntity| and 
> |AbstractGroup.MEMBER_DELEGATE_*| (following a deprecation cycle as 
> none are |@Beta|). Might also be a good time to remove the 
> ill-conceived |FIRST_MEMBER| from |AbstractGroup|.
>
> —
> <https://github.com/apache/incubator-brooklyn/pull/929#issuecomment-146541901>.
>


-- 
 
message after it was sent.

morganbrooke:
SO I think you're saying that you want that 2nd parent element in BasicStartables to stand out more?

You put a 1px dark gray border around each group but I think the background color is a pretty good indication of what belongs to what.

morganbrooke:
+1

alasdairhodge:
Paraphrasing: for regular nodes, members ("indirect" subordinates) are visually distinguishable from children ("direct" subordinates) due to the indirection arrow. However, for indirect nodes, children and members are indistinguishable. One could argue that indirect subs of an indirect node are "doubly-indirect" and style accordingly, but that could get ridiculous.

Alternatively, we could add *another* indicator to distinguish "child" vs "member" (vs "other relationship" in the future?), somehow in combination with the arrow, but I'd be very wary about overcooking it. Happy to review any suggestions.

PR #931, opened 2015-10-02
Fix VanillaSoftwareProcess install

aledsage:
Previously, if download and install-cmd were supplied, only the 
download would be executed.

Previously if two VanillaSoftwareProcesses were used on the same
machine where neither had a downloadUrl, it would have skipped the
install of the second (sharing the same install folder).

Previously there was poor test coverage of VanillaSoftwareProcess.

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaSoftwareProcessSshDriver.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaSoftwareProcessSshDriver.java
@@ -21,6 +21,7 @@ package org.apache.brooklyn.entity.software.base;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import org.apache.brooklyn.api.entity.EntityLocal;
 import org.apache.brooklyn.api.entity.drivers.downloads.DownloadResolver;
@@ -36,6 +37,8 @@ import org.apache.brooklyn.util.ssh.BashCommands;
 import org.apache.brooklyn.util.text.Identifiers;
 import org.apache.brooklyn.util.text.Strings;
 
+import com.google.common.collect.ImmutableMap;
+
 public class VanillaSoftwareProcessSshDriver extends AbstractSoftwareProcessSshDriver implements VanillaSoftwareProcessDriver {
 
     public VanillaSoftwareProcessSshDriver(EntityLocal entity, SshMachineLocation machine) {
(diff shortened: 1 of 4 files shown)


software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaSoftwareProcessSshDriver.java
```diff
@@ -44,13 +47,20 @@ public VanillaSoftwareProcessSshDriver(EntityLocal entity, SshMachineLocation ma
 
     String downloadedFilename = null;
 
-    /** needed because the download url might be different! */
+    /**
+     * Needed because the download url and install commands are likely different for different VanillaSoftwareProcesses!
+     * This is particularly true for YAML entities. We take a hash of the download_url, instal_command and environment variables.
```
neykov: Typo - missing l in `instal_command`

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaSoftwareProcessSshDriver.java
```diff
@@ -44,13 +47,20 @@ public VanillaSoftwareProcessSshDriver(EntityLocal entity, SshMachineLocation ma
...
+     */
     @Override
     protected String getInstallLabelExtraSalt() {
+        
         Maybe<Object> url = getEntity().getConfigRaw(SoftwareProcess.DOWNLOAD_URL, true);
-        if (url.isAbsent()) return null;
+        String installCommand = getEntity().getConfig(VanillaSoftwareProcess.INSTALL_COMMAND);
+        Map<String, Object> shellEnv = getEntity().getConfig(VanillaSoftwareProcess.SHELL_ENVIRONMENT);
+        
         // TODO a user-friendly hash would be nice, but tricky since we don't want it to be too long or contain path chars
```
neykov: Can remove this comment now, no longer applicable with the new hash content?

neykov:
Minor comments only, good to merge.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaSoftwareProcessSshDriver.java
aledsage: The hash is just a 4 (?) character hash generated from these fields, so does not look user-friendly when someone lists the contents of the `apps` directory. I'll leave this comment in for now.

aledsage:
Thanks @neykov - merging now.

PR #932, opened 2015-10-02
Fix BrooklynNode isRunning check

neykov:
`isRunning` wasn't set for BrooklynNode due to the ssh check being skipped in favour of http check (see https://github.com/apache/incubator-brooklyn/pull/790)

Infer `isRunning` from the `brooklynnode.webconsole.up` value.


sjcorbett:
Looks good.

PR #933, opened 2015-10-05
Riak Persistence Documentation

iyovcheva:


ahgittin:
i don't think apache brooklyn wants the responsibility for describing how to configure S3-compatible products -- there are lots of them.  this should maybe be something maintained elsewhere if useful to you.

suggest closing this?

PR #934, opened 2015-10-05
MySQL Cluster blueprints

ygy:


software/database/src/main/resources/org/apache/brooklyn/entity/database/mysql/mysql-cluster.yaml
```diff
@@ -0,0 +1,24 @@
+#
```
aledsage: I'm not certain where we should put these blueprint files. This is in `incubator-brooklyn/software/database/src/main/resources/org/apache/brooklyn/entity/database/mysql/mysql-cluster.yaml`, which sits alongside the configuration files for mysql.

Quite a few blueprints sit in `/Users/aled/repos/apache/incubator-brooklyn/usage/launcher/src/test/resources`. That doesn't seem like a good place either, as it doesn't fit well with populating the community catalog.

@ahgittin @tbouron any preference? What will work well with the community catalog? Or do we figure that out later, when we extract all this stuff into its own repos?!

neykov: Maybe https://github.com/brooklyncentral/blueprint-library/?

tbouron:
If the purpose is to show how to use a MySQL cluster then I would put that on the blueprint-library repository that @neykov was talking about, until the point that we create its own repo for the community catalog.

ygy:
@aledsage @neykov @tbouron Thanks for the feedback

I have created a new PR in the suggested repo - https://github.com/brooklyncentral/blueprint-library/pull/6

PR #935, opened 2015-10-05
Use entity.sensors().* etc, instead of deprecated methods

aledsage:


core/src/test/java/org/apache/brooklyn/core/entity/PolicyRegistrationTest.java
```diff
@@ -79,7 +80,7 @@ public void setUp() throws Exception {
     
     @Test
     public void testGetPoliciesIsInitiallyEmpty() {
-        assertEquals(entity.getPolicies(), ImmutableList.of());
+        assertEquals(ImmutableList.copyOf(entity.policies()), ImmutableList.of());
```
alasdairhodge: Maybe suggest Guava's `Iterables.elementsEqual()` here (and elsewhere) but no strong feelings.

core/src/test/java/org/apache/brooklyn/entity/group/DynamicGroupTest.java
```diff
@@ -476,18 +473,24 @@ public void testDoesNotDeadlockOnUnmanageWhileOtherMemberBeingAdded() throws Exc
...
+                            throw Exceptions.propagate(e);
+                        }
                     }
+                    super.emit(sensor, val);
                 }
-                super.emit(sensor, val);
+            };
+            @Override
+            public BasicSensorSupport sensors() {
+                return interceptedSensors;
```
alasdairhodge: Good catch!

alasdairhodge:
Lovely jubbly: goodbye, compiler warnings! One teeny-tiny suggestion, but good to merge as it.

aledsage:
Thanks; merging now and will address comments in separate PR.

PR #939, opened 2015-10-07
Fix/anonymous inner classes in predicates

aledsage:
Use of anonymous inner classes (or non-static inner classes) makes persistence very brittle.

Changes our `*Predicates` and `*Functions` to use static inner classes (keeping the old classes around for backwards compatibility), and adds more tests for these.


hzbarcea:
GTM. Tested, merging.

PR #940, opened 2015-10-07
VanillaWindowsProcess CatalogConfig: add priority

aledsage:
Re-orders the config keys into a sensible order, and adds priority
to the CatalogConfig.

Removes the default from LAUNCH_COMMAND.

Addresses the comments in https://github.com/apache/incubator-brooklyn/pull/927


bostko:
Good changes!

aledsage:
Thanks @bostko - failing test is unrelated. I'll look at that for a separate PR (`DslAndRebindYamlTest.testDslAttributeWhenReadyRebind`).

PR #941, opened 2015-10-07
Add sequential ID number to each DynamicCluster member

mikezaccardo:
Each `DynamicCluster` member entity now has a `cluster.member.id` integer config value representing the order in which it was added to the cluster, starting from 0 and increasing sequentially.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
@@ -28,16 +28,17 @@ import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.annotation.Nullable;
 
 import org.apache.brooklyn.api.entity.Entity;
-import org.apache.brooklyn.api.entity.EntityLocal;
 import org.apache.brooklyn.api.entity.EntitySpec;
 import org.apache.brooklyn.api.location.Location;
 import org.apache.brooklyn.api.location.MachineProvisioningLocation;
 import org.apache.brooklyn.api.mgmt.Task;
 import org.apache.brooklyn.api.policy.Policy;
+import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.core.config.render.RendererHints;
 import org.apache.brooklyn.core.effector.Effectors;
 import org.apache.brooklyn.core.entity.Entities;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -775,6 +779,10 @@ public Entity addNode(Location loc, Map<?,?> extraFlags) {
         entity.sensors().set(CLUSTER_MEMBER, true);
         entity.sensors().set(CLUSTER, this);
 
+        int nextClusterMemberId = getConfig(NEXT_CLUSTER_MEMBER_ID);
+        entity.sensors().set(CLUSTER_MEMBER_ID, nextClusterMemberId);
```
aledsage: Is this the wrong way round - should we set config on the new member to indicates its memberId, and have a sensor on the cluster that says what the next id should be? Or alternatively they could both be sensors (to be the same as `CLUSTER` and `CLUSTER_MEMBER` sensors already being set on the member)?

Is there a race if multiple members are being created concurrently? Perhaps we should extract out to a protected method something like `int obtainNextMemberId()`, which would do some synchronization and would thus increment the sensor atomically.

aledsage:
Thanks @mikezaccardo - looks good. See what you think about my comment and let me know when it's ready to merge.

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
neykov: I've implemented this [elsewhere](https://github.com/apache/incubator-brooklyn/blob/master/software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlClusterImpl.java#L83) using an `AttributeSensor<Supplier<Integer>>` which [increments an `AtomicBoolean`](https://github.com/apache/incubator-brooklyn/blob/master/software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlClusterImpl.java#L273-L280).

mikezaccardo:
@aledsage I believe my last commit addresses your comments.

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -775,6 +793,9 @@ public Entity addNode(Location loc, Map<?,?> extraFlags) {
         entity.sensors().set(CLUSTER_MEMBER, true);
         entity.sensors().set(CLUSTER, this);
 
+        int nextClusterMemberId = getAttribute(NEXT_CLUSTER_MEMBER_ID).get();
+        entity.sensors().set(CLUSTER_MEMBER_ID, nextClusterMemberId);
```
neykov: I'd say this is a config value, could even be passed in `createFlags`, to be accessible in the `init` method, but no strong feelings.

sjcorbett: +1 I think `CLUSTER_MEMBER_ID` should be a config key applied in `createNode`.

sjcorbett: Or in `createFlags` as suggested by @neykov.

mikezaccardo:
Thanks @neykov and @sjcorbett, I've addressed your comments.

sjcorbett:
Squash away. And slip a unit test in if you have a moment.

mikezaccardo:
Great, thanks @sjcorbett -- I'll add a unit test and squash today so it should be good to merge on Monday morning.

mikezaccardo:
Unit tests added + squashed.

sjcorbett:
Nice one, thanks.

PR #942, opened 2015-10-07
Fix add-app-wizard for enums (WIP)

aledsage:
Attempting to fix:

screen shot 2015-10-07 at 22 53 55

The enum drop-downs are not supposed to start with ">", and should say "Virtualization" rather than the selected= stuff.

My change reverts a change that sneaked in as part of the commit "[BROOKLYN-162] package rename to org.apache.brooklyn: software/webapp"!

However, when I just ran with it in a downstream project, it seemed not to alter the behaviour at all.

Maybe my javascript was cached or maybe the downstream build actually downloaded the latest snapshot Brooklyn (rather than taking my local build)?!


sjcorbett:
@aledsage Your fix appears to work for me.

screen shot 2015-10-08 at 10 50 47

sjcorbett:
And when I revert your fix the problem returns. Merge away.

PR #943, opened 2015-10-08
Tests jclouds rebind

aledsage:


hzbarcea:
Easy one, GTM. Tested, merging.

PR #944, opened 2015-10-09
Fix osname check when PostgreSQL is installed on RedHat BYON location

ygy:
- RedHat BYON returns osname as "Red Hat Enterprise Linux", instead of
  the expected "rhel"


bostko:
Looks good.

aledsage:
Looks good; merging.

PR #945, opened 2015-10-10
More on karaf feature.xml for brooklyn-core

CMoH:
The current showstopper is brooklyn-core wanting to instantiate apache felix, again using code from the private packages. This time the fix is not as easy as for xstreams.


CMoH:
Please ignore this - was meant for another branch.

PR #947, opened 2015-10-12
WinRM commands

bostko:
Use executePsCommand and executeCmdCommand instead of executeScript

It is a small refactoring which cleans up the too simple executeScript functionality. 

The PR is dependent on https://github.com/cloudsoft/winrm4j/pull/2


aledsage:
Note we'll need a new version of https://github.com/cloudsoft/winrm4j to be released (and pushed to maven central), and to bump the version number to 0.2.0 in `parent/pom.xml`. Therefore don't merge this until that has been done.

bostko:
Closing in favour of #950

PR #948, opened 2015-10-12
Extend brooklynnode.classpath config options

googlielmo:
classpath can be a list of
- URLs or file names or directory names as before,
- maps specifying a URL and a filename,
   e.g. { "url": "http://...", "filename": "myfile.jar" }

This feature is needed to fix incorrect filename deduction in
situations where the URL does not contain an explicit jar name

Finally, the destination directory has been changed from ./lib to
./lib/dropins


neykov:
Note that `BrooklynNode` will still work as expected when uploading files to the `lib` folder because it's [adding it to the classpath] (https://github.com/apache/incubator-brooklyn/blob/master/software/base/src/main/java/org/apache/brooklyn/entity/brooklynnode/BrooklynNodeSshDriver.java#L333). But in general +1 for the cleanup. LGTM.

sjcorbett:
@googlielmo @neykov Thanks. Will merge.

PR #949, opened 2015-10-12
Add support for correct configkey conversion from the app wizard UI

tbouron:
Right now, the app wizard modal pass all additional configurations as `String`. If the entities have configkey with different types, an array of integers for example, Brooklyn will fail right after hitting the `Finish` button because it won't be able to convert `"[]"` as an array.

This will correctly convert string value, entered within the app wizard, into their corresponding Javascript type. For instance:
- `[1,2,3,4]` will be converted as a JS `Array` that contains 1, 2, 3 and 4 integers (`Number` type)
- `{"test": true}` will be converted as a JS `Object` containing the field `test` with the boolean value `true`
- `true` will be converted as a `Boolean`
- `0.123456` will be converted as a `Number`

If the conversion fails, it will fallback to the previous implementation which is taking the value as a `String`.

### DIFF
--- a/usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
+++ b/usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
@@ -85,6 +85,13 @@ define([
         if (entity.config && _.size(entity.config)) result["brooklyn.config"] = entity.config;
         return result;
     }
+    function getConvertedConfigValue(value) {
+        try {
+            return $.parseJSON(value);
+        } catch (e) {
+            return value;
+        }
+    }
     
     var ModalWizard = Backbone.View.extend({
         tagName:'div',
@@ -510,7 +517,7 @@ define([
         getConfigMap:function (root) {
             var map = {}
             $('.app-add-wizard-config-entry',root).each( function (index,elt) {
-                map[$('#key',elt).val()] = $('#value',elt).val()
+                map[$('#key',elt).val()] = getConvertedConfigValue($('#value',elt).val())
             })
             return map;
         },
@@ -777,21 +784,13 @@ define([
             }
         },
         getConfigMap:function() {
-            var that = this;
             var map = {};
             $('.app-add-wizard-config-entry').each( function (index,elt) {
                 map[$('#key',elt).val()] = 
                     $('#checkboxValue',elt).length ? $('#checkboxValue',elt).is(':checked') :
-                    that.getConvertedConfigValue($('#value',elt).val())
+                    getConvertedConfigValue($('#value',elt).val())
             })
             return map;
-        },
-        getConvertedConfigValue:function (value) {
-            try {
-                return $.parseJSON(value);
-            } catch (e) {
-                return value;
-            }
         },
         selectionVersion:function (event) {
             this.model.spec.set("version", $(event.currentTarget).val())
--- a/usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
+++ b/usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
@@ -777,13 +777,21 @@ define([
             }
         },
         getConfigMap:function() {
-            var map = {}
+            var that = this;
+            var map = {};
             $('.app-add-wizard-config-entry').each( function (index,elt) {
                 map[$('#key',elt).val()] = 
                     $('#checkboxValue',elt).length ? $('#checkboxValue',elt).is(':checked') :
-                    $('#value',elt).val()
+                    that.getConvertedConfigValue($('#value',elt).val())
             })
             return map;
+        },
+        getConvertedConfigValue:function (value) {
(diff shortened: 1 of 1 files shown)


usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -777,14 +777,22 @@ define([
...
         },
         getConfigMap:function() {
-            var map = {}
+            var that = this;
+            var map = {};
             $('.app-add-wizard-config-entry').each( function (index,elt) {
                 map[$('#key',elt).val()] = 
                     $('#checkboxValue',elt).length ? $('#checkboxValue',elt).is(':checked') :
-                    $('#value',elt).val()
+                    that.getConvertedConfigValue($('#value',elt).val())
```
sjcorbett: You should make the same change to the other function called `getConfigMap` that's defined in the `ModalWizard.StepCreate` object (currently line 510).

sjcorbett: Suggest hoisting `getConvertedConfigValue` out of this view to be a separate function.

sjcorbett:
I was going to ask you to add a test until I looked at application-add-wizard-spec.js and realised it's a job for another pull request.

tbouron:
I addressed your comments @sjcorbett

usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
```diff
@@ -85,6 +85,13 @@ define([
         if (entity.config && _.size(entity.config)) result["brooklyn.config"] = entity.config;
         return result;
     }
+    function getConvertedConfigValue(value) {
+        try {
+            return $.parseJSON(value);
+        } catch (e) {
+            return value;
```
neykov: Theoretically a user could've declared his config key as String only to be able to use UI provided values, while treating it as an object under the hood. I wasn't able to find such cases though.

To be more robust you could check the key type and convert only if != enum, string.

tbouron: I have examples of that. One of my clients uses `String` configkey only because of this issue.

Does it really matter to check at this stage the config key is != enum, string ? The CAMP Java parser will convert to the correct type afterward. 

For example, I have the following enum:
```
public enum Product {
    A,
    B,
    C;
}
```
and configuration key:
```
ConfigKey<Product> PRODUCT = ConfigKeys.newConfigKey(Product.class, "product", "My product", Product.A);
```
If I put this in the UI `product: B`, the method `getConvertedConfigValue()` will return 'B' as a JS `String`. But the CAMP parser will still try to match that to my enum by calling `Product.valueOf("B");` which will return the valid result.

aledsage:
Nice - thanks @tbouron 

@neykov are you happy with that description of handling enum? If you don't have any more comments then can you merge this please?

usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
neykov: You'll hit a problem if you have ConfigKey<String> ITEMS, but pass an array ([1, 2, 3, 4]). In this case:
  * the client will pass the array as is to the server (not escaping it)
  * the YAML parser will deserialize it to an array
  * when the user does config().get(ITEMS) Brooklyn will try to convert from an array to a string, calling toString on the object which is not clear whether will lead to the original string or not. In the worst case it will be completely invalid value (i.e. object@adfh34ur), in the best it will still be a parseable array, but won't match the original string, leading to unexpected results.

neykov:
@aledsage Still some edge cases to discuss, will merge it afterwards.

usage/jsgui/src/main/webapp/assets/js/view/application-add-wizard.js
tbouron: @neykov The example you gave seems very convoluted. If you want to pass a list from a `String`, chances are you will use a comma-separated list (i.e. `1,2,3,4`) which is simpler to parse than `[1, 2, 3, 4]`, don't you think? This is actually what my client does, he uses comma-separated string to pass a list of options.

Now, I do get you point about the parsing but I think this is a use case that we will never reach because of this PR and its ability to pass a real array.

neykov: I would say specific, not convoluted - the behaviour for String config keys changes (i.e. getting objects instead, stripping quotes, etc.).

I don't mind merging as is as this can easily be amended in a future PR.

tbouron: @neykov My bad, my wording was wrong in is case - you can force `getConvertedValue()` to spit out `String`value by warping the input value with double quote. Is that acceptable enough?

PR #950, opened 2015-10-12
Windows script files

bostko:
The PR includes #947 and it is
Dependent on cloudsoft/winrm4j#2

Introduces support for executing multiline commands from a script. This is needed so we can have more control over how commands are executed and catch properly their exit code.
Just like in ssh it uploads multiline commands into a script file on the machine and it is executing it from there.
Multiline commands into scripts is available only in WinRmMachineLocation. If we add it to AbstractSoftwareProcessWinRmDriver it will allow writing more complicated Windows Entities.

The most important part here is that I wrote Live tests for executing scripts so we can advice users better how to write Windows scripts properly.

### DIFF
--- a/brooklyn-server/software/winrm/src/main/java/org/apache/brooklyn/location/winrm/NaiveWindowsScriptRunner.java
+++ b/brooklyn-server/software/winrm/src/main/java/org/apache/brooklyn/location/winrm/NaiveWindowsScriptRunner.java
@@ -16,14 +16,20 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.brooklyn.entity.software.base.lifecycle;
+package org.apache.brooklyn.location.winrm;
 
+import java.io.InputStream;
 import java.util.Map;
 
-public interface NativeWindowsScriptRunner {
+/** Marks something which can run scripts. Called "Naive" because it hides too much of the complexity,
+ * about script execution and other command-related tasks (put, etc).
+ */
+public interface NaiveWindowsScriptRunner {
 
     /** Runs a command and returns the result code */
     int executeNativeCommand(Map flags, String windowsCommand, String summaryForLogging);
     int executePsCommand(Map flags, String powerShellCommand, String summaryForLogging);
-    Integer executeNativeOrPsCommand(Map flags, String regularCommand, String powershellCommand, String phase, Boolean allowNoOp);
+    Integer executeNativeOrPsCommand(Map flags, String regularCommand, String powershellCommand, String summaryForLogging, Boolean allowNoOp);
+
+    int copyTo(InputStream source, String destination);
 }
--- a/brooklyn-server/software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/WinRmTool.java
+++ b/brooklyn-server/software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/WinRmTool.java
@@ -67,7 +67,26 @@ public interface WinRmTool {
             1024);
 
     /**
-     * @deprecated since 0.9.0; use {@link #executeCommand(List)} to avoid ambiguity between native command and power shell.
+     * Executes a Native Windows command.
+     * It is creating a new Shell on the destination host each time it is being called.
+     * @param command
+     * @since 0.2
+     */
+    WinRmToolResponse executeCommand(String command);
+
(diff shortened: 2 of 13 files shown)


core/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
```diff
@@ -18,49 +18,54 @@
...
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.InetAddress;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
```
sjcorbett: Please configure your IDE to not reorder imports.

ahgittin:
@bostko any update on this?  it looks like we are still waiting on https://github.com/cloudsoft/winrm4j/pull/2 -- it looks like you addressed @aledsage 's remaining comment about `@since` there so maybe @rdowner or @alasdairhodge could take a quick look and merge that, cut a new release of that, then bump the version of that used in brooklyn (in this PR?) and then merge this?

bostko:
@ahgittin Yes it is still waiting for cloudsoft/winrm4j#2
I've upgraded the winrm4j version in this PR as well.

bostko:
Can you review it again? @aledsage

bostko:
The current implementation of the code and `WindowsScriptsWinrmExitStatusLiveTest` the test is failing to instantiate `runner` in other words the driver which is using `WinRmScriptTool`.
Any ideas how to refactor it so I can have a live test for `WinRmScriptTool`?

bostko:
Done.
`VanillaWindowsProcessWinrmStreamsLiveTest`, `VanillaWindowsProcessWinrmStreamsLiveTest`, `WindowsScriptsWinrmExitStatusLiveTest` are passing.
Can you review?

brooklyn-server/software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/WinRmTool.java
```diff
@@ -66,8 +66,38 @@
...
+    WinRmToolResponse executePsCommand(String psCommand);
+
+    /**
+     * Execute a list of Windows Native commands as one command.
+     * The method translates the list of commands to a single String command with a <code>"\r\n"</code> delimiter and a terminating one.
+     * @param commands
+     * @deprecated since 0.2; Use the {@link #executeCommand(String)} instead and transform your commands list explicitly
+     */
+    @Deprecated
     WinRmToolResponse executeScript(List<String> commands);
```
aledsage, line 92: @bostko see @neykov 's comments in https://github.com/cloudsoft/winrm4j/pull/2, saying "we should keep the `(List<String>)`...".

bostko, line 92: @aledsage I kept it. It is here

brooklyn-server/software/winrm/src/main/java/org/apache/brooklyn/location/winrm/NaiveWindowsScriptRunner.java
```diff
@@ -16,14 +16,17 @@
...
  * under the License.
  */
-package org.apache.brooklyn.entity.software.base.lifecycle;
+package org.apache.brooklyn.location.winrm;
 
+import java.io.InputStream;
 import java.util.Map;
 
-public interface NativeWindowsScriptRunner {
+public interface NaiveWindowsScriptRunner {
```
aledsage, line 27: Why rename this to "Naive" instead of "Native"?

aledsage:
https://github.com/apache/incubator-brooklyn/pull/1158 contains some of these same changes (it switches to using the pure-java winrm4j client v0.2.0).

brooklyn-server/software/winrm/src/main/java/org/apache/brooklyn/location/winrm/NaiveWindowsScriptRunner.java
bostko, line 27: I named it similar to `NaiveScriptRunner`.
Docs added.

PR #952, opened 2015-10-13
Fix openiptables to show stdout/stderr

iyovcheva:


software/base/src/test/java/org/apache/brooklyn/entity/software/base/EmptySoftwareProcessStreamsLiveTest.java
```diff
@@ -0,0 +1,113 @@
...
+                .put("osFamily", "centos")
+                .put("osVersionRegex", "6\\..*")
+                .put("inboundPorts", ImmutableList.of(22, 31880, 31001, 8080, 8443, 1099))
+                .build());
+
+        app = ApplicationBuilder.newManagedApp(newAppSpec(), mgmt);
+    }
+
+    @Test(groups = "Live")
+    public void testGetsStreams() {
```
aledsage: The class name and test method name don't really describe what is being tested - no-one would guess that it's just about openIptables from the name.

Also this is not just about `EmptySoftwareProcess` - it applies to all `SoftwareProcess`es. We already have a `VanillaSoftwareProcessStreamsIntegrationTest` - I wonder if we should add more tests there, or whether we should have a new class like this. Not sure.

software/base/src/main/java/org/apache/brooklyn/entity/machine/MachineInitTasks.java
```diff
@@ -105,19 +108,24 @@ protected void openIptablesImpl(Iterable<Integer> inboundPorts, SshMachineLocati
                 iptablesRules.add(IptablesCommands.saveIptablesRules());
             }
             List<String> batch = Lists.newArrayList();
+
+            ByteArrayOutputStream outStream = new ByteArrayOutputStream();
```
aledsage: Thanks @iyovcheva - this certainly fixes it.

My concern is though that it only fixes it for this one command. If someone calls another command (e.g. uses `STOP_IPTABLES`) then that command's stdout/stderr won't get logged. I don't want us to have to include this in every method that wants to execute an ssh command.

It therefore feels like a more holistic approach would be good.

However, I'm fine with us merging this part for now and us thinking more about it.

aledsage:
Thanks @iyovcheva 

If you can address my comment about the test naming, then I think we can merge this. But I'd like us to revisit this in a future PR so that we get the stdout/stderr captured "for free" without having to include these lines in every such method.

sjcorbett:
@aledsage Are you happy for this to be merged? `EmptySoftwareProcessStreamsLiveTest` has been renamed `SoftwareProcessOpenIptablesStreamsLiveTest` rather than folded into `VanillaSoftwareProcessStreamsIntegrationTest`.

aledsage:
Thanks @iyovcheva - will merge now.
@sjcorbett I'm fine with this, as long as we revisit it more holistically soon.

PR #954, opened 2015-10-14
Amend text in creating-yaml.md

geomacy:
Amended paragraph in section "The Basic Structure" to clarify reference to the copy icon and to add some apparently missing text.


sjcorbett:
@geomacy Thanks. Test failure is a separate problem. Will merge.

PR #955, opened 2015-10-14
Camp parsing cleanup and fixes

neykov:
Modularizes the camp logic, allowing it to be shared with the TOSCA implementation. Also fixes a couple of problems:
  * The recursive check wasn't working for `brooklyn.child` items.
  * The recursive check was too strict, including all items already seen (including sibling items and their children) in the black list, making it impossible to use the same catalog item in multiple places.


core/src/main/java/org/apache/brooklyn/core/resolve/CatalogServiceSpecResolver.java
```diff
@@ -0,0 +1,108 @@
...
+        if (nonRecursiveCall) {
+            // Make a copy of the encountered types, so that we add the item being resolved for
+            // dependency items only. Siblings must not see we are resolving this item.
+            Set<String> encounteredTypes = ImmutableSet.<String>builder()
+                    .addAll(parentEncounteredTypes)
+                    .add(item.getSymbolicName())
+                    .build();
+
+            // CatalogItem generics are just getting in the way, better get rid of them, we
+            // are casting anyway.
```
sjcorbett: Is this a note to your future self? One for the mailing list?

core/src/main/java/org/apache/brooklyn/core/resolve/JavaServiceSpecResolver.java
```diff
@@ -0,0 +1,91 @@
...
+                throw e;
+            }
+        }
+    }
+
+    private EntitySpec<?> resolveInternal(String localType, BrooklynClassLoadingContext loader) {
+        Maybe<Class<? extends Entity>> javaTypeMaybe = tryLoadJavaType(localType, loader);
+        if (javaTypeMaybe.isAbsent())
+            throw new IllegalStateException("Could not find "+localType, ((Maybe.Absent<?>)javaTypeMaybe).getException());
+        Class<? extends Entity> javaType = javaTypeMaybe.get();
```
sjcorbett: Is `isAbsent` sufficient or should we check `isAbsentOrNull`?

core/src/main/java/org/apache/brooklyn/core/resolve/ServiceSpecResolver.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * @return if the resolver can create a spec for the service type
+     */
+    boolean accepts(String type, BrooklynClassLoadingContext loader);
+
+    /**
+     * Create a spec for the service type
+     */
+    @Nullable EntitySpec<?> resolve(String type, BrooklynClassLoadingContext loader, Set<String> encounteredTypes);
```
sjcorbett: Might be useful to note in the documentation what implementors should do with `encounteredTypes` and what a `null` return value indicates.

core/src/main/java/org/apache/brooklyn/core/plan/PlanToSpecTransformer.java
```diff
@@ -56,6 +57,6 @@
      * implementations will typically look at the {@link CatalogItem#getCatalogItemType()} first.
      * <p>
      * should throw {@link PlanNotRecognizedException} if this transformer does not know what to do with the plan. */
-    <T,SpecT extends AbstractBrooklynObjectSpec<? extends T, SpecT>> SpecT createCatalogSpec(CatalogItem<T, SpecT> item) throws PlanNotRecognizedException;
+    <T,SpecT extends AbstractBrooklynObjectSpec<? extends T, SpecT>> SpecT createCatalogSpec(CatalogItem<T, SpecT> item, Set<String> encounteredTypes) throws PlanNotRecognizedException;
```
sjcorbett: Worth noting in the javadoc how the additional argument should be used by implementors.

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
```diff
@@ -168,156 +139,60 @@ public static String getDeclaredType(String knownServiceType, AbstractResource o
...
+            CatalogItem<?, ?> item = CatalogUtils.getCatalogItemOptionalVersion(mgmt, Strings.removeFromStart(type, "catalog:"));
+            String proto = Urls.getProtocol(type);
+            if (item != null && encounteredCatalogTypes.contains(item.getSymbolicName())) {
+                msgDetails = "Cycle between catalog items detected, starting from " + type +
+                        ". Other catalog items being resolved up the stack are " + encounteredCatalogTypes +
+                        ". Tried loading it as a Java class instead but failed.";
+            } else if (proto != null) {
+                msgDetails = "The reference " + type + " looks like a URL (running the CAMP Brooklyn assembly-template instantiator) but the protocol " +
+                        proto + " isn't white listed (" + BrooklynCampConstants.YAML_URL_PROTOCOL_WHITELIST + "). " +
+                        "Not a catalog item or java type as well.";
```
sjcorbett: I'd phrase this something like "It's also neither a catalog item nor a java type".

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
```diff
@@ -350,35 +226,19 @@ protected String getJavaType() {
...
-        if (planId != null) {
-            entity.config().set(BrooklynCampConstants.PLAN_ID, planId);
-        }
-
-        if (spec.getLocations().size() > 0) {
-            ((AbstractEntity)entity).addLocations(spec.getLocations());
-        }
-
-        if (spec.getParent() != null) entity.setParent(spec.getParent());
+    private <T extends Entity> void decoreateSpec(EntitySpec<T> spec) {
```
sjcorbett: Typo in method name: `decoreate`.

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/service/UrlServiceSpecResolver.java
```diff
@@ -0,0 +1,72 @@
...
+            BrooklynCampConstants.YAML_URL_PROTOCOL_WHITELIST.contains(protocol);
+    }
+
+    @Override
+    public EntitySpec<?> resolve(String type, BrooklynClassLoadingContext loader, Set<String> encounteredTypes) {
+        String yaml;
+        try {
+            yaml = ResourceUtils.create(this).getResourceAsString(type);
+        } catch (Exception e) {
+            log.warn("AssemblyTemplate type " + type + " which looks like a URL can't be fetched.", e);
```
sjcorbett: I'd reword as "AssemblyTemplate type .. looks like a URL that can't be fetched".

sjcorbett:
This is a great set of changes. All of my comments are insubstantial. I'm happy for this to be merged.

@ahgittin @aledsage Any comments?

core/src/main/java/org/apache/brooklyn/core/resolve/CatalogServiceSpecResolver.java
neykov: I think it's too late to remove them - it's already baked in the public API and would be too hard to remove them in a backwards compatible way. Maybe just a reminder not to overuse generics in the future.

core/src/main/java/org/apache/brooklyn/core/resolve/JavaServiceSpecResolver.java
neykov: Present means non-null (at least in this case)

neykov:
@sjcorbett Thanks for the great comments, was able to clean up the logic even further when having to spell out the API expectations in the javadoc.

neykov:
Note that the changes here focus on entities (a.k.a. services in CAMP). Probably worth extending them to include the rest of the supported types (policies, locations, etc.) so that API changes are done at once and TOSCA can reuse those as well.

neykov:
Move policy & location resolving to core as well.

neykov:
A lot of the logic in `org.apache.brooklyn.camp.brooklyn.spi.creation.BrooklynComponentTemplateResolver.configureEntityConfig(EntitySpec<?>)` seems to be generic, not specific to CAMP. Probably makes sense to let the resolvers do the configuration, passing them the attributes/properties. This will require changing the interface once again, so better do it sooner rather than later, but better not wait and merge this PR as is, already has more than enough changes.

PR #956, opened 2015-10-14
BROOKLYN-175 Fixing MongoDB restart() phase

ygy:
- isRunning() check uses ping() during the restart phase
- ping() must complete in a reasonable amount of time since
  it is invoked against stopped entities too
- lowering the MongoClient timeouts, because default values are too 
  high when used inside ping()


sjcorbett:
Looks fine. Will run integration tests and merge.

PR #957, opened 2015-10-14
Move winrm base from brooklyn-core into brooklyn-software-winrm

CMoH:
Related to BROOKLYN-182

More cleanup will be needed later on, but this is a step that eliminates the winrm dependency from brooklyn-core


hzbarcea:
@CMoH I wanted to assign [BROOKLYN-182] to you, but I couldn't find your jira account. Do you have one?

hzbarcea:
@cmoh, looks good and tests pass.
Let's wait until tomorrow for (hopefully) an extra pair of eyes to look at this before merging.

software/winrm/pom.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bostko: What is the point creating a new module?
I see only several tests which were being moved to this new module.

The implementation of the winrm protocol is available in a separate repo.
If we want to rewrite it, we can contribute to the same old repository. Implementing OSGI could be done in the old repository as well.

CMoH:
The idea behind OSGi is that location types will be plugged into the brooklyn core. From this perspective it makes sense that code related to WinRM locations should be pulled out of brooklyn-core, as a first step.

software/winrm/pom.xml
hzbarcea: @bostko This was discussed on the mailing list on the Subject: "[PROPOSAL] Split brooklyn-core during OSGification" thread. The ./software/winrm location was suggested by @ahgittin and it's a tad less than ideal, because software/based now depends on software/winrm, but it's good enough.

This change is good regardless because it allows splitting the brooklyn-core into services that are optional (like support for winrm in this case), making it more modular and more manageable, but it's important because the current implementation is blocking the effort to fully OSGi-ify brooklyn, dragging in jython via winrm4j. You are right that a rewrite of winrm4j should be OSGi compliant, but that still leaves the issue of the brooklyn 'glue' which should still be a separate component.

hzbarcea:
LG, on top of the last commits too. Merging.

core/src/main/java/org/apache/brooklyn/core/location/Machines.java
```diff
@@ -102,15 +100,11 @@
     }
     
     public static Maybe<MachineLocation> findUniqueMachineLocation(Iterable<? extends Location> locations) {
-        return findUniqueElement(locations, MachineLocation.class);
+        return findUniqueMachineLocation(locations, MachineLocation.class);
     }
 
-    public static Maybe<SshMachineLocation> findUniqueSshMachineLocation(Iterable<? extends Location> locations) {
```
aledsage: We shouldn't delete things without first deprecating. I'll add this back in, and deprecate it.

hzbarcea:
@aledsage fair enough, I will follow up with a PR adding back the removed APIs and mark them as deprecated. 

I didn't see an internal public statement/policy on how long one should expect a deprecated API to be kept around. Something to discuss on the mailing list.

hzbarcea:
@aledsage ok, you beat me to it. Thanks!

PR #963, opened 2015-10-15
BROOKLYN-184: auto-manage entities

aledsage:
The initial review can focus on the commit "Brooklyn-184: Entities auto-managed". That makes the most important changes. The last commit ("remove deprecated uses of Entities.manage") is really just updating tests.

This could do with some more thorough (manual) testing, and checking of more of the integration tests.


core/src/main/java/org/apache/brooklyn/enricher/stock/Enrichers.java
```diff
@@ -520,6 +522,10 @@ public B from(Entity val) {
             this.fromEntity = checkNotNull(val);
             return self();
         }
+        public B from(Task<? extends Entity> val) {
```
aledsage: Note that I added in this PR the seemingly unrelated "Enrichers: support producer as Task for deferred supplier" so that I could re-write some of the code to use `EntitySpec`, rather than the previous way of constructing two entities and then calling `manage()`.

hzbarcea:
Looks good, test passed. Reviewed. Merging.

PR #965, opened 2015-10-19
MySqlCluster dump based replication

neykov:
When the binary log of the mysql master is truncated the only way to spin up new slaves is to do a dump of the master and initialize the slave from it - implemented here.


software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/InitSlaveTaskBody.java
```diff
@@ -0,0 +1,426 @@
...
+    private MySqlNode getRandomSlave() {
+        List<MySqlNode> slaves = getHealhtySlaves();
+        if (slaves.size() > 0) {
+            return slaves.get(new Random().nextInt(slaves.size()));
+        } else {
+            return getMaster();
+        }
+    }
+
+    private ImmutableList<MySqlNode> getHealhtySlaves() {
```
sjcorbett: Typo in method name: `Healhty`

utils/common/src/main/java/org/apache/brooklyn/util/GenericTypes.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import com.google.common.reflect.TypeToken;
+
+@SuppressWarnings("serial")
+public class GenericTypes {
```
sjcorbett: Add a private constructor.

sjcorbett:
Looks great. I've made two trivial comments. Happy for this to be merged.

utils/common/src/main/java/org/apache/brooklyn/util/GenericTypes.java
neykov: Converted to an interface instead, allows to remove all `public static final` declarations.

neykov:
Thanks @sjcorbett, addressed comments.

iyovcheva:
There is an error after last commit:
[ERROR] /incubator-brooklyn/software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlClusterImpl.java:[124,17] fromMembers() in org.apache.brooklyn.enricher.stock.Enrichers.AbstractAggregatorBuilder is defined in an inaccessible class or interface
[ERROR] /incubator-brooklyn/software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlClusterImpl.java:[125,17] build() in org.apache.brooklyn.enricher.stock.Enrichers.AbstractAggregatorBuilder is defined in an inaccessible class or interface

neykov:
Thanks @iyovcheva, seems to be a problem with Java 8 only, working on it.

neykov:
Java 8 build fixed in https://github.com/apache/incubator-brooklyn/pull/991

PR #968, opened 2015-10-19
Add description for configuring passwordless sudo to localhost location.

lloyddave:
The current Locations page provides information on setting up passwordless SSH
and says that passwordless sudo is required, but does not tell the user how to setup.


aledsage:
Looks good - merging.

PR #969, opened 2015-10-20
Docflow amendments

geomacy:
Pull request with a number of minor updates to documentation, arising from Belfast 'First Encounter' (https://docs.google.com/document/d/1bXCxrBJXEtU1bBoEXcjcSZHfu6ehYDrIQfAsUXycW1M/):

[GENEASY] Doc/flow improvements easy - non-getting started
1. Geoff - move the 'AWS VPC issues which may affect users with older AWS accounts' section into a footnote/subpage?
2. Geoff - define vCD in the locations page
4. Geoff - ops/catalog/index.html#adding-to-the-catalog uses an entities path to delete an application rather than a​pplications, both work but in the context of deleting an application appears to make more sense?

[GSEASY] Doc/flow improvements easy - getting started flow
17. Geoff - add explanation for 'quarantine' to getting started.


sjcorbett:
Looks fine. Thanks.

PR #970, opened 2015-10-20
Update ‘xml-apis’ dependency

alasdairhodge:
Resolves Eclipse build errors in `XmlPlanToSpecTransformer`. Command-line Maven build was happy, possibly due to later version accidentally getting picked up via transitive dependency.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -110,6 +110,10 @@
                 <groupId>com.google.guava</groupId>
                 <artifactId>guava</artifactId>
               </exclusion>
+              <exclusion>
+                <groupId>xml-apis</groupId>
+                <artifactId>xml-apis</artifactId>
+              </exclusion>
             </exclusions>
         </dependency>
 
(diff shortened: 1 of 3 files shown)


neykov:
The functionality is already provided in the JRE. There's a transitive dependency on `xml-apis` coming from `domj4`, but it's web site says that: `These dependencies are only required to compile and run unit tests for the application:`. Its `pom.xml` seems to be incorrectly generated (from gradle) so it includes it as a normal compile dependency.

I'd say remove the `xml-apis` dependency altogether and exclude it from the `dom4j` dependency (which comes from `org.reflections` in `core`).

Couldn't figure out why the enforcer `DependencyConvergence` rule didn't catch the version mismatch though.

alasdairhodge:
Thanks @neykov, works a treat.

alasdairhodge:
Grrr, build failure is unrelated:
```
testJmxFeedIsPersistedWithPreCreatedJmxHelper(org.apache.brooklyn.feed.jmx.RebindJmxFeedTest)  Time elapsed: 1.476 sec  <<< FAILURE!
java.rmi.server.ExportException: Port already in use: 40127;
```

PR #971, opened 2015-10-20
Fix CreateUserPolicy for SuSe Linux

ygy:
- Sudo is compiled with --without-secure-path on Suse
- Brooklyn uses sudo with the -E parameter which preserves the
  user environment
- CreateUserPolicy needs commands only available on the secure path
  locations


aledsage:
Thanks @ygy - I presume `CreateUserPolicyLiveTest` still works? Merging.

PR #972, opened 2015-10-20
With Clocker microservice only one security group on AWS for customizer

grkvlt:
Will warn if two AWS security groups not found, but tries to return the single group. Throws `IllegalArgumentException` if zero or more than two.


ahgittin:
looks reasonable, merging

PR #973, opened 2015-10-20
Prevent DynamicCluster NPE when subclasses have not called super.init()

sjcorbett:
The `NEXT_CLUSTER_MEMBER_ID` attribute was uninitialised, causing an exception in `addNode`.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
@@ -170,7 +170,15 @@ public class DynamicClusterImpl extends AbstractGroupImpl implements DynamicClus
     @Override
     public void init() {
         super.init();
-        sensors().set(NEXT_CLUSTER_MEMBER_ID, new NextClusterMemberIdSupplier());
+        initialiseMemberId();
+    }
+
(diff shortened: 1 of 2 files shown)


sjcorbett:
The test failure is irrelevant:
```
java.rmi.server.ExportException: Port already in use: 40130; nested exception is: 
```

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -170,7 +170,15 @@ public DynamicClusterImpl() {
     @Override
     public void init() {
         super.init();
-        sensors().set(NEXT_CLUSTER_MEMBER_ID, new NextClusterMemberIdSupplier());
+        initialiseMemberId();
+    }
+
+    private void initialiseMemberId() {
+        synchronized (mutex) {
```
neykov: There's `sensors().modify(...)` which will do the same, but I would probably have implemented it similarly to avoid the Java syntax overhead of anonymous classes.

neykov:
I've seen a bunch of blueprints missing the `super.init()` call, so seems to be a frequent omission. Maybe we should think of a run-time assertion which checks this case and gives early warning to the developer.

Looks good, will merge.

PR #975, opened 2015-10-20
Feature: SetHostnameCustomizer

aledsage:


neykov:
Looks good, can be merged. All the XxxTemplateModels look alike so probably can be refactored to DRY, just a thought for future changes.

software/base/src/main/java/org/apache/brooklyn/entity/machine/SetHostnameCustomizer.java
```diff
@@ -0,0 +1,201 @@
...
+    public String setLocalHostname(SshMachineLocation machine, String hostName, String domainFixed) {
+        log.info("Setting local hostname of " + machine + " to " + hostName 
+                + (Strings.isNonBlank(domainFixed) ? ", " + domainFixed : ""));
+        
+        boolean hasDomain = Strings.isNonBlank(domainFixed);
+        String fqdn = hasDomain ? hostName+"."+domainFixed : hostName;
+        
+        exec(machine, true, 
+                BashCommands.sudo(String.format("sed -i.bak -e '1i127.0.0.1 %s %s' -e '/^127.0.0.1/d' /etc/hosts", fqdn, hostName)),
+                BashCommands.sudo(String.format("sed -i.bak -e 's/^HOSTNAME=.*$/HOSTNAME=%s/' /etc/sysconfig/network", fqdn)),
```
ygy: This will fail for Debian Linux derivatives like Ubuntu. 
Debian uses /etc/hostname to preserve the hostname during reboots.

aledsage: Agreed. This is focused on CentOS and RHEL just now (there's a comment in the javadoc). We should at a later stage add support for other distros.

aledsage:
Thanks @neykov @ygy - have fixed the typo in the javadoc (and the other two places in the file with the same typo!)

Merging now.

PR #977, opened 2015-10-20
BROOKLYN-186 workaround in JcloudsLocation

aledsage:
Fixes BROOKLYN-186 by working around if jclouds gives us the wrong login user, or gives us both a password and ssh key (where only one of them works).

It's a bit messy doing this, but feels like the most robust thing to do.

(The other commit about `MutableSet` is just to make that easier to use in my JcloudsLocation fix, and to improve test coverage while I was at it).


neykov:
LGTM

aledsage:
Thanks @neykov - merging.

PR #980, opened 2015-10-21
Refactoring openIptables execution

iyovcheva:
Some changes are applied to openIptables execution:
1. "open iptables" task has sub tasks "insert rules" and "list rules" which show in the activity log.
2. "insert rules" has sub tasks - (1) insert rules, divided on commands batches of 100 rules where each batch is a sub task and shows in the activity log and (2) "save" sub task
Every leaf task has streams stdin, stdout, stderr.

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/entity/machine/MachineInitTasks.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/machine/MachineInitTasks.java
@@ -18,15 +18,12 @@
  */
 package org.apache.brooklyn.entity.machine;
 
-import java.io.ByteArrayOutputStream;
 import java.util.List;
-import java.util.concurrent.Callable;
 
-import com.google.common.collect.ImmutableMap;
 import org.apache.brooklyn.api.mgmt.Task;
 import org.apache.brooklyn.core.entity.EntityInternal;
 import org.apache.brooklyn.core.mgmt.BrooklynTaskTags;
-import org.apache.brooklyn.util.stream.Streams;
+import org.apache.brooklyn.util.core.task.system.ProcessTaskFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -61,34 +58,51 @@ public class MachineInitTasks {
     }
 
     /**
-     * Returns a queued {@link Task} which opens the given ports in iptables on the given machine.
+     * Returns a queued {@link Task} which stops iptables on the given machine.
      */
-    public Task<Void> openIptablesAsync(final Iterable<Integer> inboundPorts, final SshMachineLocation machine) {
-        return DynamicTasks.queue("open iptables "+toTruncatedString(inboundPorts, 6), new Callable<Void>() {
-            public Void call() {
-                openIptablesImpl(inboundPorts, machine);
-                return null;
+    public Task<Void> stopIptablesAsync(final SshMachineLocation machine) {
+        return DynamicTasks.queue("stop iptables", new Runnable() {
+            public void run() {
+                stopIptablesImpl(machine);
             }
         });
     }
 
-    /**
-     * Returns a queued {@link Task} which stops iptables on the given machine.
-     */
-    public Task<Void> stopIptablesAsync(final SshMachineLocation machine) {
-        return DynamicTasks.queue("stop iptables", new Callable<Void>() {
-            public Void call() {
-                stopIptablesImpl(machine);
-                return null;
-            }
-        });
+    protected void stopIptablesImpl(final SshMachineLocation machine) {
+
+        log.info("Stopping iptables for {} at {}", entity(), machine);
+
+        List<String> cmds = ImmutableList.<String>of();
+
+        Task<Integer> checkFirewall = checkLocationFirewall(machine);
+
+        if (checkFirewall.getUnchecked() == 0) {
+            cmds = ImmutableList.of(IptablesCommands.firewalldServiceStop(), IptablesCommands.firewalldServiceStatus());
+        } else {
+            cmds = ImmutableList.of(IptablesCommands.iptablesServiceStop(), IptablesCommands.iptablesServiceStatus());
+        }
+
+
+        subTaskHelperAllowingNonZeroExitCode("execute stop iptables", machine, cmds.toArray(new String[cmds.size()]));
     }
+
 
     /**
      * See docs in {@link BashCommands#dontRequireTtyForSudo()}
      */
     public Task<Boolean> dontRequireTtyForSudoAsync(final SshMachineLocation machine) {
         return DynamicTasks.queue(SshTasks.dontRequireTtyForSudo(machine, true).newTask().asTask());
+    }
+
+    /**
+     * Returns a queued {@link Task} which opens the given ports in iptables on the given machine.
+     */
+    public Task<Void> openIptablesAsync(final Iterable<Integer> inboundPorts, final SshMachineLocation machine) {
+        return DynamicTasks.queue("open iptables "+toTruncatedString(inboundPorts, 6), new Runnable() {
+            public void run() {
+                openIptablesImpl(inboundPorts, machine);
+            }
+        });
     }
 
     protected void openIptablesImpl(Iterable<Integer> inboundPorts, SshMachineLocation machine) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/brooklyn/location/ssh/SshMachineLocation.java
```diff
@@ -664,6 +655,24 @@ public int execCommands(Map<String,?> props, String summaryForLogging, List<Stri
...
+        Map flags = Maps.newLinkedHashMap();
+        if (BrooklynTaskTags.stream(Tasks.current(), BrooklynTaskTags.STREAM_STDIN)==null) {
+            Tasks.addTagDynamically(BrooklynTaskTags.tagForStreamSoft(BrooklynTaskTags.STREAM_STDIN,
+                    Streams.byteArrayOfString(Strings.join(commands, "\n"))));
+        }
+        if (BrooklynTaskTags.stream(Tasks.current(), BrooklynTaskTags.STREAM_STDOUT)==null) {
+            ByteArrayOutputStream stdout = new ByteArrayOutputStream();
+            ByteArrayOutputStream stderr = new ByteArrayOutputStream();
+            Tasks.addTagDynamically(BrooklynTaskTags.tagForStreamSoft(BrooklynTaskTags.STREAM_STDOUT, stdout));
+            Tasks.addTagDynamically(BrooklynTaskTags.tagForStreamSoft(BrooklynTaskTags.STREAM_STDERR, stderr));
```
neykov: Calling this method multiple times will result in many stdout/stderr tags for the current task which might be unexpected. See below for a better solution.

software/base/src/main/java/org/apache/brooklyn/entity/machine/MachineInitTasks.java
```diff
@@ -105,28 +103,85 @@ protected void openIptablesImpl(Iterable<Integer> inboundPorts, SshMachineLocati
...
+            insertIptablesRules(iptablesRules, iptablesInstallCommands, machine);
+            listIptablesRules(machine);
+        }
+    }
+
+    /**
+     * Returns a queued {@link Task} which inserts iptables rules.
+     */
+    public Task<Void> insertIptablesRules(final List<String> iptablesRules, final String installCommands, final SshMachineLocation machine) {
+        return DynamicTasks.queue("insert rules", new Callable<Void>() {
```
neykov: There's already API which will create a task and attach the streams to it, making it available in the log and in the UI. Use `SshTasks.newSshExecTaskFactory` when you have a machine available (as in this case), or `SshEffectorTasks.ssh` when you wish to retrieve the machine from the current context.

The helpers give you some additional goodies like:

```
SshTasks.newSshExecTaskFactory(machine, installCommands)
  .requiringExitCodeZero()  // checks the exit code, throwing if not 0, if a task fails subsequent task are not executed
  .summary(...) // short description used in logging
```

this will give you a `TaskAdaptable` which you can then `DynamicTasks.queue`.

---

When using `new Callable<Void>()` consider using `new Runnable()` if the API supports it.

neykov:
+1 for adding the cmds as tasks, very useful in the UI.

core/src/main/java/org/apache/brooklyn/location/ssh/SshMachineLocation.java
iyovcheva: Hi @neykov, I think that this is impossible, because the tags are Set<Object> and even there was no `==null` check, if the tags are already in, the set will remain unchanged.

neykov: Good point for the null check, but then you would loose subsequent streams when called multiple times. Doesn't feel right that the machine code deals with task related responsibilities.
You are basically re-implementing SshTasks so why not just use it.

iyovcheva:
Thanks @neykov, using SshTasks was a good point.

core/src/main/java/org/apache/brooklyn/location/ssh/SshMachineLocation.java
```diff
@@ -664,6 +655,7 @@ public int execCommands(Map<String,?> props, String summaryForLogging, List<Stri
     public int execCommands(String summaryForLogging, List<String> commands, Map<String,?> env) {
         return execCommands(MutableMap.<String,Object>of(), summaryForLogging, commands, env);
     }
+
```
neykov: There are no meaningful changes here, remove the file from the commit.

software/base/src/main/java/org/apache/brooklyn/entity/machine/MachineInitTasks.java
```diff
@@ -75,13 +72,19 @@ public Void call() {
...
-                stopIptablesImpl(machine);
-                return null;
-            }
-        });
+    public ProcessTaskWrapper<Integer> stopIptablesAsync(final SshMachineLocation machine) {
+
+        log.info("Stopping iptables for {} at {}", entity(), machine);
+
+        List<String> cmds = ImmutableList.<String>of();
+        if (isLocationFirewalldEnabled(machine)) {
```
neykov: Can convert this one as well to a task.

software/base/src/main/java/org/apache/brooklyn/entity/machine/MachineInitTasks.java
```diff
@@ -75,13 +72,19 @@ public Void call() {
...
+            cmds = ImmutableList.of(IptablesCommands.firewalldServiceStop(), IptablesCommands.firewalldServiceStatus());
+        } else {
+            cmds = ImmutableList.of(IptablesCommands.iptablesServiceStop(), IptablesCommands.iptablesServiceStatus());
+        }
+        ProcessTaskFactory<Integer>  stopIptables = SshTasks.newSshExecTaskFactory(machine, cmds.toArray(new String[cmds.size()]))
+                .summary("stop iptables");
+        return DynamicTasks.queue(stopIptables);
```
neykov: The above 3 lines repeat multiple times in the file, in functions `insertIptablesRulesOnCommandBatches`, `serviceIptablesSave`, `listIptablesRules`. Better create a single helper function which takes the machine, list of commands and summary for the task instead of creating a dedicated function for each call.

software/base/src/main/java/org/apache/brooklyn/entity/machine/MachineInitTasks.java
```diff
@@ -98,6 +101,7 @@ protected void openIptablesImpl(Iterable<Integer> inboundPorts, SshMachineLocati
             log.info("Opening ports in iptables for {} at {}", entity(), machine);
 
             List<String> iptablesRules = Lists.newArrayList();
+            String iptablesInstallCommands = "";
```
neykov: No point in creating a task and ssh-ing if the command is empty, better set it to null and don't do an ssh call in this case at all.

software/base/src/main/java/org/apache/brooklyn/entity/machine/MachineInitTasks.java
```diff
@@ -75,13 +72,19 @@ public Void call() {
...
      * Returns a queued {@link Task} which stops iptables on the given machine.
      */
-    public Task<Void> stopIptablesAsync(final SshMachineLocation machine) {
-        return DynamicTasks.queue("stop iptables", new Callable<Void>() {
-            public Void call() {
-                stopIptablesImpl(machine);
-                return null;
-            }
-        });
+    public ProcessTaskWrapper<Integer> stopIptablesAsync(final SshMachineLocation machine) {
```
neykov: Change all methods to return `Task` instead of `ProcessTaskWrapper`, use `.asTask()` to get the task from the `queue` return value.

software/base/src/main/java/org/apache/brooklyn/entity/machine/MachineInitTasks.java
```diff
@@ -75,13 +72,19 @@ public Void call() {
...
+        log.info("Stopping iptables for {} at {}", entity(), machine);
+
+        List<String> cmds = ImmutableList.<String>of();
+        if (isLocationFirewalldEnabled(machine)) {
+            cmds = ImmutableList.of(IptablesCommands.firewalldServiceStop(), IptablesCommands.firewalldServiceStatus());
+        } else {
+            cmds = ImmutableList.of(IptablesCommands.iptablesServiceStop(), IptablesCommands.iptablesServiceStatus());
+        }
+        ProcessTaskFactory<Integer>  stopIptables = SshTasks.newSshExecTaskFactory(machine, cmds.toArray(new String[cmds.size()]))
+                .summary("stop iptables");
```
neykov: Also do a `.requireExitCodeZero` unless the command is expected to fail.

neykov:
Almost there, some minor comments to address.

neykov:
Also squash the commits into one to keep the history clean.

core/src/main/java/org/apache/brooklyn/location/ssh/SshMachineLocation.java
```diff
@@ -20,17 +20,7 @@
...
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.io.PipedInputStream;
-import java.io.PipedOutputStream;
-import java.io.Reader;
-import java.io.StringReader;
+import java.io.*;
```
aledsage: Also recommend that you configure your IDE to not use `*` for imports. The convention in Brooklyn is to be explicit about the import classes.

iyovcheva:
@neykov I fixed the small remarks about making stopIptablesAsync to be a parent task, some renaming and removing the try-catch

neykov:
Excellent, merging. The build will fail anyway because of unrelated problems.

PR #982, opened 2015-10-21
Fix NPE if loginUser is not present and useJcloudsSshInit is false

ygy:


hzbarcea:
Pretty straight forward. Reviewed, tested, merging.

PR #983, opened 2015-10-22
Fix: SetHostnameCustomizer

aledsage:


software/base/src/test/java/org/apache/brooklyn/entity/machine/SetHostnameCustomizerTest.java
```diff
@@ -64,4 +68,20 @@ public void testGeneratedHostnameUsesPublicIpIfEmptyListOfPrivate() throws Excep
         
         assertEquals(customizer.generateHostname(machine), "ip-4-3-2-1-"+machine.getId());
     }
+    
+    @Test
+    public void testGeneratedHostnameIfNoIps() throws Exception {
+        SshMachineLocation machine = mgmt.getLocationManager().createLocation(LocationSpec.create(SshMachineLocationWithNoPublicIps.class)
+                .configure("address", "4.3.2.1"));
```
bostko: Why did you put address here? The location will have empty Set of ips anyway.

bostko:
Looks good.

software/base/src/test/java/org/apache/brooklyn/entity/machine/SetHostnameCustomizerTest.java
aledsage: It is compulsory configuration for an `SshMachineLocation` to have an address. I think that instantiation would have failed otherwise.

aledsage:
thanks @bostko - merging now.

software/base/src/test/java/org/apache/brooklyn/entity/machine/SetHostnameCustomizerTest.java
bostko: I tested it. The test was ok without it.

PR #984, opened 2015-10-22
Convert Groovy tests to Java

neykov:
This has been sitting around on my disk for quite some time,  better get it out before we do another mass renaming :)

The changes can be split in three main parts:
  * convert groovy tests to their java equivalent, mostly straightforward
  * remove deprecated groovy classes and their usages
  * reimplement GroovyJavaMethods in plain Java - this is not a natural conversion but allows to completely get rid of the groovy compiler for the project (for non-test code, see below for tests).

Why is this change useful - in addition to the added static safety we get from the converted code, it's no longer required to install the groovy plugin in Eclipse to be able to use the project. Probably most useful to beginners who can import the project in a plain Eclipse install.

The remaining groovy files (in `core` only) exercise how Brooklyn handles Groovy objects, doesn't make sense to convert them to Java. Maybe we could move them to a dedicated project in future.


aledsage:
Excellent, thanks @neykov 
Initial quick skim looks good; I'll look in more detail tomorrow.

core/pom.xml
```diff
@@ -247,6 +251,18 @@
                         </configuration>
                     </execution>
                 </executions>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.codehaus.groovy</groupId>
+                        <artifactId>groovy-eclipse-compiler</artifactId>
```
aledsage: Not for this PR (as definitely good to leave versions as-is while making this big change), but it would make sense to upgrade the groovy versions at some point:
* groovy-eclipse-compiler to 2.9.2-01 (from 2.9.0-01)
* groovy-eclipse-batch to 2.4.3-01 (from 2.3.4-01)
* groovy-all to 2.3.7 (from 2.3.4)

Note that https://github.com/groovy/groovy-eclipse/wiki says eclipse-compiler 2.9.1 only supports 2.3.7 (not 2.4.x), and also note there are no release notes for 2.9.2 (even though it seems like an official release, based on the numbering scheme in maven central).

usage/downstream-parent/pom.xml
```diff
@@ -62,7 +62,6 @@
     <jersey.version>1.18.1</jersey.version>
     <httpclient.version>4.4.1</httpclient.version>
     <commons-lang3.version>3.1</commons-lang3.version>
-    <groovy.version>2.3.4</groovy.version> <!-- Version supported by https://github.com/groovy/groovy-eclipse/wiki/Groovy-Eclipse-2.9.0-Release-Notes -->
```
aledsage: Do we really want to remove this? With OSGi, obviously we don't care about this kind of thing. But without that, someone writing entities that will be dropped into Brooklyn should not add a different version of groovy-all from the one already in Brooklyn.

aledsage:
Thanks @neykov - very happy for this to be merged; doing so now.

See what you think about the `groovy.version` in downstream-parent; no particularly strong feelings from me.

neykov:
Addressed review comments in https://github.com/apache/incubator-brooklyn/pull/989. It's not clear whether the groovy-eclipse-compiler 2.9.2 is released after all, so used `2.9.1-01` - no features from groovy `2.4` are needed.

neykov:
An unexpected consequence of merging this is that it shaves off more than a minute of a `-DskipTests` build :)

CMoH:
Nice!

PR #987, opened 2015-10-23
Introduce relationships between brooklyn objects

ahgittin:
adds BrooklynObject.relations() and relationships such as MANAGER_OF and IN_GROUP.
includes tests, also convering persistence.
also adds self-bounding generics to BrooklynObject, and lifts getConfig(key) to Configurable to avoid generics conflicts (not sure why `public <T> T getConfig(ConfigKey<T> key)` declared in two interfaces is incompatible but it is apparently).

there is a lot more we'd like to do -- i'll open a thread on the mailing list -- but this is ready to be merged.

for ease of reference the things i'd like to see following this are below

---

Make custom relations usable:
* add CAMP YAML `brooklyn.relations` field to allow callers to specify relations
* add REST API to query relations
* update UI to show relations

Replace management and group membership with relations:
* new "active_child" relationship for a managed child which is important enough to call out in the UI
* add subscriptions for relations on an entity
* replace 'members' field in Group with GROUP_CONTAINS relationship
    leave old field there for deserialization for one release cycle, but empty it and move everything to be a relation
    use subscription to update count
* replace `parent/children` fields in Entity with use of relations
   * again leaving fields in place for one release cycle in order to update count
   * add `addChild(Entity, boolean isActive)` such that last item can be set false to establish a management-only relationship
   * REST API call for "children" to show only ACTIVE children

Finally simplify persistence:
* make Policy, Enricher Feed, etc all be Entities
* make Location be Entities and replace the LocationRegistry with use of locations in catalog
* use IN_LOCATION, HAS_POLICY, HAS_FEED, etc relations for them, instead of custom persistence
* allow Catalog to support arbitrary types (i.e. it becomes a Registry of types, rather than a Catalog)
   * with warning or fail if we try to persist an unknown type
   * with serialization info (to allow migration across versions)
* persist as YAML


api/src/main/java/org/apache/brooklyn/api/objs/BrooklynObject.java
```diff
@@ -141,4 +149,11 @@
          */
         boolean unsubscribe(SubscriptionHandle handle);
     }
+    
+    public interface RelationSupport<T extends BrooklynObject> {
+        public <U extends BrooklynObject> void add(Relationship<? super T,U> relationship, U target);
```
neykov: Should we add `? super U` for the `Relationship`, for the same reason we have it for the `T`?

api/src/main/java/org/apache/brooklyn/api/objs/Configurable.java
```diff
@@ -41,6 +41,11 @@
     @Deprecated
     public <T> T setConfig(ConfigKey<T> key, T val);
 
+    /**
+     * Convenience method for {@code config().get(key)}
+     */
+    <T> T getConfig(ConfigKey<T> key);
```
neykov: `config().get()` is not much longer, so this is of questionable value. But if we have it, better have the mirroring method as well - remove `@Deprecated` from the `setConfig`.

Personally I prefer to have one true way to do things (in general, not only here).

neykov: I see you just moved it up the hierarchy now, so not the time to do any changes, but the above still holds.

Can remove it from `Entity`, `Policy`, etc. in this case.

core/src/main/java/org/apache/brooklyn/core/objs/BrooklynObjectInternal.java
```diff
@@ -31,12 +31,12 @@
 
 import com.google.common.annotations.Beta;
 
-public interface BrooklynObjectInternal extends BrooklynObject, Rebindable {
+public interface BrooklynObjectInternal<PublicSelfType extends BrooklynObject, InternalSelfType extends BrooklynObjectInternal<PublicSelfType,InternalSelfType>> extends BrooklynObject, Rebindable {
```
neykov: This interface is mostly used to cast an object to it and then use it. As such introducing generics here will make it awkward to use, having to suppress warnings (you already do `@SuppressWarnings({ "rawtypes", "unchecked" })` for each new usage). 

The generic types don't introduce value anyway, better follow the `getRebindSupport` pattern, where subclasses will override it with stronger typing.

api/src/main/java/org/apache/brooklyn/api/mgmt/rebind/mementos/Memento.java
```diff
@@ -77,7 +79,8 @@
 
     public Collection<Object> getTags();
     
+    Map<String, Set<BrooklynObject>> getRelations();
```
neykov: I'm surprised to see `BrooklynObject` reference in the memento, the existing practice is to store the IDs only. Not much of a difference in the end though since xstream will take care of it. May be we should do the same for the other references as well.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogItemDo.java
```diff
@@ -65,6 +64,19 @@ public SubscriptionSupportInternal subscriptions() {
         throw new UnsupportedOperationException();
     }
     
+    /**
+     * @throws UnsupportedOperationException; relations are not supported for catalog items
+     */
+    @Override
+    public RelationSupportInternal<CatalogItem<T,SpecT>> relations() {
+        throw new UnsupportedOperationException();
```
neykov: I think it's better to return an immutable empty RelationSupport object instead. This will allow to avoid special casing `CatalogItem` when working with `BrooklynObject`s.

api/src/main/java/org/apache/brooklyn/api/objs/BrooklynObject.java
```diff
@@ -141,4 +149,11 @@
          */
         boolean unsubscribe(SubscriptionHandle handle);
     }
+    
+    public interface RelationSupport<T extends BrooklynObject> {
```
neykov: Recently added *Support interfaces implement `Iterable`, see https://github.com/apache/incubator-brooklyn/pull/911 for discussion.

core/src/main/java/org/apache/brooklyn/core/entity/EntityRelations.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+    
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    public static <T extends BrooklynObject> Set<Relationship<? super T,? extends BrooklynObject>> getRelations(T source) {
+        return ((BrooklynObjectInternal)source).relations().getLocalBackingStore().getRelationships();
+    }
+
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    public static <T extends BrooklynObject,U extends BrooklynObject> Set<U> getRelationships(Relationship<? super T,U> relationship, T source) {
+        return ((BrooklynObjectInternal)source).relations().getLocalBackingStore().getRelations(relationship);
```
neykov: Seems you've got the backing store method names swapped. `getRelations` from above calls `getRelationships` and `getRelationships` here delegates to `getRelations`

core/src/main/java/org/apache/brooklyn/core/objs/AbstractBrooklynObject.java
```diff
@@ -55,8 +57,18 @@
     @SetFromFlag(value = "tags")
     private final Set<Object> tags = Sets.newLinkedHashSet();
 
+    private RelationSupportInternal<PublicSelfType> relations = new ByObjectBasicRelationSupport<PublicSelfType>(getPublicThis(), new RelationChangedCallback());
```
neykov: Will be `null` after rebind.

neykov: Obviously tests pass, but can't see where it gets initialized, constructor is not called during rebind.

core/src/main/java/org/apache/brooklyn/core/relations/ByObjectBasicRelationSupport.java
```diff
@@ -0,0 +1,103 @@
...
+import com.google.common.collect.Multimaps;
+
+import brooklyn.basic.relations.Relationship;
+
+public class ByObjectBasicRelationSupport<SourceType extends BrooklynObject> extends AbstractBasicRelationSupport<SourceType> {
+
+    DataForBasicRelations<SourceType> data;
+    
+    public ByObjectBasicRelationSupport(SourceType source, Runnable relationChangeCallback) { 
+        super(source);
```
neykov: You are passing the `source` only to get a reference to its `relations()`, which is the same as `this`, feels over engineered.

neykov:
All good! Only minor comments.
What I strongly feel should be removed is the `PublicSelfType` and `PrivateSelfType` generic types, unless there's a good technical reason for them to stay. Better to have casts, overrides at several fixed points (`BrooklynObject` and its inheriting interfaces) rather than at each usage point.
I see a relation as a named symbol, rather than a named item with behaviour. They are much closer to config keys/sensors, specific to the entity pairs they are used with. That's why a global namespace (the catalog) is a bit unexpected, care to explain some more how do you see it used (probably yaml usage is the main driver for this)?

api/src/main/java/org/apache/brooklyn/api/objs/BrooklynObject.java
ahgittin: yes

api/src/main/java/org/apache/brooklyn/api/objs/Configurable.java
ahgittin: I moved it because it was confusing the compiler when in multiple interfaces, even though the signature was the same it complained of a local generics `<T>` mismatch.  Have removed from a couple places.

api/src/main/java/org/apache/brooklyn/api/mgmt/rebind/mementos/Memento.java
ahgittin: Agree, string is better for consistency right now.  It means we need untyped `lookup(...)` methods during rebind, which I've added (simply pass `null` as type constraint argument).

Using `BrooklynObject` is perhaps nicer if we do that everywhere (and clean up refs to `entityProxy` instead of `entity` etc in the generated XML); however if we switch to a yaml representation here which I'd like to eventually then it's a moot point.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogItemDo.java
ahgittin: agree

api/src/main/java/org/apache/brooklyn/api/objs/BrooklynObject.java
ahgittin: Cool -- I hadn't seen that!  Very nice for e.g. `policies()` and `tags()`.

However it's unclear what that should do in the case of *maps* like relations -- and config.  I note this isn't applied to `ConfigurationSupport`.  Suggest deferring this.  It can always be added once we know what its semantics should be.  /cc @aledsage

core/src/main/java/org/apache/brooklyn/core/entity/EntityRelations.java
ahgittin: good catch!

core/src/main/java/org/apache/brooklyn/core/objs/AbstractBrooklynObject.java
ahgittin: the no-arg constructor *is* invoked during rebind.  i've added a test to confirm (and assert the sequence).

core/src/main/java/org/apache/brooklyn/core/relations/ByObjectBasicRelationSupport.java
ahgittin: At the time there was a reason I did the double dispatch -- let me confirm it's still the case and come back.

ahgittin:
Very helpful comments @neykov .  Have done most of them; still need to:

* Think about `SelfType` generics
* Check double dispatch

I saw the relationships as rather more standardized than config keys -- for instance `in_location` or `has_policy` or `has_child` or `has_network` -- and providing mixin behaviour where the target applies behavior.  Thus independent of specific entities, for the most part; though new ones can be added I think we should aim to reuse quite a bit more than config keys.

I have been in two minds about whether relationships might in time have more behaviour and configuration.  In TOSCA for instance relationships can have both.  An example could be machine entities with networks; the relationship (link) instance could specify which network(s) might be preferred for different entities, without needing to make some networks globally preferred or introduce proxy network instances.  Not terribly compelling -- the use cases are still unclear -- but I think we should be open to this possibility in the future.  So I'm going to do a third minor change to leave room for this:

* Change existing references to `Relationship` to `RelationshipType`, as in `relations().getRelationshipTypes()`.  This would allow us in the future to add `Relationship` as a `BrooklynObject` type where we could have instances supporting configuration, and we could do `relations().getRelationships()` to return those instances so that callers could filter, but without committing to that now.

WDYT?

ahgittin:
I've added a comment on the double dispatch via `EntityRelations` statics:

```
/** This abstract impl delegates to {@link EntityRelations} for all changes, routing through a local backing store.
 * This allows us to make the changes in both directions simultaneously when a relationship is bi-directional,
 * and should facilitate changing a backing datastore or remote instances when that is supported.
 * <p>
 * Currently it can be implemented without it, simplifying things a bit (avoiding the double dispatch)
 * apart from the {@link #add(Relationship, BrooklynObject)} method triggering the reverse addition
 * if it isn't already present. TBD which is better (and the internal call to get the backing store is 
 * marked as Beta). */
```

WDYT?

ahgittin:
Re generics -- your intuition is spot on.  There were relatively few places we had to add the correctly-generically-typed interface methods and trivial casting-super override methods and `@SuppressWarnings`; this is a small price to pay to remove the complex generics.  Both `PublicSelfType` and `InternalSelfType` are now gone.

ahgittin:
all changes pushed, including `RelationshipType` and removed generics, and comments for the double dispatch.  need to see if tests pass but otherwise ready for re-review @neykov .

core/src/main/java/org/apache/brooklyn/core/relations/ByObjectBasicRelationSupport.java
```diff
@@ -0,0 +1,104 @@
...
+
+    public static class DataForBasicRelations<T extends BrooklynObject> implements RelationSupport<T> {
+        
+        Runnable relationChangeCallback;
+        
+        public DataForBasicRelations(Runnable relationChangeCallback) {
+            this.relationChangeCallback = relationChangeCallback;
+        }
+        
+        // TODO for now, relationships are stored here (and persisted); ideally we'd look them up in catalog
```
neykov: Not really persisted, only the name is serialized.

neykov:
Perfect! Will merge.

neykov:
+1 for the double dispatch and `RelationshipType`

core/src/main/java/org/apache/brooklyn/core/objs/AbstractBrooklynObject.java
neykov: Went to investigate a bit deeper as it didn't seem right. The constructor of `BrooklynObject`s is called because we are asking Brooklyn to instantiate the object. But for plain java objects which happen to be config/sensor values xstream handles instantiation and it won't call any of the constructors.

neykov:
@geomacy  I believe this is on the roadmap, but out of scope for this PR.

ahgittin:
@geomacy +1 to these living in catalog, sooner rather than later.  see the email thread.

PR #989, opened 2015-10-26
Ignore Groovy plugin in Eclipse build

neykov:
The changes allow the project to be imported in vanilla "Eclipse for Java Developers" install without using additional plugins.

Also upgrades the groovy related dependencies to latest release.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -265,5 +265,34 @@
                 </dependencies>
             </plugin>
         </plugins>
+        <pluginManagement>
+            <plugins>
+                <!--This plugin's configuration is used to store Eclipse m2e settings only. It has no influence on the Maven build itself.-->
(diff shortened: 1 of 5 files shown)


hzbarcea:
Reviewed, looks good, simple actually. Opened the project in Luna, all ok.

There is a merge conflict because of the properties moving in #962 into the root pom.xml (from parent/pom.xml). That's I think the only really minor issue.

hzbarcea:
Forgot to mention, these changes seem to address all of @aledsage 's comments from #984 if I am not mistaken.

ahgittin:
LGTM.  But I suggest we figure out if #962 is being corrected or reverted (see comments there) before fixing the merge conflict, unless this one is urgent.

PR #991, opened 2015-10-27
Fix Java 8 build error

neykov:
The raw cast erases all generics, forcing the return value to the bound which is `AbstractAggregatorBuilder`. Following method calls get resolved against the inaccessible (protected) `AbstractAggregatorBuilder` which leads to compile errors. This is a problem in Java 8. Java 7 and the Eclipse compiler don't need the additional cast.

### DIFF
--- a/software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlClusterImpl.java
+++ b/software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlClusterImpl.java
@@ -18,6 +18,8 @@
  */
 package org.apache.brooklyn.entity.database.mysql;
 
+import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


ahgittin:
this is where generics go crazy.  looks good though.

software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlClusterImpl.java
```diff
@@ -119,7 +121,7 @@ protected void initEnrichers() {
         enrichers().add(Enrichers.builder()
                 .aggregating(MySqlNode.DATASTORE_URL)
                 .publishing(SLAVE_DATASTORE_URL_LIST)
-                .computing((Function)Functions.identity())
+                .computing((Function<Collection<String>, List<String>>)(Function)Functions.identity())
```
bostko: @neykov why do you have (Function) cast

neykov: Can't do the cast directly because the types are incompatible, first need to erase the generics and then do the cast.

bostko: can you try `Function.<Collection<String>, List<String>>identity()`

neykov: `identity()` takes only a single generic as parameter, which if you think about it makes sense - it's identity :)

PR #992, opened 2015-10-27
Brooklyn OSGification - step 1 fixes

CMoH:
This addresses most of the comments in #962. Also included is a fix the karaf assembly project build, which leads to problems further on with brooklyn-itest.

The current quick-fix is to disable brooklyn-itest, so that the jenkins build won't be a problem for everyone until we manage to integrate it properly.


ahgittin:
awesome.  the `rt` is cryptic and having both `Osgis` and `OsgiUtils` is not ideal but this fixes the main concerns and more importantly fixes the build!  merging.

PR #993, opened 2015-10-30
Introduce a type registry as a simplified catalog

ahgittin:
Currently simply provides a compatible stateless facade to catalog.  A lot more to do -- commit describes the plans -- but I want early review and of course to try to avoid any merge conflicts!

Idea of type registry is that this will allow us to generalize atop types (soon adding "beans" alongside the "specs" which is all that catalog previously supported) in order to:
* simplify persistence/rebind across versions
* define types for local resolution, e.g. when uploading a plan
* define types to be used in specialized YAML, e.g. listing tasks for an effector


ahgittin:
build failure due to maven OOME; have upped settings (builds.apache.org) and close/reopen to trigger rebuild

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypePredicates.java
```diff
@@ -0,0 +1,191 @@
...
+            this.disabled = disabled;
+        }
+        @Override
+        public boolean apply(@Nullable RegisteredType item) {
+            return (item != null) && item.isDisabled() == disabled;
+        }
+    }
+
+    @SuppressWarnings("unused")
+    private static final Function<RegisteredType,String> ID_OF_ITEM_TRANSFORMER_ANONYMOUS = new Function<RegisteredType, String>() {
```
neykov: Don't see the point of this one?

core/src/main/java/org/apache/brooklyn/core/resolve/ResolveUtils.java
```diff
@@ -21,18 +21,18 @@
...
 import org.apache.brooklyn.api.policy.Policy;
 import org.apache.brooklyn.api.policy.PolicySpec;
-import org.apache.brooklyn.core.catalog.internal.CatalogUtils;
-import org.apache.brooklyn.core.mgmt.EntityManagementUtils;
+import org.apache.brooklyn.api.typereg.RegisteredType;
 import org.apache.brooklyn.core.mgmt.classloading.BrooklynClassLoadingContext;
 import org.apache.brooklyn.core.objs.BrooklynObjectInternal.ConfigurationSupportInternal;
 import org.apache.brooklyn.util.guava.Maybe;
 
+@Deprecated /** @deprecated since 0.9.0 never belonged here, and not used much; new principled TypeRegistry simplifies things */
```
neykov: We can safely move this back to CAMP land if TOSCA is not using it. It was introduced only recently and never has seen a release.

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypeConstraints.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    public static RegisteredTypeConstraint alreadyVisited(Set<String> encounteredTypeSymbolicNames) {
+        BasicRegisteredTypeConstraint result = new BasicRegisteredTypeConstraint();
+        result.encounteredTypes = encounteredTypeSymbolicNames;
+        return result;
+    }
+    public static RegisteredTypeConstraint alreadyVisited(Set<String> encounteredTypeSymbolicNames, String anotherEncounteredType) {
+        BasicRegisteredTypeConstraint result = new BasicRegisteredTypeConstraint();
+        result.encounteredTypes = MutableSet.copyOf(encounteredTypeSymbolicNames);
```
neykov: Better use `ImmutableSet` to avoid accidental changes. Might as well use it for the above method as well, should be a NOP in the general case where we are passing immutable already.

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypeConstraints.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    public static RegisteredTypeConstraint spec(Class<? extends BrooklynObject> javaSuperType) {
+        return of(RegisteredTypeKind.SPEC, javaSuperType);
+    }
+
+    public static <T extends AbstractBrooklynObjectSpec<?,?>> RegisteredTypeConstraint extendedWithSpecSuperType(RegisteredTypeConstraint source, Class<T> specSuperType) {
+        Class<?> superType = lookupTargetTypeForSpec(specSuperType);
+        BasicRegisteredTypeConstraint constraint = new BasicRegisteredTypeConstraint(source);
+        if (source==null) source = constraint;
```
neykov: Mark `source` as `@Nullable`

usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlRebindTest.java
```diff
@@ -198,13 +191,7 @@ protected void runRebindWithCatalogAndApp(RebindWithCatalogTestMode mode) throws
...
                 break;
             case STRIP_DEPRECATION_AND_ENABLEMENT_FROM_CATALOG_ITEM:
-                //Make sure that the flags are actually removed from the XML instead of still being there with false values
-                appItem.setDeprecated(true);
-                appItem.setDisabled(true);
-                mgmt().getCatalog().persist(appItem);
-                locItem.setDisabled(true);
-                locItem.setDeprecated(true);
-                mgmt().getCatalog().persist(locItem);
+                // nothing here -- but below we rebind with these fields removed to ensure that we can rebind
```
neykov: Should keep these. They confirm that the disabled/deprecated elements are removed from the XML. Prior the adding them the test was successfully passing, but not testing anything because the elements weren't actually removed, so the default values of `false` remained.

core/src/main/java/org/apache/brooklyn/core/resolve/ResolveUtils.java
ahgittin: k.  it's a good idea but i think `TypeRegistry` gives us a cleaner place to (as soon as we can) move this logic

core/src/main/java/org/apache/brooklyn/core/catalog/internal/JavaCatalogToSpecTransformer.java
```diff
@@ -68,11 +73,11 @@ public boolean accepts(String planType) {
     public <T, SpecT extends AbstractBrooklynObjectSpec<? extends T, SpecT>> SpecT createCatalogSpec(
             CatalogItem<T, SpecT> item, Set<String> encounteredTypes) throws PlanNotRecognizedException {
         if (item.getJavaType() != null) {
-            log.warn("Deprecated functionality (since 0.9.0). Using old-style xml catalog items with java type attribute for " + item);
```
neykov: `getJavaType() != null` should still be deprecated for catalog items, even if introduced in the type registry. It is non-null only for old-style xml catalog files.

usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlRebindTest.java
ahgittin: Have restored them, but it looks like the following code removes those entries whether they are true or false so curious why it is necessary?

core/src/main/java/org/apache/brooklyn/core/catalog/internal/JavaCatalogToSpecTransformer.java
```diff
@@ -68,11 +73,11 @@ public boolean accepts(String planType) {
...
             Class<?> type;
             try {
-                type = mgmt.getCatalogClassLoader().loadClass(item.getJavaType());
-            } catch (ClassNotFoundException e) {
+                type = CatalogUtils.newClassLoadingContext(mgmt, item).loadClass(item.getJavaType());
```
neykov: Old-style catalog items don't support osgi, so no point in the change, but doesn't hurt either.

usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlRebindTest.java
neykov: To verify that they are actually removed, rather than there but `false`.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/JavaCatalogToSpecTransformer.java
ahgittin: thanks -- that clears it up, added comments to remind us for later

core/src/main/java/org/apache/brooklyn/core/resolve/ResolveUtils.java
neykov: :+1:

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypeConstraints.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /** returns a constraint which allows anything */
+    public static RegisteredTypeConstraint any() {
+        return new BasicRegisteredTypeConstraint();
+    }
+
+    public static RegisteredTypeConstraint alreadyVisited(Set<String> encounteredTypeSymbolicNames) {
+        BasicRegisteredTypeConstraint result = new BasicRegisteredTypeConstraint();
+        result.encounteredTypes = encounteredTypeSymbolicNames == null ? ImmutableSet.<String>of() : ImmutableSet.copyOf(encounteredTypeSymbolicNames);
```
neykov: I am pretty sure all `encounteredTypeSymbolicNames` are non-null, so better not encourage null usage. If needed can introduce a new `alreadyVisitedEmpty` method for top-level calls.

api/src/main/java/org/apache/brooklyn/api/typereg/BrooklynTypeRegistry.java
```diff
@@ -0,0 +1,68 @@
...
+    /** as {@link #get(String, String, RegisteredTypeConstraint)} but allows <code>"name:version"</code> 
+     * (the {@link RegisteredType#getId()}) in addition to the unversioned name,
+     * using a default marker if no version can be inferred */
+    RegisteredType get(String symbolicNameWithOptionalVersion, @Nullable RegisteredTypeConstraint constraint);
+    /** as {@link #get(String, RegisteredTypeConstraint)} but with no constraints */
+    RegisteredType get(String symbolicNameWithOptionalVersion);
+
+    // NB the seemingly more correct generics <T,SpecT extends AbstractBrooklynObjectSpec<T,SpecT>> 
+    // cause compile errors, not in Eclipse, but in maven (?) 
+    <SpecT extends AbstractBrooklynObjectSpec<?,?>> SpecT createSpec(RegisteredType type, @Nullable RegisteredTypeConstraint optionalConstraint, Class<SpecT> optionalSpecSuperType);
```
neykov: The generic `SpecT` as is is not of much value, might use instead:
```
AbstractBrooklynObjectSpec<?,?> createSpec(RegisteredType type, @Nullable RegisteredTypeConstraint optionalConstraint, Class<? extends AbstractBrooklynObjectSpec<?,?>> optionalSpecSuperType);
```

If the last argument is used just to fill in the generic types, then no point in having it.

neykov: Looks like it's used as a constraint so just ignore the this point.

neykov: But then can be included in the constraint already being passed :)

ahgittin: it is automatically applied to the constraint, converting from `EntitySpec` to `Entity`

i think the last argument is useful in usage because it avoids the caller having to cast the result, and hopefully will prevent confusion about when to set `Entity` and when to set `EntitySpec`

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypeConstraints.java
ahgittin: there may be places where we pass around the set of visited places and some might be null; agree callers should generally avoid nulls for this but i prefer code being defensive here

api/src/main/java/org/apache/brooklyn/api/typereg/BrooklynTypeRegistry.java
neykov: Agreed.

neykov:
Finished review - looks good. Moves us in the right direction to YAMLification.

ahgittin:
thx @neykov - think i've address them all, merging

PR #994, opened 2015-10-30
Add EntityAsserts and HttpUtils.

geomacy:
This are added to allow core code to make assertions on entities without
without having a dependency on TestNG.

For now, deprecate EntityTestUtils and HttpTestUtils, and in the future these
can be removed in favour of the new classes.


core/src/main/java/org/apache/brooklyn/core/entity/EntityAsserts.java
```diff
@@ -0,0 +1,192 @@
...
+                    Asserts.assertTrue(changed.get(), entity + " -> " + attribute + " not changed");
+                }});
+        } finally {
+            entity.subscriptions().unsubscribe(entity, handle);
+        }
+    }
+
+
+    @Beta @SafeVarargs
+    public static <T> void assertAttributeNever(final Entity entity, final AttributeSensor<T> attribute, T... disallowed) {
```
ahgittin: can we call this `assertAttributeContinuallyNotEqualTo`?

add javadoc noting that this loops checking the value, it does not currently set up a subscription so a disallowed value might come up very quickly and be missed.  (we could fix this if needed.)

cleaner would be for this to call the next method, `assertAttribute...(Map, ...)`, with `null` flags, or perhaps to introduce an `assertAttributeContinually(..., Predicate)` and build up a predicate of `not(or(equalTo(disallowed[0]), equalTo(disallowed[1]), ...))`

core/src/main/resources/org/apache/brooklyn/core/mgmt/persist/deserializingClassRenames.properties
```diff
@@ -1415,5 +1415,5 @@ brooklyn.management.internal.UsageManager
 brooklyn.policy.basic.ConfigMapImpl                                              : org.apache.brooklyn.core.objs.AdjunctConfigMap
 brooklyn.util.internal.Repeater                                                  : org.apache.brooklyn.util.core.internal.Repeater
 brooklyn.entity.nosql.redis.RedisClusterImpl                                     : org.apache.brooklyn.entity.nosql.redis.RedisClusterImpl
-brooklyn.test.TrustingSslSocketFactory                                           : org.apache.brooklyn.test.TrustingSslSocketFactory
+brooklyn.test.TrustingSslSocketFactory                                           : org.apache.brooklyn.util.http.TrustingSslSocketFactory
```
ahgittin: also add a line:

    org.apache.brooklyn.test.TrustingSslSocketFactory : org.apache.brooklyn.util.http.TrustingSslSocketFactory

usage/test-support/src/main/java/org/apache/brooklyn/test/EntityTestUtils.java
```diff
@@ -46,6 +45,12 @@
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 
+/**
+ * A utility class containing tests on Entities.
+ *
+ * @deprecated Prefer core assertions org.apache.brooklyn.entity.EntityAsserts.
```
ahgittin: missing `core` in package name - `o.a.b.core.e.EA`

usage/test-support/src/main/java/org/apache/brooklyn/test/EntityTestUtils.java
```diff
@@ -46,6 +45,12 @@
...
+ */
+@Deprecated
 public class EntityTestUtils {
 
     // TODO would be nice to have this... perhaps moving this class, or perhaps this whole project, to core/src/test ?
```
ahgittin: can remove the TODO :)

utils/common/src/main/java/org/apache/brooklyn/util/http/HttpUtils.java
```diff
@@ -0,0 +1,387 @@
...
+import com.google.common.collect.Maps;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+
+/**
+ * Utility methods to aid testing HTTP.
+ * 
+ * @author aled
+ */
+public class HttpUtils {
```
ahgittin: call this `HttpAsserts` to mirror `EntityAsserts` ?  (and move the utility methods out?)

also note there is `HttpTool` where utility methods could perhaps go or be moved from.  it would be nice not to have both `HttpUtils` and `HttpTool` although maybe that would be okay given classes the latter use.

ahgittin:
need tests for the http assertions i think.

apart from that this looks real good and only minor comments.

geomacy:
hi Alex et al.  I have made the changes related to the comments by Alex in the commits [above](https://github.com/apache/incubator-brooklyn/pull/994#commits-pushed-03dbd28/). 

The only one I haven't yet done is to add unit tests for HttpUtils.  I don't want to hold back the others so I propose to do this in a separate pull request.  I will get on to the tests for HttpUtils tomorrow morning.

Note that to nicely divide up the HttpAsserts from the non assertion methods I have moved HttpTool out of core and into the utils module.  This results in imports changing in numerous files, so change 03dbd28 looks scarier than it really is (47 changed files).

ahgittin:
For backwards compatibility we need to leave an `HttpTool` class in the old place, and same for `HttpToolResponse`.  I think these can just extend the newly created classes, so that any methods and creation still work.  And add the usual `@Deprecation` markers to these classes.

ahgittin:
very minor, good to go once those two last items are addressed

usage/test-support/src/main/java/org/apache/brooklyn/test/EntityTestUtils.java
```diff
@@ -46,20 +45,25 @@
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 
+/**
+ * A utility class containing tests on Entities.
+ *
+ * @deprecated Prefer core assertions class org.apache.brooklyn.core.entity.EntityAsserts.
```
aledsage: Can you include a "since 0.9.0" in the deprecated javadoc line. Same for anywhere else you've added `@deprecated`.

geomacy:
Thanks @ahgittin and @aledsage for the comments.  I have made some further updates which I think address them all, but let me know if I've missed anything.

ahgittin:
looks great, merging

utils/common/src/test/java/org/apache/brooklyn/util/HttpAssertsTest.java
```diff
@@ -0,0 +1,293 @@
...
+        HttpAsserts.assertContentEventuallyMatches(testUri("/simple"), "[Oo][Kk]");
+    }
+
+    @Test
+    public void shouldAssertContentEventuallyMatchesWithFlags() {
+        stopServer();
+        final String simple = testUri("/simple");
+        HttpAsserts.assertUrlUnreachable(simple);
+        startAfter(2);
+        HttpAsserts.assertContentEventuallyMatches(ImmutableMap.of("timeout", "3s"), testUri("/simple"), "[Oo][Kk]");
```
neykov: Causes test failures on the apache jenkins CI. The servers there are quite overloaded so delays tend not to be exact, usually much longer.
Also tests which have explicit delays are usually marked as integration as those seconds easily add up. Is it possible to restructure the tests so that they don't need fixed sleeps?

neykov: A couple of example failures:
  * https://builds.apache.org/job/incubator-brooklyn-pull-requests/2069/console
  * https://builds.apache.org/job/incubator-brooklyn-pull-requests/2067/console

geomacy:
The sleeps are there to allow the test HTTP server time to shut down.  That's not required in all tests, so I could speed things up a bit by changing the test condition, but it is needed in some cases (for the 'eventually becomes unavailable' type tests).  I could add a timeout to the shutdown, but it occurs to me that it might be better to mark these as integration tests, because

- (a) they are not going to run any faster (they already take , and 
- (b) the code under test is fairly 'low level' and not likely to break because of other code within Brooklyn changing

so it may not add a lot of value to run these tests on every build, and it will slow things down a bit. 

What do you think?

neykov:
Moving them to integration will address the immediate problem so +1.
At some point we will start running the integration tests on apache infrastructure (read - **busy**, we already do, but they are failing for a number of other reasons), so the same issue will crop up - but can deal with it then.
---
The code could be restructured such that it doesn't use waits at all:
  * tests which need non-running server be put in a separate test, which doesn't start the server by default
  * no sleep on server start, if the server doesn't come up immediately, then always use `eventuallySucceeds`
  * for the cases where there's `xxxxAfter(5); succeedsEventually`, could be changed to `failsNow; task = succeedsEventuallyInAnotherTask; startNow(); task.get()`. What could be a problem here is that connections to closed ports take a long time to timeout?
  * The only place where a fixed wait is really needed is tests which expect to fail and have a timeout, those go in integration.

All of the above said, I am perfectly fine with moving all tests to integration as is.

geomacy:
I will move them to integration now and add a note to restructure in future, but I will try to make that the near future while it's fresh in my head. Thanks.

core/src/main/java/org/apache/brooklyn/util/core/http/HttpTool.java
```diff
@@ -76,311 +76,10 @@
...
-    public static boolean isStatusCodeHealthy(int code) { return (code>=200 && code<=299); }
-
-    public static String toBasicAuthorizationValue(UsernamePasswordCredentials credentials) {
-        return "Basic "+Base64.encodeBase64String( (credentials.getUserName()+":"+credentials.getPassword()).getBytes() );
-    }
+/**
+ * @deprecated since 0.9.0. Prefer org.apache.brooklyn.util.http.HttpTool
+ */
+@Deprecated
+public class HttpTool extends org.apache.brooklyn.util.http.HttpTool {
```
neykov: This doesn't make the class backwards compatible because return types will be coming from the new package, while users of this will be expecting the old classes. 
HttpTool is used frequently so better handle this gracefully (had a downstream project build failure). @geomacy Can you get the old implementation of `HttpTool` back here while keeping it marked as `@Deprecated`. Same for `HttpToolResponse`.

geomacy:
hi @neykov, sorry I had indeed missed this comment.  Will make changes as you say.

geomacy:
Have added the changes in https://github.com/apache/incubator-brooklyn/pull/1032

PR #995, opened 2015-10-30
pass flags to assertContentEventuallyMatches() for setting timeout

johnmccabe:
tested this locally as part of the proposed test framework


neykov:
The change is useful. Only one minor comment - since the code is ported from Groovy it doesn't have generics. Could you add types to the Maps, including the new method.

ahgittin:
LGTM.  @geomacy can you incorporate these changes into #994 ?

geomacy:
@ahgittin will do

PR #998, opened 2015-11-02
Fix MySQL artefact name in line with what’s currently available.

alasdairhodge:
I don't know the original rationale for using the `glibc2.5` variant of the Linux bin-tar, but no such artefacts are available from the official MySQL connumity site or from the Cloudsoft mirror.


neykov:
@alasdairhodge The current MySQL download URL for "Linux - Generic" is `http://dev.mysql.com/get/Downloads/MySQL-5.7/mysql-5.7.9-linux-glibc2.5-x86_64.tar.gz`.
Perhaps it's the Cloudsoft mirror that should only be updated?

ahgittin:
@alasdairhodge ping

PR #999, opened 2015-11-02
Implementation of proposed Test Entities to allow testing directly from blueprints

m4rkmckenna:
**Proposal**
http://mail-archives.apache.org/mod_mbox/incubator-brooklyn-dev/201510.mbox/%3CE185CA37-E693-4733-8DE1-208872BC1978%40cloudsoftcorp.com%3E

This PR relies on #994


sandbox/brooklyn-sandbox-test-framework/README.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+## TestHtmlCall
+Entity that makes a HTTP Request and tests the response
+
+#### Configuration
+| Key | Description | Required |
+| --- | ----------- | -------- |
+| target | The target entity to effect | yes (no if *targetId* is supplied) |
+| targetId | The id of the target entity to effect | yes (no if *target* is supplied) |
```
johnmccabe: `target`/`targetId` aren't used, need to add `assert` - see the javadoc for the assertion keywords

sandbox/brooklyn-sandbox-test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCallImpl.java
```diff
@@ -0,0 +1,98 @@
...
+     * @param assertions The map of assertions
+     */
+    private void checkAssertions(final String url, final Map<?, ?> assertions) {
+
+        for (final Map.Entry<?, ?> entry : assertions.entrySet()) {
+            if (Objects.equal(entry.getKey(), "regex")) {
+                LOG.info("Testing if url [{}] matches regex [{}]",
+                        new Object[]{url, entry.getValue()});
+                assertContentMatches(url, TypeCoercions.coerce(entry.getValue(), String.class));
+            } else if (Objects.equal(entry.getKey(), "string")) {
```
johnmccabe: agree with @m4rkmckenna that `contains` would be more meaningful than `string` in this context

ahgittin:
Should this be moved to `usage/test-framework` ?  /cc @aledsage @rdowner

m4rkmckenna:
@ahgittin @aledsage @rdowner Do you want me to move to ```usage/test-framework``` ... or is it ok where it is in ```sandbox/brooklyn-sandbox-test-framework```?

ahgittin:
I say move it across in this PR.

Best
Alex

m4rkmckenna:
@ahgittin Moved `sandbox/brooklyn-sandbox-test-framework` to `usage/test-framework`

ahgittin:
This looks great at first blush.  I'll add comments but so far nothing looks major so I'm of a mind to merge it unless I find anything else, and we can address comments as a separate PR.

usage/test-framework/README.md
```diff
@@ -0,0 +1,109 @@
+# Entities
```
ahgittin: These docs should be moved to the integrated `docs/guide/`.

usage/test-framework/etc/exampl-catalog.bom
```diff
@@ -0,0 +1,15 @@
+brooklyn.catalog:
```
ahgittin: name of this file is misspelled -- `exampl`

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/BaseTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ *  A base interface for all tests.
+ */
+public interface BaseTest extends Entity, Startable {
+
+    /**
+     * The target entity to test (optional, use either this or targetId).
+     */
+    @SetFromFlag(nullable = false)
```
ahgittin: why `nullable=false`; it can be `null` if `TARGET_ID` is supplied, in fact it should be an error if both are non-null and they don't agree ?

ahgittin: feels like it would make more sense for this to take an `EntitySpec` (cf `DynamicCluster.memberSpec` -- and let's also find a way to remove the `$brooklyn:entitySpec` requirement), and if supplied, for it to be started *before* the defined `brooklyn.children`.

btw at present this can only be used as a more long-winded way of `targetId` right?

and we could make it polymorphic; i.e. if it's a string then it's taken as an entity ID, if it's an `EntitySpec` or a `Map` it is taken as a spec, and if it's an Entity (using `$brooklyn:component`) we take it as an entity.  then no need for `targetId`.

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/ParallelTestCaseImpl.java
```diff
@@ -0,0 +1,130 @@
...
+import org.apache.brooklyn.core.entity.Attributes;
+import org.apache.brooklyn.core.entity.lifecycle.Lifecycle;
+import org.apache.brooklyn.core.entity.trait.StartableMethods;
+import org.apache.brooklyn.util.core.task.DynamicTasks;
+import org.apache.brooklyn.util.exceptions.Exceptions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This implementation will start all child entities in parallel.
```
ahgittin: javadoc should probably be on the interface?

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/ParallelTestCaseImpl.java
```diff
@@ -0,0 +1,130 @@
...
+        sensors().set(Attributes.SERVICE_STATE_ACTUAL, Lifecycle.STARTING);
+        try {
+            // Get an unsubmitted task for starting all the children of this entity in parallel,
+            // at the same location as this entity.
+            final TaskAdaptable<?> taskAdaptable = StartableMethods.startingChildren(this);
+            logger.trace("{}, TaskAdaptable: {}", this, taskAdaptable);
+
+            // Submit the task to the ExecutionManager so that they actually get started
+            // and then wait until all the parallel child entities have completed.
+            submitTaskAndWait(taskAdaptable);
```
ahgittin: what you've done works but I'd probably do:

    DynamicTasks.queue(taskAdaptable);
    DynamicTasks.waitForLast();

because it's an effector it's guaranteed to have a queueing context, and the advantage of a queueing context is that you can build up more complicated task logic (e.g. starting `target` if required) ahead of time and that task logic shows up nicely in the UI.  obviously you want to wait for the queue to drain before setting the service state or catching, which is what `waitForLast()` does, and it throws if there's a problem.  i wish we had better guidance on tasks (though i think @sjcorbett wrote some?).

more logging than we'd usually do as all the info will be available in the effector/task logging, but no issues with that especially if it's helpful to learn.

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCall.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Entity that makes a HTTP Request and tests the respose
+ *
+ * @author johnmccabe
+ */
+@ImplementedBy(value = TestHttpCallImpl.class)
+public interface TestHttpCall extends Entity, Startable {
```
ahgittin: feels odd that `EffectorTest` extends `BaseTest` but this has no such marker; some sort of marker would be useful here -- especially if it can inherit `ASSERTIONS` and `TIMEOUT`

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCall.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @SetFromFlag(nullable = false)
+    ConfigKey<String> TARGET_URL = ConfigKeys.newStringConfigKey("url", "URL to test");
+
+    @SetFromFlag(nullable = false)
+    ConfigKey<Map> ASSERTIONS = ConfigKeys.newConfigKey(Map.class, "assert",
+            "Assertions to be evaluated", Maps.newLinkedHashMap());
+
+    ConfigKey<Duration> TIMEOUT = ConfigKeys.newConfigKey(Duration.class, "timeout",
```
ahgittin: as above, inheriting these would be good, or at least defining the constants in an interface which is either extended or referenced, as in

    ConfigKey<Duration> TIMEOUT = TestAssertionsConfig.TIMEOUT;

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCallImpl.java
```diff
@@ -0,0 +1,107 @@
...
+     * See the test/resources directory for examples.
+     *
+     * @param url        The target URL to be tested
+     * @param flags      Passed to {@link org.apache.brooklyn.util.http.HttpAsserts#assertContentEventuallyContainsText(Map, String, String, String...)},
+     *                   {@link org.apache.brooklyn.util.http.HttpAsserts#assertContentEventuallyMatches(Map, String, String)},
+     *                   {@link org.apache.brooklyn.util.http.HttpAsserts#assertHttpStatusCodeEventuallyEquals(Map, String, int)}
+     *
+     * @param assertions The map of assertions
+     */
+    private void checkAssertions(final String url, final Map<String,?> flags, final Map<?, ?> assertions) {
```
ahgittin: this logic should be pulled out to a convenience class, e.g. `TestFrameworkAssertions`

in time i'd like to make it so the set of assertions is extensible using the catalog (this is the work I've been doing with `BrooklynTypeRegister.createBean`)

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCallImpl.java
```diff
@@ -0,0 +1,107 @@
...
+
+        for (final Map.Entry<?, ?> entry : assertions.entrySet()) {
```
ahgittin: this kind of map feels like the wrong syntax as it prevents multiple tests of the same type (e.g. multiple `regex` checks)

i think we should take a list of maps, where the maps should be single-entry.  (if a map is provided we could support the current implementation but with a warning.)  this would also allow us to define more complex assertions which might require multiple settings (e.g. `ignoreCase` for `string` (aka `contains`)

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCallImpl.java
```diff
@@ -0,0 +1,107 @@
...
+            if (Objects.equal(entry.getKey(), "regex")) {
+                LOG.info("Testing if url [{}] matches regex [{}]",
+                        new Object[]{url, entry.getValue()});
+                assertContentEventuallyMatches( flags, url, TypeCoercions.coerce(entry.getValue(), String.class));
+            } else if (Objects.equal(entry.getKey(), "string")) {
```
ahgittin: call this `contains` ?

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCallImpl.java
ahgittin: e.g.

    TestFrameworkAssertions.checkAssertions(Supplier<String> dataSupplier, Map flags, Object assertions)

where `assertions` is preferred as a `List<Map>` but `Map` might also be supported

ahgittin: and the implementation could do a `getPredicate(Map)` which does a lookup on the key(s) in the maps to return e.g. `StringPredicates.containsStringLiteral(xxx)`, which is then checked with a new method `Asserts.assertThatEventually(Map flags, Supplier<T>, Predicate<T>)`

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCall.java
ahgittin: i note the implementation of this one does not permit children; that's probably a useful distinction to call out, and probably cleaner if the assertions in general do not (and they should assert on `init` that they have no children)

usage/test-framework/src/test/java/org/apache/brooklyn/test/framework/TestEntity.java
```diff
@@ -0,0 +1,55 @@
+package org.apache.brooklyn.test.framework;
```
ahgittin: move this entity and related tests to its own sub-package?

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/BaseTest.java
ahgittin: i now see it's very useful to pass the `Entity` when writing in java; we should definitely continue to support this.

usage/test-framework/src/test/java/org/apache/brooklyn/test/framework/TestSensorTest.java
```diff
@@ -0,0 +1,270 @@
...
+
+        //Add Sensor Test for STRING sensor
+        app.createAndManageChild(EntitySpec.create(TestSensor.class)
+                .configure(TestSensor.TARGET_ENTITY, app)
+                .configure(TestSensor.SENSOR_NAME, STRING_SENSOR.getName())
+                .configure(TestSensor.ASSERTIONS, ImmutableMap.of("regex", String.format(".*%s.*", UUID.randomUUID().toString()))));
+
+        try {
+            app.start(ImmutableList.of(loc));
+        } catch (final PropagatedRuntimeException pre) {
```
ahgittin: I've added a few things to `Asserts` to simplify this; you should be able to simplify many of the tests here using:

    try {
        app.start(ImmutableList.of(loc));
        Asserts.shouldHaveFailedPreviously();
    } catch (Throwable t) {
        Asserts.expectedFailureOfType(t, AssertionError.class);
        // and assert the error message has good detail, including the actual and expected values
        // (assuming we set `regexExpected` as ".*"+UUID.randomUUID()+".*" and pass that in above
        Asserts.expectedFailureContainsIgnoreCase(t, "sensor", "regex", /* actual */ "null", /* expected */ regexExpected);
    }

usage/test-framework/src/test/java/org/apache/brooklyn/test/framework/TestSensorTest.java
```diff
@@ -0,0 +1,270 @@
...
+            sensorTestFail = true;
+        } finally {
+            assertThat(sensorTestFail).isTrue().as("regex assert should have failed");
+        }
+    }
+
+    @Test
+    public void testAssertRegexOnNullSensor() {
+        boolean sensorTestFail = false;
+        final String sensorValue = String.format("%s%s%s", UUID.randomUUID().toString(), System.currentTimeMillis(), UUID.randomUUID().toString());
```
ahgittin: btw this should be `null`, or removed (in this test only)

usage/test-framework/src/test/java/org/apache/brooklyn/test/framework/TestSensorTest.java
```diff
@@ -0,0 +1,270 @@
...
+        app.start(ImmutableList.of(loc));
+
+    }
+
+
+    class TestObject {
+        private final String id;
+
+        public TestObject() {
+            id = UUID.randomUUID().toString();
```
ahgittin: FWIW we tend to use `Identifiers.makeRandomId(/* length */ 8)` instead of `UUID` -- no big reason but it makes for more easily distinguishable identifiers, shorter and more chars

usage/test-framework/etc/exampl-catalog.bom
ahgittin: the files in `etc` at present don't look like they are used for anything.

maybe move them to `src/test/resources/test-framework-examples` with a `README.md` in that dir explaining what they are used for.  and maybe in the `qa` project we could run them all as integration tests?

ahgittin:
completed review.  superb start.  as there is more than i first thought including type hierarchy changes i'm inclined to leave this open until you've addressed the ones that you think should be addresed soon.

m4rkmckenna:
@ahgittin Addressed most of the comments.
* Im currently working on the `TestFrameworkAssertions` utility class ... is it ok for this to be a fast follow or would you rather it included?
* @burkechris has already fixed the issues with `ParallelTestCase` ... is it ok if this is addressed in a separate PR?

ahgittin:
LGTM merging

ahgittin:
I look forward to the follow-on PR's.  Moving the root `README.md` into the `docs/guide` is particularly important as that makes it something people can actually find and use!

PR #1000, opened 2015-11-04
HTTP connection pooling for BrooklynApi clients

googlielmo:
This pull request includes the following modifications:

- introduce static factory method to create clients
- use PoolingHttpClientConnectionManager
- deprecate existing BrooklynApi constructors

*Rationale*

After running several long running load tests a port depletion condition was observed, which resulted in a ```java.net.NoRouteToHostException: Cannot assign requested address
        at java.net.PlainSocketImpl.socketConnect(Native Method)``` 
This was due to the BrooklynApi client creating a new connection for each request to the Brooklyn REST APIs.

The `PoolingHttpClientConnectionManager` from Apache httpclient 4.4.1 allows for connection reuse.
Each `BrooklynApi` instance thus maintains a private pool of HTTP connections.

A custom `ClientExecutor` (for cases in which the default pool implementation is not applicable) can be provided by calling a dedicated constructor or by overriding `getClientExecutor()`


usage/rest-client/src/main/java/org/apache/brooklyn/rest/client/BrooklynApiFactory.java
```diff
@@ -0,0 +1,29 @@
...
+ * under the License.
+ */
+package org.apache.brooklyn.rest.client;
+
+import java.net.URL;
+
+/**
+ * Creates {@link BrooklynApi} instances.
+ */
+public interface BrooklynApiFactory {
```
ahgittin: does this class pull its weight, as opposed to having static methods:

    BrooklynApi.newInstance(URL endpoint);
    BrooklynApi.newInstance(URL endpoint, String username, String password);
    BrooklynApi.newInstance(URL endpoint, ClientExecutor clientExecutor);

with a comment to the effect that the first two return pooling executors (ie when would someone not want that?).

@neykov wdyt?

googlielmo: Definitely possible, but then there would be only one static pool (versus one pool per factory).
How would one configure the static pool (max size etc.)?

neykov: Don't think we need the non-pooling variant.
@googlielmo makes a good point re configuration. I think it's a good balance if we keep the DefaultBrooklynApiFactory and remove the interface and the default implementation.
+1 for having non-username-password overload.

usage/rest-client/src/main/java/org/apache/brooklyn/rest/client/BrooklynApi.java
```diff
@@ -114,11 +150,23 @@ public BrooklynApi(String endpoint, @Nullable Credentials credentials) {
         }
     }
 
+    /**
+     * @deprecated since 0.9.0. Use {@link BrooklynApiFactory#getBrooklynApi(java.net.URL, String, String)} instead
+     * @see org.apache.brooklyn.rest.client.DefaultBrooklynApiFactory
+     * @see org.apache.brooklyn.rest.client.PoolingBrooklynApiFactory
+     */
+    @Deprecated
```
neykov: Should keep this one in case users need to fine-tune the executor creation - for example to use their own pooling implementation.

usage/rest-client/src/main/java/org/apache/brooklyn/rest/client/BrooklynApi.java
```diff
@@ -33,7 +33,7 @@
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.Credentials;
 import org.apache.http.auth.UsernamePasswordCredentials;
-import org.apache.http.impl.client.DefaultHttpClient;
+import org.apache.http.impl.client.*;
```
neykov: The convention in Brooklyn is to list all used classes.

usage/rest-client/src/main/java/org/apache/brooklyn/rest/client/BrooklynApiFactory.java
neykov: Could even be static methods on the factory class.

usage/rest-client/src/main/java/org/apache/brooklyn/rest/client/BrooklynApi.java
googlielmo: Will fix (bitten by "Optimize imports" in the IDE)

usage/rest-client/src/main/java/org/apache/brooklyn/rest/client/BrooklynApi.java
googlielmo: Good point. This one is deprecated but the next one is not. The idea was that if you want change implementation then probably you just have to write another factory (and use the next constructor) :)

usage/rest-client/src/main/java/org/apache/brooklyn/rest/client/BrooklynApiFactory.java
ahgittin: why would i want to create two `BrooklynApi` instances sharing a pool?  feels like that use case isn't compelling enough to justify having a factory.  if i wanted pool-sharing behaviour i'd use the same `BrooklynApi` instance.  (if that's not allowed for thread reasons then i'd rather have a wrapper which does make it thread safe, as opposed to having to keep a handle on a factory instance.)

googlielmo: Actually, the threaded situation was the use case I wanted to tackle with that factory.
No problem in switching to a static pool *somewhere* (e.g., in a static factory like the one suggested by @neykov ) – I don't expect users to instantiate more than one factory in any case :-)

usage/rest-client/src/main/java/org/apache/brooklyn/rest/client/BrooklynApi.java
neykov: But it's protected so no access to it?

googlielmo: Yup, it's a little too protective there. What if we change to `protected` and let the user subclass the `BrooklynApi` to provide an alternative implementation?

neykov: Sounds a bit too involved just to construct the object. When in doubt I try to follow the "Keep it simple" moto. Nothing wrong with using it without a factory. The factory just does some sugaring over the parameters.

googlielmo: Then probably it's best to reinstate the previous public constructor and remove the new one. The subclassing is certainly too complicated, especially if we don't follow the factory route.

googlielmo: I heartily agree the factory is just a means to an end, in this case easy configurability (independence from any particular JAX-RS implementation would be another, but it's not the point of this PR).
We can achieve that goal in many ways! I'll do an alternative implementation with the static pool as @ahgittin suggested. ^_^

grkvlt:
Congratulations on the **one thousandth** pull request for Brooklyn @googlielmo :cake: :cake: :cake:

usage/rest-client/src/main/java/org/apache/brooklyn/rest/client/BrooklynApiFactory.java
ahgittin: still not convinced the factory gives us anything over a static `BrooklynApi.newInstance(URI)`

googlielmo:
Here's to another one thousand @grkvlt! :-)

googlielmo:
Latest commit addresses comments.

Each `BrooklynApi` instance now maintains a private pool of HTTP connections.

The `newInstance()` static factory methods now create `BrooklynApi`'s using the connection pool.

A custom `ClientExecutor` (for cases in which the default pool implementation is not applicable) can be provided by calling a dedicated constructor or by overriding `getClientExecutor()`

usage/rest-client/src/main/java/org/apache/brooklyn/rest/client/BrooklynApi.java
```diff
@@ -252,7 +386,7 @@ public AccessApi getAccessApi() {
     }
 
     /**
-     * @deprecated Use {@link #getEntity(Response, GenericType)} instead.
+     * @deprecated since 0.9.0. Use {@link #getEntity(Response, GenericType)} instead.
```
sjcorbett: This method has been deprecated since `0.8.0-incubating`.

ahgittin:
LGTM.  Test failure unrelated (/cc @geomacy it's the `HttpTest`)

geomacy:
Working on that now, per the conversation at https://github.com/apache/incubator-brooklyn/pull/994#issuecomment-154074295

usage/rest-client/src/main/java/org/apache/brooklyn/rest/client/BrooklynApi.java
googlielmo: Whoops! Will fix. Thanks @sjcorbett!

PR #1001, opened 2015-11-05
YAML config keys (parameters)

neykov:
Adds parameters to specs and catalog items, which are then added as config keys to the entity's dynamic type.
What this allows is:
  * document the possible inputs for a catalog item
  * the UI can now ask the user for input when selecting a catalog item
  * setting the correct type for inputs (coercing), validating
  * add config keys to an entity by setting them in the spec

Parameters can be set at any level in the CAMP plan, including in the catalog item meta info. If not set the underlying type parameters are used (if a java class, then its config keys are used, if a catalog item, then the item's parameters). The top level service parameters become the catalog item's parameters.

For referencing any passed parameters new scopes are introduced:
  * `$brooklyn:root()` returns the application entity (top-level one)
  * `$brooklyn:scopeRoot()` returns the top level entity of the current catalog item

Example:
```
brooklyn.catalog:
  id: parameterExample
  version: 0.0.1
  brooklyn.parameters:
  - name: maxHits
    type: integer
    constraints:
    - required
  - config2
  - config3
  item:
    type: BasicApplication
    brooklyn.children:
    - type: entity1
      # can be set at deeper levels when typing is needed
      brooklyn.parameters:
      - name: entity1Config
        type: integer
      brooklyn.config:
        entity1Config: $brooklyn:scopeRoot().config("maxHits")
    - type: entity2
      brooklyn.config:
        entity2Config: $brooklyn:config("maxHits")  # using the inherited config
```

Which results in the following catalog item:

catalog item view

When using
 the Application Wizard to create an application from the above catalog item the UI asks you to fill in the parameters:

screen shot 2015-11-05 at 12 51 35

creating the following app plan:

```
location: uLXQ8KFk
services:
  - type: "parameterExample:0.0.1"
    brooklyn.config:
      maxHits: 55
```


neykov:
Lots of conflicts due to recent changes, working on it. Input & comments still appreciated.

neykov:
Resolved conflicts (Jenkins failure is unrelated). @ahgittin Do you think registry types should have parameters as well, or it's too specific to catalog items and brooklyn objects?

ahgittin:
@neykov I think registry types don't need parameters.  It's down to the individual `TypePlanTransformers` how to transform a plan (eg YAML, but possibly also ZIP etc) into a spec or instance.

ahgittin:
Oh I understand the question now -- should `RegisteredType` expose a `getDefinedConfigKeys()`.  That functionality is wanted for specs to show detail in the catalog.  However I think we can get that simply by the caller creating the spec and then asking the spec -- so still I'd say no.

Have looked over the code briefly and it looks good, nay, great.

The test failure looks unrelated to this PR; seems related to @geomacy's recent addition:

    shouldAssertContentEventuallyMatchesWithFlags(org.apache.brooklyn.util.HttpAssertsTest)
    java.lang.AssertionError: Assertion failed
	at java.net.PlainSocketImpl.socketConnect(Native Method)

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/CampUtils.java
```diff
@@ -23,48 +23,94 @@
...
         } else {
             throw new IllegalStateException("Unable to instantiate YAML; incompatible instantiator "+instantiator+" for "+at);
         }
     }
 
+    public static EntitySpec<? extends Application> createWrapperApp(AssemblyTemplate template, BrooklynClassLoadingContext loader) {
+        BrooklynComponentTemplateResolver resolver = BrooklynComponentTemplateResolver.Factory.newInstance(
+            loader, buildWrapperAppTemplate(template));
+        EntitySpec<Application> wrapperSpec = resolver.resolveSpec(ImmutableSet.<String>of());
+        if (!hasExplicitParams(template)) {
```
ahgittin: why is this needed?

neykov: The spec will always be populated with parameters - either from the template or from the apps's java class (BasicApplication has defaultDisplayName). Clear out the parameters if coming from the class (i.e. the defaults) so they don't overwrite the entity spec when unwrapping.

PR #1002, opened 2015-11-05
Update Tomcat 7 download version

sjcorbett:
Have checked integration tests pass.


sjcorbett:
The test failure is irrelevant to this change.

PR #1003, opened 2015-11-05
Hashicorp Vault external config supplier, plus documentation for external config

rdowner:
Adds an implementation of ExternalConfigSupplier for Hashicorp [Vault](https://www.vaultproject.io).

Also provides documentation for the externalized configuration system.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/config/external/vault/VaultAppIdExternalConfigSupplier.java
+++ b/core/src/main/java/org/apache/brooklyn/core/config/external/vault/VaultAppIdExternalConfigSupplier.java
@@ -1,0 +1,71 @@
+package org.apache.brooklyn.core.config.external.vault;
+
+import com.google.api.client.util.Lists;
+import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableMap;
+import com.google.gson.JsonObject;
+import org.apache.brooklyn.api.mgmt.ManagementContext;
+import org.apache.brooklyn.util.exceptions.Exceptions;
+import org.apache.brooklyn.util.text.Strings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.InetAddress;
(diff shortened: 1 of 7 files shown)


ahgittin:
awesome -- and great docs!

core/src/main/java/org/apache/brooklyn/core/config/external/vault/VaultUserPassExternalConfigSupplier.java
```diff
@@ -0,0 +1,39 @@
+package org.apache.brooklyn.core.config.external.vault;
+
+import com.google.api.client.util.Lists;
```
sjcorbett: This import should be `com.google.common.collect.Lists`.

core/src/main/java/org/apache/brooklyn/core/config/external/vault/VaultAppIdExternalConfigSupplier.java
```diff
@@ -0,0 +1,71 @@
+package org.apache.brooklyn.core.config.external.vault;
+
+import com.google.api.client.util.Lists;
```
sjcorbett: And this one.

core/src/main/java/org/apache/brooklyn/core/config/external/vault/VaultExternalConfigSupplier.java
```diff
@@ -0,0 +1,114 @@
+package org.apache.brooklyn.core.config.external.vault;
+
+import com.google.api.client.util.Lists;
```
sjcorbett: And this one!

ahgittin:
@sjcorbett good spot, i'll push a fix for those

ahgittin:
also adding apache headers!

rdowner:
Argh, rookie mistake... how long have I been here now? Thanks for fixing it up Alex.

PR #1004, opened 2015-11-05
Make the HttpAssertsTest an integration test for now.

geomacy:
Later task will be to restructure to avoid sleeps where possible, see conversation at
https://github.com/apache/incubator-brooklyn/pull/994#issuecomment-154074295.


neykov:
Looks good, will merge as soon as Jenkins confirms it likes the changes.

PR #1005, opened 2015-11-05
tidy of HttpAssertsTest

ahgittin:
@geomacy a few tidies to HttpAssertsTest

i think this should fix the failures we've seen on hosted jenkins, which i think are due to delays cleaning up ports.

however it should remain an integration test except for those which run with no delays (btw i think doing that is low priority -- good integration coverage is sufficient)


geomacy:
Thanks very much @ahgittin!   I shall treat the task of separating any "non-slow" tests from the rest as a low priority, as you say.

PR #1006, opened 2015-11-05
Disable consistently failing integration tests

neykov:
Disable all integration tests that failed consistently in the last five runs.
Lots more failing tests which succeeded at least once, TBD whether will be fixed or disabled.

Introduce new test group `Broken` which means delete the test or fix it.

The idea behind disabling the tests is to make the integration builds useful, currently they are completely ignored. Hard to notice regressions.

Tried some of the tests locally and they passed, could be an environment issue at least for part of them.


neykov:
@hzbarcea All the change does is use the existing `SYSTEM_BUNDLES` in `EmbeddedFelixFramework`. Currently the code is adding items to `EmbeddedFelixFramework.SYSTEM_BUNDLES` and checking for their existence in `Osgis.SYSTEM_BUNDLES` which leads to some integration tests failing.

So while the dependency is still there, at least the extension bundles will be identified correctly until your changes land.

ahgittin:
some of these we really should be able to fix, but agree a "broken" category is useful to not block integration successes.  we have to make sure to review broken tests before a release /cc @aledsage

hzbarcea:
Tested, got ready to commit, @ahgittin beat me to it. We can address the osgi framework issue in a follow-up pr. Thanks @neykov.

PR #1007, opened 2015-11-06
Unwrap camp plan with parent overriding child displayName

neykov:
There's an unexpected side effect presently where pointing to a catalog item and giving a top-level name to the plan will prevent unwrapping. It's surprising to users and even more so because it wasn't the case a few months ago.

For example the following catalog item:

```
brooklyn.catalog:
  id: simpleApp
  version: 1.0
  displayName: Simple App
  itemType: template
  item:
    services:
    - type: my-simple-app
```

when started in the Application Wizard and given a name, equivalent to the following plan:

```
name: My App
services:
  - type: "simpleApp:1.0"
```

will result in this hierarchy:

```
* My App (BasicApplication)
    * Simple App (my-simple-app)
```

I propose that we change the logic such that the root level name will overwrite the child's name, so the hierarchy becomes:

```
* My App (my-simple-app)
```

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/EntityManagementUtils.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/EntityManagementUtils.java
@@ -287,8 +287,10 @@ public class EntityManagementUtils {
         return isWrapperApp(spec) && hasSingleChild(spec) &&
                 //equivalent to no keys starting with "brooklyn."
                 spec.getEnrichers().isEmpty() &&
+                spec.getEnricherSpecs().isEmpty() &&
                 spec.getInitializers().isEmpty() &&
-                spec.getPolicies().isEmpty();
+                spec.getPolicies().isEmpty() &&
+                spec.getPolicySpecs().isEmpty();
     }
 
     public static boolean isWrapperApp(EntitySpec<?> spec) {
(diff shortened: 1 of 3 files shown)


bostko:
@neykov, would you like to add more complicated test like the one bellow.
You should also assert that the name which will be passed when a user selects catalog item, will be assigned to the BasicApplication node.

```
brooklyn.catalog:
  id: org.apache.webcluster.tomcat
  version: 1.0
  description: A Tomcat load-balanced cluster, with MySQL
  displayName: MySQL and Tomcat cluster
  itemType: template

- type: org.apache.brooklyn.entity.software.base.VanillaSoftwareProcess
  defaultDisplayName: Web-cluster
  displayName: extra node
  brooklyn.config:
    childSpec:
      $brooklyn:entitySpec:
        type: org.apache.brooklyn.entity.webapp.ControlledDynamicWebAppCluster
        brooklyn.config:
          initialSize: 1

          memberSpec:
            $brooklyn:entitySpec:
              type: org.apache.brooklyn.entity.webapp.tomcat.TomcatServer
```

ahgittin:
Looks good.  Agree with @bostko a catalog item test would be useful but not essential.

There is one failing test:
* org.apache.brooklyn.camp.brooklyn.AppYamlTest.testWrapsAppIfNameAtTopLevelAndOnApp

neykov:
Added some more catalog tests, also fixed another issue with unwrapping when enrichers/policies are set at root.

iyovcheva:
The changes were successfully tested in vCloud. Everything looks fine.

PR #1009, opened 2015-11-09
added missing URI/URL to String coercions

johnmccabe:
Added coercions to handle cases where a String ConfigKey is being populated from URI/URL typed sensors. For example (where `url` is a `String`, and `main.uri` is a `URI`):

```
url: $brooklyn:component("tomcat").attributeWhenReady("main.uri")
```

*Note*: I can only find Brooklyn entities using either `URI` or `String` to store URLs, have included a `URL` coercion for completeness.

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/util/core/internal/TypeCoercionsTest.java
+++ b/core/src/test/java/org/apache/brooklyn/util/core/internal/TypeCoercionsTest.java
@@ -23,6 +23,9 @@ import static org.testng.Assert.assertNull;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URL;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/brooklyn/util/core/internal/TypeCoercionsTest.java
```diff
@@ -20,9 +20,13 @@
 
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertNull;
+import static org.testng.Assert.fail;
```
neykov: This one doesn't seem to be used, otherwise perfect.

johnmccabe: Fixed and squashed, thanks @neykov

neykov:
Thanks @johnmccabe, merging.

PR #1010, opened 2015-11-09
Prevent catalog recursion in $brooklyn:memberSpec

neykov:
See test case for details, special flags handling was resetting the encountered catalog ids stack.

/cc @aledsage

### DIFF
--- a/usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlAppTest.java
+++ b/usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlAppTest.java
@@ -21,10 +21,18 @@ package org.apache.brooklyn.camp.brooklyn.catalog;
 import org.testng.annotations.Test;
 
 import org.apache.brooklyn.camp.brooklyn.AbstractYamlTest;
+import org.apache.brooklyn.core.mgmt.internal.LocalManagementContext;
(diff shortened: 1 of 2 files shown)


usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlAppTest.java
```diff
@@ -24,50 +24,32 @@
...
-     * We need to use an id that matches something else already on the classpath.
-     * Otherwise we'd get an error telling us "could not resolve item ..." when
-     * attempting to add the initial catalog item.
-     */
-    @Test(groups="WIP") // TODO Fix this!
-    public void testAddCatalogItemWithCircularReference() throws Exception {
-        // Add a catalog item with a circular reference to its own id.
+
+    @Test
+    public void testAddCatalogItemWithMemberSpecCircularReference() throws Exception {
```
ahgittin: test addition is good, but is it worth keeping the note/sketch-test for the case where the circular reference is in a child?

neykov: Don't think so, the problem is in crossing the `memberSpec` boundary, so the changes are just to make the test minimal - being a child is not a factor which triggers the bug.
Don't mind it though, so brought back the original test + note.

ahgittin:
good fix, merging

PR #1012, opened 2015-11-09
Move dependant OSGi projects out of resources

neykov:
Move OSGi test projects from `src/test/resources/dependencies` to `src/test/dependencies`.
The projects' sources shouldn't be on the classpath, only the build artifacts are used for tests. Also Eclipse gives errors, trying to compile them when in resources.
The projects' `outputDirectory` was using a relative path to `src/test/dependencies` anyway so now they will go to the correct destination when installed.

@hzbarcea, @CMoH, wdyt?


ahgittin:
nobody likes these nested projects but they still seem to be the least bad option. this seems to me a good cleanup.

hzbarcea:
Sounds reasonable. I think there are a couple of tests failing in ./usage/camp.

PR #1013, opened 2015-11-10
SshTool choice

aledsage:
Adds SshMachineLocation.SSH_TOOL_CLASS
    
- Deprecates SshTool.PROP_TOOL_CLASS
- Adds tests

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/config/ConfigKeys.java
+++ b/core/src/main/java/org/apache/brooklyn/core/config/ConfigKeys.java
@@ -157,6 +157,14 @@ public class ConfigKeys {
         return newConfigKeyRenamed(prefix+key.getName(), key);
     }
 
+    public static <T> ConfigKey<T> newConfigKeyWithPrefixRemoved(String prefix, ConfigKey<T> key) {
+        if (key.getName().startsWith(prefix)) {
+            return newConfigKeyRenamed(key.getName().substring(prefix.length()), key);
+        } else {
+            throw new IllegalArgumentException("key "+key+" does not start with prefix "+prefix);
+        }
+    }
+
     /** converts the name of the key from one case-strategy (e.g. lowerCamel) to andother (e.g. lower-hyphen) */
     public static <T> ConfigKey<T> convert(ConfigKey<T> key, CaseFormat inputCaseStrategy, CaseFormat outputCaseStrategy) {
         return newConfigKeyRenamed(inputCaseStrategy.to(outputCaseStrategy, key.getName()), key);
--- a/core/src/main/java/org/apache/brooklyn/location/localhost/LocalhostMachineProvisioningLocation.java
+++ b/core/src/main/java/org/apache/brooklyn/location/localhost/LocalhostMachineProvisioningLocation.java
@@ -245,6 +245,8 @@ public class LocalhostMachineProvisioningLocation extends FixedListMachineProvis
     }
     
     public static class LocalhostMachine extends SshMachineLocation implements HasSubnetHostname {
+        private static final Logger LOG = LoggerFactory.getLogger(LocalhostMachine.class);
+
         // declaring this here (as well as on LocalhostMachineProvisioningLocation) because:
         //  1. machine.getConfig(key) will not inherit default value of machine.getParent()'s key
         //  2. things might instantiate a `LocalhostMachine` without going through LocalhostMachineProvisioningLocation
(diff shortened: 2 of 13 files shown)


core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/SshTool.java
```diff
@@ -49,7 +50,14 @@
      * These keys are detected from entity/global config and automatically applied to ssh executions. */
     public static final String BROOKLYN_CONFIG_KEY_PREFIX = "brooklyn.ssh.config.";
     
-    public static final ConfigKey<String> PROP_TOOL_CLASS = newStringConfigKey("tool.class", "SshTool implementation to use", null);
+    /**
+     * @deprecated since 0.9.0; use {@link SshMachineLocation#SSH_TOOL_CLASS}
+     */
+    @Deprecated
+    public static final ConfigKey<String> PROP_TOOL_CLASS = newStringConfigKey(
```
bostko: Can you explain why did you deprecate this?
It is good if this key is in a generic location like `SshTool`. I was thinking about refactoring Winrm support to follow the same pattern as it is for ssh, if so we could implement this dynamic sshTool loading with the same keys and the same API.
In regard with this please check https://github.com/apache/incubator-brooklyn/pull/950

core/src/main/java/org/apache/brooklyn/location/localhost/LocalhostMachineProvisioningLocation.java
```diff
@@ -73,6 +73,7 @@
 
     private static final long serialVersionUID = -7791239672433897762L;
 
+    /** @deprecated since 0.9.0; shouldn't be public */
```
neykov: Add `@Deprecated` annotation as well?

core/src/main/java/org/apache/brooklyn/location/ssh/SshMachineLocation.java
```diff
@@ -135,13 +136,17 @@
 public class SshMachineLocation extends AbstractLocation implements MachineLocation, PortSupplier, WithMutexes, Closeable {
 
     /** @deprecated since 0.7.0 shouldn't be public */
```
neykov: Remove `@deprecated` marker now that these are private.

neykov:
LGTM, minor comments.

iyovcheva:
Looks good to me as well.

core/src/main/java/org/apache/brooklyn/core/config/ConfigKeys.java
```diff
@@ -157,6 +157,14 @@ public static PortAttributeSensorAndConfigKey newPortSensorAndConfigKey(String n
         return newConfigKeyRenamed(prefix+key.getName(), key);
     }
 
+    public static <T> ConfigKey<T> newConfigKeyWithoutPrefix(String prefix, ConfigKey<T> key) {
```
ahgittin: `newConfigKeyWithPrefixRemoved` ?

core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/SshTool.java
ahgittin: agree wtih @bostko about pattern where we share the static config key definition; but this particular key name is horrible so +1 to the deprecation.

ahgittin:
agree with others, minor comments, then good to merge

core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/SshTool.java
aledsage: We need a different key name for the class name of the WinRmTool and the SshTool. We'll use this by writing some configuration on the location (e.g. in brooklyn.properties on a named location), and will want to supply an implementation of WinRmTool and an implementation of SshTool - hence we need a different key name for each.

My reasoning for not having it in SshTool is that the config there is about config the SshTool understands. By the time you have an SshTool, then it has been instantiated. The place that instantiates the tool is SshMachineLocation, so that feels like a good place to define this config key. I also defined it in BrooklynConfigKeys, but with the common prefix "brooklyn.ssh.config.".

aledsage:
Thanks all. I've incorporated this feedback (and added comments to the code about different keys for WinRmTool versus SshTool implementations, and about why the config key is not to be defined in SshTool).

I've also rebased against master.

Waiting for jenkins to confirm that it builds, and then will merge.

PR #1016, opened 2015-11-10
expand assertions and http assertions for string containment, updating rest tests

ahgittin:
/cc @geomacy


ahgittin:
got a +1 at https://github.com/ahgittin/incubator-brooklyn/commit/b46ad724626f70b95b3631e0eeab6b4708341401#commitcomment-14306075 (before i pushed a tiny tweak to the PR); merging

PR #1017, opened 2015-11-10
Flesh out type creation using the new TypeRegistry

ahgittin:
Switch many uses of Catalog to RegisteredType and improve the API around the TypePlanTransformer.  See individual commits for detail.

Next is:
* enable CAMP to use bean creation and register some beans
* swap CAMP to use a new simpler implementation for parsing, enabling reflection-based doc generation and code completion proposals
* persist RegisteredType instances instead of CatalogItems

Note that this builds on #1016 (which should be quick to review) so ignore commit  036ba79.


api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
```diff
@@ -160,5 +172,9 @@ public boolean equals(Object obj) {
     public int hashCode() {
         return Objects.hashCode(getCatalogItemId(), getDisplayName(), getType(), getTags());
     }
+
+    /** strings inserted as flags, config keys inserted as config keys; 
+     * if you want to force one or the other, create a ConfigBag and convert to the appropriate map type */
+    public abstract SpecT configure(Map<?,?> val);
```
neykov: Configuration handling between specs seems identical these days, why not promote all related implementations?

But at least promote abstract `getFlags()/getConfig` as otherwise the javadoc doesn't make much sense, wdyt?

api/src/main/java/org/apache/brooklyn/api/mgmt/classloading/BrooklynClassLoadingContext.java
```diff
@@ -0,0 +1,50 @@
...
+    public <T> Maybe<Class<? extends T>> tryLoadClass(String className, @Nullable Class<T> supertype);
+
+    /** As {@link ClassLoader#getResource(String)} */
+    public URL getResource(String name);
+
+    /**
+     * As {@link ClassLoader#getResources(String)} but returning an {@link Iterable} rather than
+     * an {@link java.util.Enumeration}.
+     */
+    public Iterable<URL> getResources(String name);
```
neykov: Not friendly to return `Iterable`, may be it's the right time to change to `Collection`?

api/src/main/java/org/apache/brooklyn/api/typereg/BrooklynTypeRegistry.java
```diff
@@ -45,24 +45,25 @@
...
     // cause compile errors, not in Eclipse, but in maven (?) 
-    <SpecT extends AbstractBrooklynObjectSpec<?,?>> SpecT createSpec(RegisteredType type, @Nullable RegisteredTypeConstraint optionalConstraint, Class<SpecT> optionalSpecSuperType);
+    <SpecT extends AbstractBrooklynObjectSpec<?,?>> SpecT createSpec(RegisteredType type, @Nullable RegisteredTypeLoadingContext optionalConstraint, Class<SpecT> optionalSpecSuperType);
+    <SpecT extends AbstractBrooklynObjectSpec<?,?>> SpecT createSpecFromPlan(String planFormat, Object planData, @Nullable RegisteredTypeLoadingContext optionalConstraint, Class<SpecT> optionalSpecSuperType);
     
-    // TODO when we support beans
-//    <T> T createBean(RegisteredType type, @Nullable RegisteredTypeConstraint optionalConstraint, Class<T> optionalResultSuperType);
+    <T> T createBean(RegisteredType type, @Nullable RegisteredTypeLoadingContext optionalConstraint, Class<T> optionalResultSuperType);
+    <T> T createBeanFromPlan(String planFormat, Object planData, @Nullable RegisteredTypeLoadingContext optionalConstraint, Class<T> optionalBeanSuperType);
```
neykov: I'd say that all the `createXXX` methods don't belong to the registry. Makes more sense if another piece of code does this, having a reference to the registry. Even `createSpec` on the CI was out of place for me, but as a single method was a 'don't bother' case.

api/src/main/java/org/apache/brooklyn/api/typereg/BrooklynTypeRegistry.java
```diff
@@ -45,24 +45,25 @@
...
+    RegisteredType get(String symbolicName, String version, @Nullable RegisteredTypeLoadingContext constraint);
+    /** as {@link #get(String, String, RegisteredTypeLoadingContext)} with no constraints */
     RegisteredType get(String symbolicName, String version);
-    /** as {@link #get(String, String, RegisteredTypeConstraint)} but allows <code>"name:version"</code> 
+    /** as {@link #get(String, String, RegisteredTypeLoadingContext)} but allows <code>"name:version"</code> 
      * (the {@link RegisteredType#getId()}) in addition to the unversioned name,
      * using a default marker if no version can be inferred */
-    RegisteredType get(String symbolicNameWithOptionalVersion, @Nullable RegisteredTypeConstraint constraint);
-    /** as {@link #get(String, RegisteredTypeConstraint)} but with no constraints */
+    RegisteredType get(String symbolicNameWithOptionalVersion, @Nullable RegisteredTypeLoadingContext constraint);
```
neykov: It's not clear from the javadoc what does the `constraint` filter on `get` methods. Do we allow conflicting types - for example a catalog item and a bean with the same `symbolicName` where `version` is not enough of a discriminator?
   * if yes, then shouldn't the return type be a collection for the case where the `constraint

neykov: It's not clear from the javadoc what does the `constraint` filter on `get` methods. Do we allow conflicting types in the registry - for example a catalog item and a bean with the same `symbolicName` (in which case `version` is not enough of a discriminator)?
  * if yes, then shouldn't the return type be a collection for the case where the `constraint` wasn't strict enough?
  * if no, then why have a constraint at all here, I see it's not used yet in the implementation?

core/src/main/java/org/apache/brooklyn/core/plan/PlanNotRecognizedException.java
```diff
@@ -18,8 +18,14 @@
...
 package org.apache.brooklyn.core.plan;
 
+import org.apache.brooklyn.core.typereg.BrooklynTypePlanTransformer;
+import org.apache.brooklyn.core.typereg.UnsupportedTypePlanException;
+
+/** @deprecated since 0.9.0 use {@link UnsupportedTypePlanException} as part of switch to {@link BrooklynTypePlanTransformer} */
+@Deprecated 
 public class PlanNotRecognizedException extends RuntimeException {
 
+    /** {@link UnsupportedTypePlanException} */
```
neykov: Don't see the point of the comment?

core/src/main/java/org/apache/brooklyn/core/plan/PlanToSpecFactory.java
```diff
@@ -36,12 +40,30 @@
...
+/** @deprecated since 0.9.0 use {@link TypePlanTransformers} as part of switch to {@link BrooklynTypePlanTransformer};
+ * mark transformers as deprecated if there is a preferred corresponding {@link BrooklynTypePlanTransformer} */
+@Deprecated 
 public class PlanToSpecFactory {
     
     private static final Logger log = LoggerFactory.getLogger(PlanToSpecFactory.class);
 
-    private static Collection<PlanToSpecTransformer> getAll() {
-        return ImmutableList.copyOf(ServiceLoader.load(PlanToSpecTransformer.class));
+    private static Collection<PlanToSpecTransformer> getAll(boolean includeDeprecated) {
```
neykov: The `includeDeprecated` filter doesn't have good vibes :). It's all in deprecated functionality so no big deal, but still hope it doesn't stick for too long.

core/src/main/java/org/apache/brooklyn/core/typereg/BrooklynTypePlanTransformer.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.ServiceLoader;
+
+import org.apache.brooklyn.api.typereg.BrooklynTypeRegistry;
+import org.apache.brooklyn.api.typereg.BrooklynTypeRegistry.RegisteredTypeKind;
+import org.apache.brooklyn.api.typereg.RegisteredType;
+import org.apache.brooklyn.api.typereg.RegisteredTypeLoadingContext;
+import org.apache.brooklyn.core.mgmt.ManagementContextInjectable;
+
+/**
+ * Interface for use by schemes which with to be able to transform plans.
```
neykov: "which with" - is something missing?

core/src/main/java/org/apache/brooklyn/core/typereg/BrooklynTypePlanTransformer.java
```diff
@@ -0,0 +1,68 @@
...
+ * <p>
+ * To add a new plan transformation scheme, simply create an implementation and declare it
+ * as a java service (cf {@link ServiceLoader}).
+ * <p>
+ * Implementations may wish to extend {@link AbstractTypePlanTransformer} which simplifies the process.
+ */
+public interface BrooklynTypePlanTransformer extends ManagementContextInjectable {
+
+    /** @return a code to identify type implementations created specifying the use of this plan transformer. */
```
neykov: Sounds too twisted to me. Maybe something like "An identifier for the transformer. Can be used by `RegisteredType` instances to target a specific transformer".

core/src/main/java/org/apache/brooklyn/core/typereg/BrooklynTypePlanTransformer.java
```diff
@@ -0,0 +1,68 @@
...
+    String getFormatCode();
+    /** @return a display name for this transformer. */
+    String getFormatName();
+    /** @return a description for this transformer */
+    String getFormatDescription();
+
+    /** @return how appropriate is this transformer for the {@link RegisteredType#getPlan()} of the type;
+     * 0 (or less) if not, 1 for absolutely, and in some autodetect cases a value between 0 and 1 indicate a ranking.
+     * <p>
+     * The framework guarantees arguments are nonnull, and that the {@link RegisteredType#getPlan()} is also not-null.
```
neykov: Worth annotating the arguments with `@Nonnull`, more visible than the javadoc. The return values for the methods you mention as well.

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/CampTypePlanTransformer.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.brooklyn.core.typereg.AbstractTypePlanTransformer;
+import org.apache.brooklyn.core.typereg.BasicTypeImplementationPlan;
+import org.apache.brooklyn.core.typereg.RegisteredTypes;
+import org.apache.brooklyn.util.guava.Maybe;
+
+import com.google.common.collect.ImmutableList;
+
+public class CampTypePlanTransformer extends AbstractTypePlanTransformer {
+
+    private static final List<String> FORMATS = ImmutableList.of("brooklyn-camp", "camp", "brooklyn");
```
neykov: Why introduce alternative identifiers if they were never used before? It would make sense only to keep backwards compatibility, but not for new functionality.
Better stick to a single one.

core/src/main/java/org/apache/brooklyn/core/typereg/BrooklynTypePlanTransformer.java
```diff
@@ -0,0 +1,68 @@
...
+     * However the format in that plan may be null. */
+    double scoreForType(RegisteredType type, RegisteredTypeLoadingContext context);
```
neykov: Don't like this kind of non-determinism for heuristics - in particular how each implementation needs to keep an eye on the other implementations for determining the right score. Don't have a better suggestion though so will just :mute:.

core/src/main/java/org/apache/brooklyn/core/typereg/AbstractTypePlanTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+                @Override protected Object visitBean(RegisteredType type) {
+                    try { 
+                        return createBean(type, context);
+                    } catch (Exception e) { throw Exceptions.propagate(e); }
+                }
+                
+            }.visit(type), type, context);
+        } catch (UnsupportedTypePlanException e) {
+            // no logging
+            throw Exceptions.propagate(e);
```
neykov: My understanding is that this is just to prevent the next catch from taking over - worth a comment. May be even throwing the same exception instead of calling in to `Exceptions.propagate` to make it more visible (even if the end result is the same).

core/src/main/java/org/apache/brooklyn/core/typereg/AbstractTypePlanTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+        return getFormatCode()+":"+JavaClassNames.simpleClassName(this);
+    }
+    
+    @Override
+    public double scoreForType(RegisteredType type, RegisteredTypeLoadingContext context) {
+        if (getFormatCode().equals(type.getPlan().getPlanFormat())) return 1;
+        if (type.getPlan().getPlanFormat()==null)
+            return scoreForNullFormat(type.getPlan().getPlanData(), type, context);
+        else
+            return scoreForNonmatchingNonnullFormat(type.getPlan().getPlanFormat(), type.getPlan().getPlanData(), type, context);
```
neykov: (Personal preference ahead) I am on the fail-fast side of the spectrum so my thinking is that this kind of "make every last desperate attempt to produce something usable" approach just makes implementations more complex without actually working in the wild.

core/src/main/java/org/apache/brooklyn/core/typereg/AbstractTypePlanTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+    
+    protected <T> T validate(T createdObject, RegisteredType type, RegisteredTypeLoadingContext context) {
+        if (createdObject==null) return null;
+        // TODO validation based on the constraint, throw UnsupportedTypePlanException with details if not matched
+        return createdObject;
+    }
+
+    protected abstract AbstractBrooklynObjectSpec<?,?> createSpec(RegisteredType type, RegisteredTypeLoadingContext context) throws Exception;
```
neykov: `throws Exception` works for tests, but doesn't it mask failure points that should be handled in implementations?

core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
```diff
@@ -90,31 +96,95 @@ public RegisteredType get(String symbolicNameWithOptionalVersion, RegisteredType
...
         }
-        constraint = RegisteredTypeConstraints.extendedWithSpecSuperType(constraint, specSuperType);
+        constraint = RegisteredTypeLoadingContexts.withSpecSuperType(constraint, specSuperType);
 
-        // TODO look up in the actual registry
+        Maybe<Object> result = TypePlanTransformers.transform(mgmt, type, constraint);
+        if (result.isPresent()) return (SpecT) result.get();
         
         // fallback: look up in (legacy) catalog
+        // TODO remove once all transformers are available in the new style
```
neykov: Won't relying on the deprecated transformers make them stick around forever. Why not implement them the recommended way and be done with it?

core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
```diff
@@ -90,31 +96,95 @@ public RegisteredType get(String symbolicNameWithOptionalVersion, RegisteredType
...
         CatalogItem item = (CatalogItem) mgmt.getCatalog().getCatalogItem(type.getSymbolicName(), type.getVersion());
-        return (SpecT) BasicBrooklynCatalog.internalCreateSpecWithTransformers(mgmt, item, constraint.getEncounteredTypes());
+        if (item==null) {
+            // if not in catalog (because loading a new item?) then look up item based on type
+            // (only really used in tests; possibly also for any recursive legacy transformers we might have to create a CI; cross that bridge when we come to it)
+            CatalogItemType ciType = CatalogItemType.ofTargetClass( (Class)constraint.getExpectedJavaSuperType() );
```
neykov: No point in adding code to support tests, better adapt the tests?
Also don't get the recursive transformers comment, can you give an example.

core/src/main/java/org/apache/brooklyn/core/typereg/BrooklynTypePlanTransformer.java
```diff
@@ -0,0 +1,68 @@
...
+     * <p>
+     * The framework guarantees this will only be invoked when {@link #scoreForType(RegisteredType, RegisteredTypeLoadingContext)} 
+     * has returned a positive value.
+     * <p>
+     * Implementations should either return null or throw {@link UnsupportedTypePlanException} 
+     * if the {@link RegisteredType#getPlan()} is not supported. */
+    Object create(RegisteredType type, RegisteredTypeLoadingContext context);
+    
+    double scoreForTypeDefinition(String formatCode, Object catalogData);
+    List<RegisteredType> createFromTypeDefinition(String formatCode, Object catalogData);
```
neykov: Will these ever be used, the registry `xxxFromPlan` methods already wrap the plans in a type?

core/src/main/java/org/apache/brooklyn/core/typereg/BrooklynTypePlanTransformer.java
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Interface for use by schemes which with to be able to transform plans.
+ * <p>
+ * To add a new plan transformation scheme, simply create an implementation and declare it
+ * as a java service (cf {@link ServiceLoader}).
+ * <p>
+ * Implementations may wish to extend {@link AbstractTypePlanTransformer} which simplifies the process.
+ */
+public interface BrooklynTypePlanTransformer extends ManagementContextInjectable {
```
neykov: The name is confusing - it's working on `RegisteredType`s, not `BrooklynType`s?

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
```diff
@@ -66,180 +94,131 @@ public static RegisteredType of(CatalogItem<?, ?> item) {
...
-    public abstract static class AbstractRegisteredType implements RegisteredType {
 
-        final String symbolicName;
-        final String version;
+    /** returns the {@link Class} object corresponding to the given java type name,
+     * using the cache on the type and the loader defined on the type
+     * @param mgmt */
+    @Beta
+    // TODO should this be on the AbstractTypePlanTransformer ?
+    public static Class<?> loadActualJavaType(String javaTypeName, ManagementContext mgmt, RegisteredType type, RegisteredTypeLoadingContext context) throws Exception {
```
neykov: Looks like `throws Exception` is not needed.

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
```diff
@@ -66,180 +94,131 @@ public static RegisteredType of(CatalogItem<?, ?> item) {
...
+        Class<?> result = ((BasicRegisteredType)type).getCache().get(ACTUAL_JAVA_TYPE);
```
neykov: Isn't it better to have a local `WeakHashMap` cache, than assume that the parameter will be of specific type?

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
```diff
@@ -66,180 +94,131 @@ public static RegisteredType of(CatalogItem<?, ?> item) {
...
-        String displayName;
-        String description;
-        String iconUrl;
-        boolean deprecated;
-        boolean disabled;
-
-        // TODO ensure this is re-populated on rebind
-        transient Class<?> javaType;
+        result = CatalogUtils.newClassLoadingContext(mgmt, type, context==null ? null : context.getLoader()).loadClass( javaTypeName );
+        Preconditions.checkNotNull(result, "Could not load class "+javaTypeName+"; returned null (should have thrown a different exception!)");
```
neykov: `loadClass` returns non null or throws - better mark it in its contract than do defensive programming at call sites.

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypeLoadingContexts.java
```diff
@@ -0,0 +1,236 @@
...
+        
+        private BasicRegisteredTypeLoadingContext() {}
+        
+        public BasicRegisteredTypeLoadingContext(@Nullable RegisteredTypeLoadingContext source) {
+            if (source==null) return;
+            
+            this.kind = source.getExpectedKind();
+            this.expectedSuperType = source.getExpectedJavaSuperType();
+            this.encounteredTypes = source.getAlreadyEncounteredTypes();
+            this.loader = (BrooklynClassLoadingContext) source.getLoader();
```
neykov: Cast not needed.

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypeLoadingContexts.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    /** given a {@link BrooklynObject}, returns the spec class which would generate it, for instance returns {@link EntitySpec} given {@link Entity},
+     * or null if not known */
+    static <BO extends BrooklynObject> Class<? extends AbstractBrooklynObjectSpec<?,?>> lookupSpecTypeForTarget(Class<BO> targetSuperType) {
+        if (targetSuperType==null) return null;
+        BrooklynObjectType best = null;
+
+        for (BrooklynObjectType t: BrooklynObjectType.values()) {
+            if (t.getSpecType()==null) continue;
```
neykov: Should be `getInterfaceType()`

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypeLoadingContexts.java
```diff
@@ -0,0 +1,236 @@
...
+            if (!t.getInterfaceType().isAssignableFrom(targetSuperType)) continue;
+            // on equality, exit immediately
+            if (t.getInterfaceType().equals(targetSuperType)) return t.getSpecType();
+            // else pick which is best
+            if (best==null) { best = t; continue; }
+            // if t is more specific, it is better (handles case when e.g. a Policy is a subclass of Entity)
+            if (best.getSpecType().isAssignableFrom(t.getSpecType())) { best = t; continue; }
```
neykov: Again should be `getInterfaceType()`

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypePredicates.java
```diff
@@ -109,36 +109,46 @@ public boolean apply(@Nullable RegisteredType item) {
         }
     }
 
-    public static <T> Predicate<RegisteredType> javaType(final Predicate<Class<T>> filter) {
-        return new JavaTypeMatches(filter);
+    public static <T> Predicate<RegisteredType> anySuperType(final Predicate<Class<T>> filter) {
```
neykov: Generic type is not used, replace with wildcard.

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypePredicates.java
```diff
@@ -109,36 +109,46 @@ public boolean apply(@Nullable RegisteredType item) {
...
+        return anySuperType((Predicate)Predicates.assignableFrom(filter));
     }
     
-    private static class JavaTypeMatches implements Predicate<RegisteredType> {
+    private static class AnySuperTypeMatches implements Predicate<RegisteredType> {
         private final Predicate<Class<?>> filter;
         
         @SuppressWarnings({ "rawtypes", "unchecked" })
-        private <T> JavaTypeMatches(Predicate filter) {
+        private <T> AnySuperTypeMatches(Predicate filter) {
```
neykov: Generic type not used, remove.

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypePredicates.java
```diff
@@ -109,36 +109,46 @@ public boolean apply(@Nullable RegisteredType item) {
...
+                    if (apply((RegisteredType)o)) return true;
+                }
+            }
+            return false;
         }
     }
 
-    public static final Predicate<RegisteredType> IS_APPLICATION = javaTypeAssignableFrom(Application.class);
+    public static final Predicate<RegisteredType> IS_APPLICATION = assignableFrom(Application.class);
     // TODO do we need this?  introduced already deprecated in 0.9.0 so can be removed, or enabled
```
neykov: +1 to remove

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
```diff
@@ -66,180 +94,131 @@ public static RegisteredType of(CatalogItem<?, ?> item) {
...
+            ((BasicRegisteredType)type).superTypes.add(superType);
         }
+        return type;
+    }
 
-        @Override
-        public String getDisplayName() {
-            return displayName;
-        }
+    /** returns the implementation data for a spec if it is a string (e.g. plan yaml or java class name); else false */
```
neykov: ; else null

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
```diff
@@ -66,180 +94,131 @@ public static RegisteredType of(CatalogItem<?, ?> item) {
...
+    @Beta
+    public static String getImplementationDataStringForSpec(RegisteredType item) {
+        if (item==null || item.getPlan()==null) return null;
+        Object data = item.getPlan().getPlanData();
+        if (!(data instanceof String)) return null;
```
neykov: Looking at the usages better throw in this case.

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
```diff
@@ -66,180 +94,131 @@ public static RegisteredType of(CatalogItem<?, ?> item) {
...
-        public String toString() {
-            return JavaClassNames.simpleClassName(this)+"["+getId()+
-                (isDisabled() ? ";DISABLED" : "")+
-                (isDeprecated() ? ";deprecated" : "")+
-                "]";
-        }
+    /** Returns a wrapped map, if the object is YAML which parses as a map; 
+     * otherwise returns absent capable of throwing an error with more details */
+    @SuppressWarnings("unchecked")
+    public static Maybe<Map<Object,Object>> getAsYamlMap(Object planData) {
```
neykov: Return type should be `Maybe<Map<?,?>>` (it's a read-only container), callers are not supposed to mutate it.

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
```diff
@@ -66,180 +94,131 @@ public static RegisteredType of(CatalogItem<?, ?> item) {
...
         }
-
-        public TypeImplementation getImplementation() {
-            return impl;
+        for (Object st: allKnownTypes) {
+            if (st instanceof RegisteredType) {
+                if (isSubTypeOf((RegisteredType)st, superType)) return true;
+            }
         }
+        return false;
```
neykov: Looks the same as `RegisteredTypePredicates.assignableFrom(Class<?>)`? Re-use implementation and unify names.

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypeLoadingContexts.java
```diff
@@ -0,0 +1,236 @@
...
+import org.apache.brooklyn.util.collections.MutableSet;
+import org.apache.brooklyn.util.javalang.JavaClassNames;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.ImmutableSet;
+
+public class RegisteredTypeLoadingContexts {
+
+    private static final Logger log = LoggerFactory.getLogger(RegisteredTypeLoadingContexts.BasicRegisteredTypeLoadingContext.class);
```
neykov: Weird logger, is it intentional?

core/src/main/java/org/apache/brooklyn/core/typereg/TypePlanTransformers.java
```diff
@@ -0,0 +1,160 @@
...
+    private static final Logger log = LoggerFactory.getLogger(TypePlanTransformers.class);
+
+    private static Collection<BrooklynTypePlanTransformer> getAll() {
+        return ImmutableList.copyOf(ServiceLoader.load(BrooklynTypePlanTransformer.class));
+    }
+
+    private static Collection<Class<? extends BrooklynTypePlanTransformer>> OVERRIDE;
+    @SafeVarargs
+    @VisibleForTesting
+    public synchronized static void forceAvailable(Class<? extends BrooklynTypePlanTransformer> ...classes) {
```
neykov: I like the idea of tests not using the `ServiceLoader` mechanism. Suggest passing instances instead.

core/src/main/java/org/apache/brooklyn/core/typereg/TypePlanTransformers.java
```diff
@@ -0,0 +1,160 @@
...
+        OVERRIDE = Arrays.asList(classes);
+    }
+    public synchronized static void clearForced() {
+        OVERRIDE = null;
+    }
+
+    public static Collection<BrooklynTypePlanTransformer> all(ManagementContext mgmt) {
+        // TODO cache these in the TypeRegistry, looking for new ones periodically or supplying a way to register them
```
neykov: Should switch to an OSGi solution instead of trying to improve this longer term.

core/src/main/java/org/apache/brooklyn/core/typereg/TypePlanTransformers.java
```diff
@@ -0,0 +1,160 @@
...
+    public static List<BrooklynTypePlanTransformer> forType(ManagementContext mgmt, RegisteredType type, RegisteredTypeLoadingContext constraint) {
+        Multimap<Double,BrooklynTypePlanTransformer> byScoreMulti = ArrayListMultimap.create(); 
+        Collection<BrooklynTypePlanTransformer> transformers = all(mgmt);
+        for (BrooklynTypePlanTransformer transformer : transformers) {
+            double score = transformer.scoreForType(type, constraint);
+            if (score>0) byScoreMulti.put(score, transformer);
+        }
+        Map<Double, Collection<BrooklynTypePlanTransformer>> tree = new TreeMap<Double, Collection<BrooklynTypePlanTransformer>>(byScoreMulti.asMap());
+        List<Collection<BrooklynTypePlanTransformer>> highestFirst = new ArrayList<Collection<BrooklynTypePlanTransformer>>(tree.values());
+        Collections.reverse(highestFirst);
```
neykov: Could pass `Collections.reverseOrder()` to the `TreeMap` instead.

core/src/main/java/org/apache/brooklyn/core/typereg/TypePlanTransformers.java
```diff
@@ -0,0 +1,160 @@
...
+        return MutableList.copyOf(Iterables.concat(highestFirst)).asUnmodifiable();
```
neykov: Could use `ImmutableList.copyOf()` instead.
Don't understand the usage of `concat` here.

core/src/test/java/org/apache/brooklyn/core/catalog/internal/StaticTypePlanTransformer.java
```diff
@@ -0,0 +1,115 @@
...
+public class StaticTypePlanTransformer extends AbstractTypePlanTransformer {
+    
+    public StaticTypePlanTransformer() {
+        super("static-types", "Static Type", "Static transformer for use in tests");
+    }
+
+    private static final Map<String, AbstractBrooklynObjectSpec<?, ?>> REGISTERED_SPECS = new ConcurrentHashMap<>();
+
+    public static void forceInstall() {
+        TypePlanTransformers.forceAvailable(StaticTypePlanTransformer.class, JavaClassNameTypePlanTransformer.class);
```
neykov: How is this class related to `JavaClassNameTypePlanTransformer`?

core/src/test/java/org/apache/brooklyn/core/catalog/internal/SpecParameterInMetaTest.java
```diff
@@ -43,18 +49,40 @@
...
+        PlanToSpecFactory.clearForced();
+    }
+
+    @Test
+    public void testCanRetrieveWithNew() {
+        AbstractBrooklynObjectSpec<?, ?> spec = mgmt.getTypeRegistry().createSpecFromPlan(null, specId, null, null);
+        Assert.assertNotNull(spec);
+    }
+
+    // it's not actually added to the catalog; probably it would be cleaner if it is;
```
neykov: Being able to create a spec from a plan is different from adding a catalog item so don't agree, it's a separate thing. The mechanism could be used for application specs as well, it's not specific to the catalog.

Could add a utility method somewhere to add a catalog item for the registered spec, but not useful for the following tests.

core/src/test/java/org/apache/brooklyn/core/catalog/internal/SpecParameterInMetaTest.java
```diff
@@ -43,18 +49,40 @@
...
+    // but for now when we resolve in PlanToSpecFactory we make explicit reference to StaticTypePlanTransformer
```
neykov: Don't get this one.

core/src/test/java/org/apache/brooklyn/core/catalog/internal/SpecParameterInMetaTest.java
```diff
@@ -43,18 +49,40 @@
...
-    private String spec;
+    private String specId;
 
     @BeforeMethod(alwaysRun=true)
     public void setUp() {
         mgmt = LocalManagementContextForTests.newInstanceWithOsgi();
         catalog = mgmt.getCatalog();
-        spec = TestToSpecTransformer.registerSpec(EntitySpec.create(BasicEntity.class));
+        StaticTypePlanTransformer.forceInstall();
+        PlanToSpecFactory.forceAvailable(TestToSpecTransformer.class, JavaCatalogToSpecTransformer.class);
```
neykov: Part of the tests depend on StaticTypePlanTransformer, other part on PlanToSpecFactory, but not at the same time? If so it would be nice to split them in separate files so the dependencies are clear.

core/src/test/java/org/apache/brooklyn/core/catalog/internal/TestToSpecTransformer.java
```diff
@@ -20,43 +20,36 @@
...
- *    id: test.inputs
- *    version: 0.0.1
- *    item: <specId>
- * } </pre>
+ * For use in conjunction with {@link StaticTypePlanTransformer};
+ * this will lookup an item by ID or in a map "type: id".
+ * <p>
+ * Should be removed when catalog is able to function using new-style lookups.
  */
 public class TestToSpecTransformer implements PlanToSpecTransformer {
```
neykov: Still can't wrap my head around the changes in this file. `StaticTypePlanTransformer` doesn't seem to be involved at all where this is used.
Looks like it's just a minimalistic CAMP parser, delegating to the registry to resolve the type. It's used in a single test - not worth it's weight (especially implementing deprecated functionality).

core/src/test/java/org/apache/brooklyn/core/plan/XmlPlanToSpecTransformer.java
```diff
@@ -41,6 +41,8 @@
 /** Example implementation of {@link PlanToSpecTransformer} showing 
  * how implementations are meant to be written. */
 public class XmlPlanToSpecTransformer implements PlanToSpecTransformer {
+    
+    // this is REPLACED by ExampleXmlTypePlanTransformer
```
neykov: Just delete this one?

core/src/test/java/org/apache/brooklyn/core/test/BrooklynMgmtUnitTestSupport.java
```diff
@@ -0,0 +1,58 @@
...
+            mgmt = LocalManagementContextForTests.newInstance();
+        }
+    }
+
+    @AfterMethod(alwaysRun=true)
+    public void tearDown() throws Exception {
+        try {
+            if (mgmt != null) Entities.destroyAll(mgmt);
+        } catch (Throwable t) {
+            LOG.error("Caught exception in tearDown method", t);
```
neykov: I see you just copied this, but still rings an alarm in my head - aren't exceptions in tearDown part of the test failing as well? No strong opinion though.

core/src/test/java/org/apache/brooklyn/core/typereg/JavaTypePlanTransformerTest.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+package org.apache.brooklyn.core.typereg;
+
+import org.apache.brooklyn.api.typereg.RegisteredType;
+import org.apache.brooklyn.core.test.BrooklynMgmtUnitTestSupport;
+import org.testng.Assert;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+public class JavaTypePlanTransformerTest extends BrooklynMgmtUnitTestSupport {
```
neykov: Tiny, test class name should follow class being tested.

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/api/AssemblyTemplateSpecInstantiator.java
```diff
@@ -23,19 +23,23 @@
...
+    @Deprecated /** @deprecaed since 0.9.0 include encountered types */
+    EntitySpec<? extends Application> createApplicationSpec(AssemblyTemplate template, CampPlatform platform, BrooklynClassLoadingContext loader);
+    
     /**
      * Gets the single item returned by {@link #createServiceSpecs}
      * and wraps it in an Application if needed, applying top-level
      * attributes and locations to the root entity.
      */
-    EntitySpec<? extends Application> createApplicationSpec(AssemblyTemplate template, CampPlatform platform, BrooklynClassLoadingContext loader);
+    EntitySpec<? extends Application> createApplicationSpec(AssemblyTemplate template, CampPlatform platform, BrooklynClassLoadingContext loader, Set<String> encounteredCatalogTypes);
```
neykov: Doesn't look right, aren't application specs always top level?

On another note, I haven't removed this method just for backwards compatibility. It belongs to an utils class with the `createServiceSpecs` doing the heavy lifting.

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/api/AssemblyTemplateSpecInstantiator.java
```diff
@@ -23,19 +23,23 @@
...
+import org.apache.brooklyn.api.mgmt.classloading.BrooklynClassLoadingContext;
 import org.apache.brooklyn.camp.CampPlatform;
 import org.apache.brooklyn.camp.spi.AssemblyTemplate;
 import org.apache.brooklyn.camp.spi.instantiate.AssemblyTemplateInstantiator;
-import org.apache.brooklyn.core.mgmt.classloading.BrooklynClassLoadingContext;
 
 public interface AssemblyTemplateSpecInstantiator extends AssemblyTemplateInstantiator {
 
+    @Deprecated /** @deprecaed since 0.9.0 include encountered types */
+    EntitySpec<? extends Application> createApplicationSpec(AssemblyTemplate template, CampPlatform platform, BrooklynClassLoadingContext loader);
```
neykov: No point in keeping this method. There will be compile errors anyway because of the new one and it doesn't even get called (false sense of security).

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/CampResolver.java
```diff
@@ -0,0 +1,144 @@
...
+                .add(item.getSymbolicName())
+                .build();
+        } else {
+            encounteredTypes = parentEncounteredTypes;
+        }
+
+        AbstractBrooklynObjectSpec<?, ?> spec;
+        String planYaml = RegisteredTypes.getImplementationDataStringForSpec(item);
+        MutableSet<Object> supers = MutableSet.copyOf(item.getSuperTypes());
+        supers.addIfNotNull(expectedType);
```
neykov: Check `item` and `expectedType` compatibility?

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/CampResolver.java
```diff
@@ -0,0 +1,144 @@
...
+
+    /** whether to allow parsing of the legacy 'full' syntax, 
+     * where a non-application items are wrapped:
+     * <li> in a "services:" block for entities,
+     * <li> in a "brooklyn.locations" or "brooklyn.policies" block for locations and policies */
+    boolean allowLegacyFullSyntax = true;
+
+    /** whether to allow parsing of the type syntax, where an item is a map with a "type:" field,
+     * i.e. not wrapped in any "services:" or "brooklyn.{locations,policies}" block */
+    boolean allowTypeSyntax = true;
```
neykov: Are these a part of future changes, no used at the moment?

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/CampResolver.java
```diff
@@ -0,0 +1,144 @@
...
+    private static EntitySpec<?> createEntitySpecFromServicesBlock(String plan, BrooklynClassLoadingContext loader, Set<String> encounteredTypes, boolean isApplication) {
+        CampPlatform camp = CampInternalUtils.getCampPlatform(loader.getManagementContext());
+
+        AssemblyTemplate at = CampInternalUtils.registerDeploymentPlan(plan, loader, camp);
+        AssemblyTemplateInstantiator instantiator = CampInternalUtils.getInstantiator(at);
+        if (instantiator instanceof AssemblyTemplateSpecInstantiator) {
+            EntitySpec<? extends Application> appSpec = ((AssemblyTemplateSpecInstantiator)instantiator).createApplicationSpec(at, camp, loader, encounteredTypes);
+
+            if (!isApplication && EntityManagementUtils.canPromoteChildrenInWrappedApplication(appSpec) && appSpec.getChildren().size()==1) {
+                CampInternalUtils.resetSpecIfTemplateHasNoExplicitParameters(at, appSpec);
```
neykov: Already happened in `createWrapperApp`

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/CampResolver.java
```diff
@@ -0,0 +1,144 @@
...
+ 
+    private static EntitySpec<?> createEntitySpecFromServicesBlock(String plan, BrooklynClassLoadingContext loader, Set<String> encounteredTypes, boolean isApplication) {
+        CampPlatform camp = CampInternalUtils.getCampPlatform(loader.getManagementContext());
+
+        AssemblyTemplate at = CampInternalUtils.registerDeploymentPlan(plan, loader, camp);
+        AssemblyTemplateInstantiator instantiator = CampInternalUtils.getInstantiator(at);
+        if (instantiator instanceof AssemblyTemplateSpecInstantiator) {
+            EntitySpec<? extends Application> appSpec = ((AssemblyTemplateSpecInstantiator)instantiator).createApplicationSpec(at, camp, loader, encounteredTypes);
+
+            if (!isApplication && EntityManagementUtils.canPromoteChildrenInWrappedApplication(appSpec) && appSpec.getChildren().size()==1) {
```
neykov: `appSpec.getChildren().size()==1` already included in the `canPromoteChildrenInWrappedApplication` check.
Also (minor) can move this block in a wrapper function instead of using a flag argument.

usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogParametersTest.java
```diff
@@ -273,7 +273,54 @@ public void testAppSpecInheritsCatalogParameters() {
...
+        assertEquals(c1.config().get(refConfigKey), testValue);
+        Entity c2 = childIter.next();
+        assertEquals(c2.config().get(refConfigKey), testValue);
+        assertFalse(childIter.hasNext());
+    }
+    
+    // XXX TODO parameters on the root don't work with new type registry; 
+    // they require the CI being able to keep them,
+    // or else modifying the plan. TODO should they be permitted as metadata in this way?
+    // or treaded like a declaration of config keys on the entity?  i (alex) prefer the latter.
```
neykov: Do you plan on supporting this short term? If not remove/update mention in the catalog docs.

I have a slight preference for parameters in meta, but don't mind removing support.

utils/common/src/main/java/org/apache/brooklyn/util/text/Identifiers.java
```diff
@@ -26,6 +26,9 @@
     
     public static final String JAVA_GOOD_START_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz_";
     public static final String JAVA_GOOD_NONSTART_CHARS = JAVA_GOOD_START_CHARS+"1234567890";
+    public static final String JAVA_SEGMENT_REGEX = "["+JAVA_GOOD_START_CHARS+"]"+"["+JAVA_GOOD_NONSTART_CHARS+"]*";
+    public static final String JAVA_PACKAGE_OR_CLASS_REGEX = "("+JAVA_SEGMENT_REGEX+"\\."+")*"+JAVA_SEGMENT_REGEX;
+    public static final String JAVA_BINARY_REGEX = JAVA_PACKAGE_OR_CLASS_REGEX+"(\\$["+JAVA_GOOD_NONSTART_CHARS+"]+)*";
```
neykov: This includes a tiny subset of valid java identifiers, see http://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.8. Could lead to subtle bugs.

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
```diff
@@ -66,180 +94,131 @@ public static RegisteredType of(CatalogItem<?, ?> item) {
...
+    /** Preferred mechanism for defining a bean {@link RegisteredType} */
+    public static RegisteredType bean(String symbolicName, String version, TypeImplementationPlan plan, @Nullable Class<?> superType) {
+        return addSuperType(new BasicRegisteredType(RegisteredTypeKind.BEAN, symbolicName, version, plan), superType);
     }
     
-    public static RegisteredTypeKind getKindOf(RegisteredType type) {
-        return new RegisteredTypeKindVisitor<RegisteredTypeKind>() {
-            @Override protected RegisteredTypeKind visitSpec(RegisteredSpecType type) { return RegisteredTypeKind.SPEC; }
-        }.visit(type);
+    public static RegisteredType spec(String symbolicName, String version, TypeImplementationPlan plan, @Nullable Class<?> superType) {
```
neykov: `symbolicName`, `version` are `@Nullable` as well

api/src/main/java/org/apache/brooklyn/api/typereg/BrooklynTypeRegistry.java
```diff
@@ -45,24 +45,25 @@
...
+    /** as {@link #get(String, RegisteredTypeLoadingContext)} but with no constraints */
     RegisteredType get(String symbolicNameWithOptionalVersion);
 
     // NB the seemingly more correct generics <T,SpecT extends AbstractBrooklynObjectSpec<T,SpecT>> 
     // cause compile errors, not in Eclipse, but in maven (?) 
-    <SpecT extends AbstractBrooklynObjectSpec<?,?>> SpecT createSpec(RegisteredType type, @Nullable RegisteredTypeConstraint optionalConstraint, Class<SpecT> optionalSpecSuperType);
+    <SpecT extends AbstractBrooklynObjectSpec<?,?>> SpecT createSpec(RegisteredType type, @Nullable RegisteredTypeLoadingContext optionalConstraint, Class<SpecT> optionalSpecSuperType);
+    <SpecT extends AbstractBrooklynObjectSpec<?,?>> SpecT createSpecFromPlan(String planFormat, Object planData, @Nullable RegisteredTypeLoadingContext optionalConstraint, Class<SpecT> optionalSpecSuperType);
```
neykov: `planFormat` is `@Nullable`, the optional arguments obviously as well.

core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
```diff
@@ -90,31 +96,95 @@ public RegisteredType get(String symbolicNameWithOptionalVersion, RegisteredType
...
 
-        // TODO look up in the actual registry
+        Maybe<Object> result = TypePlanTransformers.transform(mgmt, type, constraint);
+        if (result.isPresent()) return (SpecT) result.get();
         
         // fallback: look up in (legacy) catalog
+        // TODO remove once all transformers are available in the new style
         CatalogItem item = (CatalogItem) mgmt.getCatalog().getCatalogItem(type.getSymbolicName(), type.getVersion());
-        return (SpecT) BasicBrooklynCatalog.internalCreateSpecWithTransformers(mgmt, item, constraint.getEncounteredTypes());
+        if (item==null) {
```
neykov: You are mixing two types of transformations in this method:
  * plan (i.e. camp) to spec transformers - `BrooklynTypePlanTransformer` (former `PlanToSpecTransformer), the part up to this line
  * type (symbolicName, java identifier) to spec transformers - `EntitySpecResolver`,  the part from this line until the end. This type of transformation is not needed at this level, it shoud be nested in the plan transformers. Looks like it's needed here just to get `SpecParameterInMetaTest` from passing.

Instead of this second part, `StaticTypePlanTransformer` should be taught how to handle the plans from the test (essentially same as pre-changes `TestToSpecTransformer`). The commit https://github.com/neykov/incubator-brooklyn/commit/8e9d69dbaeb6b15f8351f4bf5f2928c379a86b3d shows what I mean. The test `testOsgiClassScanned` is moved to `brooklyn-camp` as the camp parser is needed to create specs out of the catalog items.
With these changes the legacy transformers are used only during catalog `addItems`. Should be easy to move to the new API, I suppose part of the next PR.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogUtils.java
```diff
@@ -85,16 +89,23 @@ public static BrooklynClassLoadingContext getClassLoadingContext(Entity entity)
...
+    public static BrooklynClassLoadingContext newClassLoadingContext(@Nullable ManagementContext mgmt, String catalogItemId, Collection<? extends OsgiBundleWithUrl> libraries, BrooklynClassLoadingContext loader) {
         BrooklynClassLoadingContextSequential result = new BrooklynClassLoadingContextSequential(mgmt);
 
         if (libraries!=null && !libraries.isEmpty()) {
             result.add(new OsgiBrooklynClassLoadingContext(mgmt, catalogItemId, libraries));
         }
 
-        BrooklynClassLoadingContext loader = BrooklynLoaderTracker.getLoader();
-        if (loader != null) {
+        if (loader !=null) {
```
neykov: Shouldn't be allowed to stack loaders. A new loader barrier starts at a catalog item and the same loader is used for all types in the item, including specs referenced in URLs.
A loader delegates to a single item's osgi bundles, falling back to the catalog class loader. The root application loader uses only the catalog class loader.
I believe that even adding the thread local loader is wrong (in the sense that shouldn't be needed, but added just in case).

neykov:
Finished review, the changes look great and functionally correct. No need to reply to the copious amount of comments, just take what you think is useful.
The last two comments deserve special attention though as they touch on the high level interplay of the components.

api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
ahgittin: you're right, that simplifies things -- have promoted all flag/config methods, and done related tidy

api/src/main/java/org/apache/brooklyn/api/mgmt/classloading/BrooklynClassLoadingContext.java
ahgittin: what's unfriendly about `Iterable`?  personally i ilke it, and it means implementations are more efficient.  (it's better than `Enumeration` which java's `ClassLoader` returns, and presumably implementations there support lazy eval.)

api/src/main/java/org/apache/brooklyn/api/typereg/BrooklynTypeRegistry.java
ahgittin: Interesting point.  Logically they could be moved.  But where would we move them to?  Would we have another `managementContext.getTypeInstantiator()` which uses the `TypeRegistry` ?  That doesn't seem so bad.  Let's think on this.

api/src/main/java/org/apache/brooklyn/api/typereg/BrooklynTypeRegistry.java
ahgittin: You're right.

In general it serves a few purposes:
* it supplies class-loading context
* it helps with recursive resolution (`alreadyEncounteredType`)
* it prevents returning incompatible classes

But none of them are needed for a lookup.  Maybe we should remove; I'll look at that.

BTW correct that we don't allow different instances of the same symbolic name+version.  However i have in my mind a concept of `aliases` which types could declare, and those could conflict, so if looking up an alias the context could be useful.

core/src/main/java/org/apache/brooklyn/core/plan/PlanToSpecFactory.java
ahgittin: +1

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/CampTypePlanTransformer.java
ahgittin: agree, for now

core/src/main/java/org/apache/brooklyn/core/typereg/BrooklynTypePlanTransformer.java
ahgittin: have expanded comments to try to help determine right score.  if we store formats right this won't be used very often (or possibly even ever in which case we can get rid of it)

core/src/main/java/org/apache/brooklyn/core/typereg/AbstractTypePlanTransformer.java
ahgittin: correct.  have combined, with type check to keep same logging behaviour.

core/src/main/java/org/apache/brooklyn/core/typereg/AbstractTypePlanTransformer.java
ahgittin: That wasn't the intention.  The idea is to simplify subclass implementations which typically do different things if the format is specified.  Have added javadoc to the class.

core/src/main/java/org/apache/brooklyn/core/typereg/AbstractTypePlanTransformer.java
ahgittin: there are so many exceptions when working with classes and loading, i don't think that's likely.  happy to revert in time however; i figure this will be beta for a few releases.

core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
ahgittin: that's the plan.  but downstream projects may have them so keep them for at least a release or two.

core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
ahgittin: this is a port of old logic; the TODO above notes we want to remove the block

core/src/main/java/org/apache/brooklyn/core/typereg/BrooklynTypePlanTransformer.java
ahgittin: TBD.  My thinking is that this is where `brooklyn.catalog` syntax could be handled.  But it is very much a beta placeholder until we have implementations.

There is duplication with `BrooklynTypeRegistery.createXxxFromPlan` which will naturally be resolved as part of that work.

core/src/main/java/org/apache/brooklyn/core/typereg/BrooklynTypePlanTransformer.java
ahgittin: It's unfortunate we have `BrooklynType`.  I've proposed in comment there that *that* be renamed.

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
ahgittin: interesting idea.  have added `@Beta // TODO` on `getCache()` suggesting that, if it ends up that the cache is not wanted elsewhere

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypeLoadingContexts.java
ahgittin: good spots

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypePredicates.java
ahgittin: i don't think nested generic wildcard works like that -- `Predicate<Class<?>>` does not match any `Predicate<Class<T>>`.

core/src/main/java/org/apache/brooklyn/core/typereg/TypePlanTransformers.java
ahgittin: I liked the idea of instances but because instances need the management context injected that would prevent their use for rebind tests.

core/src/main/java/org/apache/brooklyn/core/typereg/TypePlanTransformers.java
ahgittin: +1

core/src/main/java/org/apache/brooklyn/core/typereg/TypePlanTransformers.java
ahgittin: agree immutable.
concat is to flatten the (collection) entries in the treemap.
reverse as an explicit line makes the logic clearer imho.

core/src/test/java/org/apache/brooklyn/core/catalog/internal/StaticTypePlanTransformer.java
ahgittin: different.  this one requires the caller to provide the specs.  `JCNTPT` does a `Class.forName(typeName)` to create the spec.  added javadoc to this effect, plus a few other cleanups (clears the `REGISTERED_SPECS` map!)

core/src/test/java/org/apache/brooklyn/core/catalog/internal/SpecParameterInMetaTest.java
ahgittin: I don't quite follow.  In general if a plan refers to a type, I'd expect that type in the catalog (or a java class).  While a transformer can define other rules for instantiating types, I'm not sure that's good practice.  (Except it's okay for tests.)

(have captured this discussion in the code)

core/src/test/java/org/apache/brooklyn/core/catalog/internal/SpecParameterInMetaTest.java
ahgittin: Intention is to get rid of the `TestToSpecTransformer` but we need a bit more machinery first.  Have added TODO to that effect.

core/src/test/java/org/apache/brooklyn/core/catalog/internal/TestToSpecTransformer.java
ahgittin: It's a good exercise to make this work.  Should be cleaner when we're completely migrated.

core/src/test/java/org/apache/brooklyn/core/plan/XmlPlanToSpecTransformer.java
ahgittin: when we remove the PlanToSpecTransformer, yes

core/src/test/java/org/apache/brooklyn/core/test/BrooklynMgmtUnitTestSupport.java
ahgittin: // we should fail here, except almost always that masks a primary failure in the test itself,
            // so it would be extremely unhelpful to do so. if we could check if test has not already failed,
            // that would be ideal, but i'm not sure if that's possible with TestNG. ?

ahgittin: (agree, have added comment)

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/api/AssemblyTemplateSpecInstantiator.java
ahgittin: agree, but the `CampResolver` methods use it, currently calling to this when they should be calilng services.  rather than repair that i hope we can deprecate all of this in favour of a new much simpler `CampTypePlanTransformer`.

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/api/AssemblyTemplateSpecInstantiator.java
ahgittin: agree

usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/CampResolver.java
ahgittin: good spot

usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogParametersTest.java
ahgittin: i think we've agreed we'll remove this?  we can remove the test (and this comment) when we do

utils/common/src/main/java/org/apache/brooklyn/util/text/Identifiers.java
ahgittin: agree, added the word `GOOD` and javadoc

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
ahgittin: i'd like to be able to make them non-null, when the creation API is good enough

core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
ahgittin: Yes, this is what I'm trying to clear up in the next PR.  Your reference is helpful!

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogUtils.java
ahgittin: +1 to tightening this; i've added some TODO comments and deprecation to move us in this direction

api/src/main/java/org/apache/brooklyn/api/mgmt/classloading/BrooklynClassLoadingContext.java
neykov: I guess it's down to personal preference. My view is take `Iterable` and return the most specific type possible, `Iterable` is for lazy containers only. What I don't like about `Iterable` is that it's not of much use outside of guava, most libraries take `Collection`s.
The implementations are all eager in this case.

api/src/main/java/org/apache/brooklyn/api/typereg/BrooklynTypeRegistry.java
neykov: No need for a new entry point on the context, a `new TypeInstantiator(mgmt).createBean(...)` will do.

api/src/main/java/org/apache/brooklyn/api/typereg/BrooklynTypeRegistry.java
neykov: All three points above apply to the instantiator code path only, not the lookup.

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypePredicates.java
neykov: Good point.

core/src/test/java/org/apache/brooklyn/core/catalog/internal/SpecParameterInMetaTest.java
neykov: The idea of `TestToSpecTransformer` is to convert a plan to a pre-registered spec. It's not working on types but on plans. This means that it's able to convert a plan it understands (be it coming from a catalog item or used as an app plan) to a spec. The plan itself may be added as a catalog item, but it's a separate matter.

A plan refers to a type, but is not itself a type, it's a composition of types. Calling `registerSpec` tells the transformer how to interpret the plan, not what the spec for a specific type is.

core/src/test/java/org/apache/brooklyn/core/catalog/internal/SpecParameterInMetaTest.java
neykov: btw removing parameters in meta support made this class obsolete, see https://github.com/apache/incubator-brooklyn/pull/1036. Some other conflicts in the tests between the PRs unfortunately due to a renaming which was needed.

usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogParametersTest.java
neykov: Yes, I commented before we did so :)

api/src/main/java/org/apache/brooklyn/api/typereg/BrooklynTypeRegistry.java
ahgittin: Yup I reached the same conclusion.  Testing your change now.

ahgittin:
Awesome thorough review.  I've addressed all comments except where I'm commented.  Much better now.  Going to merge #1036 however and then clean this up once more before merging.

api/src/main/java/org/apache/brooklyn/api/typereg/RegisteredType.java
```diff
@@ -38,19 +46,24 @@
...
      * or if precise type details are unavailable and all that is known at creation is some higher level interface/supertype
      * (e.g. this may return {@link Entity} even though the spec points at a specific subclass,
-     * for instance because the YAML has not yet been parsed or OSGi bundles downloaded). 
+     * for instance because the YAML has not yet been parsed or OSGi bundles downloaded).
+     * <p>
+     * This may include other registered types such as marker interfaces.
      */
-    Class<?> getJavaType();
+    @Beta
+    @Nullable Set<Object> getSuperTypes();
```
sjcorbett: Any particular reason this is `@Nullable`? Feels better for it to be the opposite. Its callers already seem to assume otherwise (e.g. https://github.com/apache/incubator-brooklyn/pull/1017/files#diff-2c1cbcb42656df77efc28ab3153cb1e5R250).

ahgittin:
All addressed, and merged/resolved with #1036. Ready for merging if tests pass on the server.

@neykov the simpler signatures `TypeRegistry.get` are nice, and i've added validation and a simple test for the `StaticTypePlanTransformer`.  The old `Test...Transformer` is gone as I see you've moved the tests to `usage/camp` where it's not needed.

api/src/main/java/org/apache/brooklyn/api/typereg/RegisteredType.java
ahgittin: shouldn't be, will remove

PR #1018, opened 2015-11-10
Add ManagementContext.getConfigWithExternalConfigResolved()

rdowner:
I'm not really sure about this. Throwing it out there to see what people think.

The external configuration supplier feature is great, but it doesn't work with site-wide configuration defined in brooklyn.properties. This PR is an attempt to add this function.

It, unfortunately, is not a simple matter. Because external configuration is *defined* in brooklyn.properties, there's a possible recursion problem. We could resolve this during initialisation of the management context, but we want to keep the behaviour that the configuration suppliers are invoked whenever required and always return fresh values, rather than resolving them at startup and leaving them to go stale. Some of my attempted solutions exposed weird bugs.

So the simplest possible thing is to not attempt to retro-fit this but instead add a new method which explicitly retrieves the brooklyn.properties values, with configuration resolved there and then.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/AbstractManagementContext.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/AbstractManagementContext.java
@@ -29,9 +29,12 @@ import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.annotation.Nullable;
 
+import com.google.common.collect.Maps;
 import org.apache.brooklyn.api.catalog.BrooklynCatalog;
 import org.apache.brooklyn.api.effector.Effector;
 import org.apache.brooklyn.api.entity.Entity;
@@ -89,6 +92,8 @@ import com.google.common.collect.ImmutableSet;
 
 public abstract class AbstractManagementContext implements ManagementContextInternal {
     private static final Logger log = LoggerFactory.getLogger(AbstractManagementContext.class);
+
+    private static final Pattern EXTERNAL_CONFIG_PATTERN = Pattern.compile("^\\$brooklyn:external:([^:]+):([^:]+)$");
 
     private static DataGridFactory loadDataGridFactory(BrooklynProperties properties) {
         String clazzName = properties.getFirst(DataGridFactory.class.getName());
@@ -370,6 +375,27 @@ public abstract class AbstractManagementContext implements ManagementContextInte
     @Override
     public BrooklynProperties getBrooklynProperties() {
         return configMap;
+    }
+
+    public StringConfigMap getConfigWithExternalConfigResolved() {
+        final ExternalConfigSupplierRegistry externalConfigProviderRegistry = getExternalConfigProviderRegistry();
+        if (externalConfigProviderRegistry == null)
+            return configMap; // no external config available yet, so return config without resolving
+
+        Map extConfigResolved = Maps.transformValues(configMap, new Function() {
+            @Nullable
+            @Override
+            public Object apply(Object input) {
+                if (input == null) return null;
+                if (!(input instanceof String)) return input;
+                Matcher m = EXTERNAL_CONFIG_PATTERN.matcher((String) input);
+                if (!(m.matches())) return input;
+                String providerName = m.group(1);
+                String key = m.group(2);
+                return externalConfigProviderRegistry.getConfig(providerName, key);
+            }
+        });
+        return BrooklynProperties.Factory.newEmpty().addFromMap(extConfigResolved);
     }
 
     @Override
--- a/api/src/main/java/org/apache/brooklyn/api/mgmt/ManagementContext.java
+++ b/api/src/main/java/org/apache/brooklyn/api/mgmt/ManagementContext.java
@@ -180,7 +180,9 @@ public interface ManagementContext {
      * Defaults to reading ~/.brooklyn/brooklyn.properties but configurable in the management context.
      */
     StringConfigMap getConfig();
-    
+
+    StringConfigMap getConfigWithExternalConfigResolved();
+
     /**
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/apache/brooklyn/core/mgmt/internal/AbstractManagementContext.java
```diff
@@ -372,6 +377,27 @@ public BrooklynProperties getBrooklynProperties() {
         return configMap;
     }
 
+    public StringConfigMap getConfigWithExternalConfigResolved() {
+        final ExternalConfigSupplierRegistry externalConfigProviderRegistry = getExternalConfigProviderRegistry();
+        if (externalConfigProviderRegistry == null)
+            return configMap; // no external config available yet, so return config without resolving
```
neykov: May be set external config values to null to prevent using the raw string or even throw as the caller would expect them to be resolved already?

neykov:
The changes look good. I like the approach because there's no hidden magic, easy to reason about. On the other hand what I don't like is that it would be very hard to figure out which config keys support external config, leaving users confused.

What problems surfaced when you tried resolving the values when requested (what you mention in the description)?

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/AbstractManagementContext.java
rdowner: This check is probably superfluous. It was required in an earlier iteration of this code where the constructor had possibly not finished, but that should not be the case now. I will either remove it, or (to be safe) make it throw an `IllegalStateException`.

rdowner:
@neykov I saw a problem if I modified `LocalManagementContext.getConfig()` to do the resolution. If this method returns something that is not the original `configMap` - even if it is an identical copy (such as by replacing the method body with `return BrooklynProperties.Factory.newEmpty().addFrom(configMap);`) then I got this exception during launch:

```
2015-11-11 11:10:20,534 WARN  Error importing catalog from classpath:/brooklyn/default.catalog.bom: org.apache.brooklyn.util.exceptions.CompoundRuntimeException: Could not resolve item server, 2 errors including: Transformer for Brooklyn OASIS CAMP interpreter gave an error creating this plan: No CAMP Platform is registered with this Brooklyn management context.
org.apache.brooklyn.util.exceptions.CompoundRuntimeException: Could not resolve item server, 2 errors including: Transformer for Brooklyn OASIS CAMP interpreter gave an error creating this plan: No CAMP Platform is registered with this Brooklyn management context.
        at org.apache.brooklyn.util.exceptions.Exceptions.create(Exceptions.java:306) ~[org.apache.brooklyn-brooklyn-utils-common-0.9.0-SNAPSHOT.jar:0.9.0-SNAPSHOT]
Caused by: java.lang.IllegalStateException: Transformer for Brooklyn OASIS CAMP interpreter gave an error creating this plan: No CAMP Platform is registered with this Brooklyn management context.
        at org.apache.brooklyn.util.guava.IllegalStateExceptionSupplier.get(IllegalStateExceptionSupplier.java:52) ~[org.apache.brooklyn-brooklyn-utils-common-0.9.0-SNAPSHOT.jar:0.9.0-SNAPSHOT]
Caused by: org.apache.brooklyn.util.exceptions.PropagatedRuntimeException:
        at org.apache.brooklyn.util.exceptions.Exceptions.create(Exceptions.java:298) ~[org.apache.brooklyn-brooklyn-utils-common-0.9.0-SNAPSHOT.jar:0.9.0-SNAPSHOT]
Caused by: org.apache.brooklyn.util.exceptions.PropagatedRuntimeException: Transformer for Brooklyn OASIS CAMP interpreter gave an error creating this plan: No CAMP Platform is registered with this Brooklyn management context.
        at org.apache.brooklyn.core.plan.PlanToSpecFactory.attemptWithLoaders(PlanToSpecFactory.java:110) ~[org.apache.brooklyn-brooklyn-core-0.9.0-SNAPSHOT.jar:0.9.0-SNAPSHOT]
Caused by: java.lang.IllegalStateException: No CAMP Platform is registered with this Brooklyn management context.
        at org.apache.brooklyn.camp.brooklyn.spi.creation.CampUtils.getCampPlatform(CampUtils.java:217) ~[org.apache.brooklyn-brooklyn-camp-0.9.0-SNAPSHOT.jar:0.9.0-SNAPSHOT]
```

Interestingly if I modify the method to use a MemoizedSupplier - meaning that it will consistently return the same object - the error disappears. This suggests that there is something that gets upset if `getConfig()` returns a different object, even if the contents are the same.

[A thought occurs to me. Is something *modifying* the config object that is returned?]

rdowner:
Bingo. https://github.com/apache/incubator-brooklyn/blob/master/usage/camp/src/main/java/org/apache/brooklyn/camp/brooklyn/BrooklynCampPlatform.java#L73:

```java
        ((BrooklynProperties)bmc.getConfig()).put(BrooklynCampConstants.CAMP_PLATFORM, this);
```

Downcasts to `BrooklynProperties` and changes the config. And if it's changing it on a copy of the config, then that change is ignored, and we end up with no CAMP platform set.

neykov:
So we need to resolve external values at `getConfig` time, but at the same time allow mutations of the properties. What if you patch the config in the `AbstractManagementContext` constructor, but lazily - here's what I mean. In the constructor wrap the properties in a delegate, including a reference to the mgmt context. On `get`s check the value and if needed resolve. Alternatively convert any external values to suppliers, resolving on `get` (i.e. make `BrooklynProperties` handle suppliers).

It's obviously hackish to mutate the `ManagementContext.getConfig()` - might be a good reason to find a generic way to do it, but out of scope for this.

aledsage:
@neykov @rdowner I'm having a look at this - looking at neykov's suggestion.

aledsage:
@rdowner @neykov See https://github.com/apache/incubator-brooklyn/pull/1112, which gives an alternative implemention. I believe that replaces this PR, so this can be closed.

ahgittin:
@rdowner can we close?

rdowner:
Aled's implementation is better :-) closing this one.

PR #1022, opened 2015-11-12
Config supplied to SshTool’s constructor

aledsage:
When overriding the SshTool implementation to use, allows extra config to be supplied that will be passed to the constructor.


sjcorbett:
Looks good. Will merge.

PR #1023, opened 2015-11-12
location is not serializable

aledsage:
Deletes `extends Serializable` from location interface, and removes all declared serialVersionUID from location implementations.
    
Location was not really serialisable anyway - for example it had non-transient references to `ManagementContext`. The declaration of “Serializable” was a throwback to very old days when persistence was imagined to be very different.

Also has some other very minor tidy up (in separate commits).


ahgittin:
LGTM

PR #1026, opened 2015-11-12
BROOKLYN-144: avoid err when persisting ha-record

aledsage:
- Adds Exception.propagate(msg, throwable) so can include info about
  the file that could not be created.
- Fix FileUtil.setFilePermissions so that createNewFile creates the
  parent directory if necessary.
- HighAvailabilityManager’s polling task: log.err on first exception,
  then log.debug for subsequent consecutive exceptions.

### DIFF
--- a/utils/common/src/test/java/org/apache/brooklyn/util/exceptions/ExceptionsTest.java
+++ b/utils/common/src/test/java/org/apache/brooklyn/util/exceptions/ExceptionsTest.java
@@ -59,6 +59,28 @@ public class ExceptionsTest {
     }
     
     @Test
+    public void testPropagateCheckedExceptionWithMessage() throws Exception {
+        String extraMsg = "my message";
+        Exception tothrow = new Exception("simulated");
+        try {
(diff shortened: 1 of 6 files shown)


utils/common/src/test/java/org/apache/brooklyn/util/exceptions/ExceptionsTest.java
```diff
@@ -59,6 +59,28 @@ public void testPropagateCheckedException() throws Exception {
...
+            assertEquals(e.getMessage(), "my message");
+            assertEquals(e.getCause(), tothrow);
+        }
+    }
+    
+    @Test
+    public void testPropagateRuntimeExceptionIgnoresMessage() throws Exception {
+        NullPointerException tothrow = new NullPointerException("simulated");
+        try {
+            throw Exceptions.propagate("my message", tothrow);
```
ahgittin: see `Asserts.shouldHaveFailedPreviously()` (useless in this simple case but i find it makes this type of test more readable)

ahgittin:
LGTM

PR #1027, opened 2015-11-12
Improve performance tests

aledsage:
Our performance tests need a much bigger overhaul. However, these changes make them a bit more usable - and makes them record performance results in a local file which is better than nothing.

- Extracts code from AbstractPerformanceTest into PerformanceMeasurer, making it more configurable and usable from more places.
- Adds a default persister to write files to `~/brooklyn-performance/` (while we figure out where to write those to in anger).

### DIFF
--- a/usage/test-support/src/main/java/org/apache/brooklyn/test/performance/PerformanceMeasurer.java
+++ b/usage/test-support/src/main/java/org/apache/brooklyn/test/performance/PerformanceMeasurer.java
@@ -1,0 +1,156 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.test.performance;
+
+import static org.testng.Assert.fail;
+
+import java.util.Date;
+import java.util.List;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.brooklyn.util.exceptions.Exceptions;
+import org.apache.brooklyn.util.time.Duration;
+import org.apache.brooklyn.util.time.Time;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.annotations.Beta;
+import com.google.common.base.Stopwatch;
+import com.google.common.collect.Lists;
+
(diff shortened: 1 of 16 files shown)


usage/test-support/src/main/java/org/apache/brooklyn/test/performance/MeasurementOptions.java
```diff
@@ -0,0 +1,135 @@
...
+import com.google.common.annotations.Beta;
+import com.google.common.base.Objects;
+
+/**
+ * For building up a description of what to measure.
+ * 
+ * See {@link PerformanceTestUtils#measure(MeasurementOptions)}.
+ */
+@Beta
+public class MeasurementOptions {
```
ahgittin: `MeasurementOptions` is a cryptic name.  `PerformanceTestOptions` ? or `PerformanceTestDescriptor` (as some aren't optional) ?

usage/test-support/src/main/java/org/apache/brooklyn/test/performance/MeasurementResult.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.common.annotations.Beta;
+import com.google.common.base.Objects;
+
+/**
+ * The results of a performance test.
+ * 
+ * See {@link PerformanceTestUtils#measure(MeasurementOptions)}.
+ */
+@Beta
+public class MeasurementResult {
```
ahgittin: and `PerformanceTestResult` ?

usage/test-support/src/main/java/org/apache/brooklyn/test/performance/PerformanceMeasurer.java
```diff
@@ -0,0 +1,154 @@
...
+ * horrendously wrong. 
+ * 
+ * However, good performance measurement involves much more warm up (e.g. to ensure java HotSpot 
+ * optimisation have been applied), and running the test for a reasonable length of time.
+ * 
+ * Longevity tests are also important for to check if object creation is going to kill
+ * performance in the long-term, etc.
+ */
+@Beta
+public class PerformanceMeasurer {
```
ahgittin: `PerformanceTestRunner` ?

usage/test-support/src/main/java/org/apache/brooklyn/test/performance/PerformanceMeasurer.java
```diff
@@ -0,0 +1,154 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(PerformanceMeasurer.class);
+
+    /**
+     * Runs a performance test. Repeatedly executes the given job. It measuring either the time it takes for
+     * many iterations, or the number of iterations it can execute in a fixed time.
+     */
+    public static MeasurementResult measure(MeasurementOptions options) {
```
ahgittin: `run(PerformanceTestDescriptor)` ?

ahgittin:
good improvement; optional suggestion on the naming strategy

usage/test-support/src/main/java/org/apache/brooklyn/test/performance/PerformanceMeasurer.java
aledsage: Agree with your other renames.

However `TestRunner` is a technical term in testng (http://testng.org/javadoc/org/testng/TestRunner.html) so that doesn't seem like the right name. I'll stick with the clumsy-but-accurate `PerformanceMeasurer` for now, marked as `@Beta`.

aledsage:
Incorporated comments (except left the `PerformanceMeasurer` class name). Merging now.

PR #1028, opened 2015-11-13
Adding missing ASF license headers to test-framework

m4rkmckenna:
The `test-framework` was using an incorrect parent pom so the `apache-rat-plugin` did not fail the build


ahgittin:
Looks like there are other files missing headers:

https://builds.apache.org/job/incubator-brooklyn-pull-requests/2126/

m4rkmckenna:
@ahgittin The markdown files caught me out :)

ahgittin:
happens to me all the time :) .  it especially annoys me that the rat plugin doesn't log the offending files to stdout.

aledsage:
Thanks @m4rkmckenna - that was an extremely quick response!

PR #1029, opened 2015-11-13
ParallelTestCase change and documentation updates

burkechris:
Addressed review comments.

* Changed how ParallelTestCase task is submitted and waited upon. Functionally no change.
* Also added ParallelTestCase documentation to the docs.
* Moved the Test Entities documentation to within the Brooklyn docs guide.


aledsage:
Looks good; merging.

PR #1031, opened 2015-11-13
yaml testing docs

johnmccabe:
Added details of the Test Entities to the YAML Blueprints guide.

Have explicitly *not* added a reference to the Java Blueprints section at this point, shout if you'd like that added.


docs/guide/yaml/test/example_yaml/paralleltestcase-example.yaml
```diff
@@ -0,0 +1,46 @@
...
+    brooklyn.config:
+      simple.confg: someValue
+    services:
+    - type: org.apache.brooklyn.test.framework.TestCase
+      name: Parallel Testcase
+      brooklyn.children:
+      - type: org.apache.brooklyn.test.framework.TestCase
+        name: Parallel Testcase
+        brooklyn.children:
+        - type: org.apache.brooklyn.entity.webapp.ControlledDynamicWebAppCluster
```
aledsage: I think I missed some earlier discussion perhaps - why do we put the entity-under-test as a child of the `TestCase` entity?

docs/guide/yaml/test/example_yaml/paralleltestcase-example.yaml
```diff
@@ -0,0 +1,46 @@
...
+  license: Apache-2.0
+  item:
+    brooklyn.config:
+      simple.confg: someValue
+    services:
+    - type: org.apache.brooklyn.test.framework.TestCase
+      name: Parallel Testcase
+      brooklyn.children:
+      - type: org.apache.brooklyn.test.framework.TestCase
+        name: Parallel Testcase
```
aledsage: Why a TestCase inside a TestCase, both with name "Parallel Testcase", and then a child of that being the entity-under-test?

docs/guide/yaml/test/example_yaml/simple-tomcat-app-sensor-test.yaml
```diff
@@ -0,0 +1,24 @@
+brooklyn.catalog:
+  id: simple-tomcat-app-sensor-test
+  version: 1.0
+  itemType: template
+  iconUrl: http://tomcat.apache.org/images/tomcat.png
+  name: Simple Tomcat Application Sensor Test
+  license: Apache-2.0
+  item:
+    brooklyn.config:
+      simple.confg: someValue
```
aledsage: Can we delete this (from here and from all the other examples)? I'm a fan of keeping the examples as simple as possible (so people don't get a false impression of how long it has to be). If we want to demonstrate other thing such as config, we can give a separate example for that.

docs/guide/yaml/test/index.md
```diff
@@ -0,0 +1,24 @@
+---
+title: Testing YAML Blueprints
+layout: website-normal
+children:
+- test-entities.md
+- usage-examples.md
+---
+
+Brooklyn provides a selection of basic test entities which can be used to validate Blueprints via YAML. These are divided into two groups, structural, which effect the order in which child entities are started, and validation, which are used to confirm the application is deployed as intended:
```
aledsage: I'd fo for a colon after the "two groups", a semi-colon after the "are started", and a full stop at the end of the sentence.

docs/guide/yaml/test/example_yaml/paralleltestcase-example.yaml
johnmccabe: Both having the name 'Parallel Testcase' is a mistake, the second should be something along the lines of 'Deploy Application'.

The first TestCase is used to allow sequential execution of the children, ie deploy app, then run tests. The ParallelTestCase starts the `ControlledDynamicWebAppCluster` and `MySqlNode` in parallel to reproduce the behaviour without tests (ie just nested under `-services:`.

docs/guide/yaml/test/example_yaml/paralleltestcase-example.yaml
johnmccabe: see response above

docs/guide/yaml/test/example_yaml/simple-tomcat-app-sensor-test.yaml
johnmccabe: Agreed, specifics of packaging up tests can be dealt with in the context of a test runner.

docs/guide/yaml/test/test-entities.md
```diff
@@ -0,0 +1,72 @@
...
+## Structural Test Entities
+
+### TestCase
+The `TestCase` entity acts as a container for a list child entities which are started *sequentially*.
+{% highlight yaml %}
+{% readj example_yaml/testcase-entity.yaml %}
+{% endhighlight %}
+The ability to sequentially run entities is used to allow applications to fully deploy before attempting to start any test entities.
+
+If your application blueprint consists of multiple services then you will also require the `ParallelTestCase` entity.
```
aledsage: I was looking at it as the entity-under-test and the TestCase could execute concurrently. The first assertion could then be an assertion that within 10 minutes the serviceUp was true.

The sequential TestCase also allows a series of assertions, such as invoking an effector and then asserting that a sensor has the expected value (or that it is reachable using the http test, or whatever).

docs/guide/yaml/test/test-entities.md
```diff
@@ -0,0 +1,72 @@
...
+The `TestHttpCall` entity performs a HTTP GET on the specified URL and performs an assertion on the response.
+{% highlight yaml %}
+{% readj example_yaml/testhttpcall-entity.yaml %}
+{% endhighlight %}
+
+#### Parameters
+- `url` - URL to perform GET request on, for example `$brooklyn:component("tomcat").attributeWhenReady("webapp.url")``.
+- `timeout` - time to wait for a HTTP response. For example `10s`, `10m`, etc
+- `assertions` - perform one of the following assertions.
+  - `status` - response must match the specified status code. The example above shows an assertions on a `404` response.
```
aledsage: Longer term, would be nice to also allow a range (e.g. "200-299") (similar to `PortRanges.fromString()`)

docs/guide/yaml/test/usage-examples.md
```diff
@@ -0,0 +1,57 @@
...
+layout: website-normal
+---
+
+{% include fields.md %}
+
+## Introduction
+This section provides example tests carried out on the following `simple-tomcat-app` web application catalog item:
+
+{% highlight yaml %}
+{% readj example_yaml/simple-tomcat-app.yaml %}
```
aledsage: Do we want to show all these examples as catalog items, or just as YAML for deploying an app directly? The latter would be shorter examples.

docs/guide/yaml/test/test-entities.md
m4rkmckenna: @aledsage I have a change locally based from comments on #999 by @ahgittin that will address this. Basically there will be a common set of assertions so the above would be covered by a regular expression
```
assert:
  matches: 2[0-9][0-9]
```

docs/guide/yaml/test/test-entities.md
johnmccabe: Both approaches are feasible, I had documented the approach that seemed simplest to me. We could take an agnostic approach and describe both test approaches, I'd prefer to have one canonical way of approaching a test though - both sequential and parallel have their pros and cons.

For example with the parallel approach this is what an effector test would look like, is this more like you were thinking?
```
name: Parallel Effector Test Example
services:
- type: org.apache.brooklyn.entity.webapp.ControlledDynamicWebAppCluster
  name: My Web
  id: webappcluster
  brooklyn.config:
    wars.root: http://search.maven.org/remotecontent?filepath=io/brooklyn/example/brooklyn-example-hello-world-sql-webapp/0.6.0/brooklyn-example-hello-world-sql-webapp-0.6.0.war
    java.sysprops:
      brooklyn.example.db.url: >
        $brooklyn:formatString("jdbc:%s%s?user=%s&password=%s",
        component("db").attributeWhenReady("datastore.url"),
        "visitors", "brooklyn", "br00k11n")
- type: org.apache.brooklyn.entity.database.mariadb.MariaDbNode
  id: db
  name: My DB
  brooklyn.config:
    creationScriptUrl: https://bit.ly/brooklyn-visitors-creation-script
- type: org.apache.brooklyn.test.framework.TestCase
  name: Test Deploy Effector
  brooklyn.children:
  - type: org.apache.brooklyn.test.framework.TestSensor
    name: webappcluster service.isUp
    targetId: webappcluster
    sensor: service.isUp
    timeout: 15m
    assert:
      equals: true
  - type: org.apache.brooklyn.test.framework.TestSensor
    name: db service.isUp
    targetId: db
    sensor: service.isUp
    timeout: 5m
    assert:
      equals: true
  - type: org.apache.brooklyn.test.framework.TestEffector
    name: Deploy WAR in /newcontext
    targetId: webappcluster
    effector: deploy
    params:
      url: http://search.maven.org/remotecontent?filepath=io/brooklyn/example/brooklyn-example-hello-world-sql-webapp/0.6.0/brooklyn-example-hello-world-sql-webapp-0.6.0.war
      targetName: newcontext
  - type: org.apache.brooklyn.test.framework.TestHttpCall
    name: /newcontext Status Code 200
    timeout: 5m
    url: $brooklyn:formatString("%s/newcontext/", component("webappcluster").attributeWhenReady("webapp.url"))
    assert:
      status: 404
```

docs/guide/yaml/test/example_yaml/paralleltestcase-example.yaml
johnmccabe: Switching to the parallel example mentioned [here](https://github.com/apache/incubator-brooklyn/pull/1031#discussion_r44918698) and rewording the structural test entity descriptions to be less prescriptive on their usage.

docs/guide/yaml/test/test-entities.md
johnmccabe: I'll give you access to this PR branch so you can push in your assertion related updates @m4rkmckenna

docs/guide/yaml/test/usage-examples.md
johnmccabe: Dropping the catalogs from these examples

docs/guide/yaml/test/index.md
johnmccabe: Done

docs/guide/yaml/test/test-entities.md
aledsage: Yes, that's what I was thinking of. That feels more like you are adding the test assertions, rather than having to "restructure" their app by putting it inside a `ParallelTest entity`.

docs/guide/yaml/test/example_yaml/testcases/getting-started-test-example.yaml
```diff
@@ -0,0 +1,68 @@
...
+        "visitors", "brooklyn", "br00k11n")
+- type: org.apache.brooklyn.entity.database.mysql.MySqlNode
+  id: db
+  name: My DB
+  brooklyn.config:
+    creationScriptUrl: https://bit.ly/brooklyn-visitors-creation-script
+- type: org.apache.brooklyn.test.framework.TestHttpCall
+  name: Check HTTP Response Status Code
+  url: >
+    $brooklyn:formatString("http://%s:%s",
```
aledsage: Do we not advertise this combined root URI as a sensor? The code looks like it should propagate "main.uri" to the ControlledDynamicWebAppCluster instance.

docs/guide/yaml/test/index.md
```diff
@@ -0,0 +1,24 @@
+---
+title: Testing YAML Blueprints
+layout: website-normal
+children:
+- test-entities.md
+- usage-examples.md
+---
+
+Brooklyn provides a selection of basic test entities which can be used to validate Blueprints via YAML. These are divided into two groups; structural, which effect the order in which child entities are started; and validation, which are used to confirm the application is deployed as intended.
```
aledsage: Instead of "order in which child entities are started", perhaps relate it to the order in which child test-things are run. But obviously not with that wording! I'm fine with leaving it as-is if we can't come up with a better term for "child test-things".

aledsage:
Thanks @johnmccabe looks good. Only a couple of additional minor comments. Ping me when you've taken a look at those, and I'll merge it.

docs/guide/yaml/test/example_yaml/testcases/getting-started-test-example.yaml
johnmccabe: We do, but its current being assigned a non-public address for the NginxController (on SL), I would have expected the webapp.url to be a public address (per the change in #959), had only spotted this last night and have not gotten round to raising it - will do so now - wanted to make sure the examples would be functional in the immediate term so pushed in this workaround.

docs/guide/yaml/test/index.md
johnmccabe: Change "_structural, which effect the order in which child entities are started_" to "_structural,  which control the sequencing of service entity deployment and test entity execution_"

PR #1032, opened 2015-11-13
Restore old HttpTool / Response  for backward compatibility

geomacy:
PR follow-up to comment https://github.com/apache/incubator-brooklyn/pull/994#discussion_r44706399

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/util/core/http/HttpTool.java
+++ b/core/src/main/java/org/apache/brooklyn/util/core/http/HttpTool.java
@@ -18,68 +18,11 @@
  */
 package org.apache.brooklyn.util.core.http;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-import static com.google.common.base.Preconditions.checkState;
 
-import java.net.URI;
-import java.security.cert.CertificateException;
-import java.security.cert.X509Certificate;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Map;
-import java.util.Map.Entry;
-
-import org.apache.brooklyn.util.crypto.SslTrustUtils;
-import org.apache.brooklyn.util.exceptions.Exceptions;
-import org.apache.brooklyn.util.net.URLParamEncoder;
-import org.apache.brooklyn.util.text.Strings;
-import org.apache.brooklyn.util.time.Duration;
-import org.apache.commons.codec.binary.Base64;
-import org.apache.http.ConnectionReuseStrategy;
-import org.apache.http.HttpEntity;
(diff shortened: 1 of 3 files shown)


neykov:
Thanks @geomacy, merging.

neykov:
Did some more testing, unfortunately this doesn't fully solve the problem. Will hold off merging so we can further discuss it. The changes here are perfectly good though, but not enough.

There's an API that's heavilty used in entities - `HttpFeed` and related (i.e. `HttpValueFunction`, `HttpPollConfig`). The package renaming changes the signature of the methods in these classes so that already compiled entities against pre-changes version will not be compatible with post-changes versions. Probably worth moving to the dev list.

aledsage:
Thanks @geomacy - I'd also include a comment at the top of each of those classes to explain why the code is still there (rather than the "extends" mechanism). For example, copy-paste part of Svet's comment from https://github.com/apache/incubator-brooklyn/pull/994#discussion_r44706399.

geomacy:
Will do that,  thanks Aled.

G

ahgittin:
As discussed on mailing list, I'm against.  Short of introducing alternates for every call which references these classes, it's impossible to completely preserve backwards binary compatibility for these changes.  The halfway house of leaving it as was (closing this) and putting a comment in the release notes seems the pragmatic choice.

geomacy:
Ok, I shall leave this as-is then.

geomacy:
Or rather, should I just close this pull request?

ahgittin:
My suggestion is to close this.  @neykov @aledsage you agree?

neykov:
No strong opinion on the backwards compatibility. What I'd like though is to either merge this or completely remove HttpTool in core (just leaving a shell class with a javadoc which points to the new one). This way we won't have an HttpTool in core, returning HttpToolResponse from utils.

ahgittin:
@neykov fine with your plans to gut `HttpTool` core, rather than allow it to (surprisingly) return the new non-core `HttpToolResponse`

@geomacy can you do that?

geomacy:
i.e. remove it altogether?  Yes, I'll update the pull request accordingly.

geomacy:
I've updated the core tool and response classes in [14945f7](https://github.com/apache/incubator-brooklyn/commit/14945f75b1e0a4a5cc17dbb30459d092649bf770)

neykov:
Can you squash. Are you updating the release notes separately?

geomacy:
Release notes updated and changes squashed.

neykov:
LGTM, will merge.

PR #1033, opened 2015-11-13
BROOKLYN-193: Add configuration for skipping invalid configs

bostko:
Add configuration option for skipping invalid config values

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindManagerImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindManagerImpl.java
@@ -92,6 +92,9 @@ public class RebindManagerImpl implements RebindManager {
     public static final ConfigKey<RebindFailureMode> REBIND_FAILURE_MODE =
             ConfigKeys.newConfigKey(RebindFailureMode.class, "rebind.failureMode.rebind",
                     "Action to take if a failure occurs during rebind", RebindFailureMode.FAIL_AT_END);
+    public static final ConfigKey<RebindFailureMode> ADD_CONFIG_FAILURE_MODE =
+            ConfigKeys.newConfigKey(RebindFailureMode.class, "rebind.failureMode.addConfig",
+                    "Action to take if a failure occurs when setting a config value. It could happen coercion of the value type to fail.", RebindFailureMode.FAIL_AT_END);
     public static final ConfigKey<RebindFailureMode> ADD_POLICY_FAILURE_MODE =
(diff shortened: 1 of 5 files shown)


aledsage:
Looks good. Can we get a unit test?

Should the default be `addConfigFailureMode = RebindManager.RebindFailureMode.FAIL_AT_END`? If so, we need to ensure customers are aware of how to configure it so as to skip this error (great that you've already added it to `docs/guide/ops/persistence/index.md` - this is more a reminder to us to point it out to customers explicitly).

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindManagerImpl.java
```diff
@@ -92,6 +92,9 @@
     public static final ConfigKey<RebindFailureMode> REBIND_FAILURE_MODE =
             ConfigKeys.newConfigKey(RebindFailureMode.class, "rebind.failureMode.rebind",
                     "Action to take if a failure occurs during rebind", RebindFailureMode.FAIL_AT_END);
+    public static final ConfigKey<RebindFailureMode> ADD_CONFIG_FAILURE_MODE =
+            ConfigKeys.newConfigKey(RebindFailureMode.class, "rebind.failureMode.addConfig",
+                    "Action to take if a failure occurs when setting a config value. It could happen coercion of the value type to fail.", RebindFailureMode.FAIL_AT_END);
```
neykov: The default for isolated problems is to `.CONTINUE` which will let Brooklyn complete the startup, but display an error message in the console and deny REST requests. Better apply the same default here.

bostko: I just preserved it like the behavior it was before this change. Startup was failing the same way.
Svet, do you really want me to change the default to CONTINUE ?

neykov: Good point, keep it as is. I confused it for a different functionality (`--startupFailOnPersistenceErrors` startup option).

aledsage: As long as we can definitely supply the configuration on Brooklyn startup to override it to CONTINUE, so that a customer can rebind with this kind of error.

@bostko can we definitely do that?

aledsage:
Thanks; merging.

bostko:
Yes I tested with `rebind.failureMode.addConfig=continue` and without it and it works as expected.
I am working on providing tests now. It will take me several hours, they can be merged in a separate PR if you'd like.

PR #1034, opened 2015-11-16
Fix jclouds sshable

aledsage:
(Unrelated to other changes, fixes `NetworkingUtilsTest.testIsPortAvailableReportsTrueWhenPortIsFree`).

Lots of fixes to get the `JcloudsLoginLiveTest` working again.

- registerJcloudsSshMachineLocation: use the userCredentials that are passed into the method, rather than looking up what we expect from the config.  
  (Hopefully fixes testSpecifyingPasswordAndSshKeysPrefersKeys).
- Fix waitForSshable to when password supplied on JcloudsLocation config, and using useJcloudsSshInit=false.
- Fix use of DISABLE_ROOT_AND_PASSWORD_SSH: if for root user with password-based login, then explicitly set PasswordAuthentication=yes.  
  If root user, then explicitly set PermitRootLogin=true
- Fix getPublicHostnameAws so pass in the credentials (some code-paths will not have configured the credentials in the generic config).  
  Same for getPrivateHostnameAws.
- Tidy up the createTemporarySshMachineLocaton
- Logging of timing: include how long to get semaphore
- Logging of failure: include how long for each stage
    
Previously hit a problem when the jclouds location was configured with
a password + a blank ssh key, and using useJcloudsSshInit=false.
It failed to ssh to the newly provisioned AWS VM in waitForSshable. It
got the node.getCredentials, which gives us the ssh-key from the AWS
key-pair. But then we instantiate an SshMachineLocation to check if we
can connection (before continuing with the user-setup). That
machineLocation inherits the password (which has not been set on the
machine), so it fails to ssh - the password takes precedence over the
ssh-key.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -645,6 +645,12 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
         String groupId = elvis(setup.get(GROUP_ID), cloudMachineNamer.generateNewGroupId(setup));
         NodeMetadata node = null;
         JcloudsMachineLocation machineLocation = null;
+        Duration semaphoreTimestamp = null;
+        Duration templateTimestamp = null;
+        Duration provisionTimestamp = null;
+        Duration usableTimestamp = null;
+        Duration customizedTimestamp = null;
+        Stopwatch provisioningStopwatch = Stopwatch.createStarted();
         
         try {
             LOG.info("Creating VM "+setup.getDescription()+" in "+this);
@@ -659,9 +665,7 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
             } else {
                 LOG.debug("Acquired in {} machine-creation permit immediately", this);
             }
-
-            Stopwatch provisioningStopwatch = Stopwatch.createStarted();
-            Duration templateTimestamp, provisionTimestamp, usableTimestamp, customizedTimestamp;
+            semaphoreTimestamp = Duration.of(provisioningStopwatch);
 
             LoginCredentials userCredentials = null;
             Set<? extends NodeMetadata> nodes;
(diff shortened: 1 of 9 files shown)


utils/common/src/main/java/org/apache/brooklyn/util/net/Networking.java
```diff
@@ -486,6 +487,12 @@ public static boolean isLocalhost(String remoteAddress) {
     }
     
     public static boolean isReachable(HostAndPort endpoint) {
+        // TODO Should we create an unconnected socket, and then use the calls below (see jclouds' InetSocketAddressConnect):
+        //      socket.setReuseAddress(false);
+        //      socket.setSoLinger(false, 1);
+        //      socket.setSoTimeout(timeout);
+        //      socket.connect(socketAddress, timeout);
```
neykov: Don't see the point of the above, but +1 for the `,conntect` timeout.

utils/common/src/test/java/org/apache/brooklyn/util/net/ReachableSocketFinderTest.java
```diff
@@ -0,0 +1,165 @@
...
+            fail("Expected failure, but got "+result);
+        } catch (NoSuchElementException e) {
+            // success
+        }
+    }
+
+    private ServerSocket connectToPort() throws Exception {
+        int port = findAvailablePort();
+        LOG.info("acquired port on "+port+" for test "+JavaClassNames.niceClassAndMethod());
+        return new ServerSocket(port);
```
neykov: Could use port 0 to avoid someone else stealing the port.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1653,8 +1662,10 @@ protected LoginCredentials createUser(ComputeService computeService, NodeMetadat
                         LOG.warn("exit code {} when creating user for {}; usage may subsequently fail", exitcode, node);
                     }
                 } finally {
-                    getManagementContext().getLocationManager().unmanage(sshLoc);
-                    Streams.closeQuietly(sshLoc);
+                    if (sshLoc != null) {
```
neykov: Can't be null

neykov:
Good to merge after addressing the potential NPE error, others are minor.

utils/common/src/main/java/org/apache/brooklyn/util/net/Networking.java
aledsage: Let's do the actual change in a different PR.

aledsage:
Thanks @neykov - merging now.

I didn't address your comment "Will be able to get rid of compoundFuture as well, leaving out the need for ListeningExecutorService (and could even use Brooklyn's task manager)." We could look at that in a future PR, but I'm not quite sure what you're proposing.

PR #1036, opened 2015-11-17
Remove catalog item parameters in line with type registry changes

neykov:
First commit somehow got excluded from the original PR.
Second commit removes the catalog item getParameters API as discussed in type registry PRs.  It was merged 8 days ago, so no point in going through a deprecation cycle.


ahgittin:
looks good; i'll sort out conflicts with #1017

PR #1037, opened 2015-11-17
Add quorum check enricher and aggregator

mikezaccardo:
This addition enables the creation of a sensor, defined in YAML, to set its value based on whether a quorum of entities has has the necessary source sensor value.

Example:

```
name: Quorum Check Test
location: localhost
services:
  - type: org.apache.brooklyn.entity.group.DynamicCluster
    name: Cluster
    
    brooklyn.config:
      initialSize: 5
    
    memberSpec:
        $brooklyn:entitySpec:
            type: org.apache.brooklyn.entity.software.base.EmptySoftwareProcess
            name: Cluster Member
            
    brooklyn.enrichers:
    - type: org.apache.brooklyn.enricher.stock.Aggregator
      brooklyn.config:
        uniqueTag: service-all-is-up-aggregator
        enricher.transformation.untyped: "isQuorate"
        quorum.check.type: "allAndAtLeastOne"
        quorum.total.size: $brooklyn:config("cluster.initial.size")
        enricher.aggregator.excludeBlank: true
        enricher.aggregating.fromMembers: true
        enricher.sourceSensor: $brooklyn:sensor("service.isUp")
        enricher.targetSensor: $brooklyn:sensor("cluster.isQuorate")
```

This creates a `cluster.isQuorate` sensor whose value depends on whether a quorum of cluster members have `service.isUp` equal to true.  In this example, all members must have a true value since `allAndAtLeastOne` is the specified `quorum.check.type`.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/enricher/stock/Aggregator.java
+++ b/core/src/main/java/org/apache/brooklyn/enricher/stock/Aggregator.java
@@ -35,6 +35,7 @@ import org.apache.brooklyn.core.BrooklynLogging;
 import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.util.collections.MutableList;
 import org.apache.brooklyn.util.collections.MutableMap;
+import org.apache.brooklyn.util.collections.QuorumCheck.QuorumChecks;
 import org.apache.brooklyn.util.core.flags.SetFromFlag;
 import org.apache.brooklyn.util.exceptions.Exceptions;
 import org.apache.brooklyn.util.text.StringPredicates;
@@ -63,6 +64,13 @@ public class Aggregator<T,U> extends AbstractAggregator<T,U> implements SensorEv
     public static final ConfigKey<Function<? super Collection<?>, ?>> TRANSFORMATION = ConfigKeys.newConfigKey(new TypeToken<Function<? super Collection<?>, ?>>() {}, "enricher.transformation");
     
     public static final ConfigKey<Boolean> EXCLUDE_BLANK = ConfigKeys.newBooleanConfigKey("enricher.aggregator.excludeBlank", "Whether explicit nulls or blank strings should be excluded (default false); this only applies if no value filter set", false);
+
+    /**
+     * @see QuorumChecks
+     */
+    public static final ConfigKey<String> QUORUM_CHECK_TYPE = ConfigKeys.newStringConfigKey("quorum.check.type", "The requirement to be considered quorate -- possible values: 'all', 'allAndAtLeastOne', 'atLeastOne', 'atLeastOneUnlessEmpty', 'alwaysHealthy'", "allAndAtLeastOne");
+
+    public static final ConfigKey<Integer> QUORUM_TOTAL_SIZE = ConfigKeys.newIntegerConfigKey("quorum.total.size", "The total size to consider when determining if quorate", 1);
 
     protected Sensor<T> sourceSensor;
     protected Function<? super Collection<T>, ? extends U> transformation;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/brooklyn/enricher/stock/Aggregator.java
```diff
@@ -64,6 +65,10 @@
     
     public static final ConfigKey<Boolean> EXCLUDE_BLANK = ConfigKeys.newBooleanConfigKey("enricher.aggregator.excludeBlank", "Whether explicit nulls or blank strings should be excluded (default false); this only applies if no value filter set", false);
 
+    public static final ConfigKey<String> QUORUM_CHECK_TYPE = ConfigKeys.newStringConfigKey("quorum.check.type", "The requirement to be considered quorate", "allAndAtLeastOne");
```
sjcorbett: Worth indicating the possible values in the description? Appreciate there's a chance it could get out of sync with the strings in QuorumCheck.

core/src/main/java/org/apache/brooklyn/enricher/stock/Enrichers.java
```diff
@@ -891,6 +892,39 @@ protected static int count(Iterable<? extends Object> vals, boolean includeNullV
...
+        public ComputingIsQuorate(TypeToken<T> typeToken, QuorumCheck quorumCheck, int totalSize) {
+            this.quorumCheck = quorumCheck;
+            this.totalSize = totalSize;
+
+            if (typeToken!=null && TypeToken.of(Boolean.class).isAssignableFrom(typeToken.getType())) {
+                this.typeToken = typeToken;
+            } else if (typeToken==null || typeToken.isAssignableFrom(Boolean.class)) {
+                this.typeToken = (TypeToken)TypeToken.of(Boolean.class);
+            } else {
+                throw new IllegalArgumentException("Type "+typeToken+" is not valid for "+this);
```
sjcorbett: Should indicate what was expected.

core/src/main/java/org/apache/brooklyn/enricher/stock/Enrichers.java
```diff
@@ -891,6 +892,39 @@ protected static int count(Iterable<? extends Object> vals, boolean includeNullV
         return result;
     }
     
+    @Beta
+    public static class ComputingIsQuorate<T extends Object> implements Function<Collection<T>, T> {
```
sjcorbett: Had to compare this to existing enrichers to clock the logic. Why `T extends Object` and not just `Boolean`? You've got `(T) Boolean.valueOf()` in `apply` but the Boolean class is `final` so `T` can't be anything but a Boolean.

mikezaccardo: Changing this definition to `public static class ComputingIsQuorate implements Function<Collection<Boolean>, Boolean>` is certainly sensible.

However, doing this causes problems in `Aggregator`.  The return value must ultimately be assigned to `Function<? super Collection<T>, ? extends U> transformation` but I cannot seem to escape compile errors when attempting to cast / assign the value.

sjcorbett: I can't escape the generics either. However, I can cast `ComputingIsQuorate` to `(Function<? super Collection<?>, ?>)` in Aggregator! Personally I prefer keeping ComputingIsQuorate simple and the ugliness in `lookupTransformation`.

mikezaccardo:
@sjcorbett I've updated the signature to `public static class ComputingIsQuorate<T> implements Function<Collection<Boolean>, Boolean>` without any errors.

The generic `T` remains for the `TypeToken` constructor parameter -- we check to see if it is compatible with `Boolean` there.

PR #1038, opened 2015-11-17
Add YAML port range support

mikezaccardo:
All values supported by `org.apache.brooklyn.core.location.PortRanges.fromString(String)` can now be included in `inboundPorts` in YAML blueprints.

Example:

```
name: "Port Range Example"
location: <cloud provider>

services:
- type: org.apache.brooklyn.entity.software.base.EmptySoftwareProcess
  name: VM
  
  provisioning.properties:
    inboundPorts:
    - 88
    - 100-200
    - 8080+
```

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -24,6 +24,8 @@ import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.apache.brooklyn.util.JavaGroovyEquivalents.elvis;
 import static org.apache.brooklyn.util.JavaGroovyEquivalents.groovyTruth;
 import static org.apache.brooklyn.util.ssh.BashCommands.sbinPath;
+import io.cloudsoft.winrm4j.pywinrm.Session;
+import io.cloudsoft.winrm4j.pywinrm.WinRMFactory;
 
 import java.io.ByteArrayOutputStream;
 import java.io.File;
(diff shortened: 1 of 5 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -3004,6 +3006,14 @@ protected static double toDouble(Object v) {
         }
     }
 
+    protected static int[] toIntPortArray(Object v) {
+        Collection<String> inboundPortEntries = toListOfStrings(v);
```
neykov: Not fully backwards compatible. `toIntArray` supports the format `[1,2,3]`. Also accepts int arrays if passed from Java code. The first one should definitely be supported, no hard opinion on the second one.
Would be nice to have a test for the new method.

Otherwise looks very useful.

ahgittin:
also see:

    PortRanges.fromString
    StringEscapes.unwrapJsonishListIfPossible

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
mikezaccardo: Thanks, @neykov.  I will add support for the `[1, 2, 3]` format and Java int arrays.  I'll also add test(s) in `JcloudsLocationTest`.

mikezaccardo:
@neykov @ahgittin I expanded the `fromIterable` method to support all of the previously supported formats, added some additional formats that previously would not work, and of course preserved port ranges.  There are tests that verify all of these.

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/JcloudsLocationTest.java
```diff
@@ -47,6 +49,8 @@
 import org.jclouds.scriptbuilder.domain.OsFamily;
 import org.jclouds.scriptbuilder.domain.StatementList;
 import org.mockito.Mockito;
+import org.python.google.common.collect.Lists;
+import org.python.google.common.primitives.Ints;
```
sjcorbett: Wrong imports. Use `com.google.common.collect`.

sjcorbett: And `com.google.common.primitives.Ints`

sjcorbett:
Looks good, will merge.

PR #1039, opened 2015-11-18
Adds support for interfaces tag on catalog items

tbouron:
This adds, for each catalog item, all its implemented interfaces as a tag. It also exposed catalog items' tags on the REST API.

This is based on the code provided by #1020

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
@@ -558,7 +558,7 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
             throw new IllegalStateException("Could not resolve plan once id and itemType are known (recursive reference?): "+sourceYaml);
         }
         String sourcePlanYaml = planInterpreter.getPlanYaml();
-        
+
         CatalogItemDtoAbstract<?, ?> dto = createItemBuilder(itemType, symbolicName, version)
             .libraries(libraryBundles)
             .displayName(displayName)
@@ -859,6 +859,7 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
         log.debug("Adding manual catalog item to "+mgmt+": "+yaml);
         checkNotNull(yaml, "yaml");
         List<CatalogItemDtoAbstract<?, ?>> result = collectCatalogItems(yaml);
+
         // do this at the end for atomic updates; if there are intra-yaml references, we handle them specially
         for (CatalogItemDtoAbstract<?, ?> item: result) {
             addItemDto(item, forceUpdate);
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
@@ -19,10 +19,12 @@
 package org.apache.brooklyn.core.mgmt;
 
 import java.io.Serializable;
+import java.util.List;
 
 import org.codehaus.jackson.annotate.JsonProperty;
 
 import com.google.common.annotations.Beta;
+import com.google.common.base.Objects;
 
 /** @since 0.7.0 some strongly typed tags for reference; note these may migrate elsewhere! */
 @Beta
@@ -51,9 +53,42 @@ public class BrooklynTags {
             return contents;
         }
     }
+
+    public static class InterfacesTag {
+        @JsonProperty
+        final List<Class<?>> interfaces;
+
+        public InterfacesTag(List<Class<?>> interfaces) {
+            this.interfaces = interfaces;
+        }
+
+        public List<Class<?>> getInterfaces() {
+            return interfaces;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+
+            InterfacesTag that = (InterfacesTag) o;
+
+            return interfaces == null
+                    ? that.interfaces == null
+                    : interfaces.equals(that.interfaces);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hashCode(interfaces);
+        }
+    }
     
     public static NamedStringTag newYamlSpecTag(String contents) { return new NamedStringTag(YAML_SPEC_KIND, contents); }
     public static NamedStringTag newNotesTag(String contents) { return new NamedStringTag(NOTES_KIND, contents); }
+    public static InterfacesTag newInterfacesTag(List<Class<?>> interfaces) {
+        return new InterfacesTag(interfaces);
+    }
     
     public static NamedStringTag findFirst(String kind, Iterable<Object> tags) {
         for (Object object: tags) {
--- a/usage/rest-api/src/main/java/org/apache/brooklyn/rest/domain/CatalogItemSummary.java
+++ b/usage/rest-api/src/main/java/org/apache/brooklyn/rest/domain/CatalogItemSummary.java
@@ -20,7 +20,10 @@ package org.apache.brooklyn.rest.domain;
 
 import java.io.Serializable;
 import java.net.URI;
+import java.util.Collection;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.apache.commons.lang.builder.EqualsBuilder;
 import org.codehaus.jackson.annotate.JsonIgnoreProperties;
@@ -29,6 +32,7 @@ import org.codehaus.jackson.map.annotate.JsonSerialize;
 import org.codehaus.jackson.map.annotate.JsonSerialize.Inclusion;
 
 import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
 /** variant of Catalog*ItemDto objects for JS/JSON serialization;
@@ -55,6 +59,8 @@ public class CatalogItemSummary implements HasId, HasName, Serializable {
     @JsonSerialize(include=Inclusion.NON_EMPTY)
     private final String iconUrl;
     private final String planYaml;
+    @JsonSerialize(include=Inclusion.NON_EMPTY)
+    private final List<Object> tags;
     private final boolean deprecated;
     
     private final Map<String, URI> links;
@@ -67,6 +73,7 @@ public class CatalogItemSummary implements HasId, HasName, Serializable {
             @JsonProperty("planYaml") String planYaml,
             @JsonProperty("description") String description,
             @JsonProperty("iconUrl") String iconUrl,
+            @JsonProperty("tags") Set<Object> tags,
             @JsonProperty("deprecated") boolean deprecated,
             @JsonProperty("links") Map<String, URI> links
             ) {
@@ -79,6 +86,7 @@ public class CatalogItemSummary implements HasId, HasName, Serializable {
         this.planYaml = planYaml;
         this.description = description;
         this.iconUrl = iconUrl;
+        this.tags = (tags == null) ? ImmutableList.of() : ImmutableList.copyOf(tags);
         this.links = (links == null) ? ImmutableMap.<String, URI>of() : ImmutableMap.copyOf(links);
         this.deprecated = deprecated;
     }
@@ -121,6 +129,10 @@ public class CatalogItemSummary implements HasId, HasName, Serializable {
         return iconUrl;
     }
 
+    public Collection<Object> getTags() {
+        return tags;
+    }
+
     public Map<String, URI> getLinks() {
         return links;
     }
@@ -140,7 +152,7 @@ public class CatalogItemSummary implements HasId, HasName, Serializable {
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(symbolicName, version, name, javaType, deprecated);
+        return Objects.hashCode(symbolicName, version, name, javaType, tags, deprecated);
     }
     
     @Override
--- a/usage/rest-server/src/main/java/org/apache/brooklyn/rest/transform/CatalogTransformer.java
+++ b/usage/rest-server/src/main/java/org/apache/brooklyn/rest/transform/CatalogTransformer.java
@@ -35,6 +35,7 @@ import org.apache.brooklyn.api.policy.Policy;
 import org.apache.brooklyn.api.policy.PolicySpec;
 import org.apache.brooklyn.api.sensor.Sensor;
 import org.apache.brooklyn.core.entity.EntityDynamicType;
+import org.apache.brooklyn.core.mgmt.BrooklynTags;
 import org.apache.brooklyn.core.objs.BrooklynTypes;
 import org.apache.brooklyn.rest.domain.CatalogEntitySummary;
 import org.apache.brooklyn.rest.domain.CatalogItemSummary;
@@ -48,7 +49,9 @@ import org.apache.brooklyn.rest.domain.SensorSummary;
 import org.apache.brooklyn.rest.domain.SummaryComparators;
 import org.apache.brooklyn.rest.util.BrooklynRestResourceUtils;
 import org.apache.brooklyn.util.collections.MutableMap;
+import org.apache.brooklyn.util.collections.MutableSet;
 import org.apache.brooklyn.util.exceptions.Exceptions;
+import org.apache.commons.lang.ClassUtils;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.ImmutableSet;
@@ -63,9 +66,10 @@ public class CatalogTransformer {
         Set<SensorSummary> sensors = Sets.newTreeSet(SummaryComparators.nameComparator());
         Set<EffectorSummary> effectors = Sets.newTreeSet(SummaryComparators.nameComparator());
 
+        EntitySpec<?> spec = null;
+
         try {
-            @SuppressWarnings({ "unchecked", "rawtypes" })
-            EntitySpec<?> spec = (EntitySpec<?>) b.getCatalog().createSpec((CatalogItem) item);
+            spec = (EntitySpec<?>) b.getCatalog().createSpec((CatalogItem) item);
             EntityDynamicType typeMap = BrooklynTypes.getDefinedEntityType(spec.getType());
             EntityType type = typeMap.getSnapshot();
 
@@ -75,7 +79,7 @@ public class CatalogTransformer {
                 sensors.add(SensorTransformer.sensorSummaryForCatalog(x));
             for (Effector<?> x: type.getEffectors())
                 effectors.add(EffectorTransformer.effectorSummaryForCatalog(x));
-            
+
         } catch (Exception e) {
             Exceptions.propagateIfFatal(e);
             
@@ -91,7 +95,7 @@ public class CatalogTransformer {
         return new CatalogEntitySummary(item.getSymbolicName(), item.getVersion(), item.getDisplayName(),
             item.getJavaType(), item.getPlanYaml(),
             item.getDescription(), tidyIconLink(b, item, item.getIconUrl()),
-            config, sensors, effectors,
+            makeTags(spec, item), config, sensors, effectors,
             item.isDeprecated(), makeLinks(item));
     }
 
@@ -115,7 +119,7 @@ public class CatalogTransformer {
         }
         return new CatalogItemSummary(item.getSymbolicName(), item.getVersion(), item.getDisplayName(),
             item.getJavaType(), item.getPlanYaml(),
-            item.getDescription(), tidyIconLink(b, item, item.getIconUrl()), item.isDeprecated(), makeLinks(item));
+            item.getDescription(), tidyIconLink(b, item, item.getIconUrl()), item.tags().getTags(), item.isDeprecated(), makeLinks(item));
     }
 
     public static CatalogPolicySummary catalogPolicySummary(BrooklynRestResourceUtils b, CatalogItem<? extends Policy,PolicySpec<?>> item) {
@@ -123,7 +127,7 @@ public class CatalogTransformer {
         return new CatalogPolicySummary(item.getSymbolicName(), item.getVersion(), item.getDisplayName(),
                 item.getJavaType(), item.getPlanYaml(),
                 item.getDescription(), tidyIconLink(b, item, item.getIconUrl()), config,
-                item.isDeprecated(), makeLinks(item));
+                item.tags().getTags(), item.isDeprecated(), makeLinks(item));
     }
 
     public static CatalogLocationSummary catalogLocationSummary(BrooklynRestResourceUtils b, CatalogItem<? extends Location,LocationSpec<?>> item) {
@@ -131,7 +135,7 @@ public class CatalogTransformer {
         return new CatalogLocationSummary(item.getSymbolicName(), item.getVersion(), item.getDisplayName(),
                 item.getJavaType(), item.getPlanYaml(),
                 item.getDescription(), tidyIconLink(b, item, item.getIconUrl()), config,
-                item.isDeprecated(), makeLinks(item));
+                item.tags().getTags(), item.isDeprecated(), makeLinks(item));
     }
 
     protected static Map<String, URI> makeLinks(CatalogItem<?,?> item) {
@@ -158,6 +162,20 @@ public class CatalogTransformer {
         if (b.isUrlServerSideAndSafe(iconUrl))
             return "/v1/catalog/icon/"+item.getSymbolicName() + "/" + item.getVersion();
         return iconUrl;
+    }
+
+    private static Set<Object> makeTags(EntitySpec<?> spec, CatalogItem<?, ?> item) {
+        Class<?> type = null;
+        if (spec.getImplementation() != null) {
+            type = spec.getImplementation();
+        } else {
+            type = spec.getType();
+        }
+
+        return MutableSet.builder()
+                .addAll(item.tags().getTags())
+                .add(new BrooklynTags.InterfacesTag(ClassUtils.getAllInterfaces(type)))
+                .build();
     }
     
 }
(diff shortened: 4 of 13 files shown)


sjcorbett:
The test failures appear valid. When running locally:
```
Failed tests:
  CatalogYamlEntityTest.testUpdatingItemAllowedIfSame:504->addCatalogOSGiEntity:862->addCatalogOSGiEntity:866->addCatalogOSGiEntity:870->AbstractYamlTest.addCatalogItems:144->AbstractYamlTest.addCatalogItems:148 » IllegalState
  CatalogYamlVersioningTest.testAddSameVersionFailsWhenIconIsDifferent:73->addCatalogEntity:249->addCatalogEntity:253->addCatalogEntity:257->AbstractYamlTest.addCatalogItems:144->AbstractYamlTest.addCatalogItems:148 » IllegalState
```

tbouron:
Tests have been fixed @sjcorbett

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
```diff
@@ -51,9 +54,35 @@ public String getContents() {
...
+            this.interfaces = interfaces;
+        }
+
+        public List<Class<?>> getInterfaces() {
+            return interfaces;
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            return EqualsBuilder.reflectionEquals(this, obj);
```
sjcorbett: Surely this is overkill for what can be just written as ~ `obj == this || (obj.getClass().equals(getClass()) && InterfacesTag.class.cast(obj).interfaces.equals(this.interfaces))`?

usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityTest.java
```diff
@@ -18,15 +18,14 @@
  */
 package org.apache.brooklyn.camp.brooklyn.catalog;
 
-import static org.testng.Assert.assertEquals;
-import static org.testng.Assert.assertTrue;
-import static org.testng.Assert.fail;
+import static org.testng.Assert.*;
```
sjcorbett: Please configure your IDE to not do this.

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
tbouron: `interfaces` is a `List`, doing a simple equals like the one you suggest don't work.

sjcorbett: Why not? Javadoc says "In other words, two lists are defined to be equal if they contain the same elements in the same order. This definition ensures that the equals method works properly across different implementations of the List interface."

tbouron: Well, this is not what I saw on my machine, that's why I used this `EqualsBuilder`

usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityTest.java
tbouron: Can't find the required configuration, do you have the link somewhere?

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
sjcorbett: Can you give an example that fails?

usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityTest.java
sjcorbett: What IDE do you use? Here's IntelliJ: http://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
tbouron: The same you gave earlier
```
Failed tests:
  CatalogYamlEntityTest.testUpdatingItemAllowedIfSame:504->addCatalogOSGiEntity:862->addCatalogOSGiEntity:866->addCatalogOSGiEntity:870->AbstractYamlTest.addCatalogItems:144->AbstractYamlTest.addCatalogItems:148 » IllegalState
  CatalogYamlVersioningTest.testAddSameVersionFailsWhenIconIsDifferent:73->addCatalogEntity:249->addCatalogEntity:253->addCatalogEntity:257->AbstractYamlTest.addCatalogItems:144->AbstractYamlTest.addCatalogItems:148 » IllegalState
```

usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityTest.java
tbouron: Thanks, I'll change my configuration

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
sjcorbett: I meant the contents of the lists.

tbouron: No I don't have it unfortunately.

tbouron:
Previous comments addressed @sjcorbett

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
```diff
@@ -51,9 +53,40 @@ public String getContents() {
...
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+
+            InterfacesTag that = (InterfacesTag) o;
+
+            return !(interfaces != null ? !interfaces.equals(that.interfaces) : that.interfaces != null);
```
sjcorbett: All that negation is a bit of a confusing way to phrase it ☺ How about:
```
            return interfaces == null 
                            ? that.interfaces == null 
                            : interfaces.equals(that.interfaces);
```

usage/rest-api/src/main/java/org/apache/brooklyn/rest/domain/CatalogItemSummary.java
```diff
@@ -121,6 +132,20 @@ public String getIconUrl() {
         return iconUrl;
     }
 
+    public Collection<Object> getTags() {
+        List<Object> result = new ArrayList<Object>();
+        for (Object t : tags) {
+            // TODO if we had access to a mapper we could use it to give better json
```
sjcorbett: Can you expand on your comment? It's not really the job of a summary class to configure mappers but technically it does have access to one. Should you be adding another class to `org.apache.brooklyn.rest.transform`?

usage/rest-api/src/main/java/org/apache/brooklyn/rest/domain/CatalogItemSummary.java
```diff
@@ -121,6 +132,20 @@ public String getIconUrl() {
...
+            result.add(Jsonya.convertToJsonPrimitive(t));
+        }
+        return result;
+    }
+
+    @JsonIgnore
+    public Collection<Object> getRawTags() {
```
sjcorbett: Is this used at all?

usage/rest-api/src/main/java/org/apache/brooklyn/rest/domain/CatalogItemSummary.java
```diff
@@ -132,6 +157,7 @@ public boolean isDeprecated() {
     @Override
     public String toString() {
         return Objects.toStringHelper(this)
+                .add("tags", tags)
```
sjcorbett: Unsure this addition is useful.

usage/rest-api/src/main/java/org/apache/brooklyn/rest/domain/CatalogItemSummary.java
```diff
@@ -140,7 +166,7 @@ public String toString() {
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(symbolicName, version, name, javaType, deprecated);
+        return Objects.hashCode(symbolicName, version, name, javaType, tags, deprecated);
```
sjcorbett: Ah, I see where EqualsBuilder came from. I guess when a class uses `EqualsBuilder` we should also use `HashCodeBuilder`. (Or, in my preference, not use `EqualsBuilder`!) This class disobeys the equals/hashcode contract. Not your fault, of course.

usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityTest.java
```diff
@@ -800,6 +803,49 @@ public void testHardcodedCatalog() throws Exception {
...
+
+        assertEquals(catalogItem.tags().getTags().size(), 1);
+        for (Object tag : catalogItem.tags().getTags()) {
+            assertTrue(tag instanceof BrooklynTags.InterfacesTag);
+            assertEquals(((BrooklynTags.InterfacesTag) tag).getInterfaces(), ClassUtils.getAllInterfaces(BasicEntity.class));
+        }
+
+        deleteCatalogEntity(symbolicName);
+    }
+
```
sjcorbett: Can you include a test that checks the interfaces of an entity that is defined in a bundle? You can follow the pattern in `CatalogOsgiVersionMoreEntityTest` and check for the `MoreEntity` interface.

sjcorbett: MoreEntity: https://github.com/apache/incubator-brooklyn/blob/master/utils/rt-osgi/src/test/dependencies/osgi/more-entities-v2/src/main/java/org/apache/brooklyn/test/osgi/entities/more/MoreEntity.java

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -558,15 +561,34 @@ private void collectCatalogItems(String sourceYaml, Map<?,?> itemMetadata, List<
...
+            .plan(sourcePlanYaml);
+        if (itemType.equals(CatalogItemType.ENTITY) || itemType.equals(CatalogItemType.TEMPLATE)) {
+            String stringServiceType = getFirstAs(itemAsMap, String.class, "type").orNull();
+            // This is not a single item but a services based YAML
+            if (stringServiceType == null) {
+                final List<Map<?, ?>> services = getFirstAs(itemAsMap, List.class, "services").orNull();
+                if (services.isEmpty()) {
+                    throw new IllegalStateException("Services are empty: " + sourceYaml);
+                }
+                stringServiceType = services.get(0).get("type").toString();
```
sjcorbett: Why only look at the first service? We could argue between all or none but just one surprises me.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -558,15 +561,34 @@ private void collectCatalogItems(String sourceYaml, Map<?,?> itemMetadata, List<
...
+            }
+            Class<?> serviceType = JavaBrooklynClassLoadingContext.create(mgmt).tryLoadClass(stringServiceType).orNull();
+            if (serviceType == null) {
+                final CatalogItem<?, ?> relatedCatalogItem = getCatalogItem(stringServiceType, "");
+                if (relatedCatalogItem != null) {
+                    serviceType = JavaBrooklynClassLoadingContext.create(mgmt).tryLoadClass(relatedCatalogItem.getJavaType()).orNull();
+                }
+            }
+            itemBuilder.tag(BrooklynTags.newInterfacesTag(ClassUtils.getAllInterfaces(serviceType)));
```
sjcorbett: Technically Brooklyn gives you `Reflections.getAllInterfaces(Class<?>)`. I don't know what its benefit over `ClassUtils` is!

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -558,15 +561,34 @@ private void collectCatalogItems(String sourceYaml, Map<?,?> itemMetadata, List<
...
+                if (services.isEmpty()) {
+                    throw new IllegalStateException("Services are empty: " + sourceYaml);
+                }
+                stringServiceType = services.get(0).get("type").toString();
+            }
+            Class<?> serviceType = JavaBrooklynClassLoadingContext.create(mgmt).tryLoadClass(stringServiceType).orNull();
+            if (serviceType == null) {
+                final CatalogItem<?, ?> relatedCatalogItem = getCatalogItem(stringServiceType, "");
+                if (relatedCatalogItem != null) {
+                    serviceType = JavaBrooklynClassLoadingContext.create(mgmt).tryLoadClass(relatedCatalogItem.getJavaType()).orNull();
```
sjcorbett: Not convinced by the use of `JavaBrooklynClassLoadingContext`. What happens when the service is in a bundle? @neykov Is a suitable `OsgiBrooklynClassLoadingContext` available in this method (BasicBrooklynCatalog.collectCatalogItems)? Or perhaps you could check the catalogue item's bundles too.

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
tbouron: Just FYI, that's the standard Intellij `equals` code generation. I'll change it.

usage/rest-api/src/main/java/org/apache/brooklyn/rest/domain/CatalogItemSummary.java
tbouron: This method was taken [from there](https://github.com/apache/incubator-brooklyn/blob/master/usage/rest-api/src/main/java/org/apache/brooklyn/rest/domain/TaskSummary.java#L137-L144). Although, simply returning the tags seems to work as expected. I'll change it.

usage/rest-api/src/main/java/org/apache/brooklyn/rest/domain/CatalogItemSummary.java
tbouron: Not directly in this project but in [brooklyn-tosca](https://github.com/cloudsoft/brooklyn-tosca/pull/30/files#diff-7bb08b08e5b7ef400557dbcb45441969R157). Although, it can be removed as per as my previous comment.

usage/rest-api/src/main/java/org/apache/brooklyn/rest/domain/CatalogItemSummary.java
tbouron: I thought it would be useful to have it in the `toString()` method but I don't have a strong feeling about it. I can remove it if you want.

usage/rest-api/src/main/java/org/apache/brooklyn/rest/domain/CatalogItemSummary.java
tbouron: Not sure how to fix it then

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
tbouron: The YAML let you add more than one services but Brooklyn will ultimately [deploys the entity only if it contains only one service](https://issues.apache.org/jira/browse/BROOKLYN-195?jql=project%20%3D%20BROOKLYN). That's why I'm checking only the first one.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
tbouron: [The code seems to be the same](https://git-wip-us.apache.org/repos/asf?p=commons-lang.git;a=blob;f=src/main/java/org/apache/commons/lang3/ClassUtils.java;h=0a7480ff8d991e555be679f86377dc19d3eb64d7;hb=HEAD#l433) as `Reflections.getAllInterfaces(Class<?>)`. Although, the `Reflections` class says in its headers:
>  Reflection utilities ("borrowed" from cloudsoft monterey).

So I would still use `ClassUtils`.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
tbouron: There is also another issue: If the `stringServiceType` is a reference to another catalog item defined in the same YAML **but after** the one been processed, we cannot get its type and therefore cannot get its interfaces. Do you have a solution for that?

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
tbouron: You suggestion cannot work, `that` must be checked to be the same class as `this`

usage/rest-api/src/main/java/org/apache/brooklyn/rest/domain/CatalogItemSummary.java
sjcorbett: It's one for a separate pull request.

usage/rest-api/src/main/java/org/apache/brooklyn/rest/domain/CatalogItemSummary.java
sjcorbett: In the spirit of keeping things concise I'd remove it but it's a minor one. Keep it if you feel strongly.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
sjcorbett: Sounds like a good test case. Your best shot would be to set the tags when all items have been collected - e.g. at the end of `addItems(String, boolean)`.

tbouron:
@sjcorbett Comments addressed

sjcorbett:
@tbouron Looks good. Would you mind squashing? Will then double check tests and merge.

tbouron:
@sjcorbett Squash done!

api/src/main/java/org/apache/brooklyn/api/catalog/CatalogItem.java
```diff
@@ -103,6 +103,11 @@ public static CatalogItemType ofTargetClass(Class<? extends BrooklynObject> type
     @Deprecated
     @Nullable public String getJavaType();
 
+    /**
+     * @return The underlying java type of the item represented or the catalog item ID, described via a YAML spec.
+     */
+    @Nullable public String getReferenceType();
```
neykov: How is this different from `getJavaType` which is being removed?

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -566,6 +575,7 @@ private void collectCatalogItems(String sourceYaml, Map<?,?> itemMetadata, List<
             .deprecated(catalogDeprecated)
             .iconUrl(catalogIconUrl)
             .plan(sourcePlanYaml)
+            .referenceType(referenceType)
```
neykov: What if the type is another catalog item, my understanding is that this must be a class name?

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -859,6 +869,34 @@ private String makeAsIndentedObject(String yaml) {
         log.debug("Adding manual catalog item to "+mgmt+": "+yaml);
         checkNotNull(yaml, "yaml");
         List<CatalogItemDtoAbstract<?, ?>> result = collectCatalogItems(yaml);
+
+        // Iterate over the results to add the interfaces tag. We are doing this here to be able to get the backward-types.
+        for (final CatalogItemDtoAbstract<?, ?> item : result) {
+            if (item.getCatalogItemType().equals(CatalogItemType.ENTITY) || item.getCatalogItemType().equals(CatalogItemType.TEMPLATE)) {
+                Class<?> type = CatalogUtils.newClassLoadingContext(mgmt, item).tryLoadClass(item.getReferenceType()).orNull();
```
neykov: I think it would be easier to create a spec out of the catalog item and get the type from there. What if referenceType is another catalog item.

api/src/main/java/org/apache/brooklyn/api/catalog/CatalogItem.java
nakomis: The difference is that it may refer either to a java type, or to a catalog item, so getJavaType is the wrong name for it, so it makes sense to deprecate getJavaType and replace it with getReferenceType (this also addresses the other comment below)

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
tbouron: See @nakomis comment above

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
tbouron: That is the purpose of the lines right below 877 to 892

usage/rest-server/src/main/java/org/apache/brooklyn/rest/transform/CatalogTransformer.java
```diff
@@ -91,7 +91,7 @@
         return new CatalogEntitySummary(item.getSymbolicName(), item.getVersion(), item.getDisplayName(),
             item.getJavaType(), item.getPlanYaml(),
             item.getDescription(), tidyIconLink(b, item, item.getIconUrl()),
-            config, sensors, effectors,
+            item.tags().getTags(), config, sensors, effectors,
```
neykov: The tags are generated at the time the catalog item is created. In the mean time the dependent catalog items may change, leading to changes in the interfaces implemented by this catalog item. Is this acceptable? Could generate them on the fly here.

nakomis: We did consider that, but creating them at the point when the catalog item is generated seemed like an overhead for something that's unlikely to change, and this way the data stored in the catalog will be consistent. Happy to change, but for now I think it's acceptable

neykov:
Thanks for the quick responses @tbouron, @nakomis. I still think that you should create a spec out of the catalog item and use the java type from there. You are essentially replicating the same logic, but doing it ad-hoc. Just a couple of examples where it would fail - any references to the newly introduced type registry won't be handled properly; there are various key words to specify the service type (type, serviceType, service_type).

As for the eager vs lazy evaluation a valid case where it would change is if I it's depeneding on the catalog item `tomcat-server` and a new version of the catalog item is introduced, implementing a new interface. This won't get reflected in all catalog items using it. I agree it's a matter of personal opinion to some extent (weighting pros and cons of eagor vs lazy) so I can agree to the current approach as well.

tbouron:
Ok, I have an idea that might suit everyone. We keep the newly introduced `referenceType` in the `CatalogItem` and we populate it with the fully qualified class name or the catalog item ID (same as what this PR do currently). For every catalog item, we still add the `InterfacesTags` **but only** if the `referenceType` is a fully qualified class name.

Now, we modify the `CatalogTransfomer` to get all underlying interfaces if the current item has a reference to another catalog item, and so on. Like that, we keep the best of both worlds:
- Interfaces are stored upon catalog item addition.
- REST API returns a consolidated list of all interfaces, build at runtime.

WDYT @neykov @sjcorbett @nakomis ?

tbouron:
FYI, I have a working version on my machine about the changes I described in my previous comment. I just want to make sure that this approach is acceptable before pushing it.

neykov:
@tbouron Seems I'm missing some context here - I don't understand why you prefer do load the class yourself instead of re-using the existing infrastructure to do it, which would be much easier. If that's the case, happy to be ignored.

tbouron:
@neykov I cannot use the entitySpec as the catalog item has not been created yet at that point, which is why I need to go through ` CatalogUtils.newClassLoadingContext(mgmt, dto).tryLoadClass(referenceType)`

tbouron:
I talked privately with @neykov and it turns out that creating the `InterfacesTag` and adding it to the `CatalogItemSummary` at "REST API call time" is:
* really cheap as we just iterate over the supertypes, don't use reflection.
* simpler as we don't need to do extra parsing for the type or the referenced catalog item.

Therefore, I'll change the code to do that.

Hopefully, this time will be the last time! :+1:

tbouron:
Commit squashed and pushed. Poke @sjcorbett @neykov

sjcorbett:
Scanned through, looks ok to me. Can you rebase master please?

usage/rest-server/src/main/java/org/apache/brooklyn/rest/transform/CatalogTransformer.java
```diff
@@ -91,7 +95,7 @@
         return new CatalogEntitySummary(item.getSymbolicName(), item.getVersion(), item.getDisplayName(),
             item.getJavaType(), item.getPlanYaml(),
             item.getDescription(), tidyIconLink(b, item, item.getIconUrl()),
-            config, sensors, effectors,
+            makeTags(spec, item), config, sensors, effectors,
```
neykov: spec could be null here

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
```diff
@@ -51,9 +53,42 @@ public String getContents() {
             return contents;
         }
     }
+
+    public static class InterfacesTag {
+        @JsonProperty
+        final List<Class<?>> interfaces;
```
neykov: Have you looked at the persistence files, how are classes serialized? Could be leaking a lot of state. Does it rebind cleanly if classes are coming from osgi bundles?
Consider converting to strings.

PR #1040, opened 2015-11-18
[BROOKLYN-185] Upgrade swagger to version 1.5.3

CMoH:


hzbarcea:
Build ok, tests pass.
Manually tested the swagger rest ui on [localhost](http://localhost:8081/assets/html/swagger-ui.html#/), all good.

@CMoH, @m4rkmckenna , excellent work.

hzbarcea:
Reviewed. Looks good.

One minor comment is that the "@author" tags are discouraged, as in time they tend to not reflect the collaborative reality. They are everywhere, so at some point we should probably get rid of them, like many other projects did. Not a priority.

CMoH:
I've added a detailed description in the commit where I disabled the tests (cc56bb27cda7240fc5db8d992c49f6b9873da9fb)

hzbarcea:
Doh, makes sense. I will create a jira to track that. Merging shortly...

PR #1041, opened 2015-11-18
Updates TomcatServer download.url as nextag URL is no longer available

nakomis:


sjcorbett:
Nextag is still available: http://download.nextag.com/apache/tomcat/tomcat-7/v7.0.65/bin/apache-tomcat-7.0.65.tar.gz.

Don't know whether we should prefer this or its replacement.

nakomis:
Hmm, that link wasn't available when I created this PR. Happy to close now it's up again

sjcorbett:
Could you close this?

PR #1042, opened 2015-11-18
Updating Validation api version

kiuby88:
Updating validation api.
I am adding [brooklyn-tosca-transformer](https://github.com/cloudsoft/brooklyn-tosca/tree/master/brooklyn-tosca-transformer) to incubator-brooklyn. I built a uber jar and I added it to dropins folder, but I found a problem with this dependency so I have updated it.


aledsage:
Looks good; thanks @kiuby88 - merging.

kiuby88:
Thanks a lot

PR #1043, opened 2015-11-18
Allows ScheduledTasks to be resubmitted even when the task throws an exception.

sjcorbett:
I had to spend some time last week debugging an entity that was mysteriously on fire even though its driver's `isRunning` method returned true (verified via Groovy console).

After some log searching I realised that roughly the following scenario had transpired:
* Entity running
* Network connectivity began affecting Brooklyn VM. Several entities were on fire.
* One entity's SoftwareProcessImpl.serviceProcessIsRunning feed SSHed to its VM. The SSH task threw an exception because of the network issue.
* The entity's serviceProcessIsRunning feed was cancelled.
* The network problem finished. The entity's status could not recover.

This PR adds a flag to `ScheduledTask` to indicate that its task should be resubmitted even after throwing an exception. The `Poller` class has been configured use this flag.


usage/jsgui/src/test/java/org/apache/brooklyn/rest/jsgui/BrooklynJavascriptGuiLauncher.java
```diff
@@ -49,7 +49,7 @@
     public static void main(String[] args) throws Exception {
         // NOTE: When running Brooklyn from an IDE (i.e. by launching BrooklynJavascriptGuiLauncher.main())
         // you will need to ensure that the working directory is set to the jsgui folder. For IntelliJ,
-        // set the 'Working directory' of the Run/Debug Configuration to $MODULE_DIR/../jsgui.
+        // set the 'Working directory' of the Run/Debug Configuration to $MODULE_DIR$/../jsgui.
```
aledsage: Not sure I follow why you make this change. The previous syntax in the comment was assuming that MODULE_DIR was an environment variable, so why surround it with a `$` on each side?

But I don't care :-)

sjcorbett: It's an IntelliJ specific thing: https://www.jetbrains.com/idea/help/path-variables.html

core/src/main/java/org/apache/brooklyn/util/core/task/ScheduledTask.java
```diff
@@ -84,12 +99,15 @@ public ScheduledTask(Map flags, Callable<Task<?>> taskFactory) {
         delay = Duration.of(elvis(flags.remove("delay"), 0));
         period = Duration.of(elvis(flags.remove("period"), null));
         maxIterations = elvis(flags.remove("maxIterations"), null);
+        Object cancelFlag = flags.remove("cancelOnException");
+        cancelOnException = cancelFlag == null || Boolean.TRUE.equals(cancelFlag);
```
aledsage: Should we do a type coercion? No strong feelings. This deep in the code, we can assume folk are using the right types.

core/src/main/java/org/apache/brooklyn/util/core/task/ScheduledTask.java
```diff
@@ -121,13 +145,17 @@ public ScheduledTask maxIterations(int val) {
...
         StringBuilder rv = new StringBuilder("Scheduler");
-        if (runCount>0) rv.append(", iteration "+(runCount+1));
-        if (recentRun!=null) rv.append(", last run "+
-            Duration.sinceUtc(recentRun.getStartTimeUtc())+" ms ago");
+        if (runCount > 0) {
+            rv.append(", iteration ").append(runCount + 1);
+        }
+        if (recentRun != null) {
+            Duration start = Duration.sinceUtc(recentRun.getStartTimeUtc());
+            rv.append(", last run ").append(start).append(" ms ago");
```
aledsage: The toString of the duration will include the units, so remove the "ms"

aledsage:
@sjcorbett looks really good! Only some very minor comments. Please merge once you've looked at those, and decided whether to incorporate/ignore.

core/src/main/java/org/apache/brooklyn/util/core/task/ScheduledTask.java
sjcorbett: Yes, I assume callers are going to do the right thing. They also have the option of calling `cancelOnException(boolean)` if they like the strongly-typed side of life.

sjcorbett:
@aledsage Thanks for your comments. Will merge.

PR #1044, opened 2015-11-18
Review rest-client's dependencies

sjcorbett:
Removes a few that are unused, marks `org.eclipse.jetty:jetty-server` as a test dependency and excludes `brooklyn-utils-rest-swagger` (which transitively means Scala is not pulled in). Downstream clients should be much lighter weight (in the order of 10Mb if they package rest-client's dependencies).

### DIFF
--- a/usage/rest-client/pom.xml
+++ b/usage/rest-client/pom.xml
@@ -65,32 +65,17 @@
             </exclusions>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


sjcorbett:
@robertgmoss Could you double check this against the CF broker please?

sjcorbett:
I've confirmed the Maven plugin continues to work as expected.

hzbarcea:
Looks good.

PR #1047, opened 2015-11-19
Remove no longer used annotation and enclosing project

hzbarcea:
After the recent swagger 2.0 swagger upgrade, the "@Apidoc" annotation is no longer used.


hzbarcea:
Doesn't seem like the error is related to this PR.

PR #1049, opened 2015-11-19
Added utility `TestFrameworkAssertions` to address comments from #999

m4rkmckenna:
TestFrameworkAssertions is a common class used by [TestSensor,TestHttpCall] instead of each entity having its own assertions

###### Previous Assert
```
assert:
  isEqualTo: SomeValue
```
###### Current Assert
```
assert:
- isEqualTo: SomeValue
```
Currently Supported assertions [isNull,notNull,isEqualTo,equalTo,equals,matches,contains]

### Changes to TestHttpCall
Added new config key `applyAssertionTo` possible values [body,status]
###### Previous Test Http Status
```
type: org.apache.brooklyn.test.framework.TestHttpCall
name: Assert 200 OK
url: http://localhost:8081
assert:
  status: 200
```
###### Current Test Http Status
```
type: org.apache.brooklyn.test.framework.TestHttpCall
name: Assert 200 OK
url: http://localhost:8081
applyAssertionTo: status
assert:
- isEqualTo: 200
```

### DIFF
--- a/usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/BaseTest.java
+++ b/usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/BaseTest.java
@@ -18,13 +18,15 @@
  */
 package org.apache.brooklyn.test.framework;
 
-import com.google.common.collect.Maps;
+import com.google.common.reflect.TypeToken;
 import org.apache.brooklyn.api.entity.Entity;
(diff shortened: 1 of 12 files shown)


hzbarcea:
Looks ok, minor comments, feel free to address them in a future commit (if at all). Tested, reviewed, merging.

PR #1050, opened 2015-11-20
Add all.members.up sensor to DynamicCluster

mikezaccardo:


core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -181,6 +189,34 @@ private void initialiseMemberId() {
...
+                        .callable(new AllMembersUpCallable()))
+                .build();
+    }
+
+    private class AllMembersUpCallable implements Callable<Boolean> {
+
+        @Override
+        public Boolean call() throws Exception {
+            if (DynamicClusterImpl.this.getMembers().isEmpty())
+                return false;
```
sjcorbett: Wouldn't no members mean that all the members are up?

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -181,6 +189,34 @@ private void initialiseMemberId() {
...
+
+            if (Lifecycle.RUNNING != DynamicClusterImpl.this.sensors().get(SERVICE_STATE_ACTUAL))
+                return false;
+
+            for (Entity member : DynamicClusterImpl.this.getMembers())
+                if (Boolean.FALSE.equals(member.sensors().get(SERVICE_UP)))
```
sjcorbett: The callable is named `AllMembersUp` but this is checking that all members are not not up. It will return true if a member's `SERVICE_UP` is null. I think you should swap the if condition for `if (!Boolean.TRUE.equals(...))`.

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -181,6 +189,34 @@ private void initialiseMemberId() {
...
+    }
+
+    private class AllMembersUpCallable implements Callable<Boolean> {
+
+        @Override
+        public Boolean call() throws Exception {
+            if (DynamicClusterImpl.this.getMembers().isEmpty())
+                return false;
+
+            if (Lifecycle.RUNNING != DynamicClusterImpl.this.sensors().get(SERVICE_STATE_ACTUAL))
```
sjcorbett: Why take the state of the cluster into account when determining whether members are up?

mikezaccardo:
Thanks for reviewing, @sjcorbett.  You're definitely right about the `null` case in my `FALSE` comparison -- I've switched to checking `!TRUE`.

As for your other comments, I think I need to elaborate on my intended use for this sensor; perhaps it is not properly named.  I want to be able to use this as a launch latch / attribute when ready sensor.

If the emptiness check is removed then this sensor will resolve to true immediately when the cluster is created (before the children are added) which is too early.  I could remove this check but then I'd need to leave in the check for the running lifecycle state -- that way even though the cluster is empty in the beginning, this sensor would not resolve to true because it is not yet running at that point.

tl;dr -- I need one or both of the checks that you flagged to remain in order for the sensor to behave the way I'd like it to.  WDYT?

sjcorbett:
I think you're right that this can just come down to nomenclature. ClusterAndMembersUpCallable? Though that doesn't indicate that it expects at least one member.

To be honest I think it would be clearer to keep the three checks separate (cluster up, members up, members.size > 0) and get your launch latch by combining them with a reducer that ANDs them all together. (Not that I know how to express this in YAML, of course!)

mikezaccardo:
@sjcorbett I believe we could actually achieve the same functionality in YAML using this feature: https://github.com/apache/incubator-brooklyn/pull/1037

sjcorbett:
@mikezaccardo What do you want to do with this now that #1037 is merged?

mikezaccardo:
@sjcorbett I still think this addition has value because it simplifies YAML blueprints (vs. having to add an aggregator from #1037), and it also better handles changes in cluster size when invoking a positive resize effector on a `DynamicCluster`.  I'd like to see it merged -- maybe a name change could assuage the concerns you raised about an empty cluster having a `false` value?

ahgittin:
changing sensor name to `CLUSTER_ONE_AND_ALL_MEMBERS_UP` `cluster.one_and_all.members.up` for clarity, and merging

PR #1051, opened 2015-11-20
EntitySpec duplication should include children

duncangrant:
If the memberspec of a dynamiccluster had a child entityspec then it would not be possible to set or expand the cluster to more than 1 node. This is because when addentity is called the parent entity is set on the childspec.  The dynamic cluster uses EntitySpec.create to duplicate the memberspec however this doesn't duplicate any children so the original and duplicate entityspecs share children.

When using EntitySpec.create to duplicate an entityspec this change duplicates children rather than attaching original child specs.

https://issues.apache.org/jira/browse/BROOKLYN-192


sjcorbett:
Looks good.

PR #1053, opened 2015-11-20
Fix eclipse 'Cannot cast' error

hzbarcea:
The swagger changes lead to the following error in eclipse: 
```Cannot cast from MutableSet<Class<capture#4-of ? extends Object>> to MutableSet<Class<?>>```

There's also no need for a MutableSet.


CMoH:
Tested. LGTM

PR #1054, opened 2015-11-23
BROOKLYN-193 Test skipping rebind failures for an entity config value

bostko:
- Exposed exceptions in RebindTestUtils


bostko:
@hzbarcea the reason I did this that it is in the tests and I think it is better to have more fine control over the exceptions in tests.
Any other who want to keep Exception?

sjcorbett:
I +1 @hzbarcea's comment. The benefits are arguable.

sjcorbett:
@bostko Thanks. Will merge.

PR #1055, opened 2015-11-23
Fix BrooklynProperties.Factory.Builder.builderEmpty

bostko:


sjcorbett:
Good spot. Will wait for Jenkins. You could also delete the comment on line ~102 that says "TODO it's always called with true here, perhaps we don't need the argument?"

PR #1056, opened 2015-11-23
BROOKLYN-197 Add jetty-schemas to launcher deps

rdowner:
Without this, Jetty will try to download XML schemas for the documents it processes. This means that Brooklyn requires a working connection to the public Internet. With this dependency, Jetty can discover the schemas it requires without an Internet connection.


sjcorbett:
Looks good.

PR #1058, opened 2015-11-24
Ensure deterministic DynamicCluster member removal order

mikezaccardo:


core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -139,18 +139,21 @@ public ZoneFailureDetector apply(final String input) {
...
         @Override public Entity apply(Collection<Entity> contenders) {
-            // choose newest entity that is stoppable, or if none are stoppable take the newest non-stoppable
-            long newestTime = 0;
+            // choose last-added (largest cluster member ID) entity that is stoppable, or if none are stoppable take the newest non-stoppable
+            int largestClusterMemberId = -1;
             Entity newest = null;
 
             for (Entity contender : contenders) {
-                boolean newer = contender.getCreationTime() > newestTime;
+                int contenderClusterMemberId = contender.config().get(CLUSTER_MEMBER_ID);
```
neykov: I'd fall back to timestamps in case the member id is missing - for example when upgrading older installations.

neykov:
LGTM

neykov:
This was haunting the CIs for quite some time, thanks for taking the time to fix.

PR #1062, opened 2015-11-24
Feature: entity monitoring disablable

aledsage:
Important if the entity is running in a locked-down location, e.g. such that only ssh access to the VM is available. This adds configuration to disable the use of jmx, http, thrift (for Cassandra), mongodb client, etc.

There is also a PR to fix java install, failing for the reason described at http://serverfault.com/questions/637549/epel-repo-for-centos-6-causing-error?newreg=7c6019c0d0ae483c8bb3af387166ce49 - i.e. adding `yum upgrade -y ca-certificates --disablerepo=epel`.

### DIFF
--- a/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/AbstractMongoDBSshDriver.java
+++ b/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/AbstractMongoDBSshDriver.java
@@ -28,6 +28,7 @@ import org.apache.brooklyn.core.entity.Entities;
 import org.apache.brooklyn.entity.software.base.AbstractSoftwareProcessSshDriver;
 import org.apache.brooklyn.entity.software.base.lifecycle.ScriptHelper;
 import org.apache.brooklyn.location.ssh.SshMachineLocation;
+import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.core.internal.ssh.SshTool;
 import org.apache.brooklyn.util.exceptions.Exceptions;
 import org.apache.brooklyn.util.net.Networking;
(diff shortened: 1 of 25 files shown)


ygy:
@aledsage  
Looks very good. Tests are passing as well.

sjcorbett:
Surprised that this has to be added to entities one by one. I suppose it's out of the scope of this pull request but I'd prefer to use a system in which an entity offers certain feeds for certain functions and indicates their requirements (e.g. detailed Tomcat stats require JMX, detailed JBoss HTTP). The superentity should piece together the constraints of the target environment with the feeds available.

software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/mongodb/AbstractMongoDBSshDriver.java
```diff
@@ -115,7 +116,12 @@ public void customize() {
     @Override
     public boolean isRunning() {
         try {
-            return MongoDBClientSupport.forServer((AbstractMongoDBServer) entity).ping();
+            if (entity instanceof MongoDBServerImpl && !((MongoDBServerImpl)entity).clientAccessEnabled()) {
```
sjcorbett: Is there a good reason to single out `MongoDBServerImpl` over the lack of support for replica sets or shards if client access is disabled? I'd prefer that it didn't. There's no reason it can't support the clients, routers and config servers too. I believe that technically the replica set and sharded entities could be written to handle such restrictive environments (by SSHing to machines and executing scripts rather than connecting with `MongoDBClientSupport`).

ahgittin:
Makes sense as a short-term solution in java-land.  @sjcorbett 's comments re Mongo will need attention but this could perhaps be deferred until we are using Mongo in such a behind-subnet environment.

In the mid-term completely agree with @sjcorbett that we want to:
(1) have the blueprints written in yaml
(2) including specifying the sensors declaratively in yaml (eg html, jmx endpoints, ssh commands, etc)
(3) define at a high-level what mechanisms are preferred+supported for sensor collection (eg brooklyn poll, or install a monitoring daemon, etc)

With this model we should be able to configure the *mechanisms* (3) e.g. to use jumphost for polling defined independent of the sources (2).

aledsage:
Thanks @sjcorbett @ahgittin 

Agree with the mid-term aims. It's tricky just now to safely just turn off jmx/http entirely in a super-class, as it depends on the entity for whether that contributes towards service-up or if it's just for usage/performance metrics.

For singling out `MongoDBServer` (versus `MongoDBRouter` and `MongoDBConfigServer`), that is a simple one to get working when there is only ssh access. If there is no replica set, then it's easy. For the methods I guarded (`initializeReplicaSet`, `addMemberToReplicaSet` and `removeMemberFromReplicaSet`) they are called by `MongoDBReplicaSetImpl`. It's hard to tell in `MongoDBServerImpl.init` if it's going to be in a replica set or stand-alone. We could try to look at the parent entity, but that seems inelegant. If you're ok with it, then I suggest leaving as-is.

sjcorbett:
MongoDBReplicaSet could also use the `clientAccessEnabled` config key and check in its own init method. Or use an `alwaysTrue` constraint on the redefined key. Anyway, I'm fine with this being merged.

sjcorbett:
Whoops - not an `alwaysTrue` constraint, an `input value must be true` constraint.

PR #1063, opened 2015-11-24
Api tags support

sjcorbett:
Supercedes #1039.

@neykov could you review commit 641247f please?


usage/rest-server/src/main/java/org/apache/brooklyn/rest/transform/CatalogTransformer.java
```diff
@@ -63,9 +66,10 @@
         Set<SensorSummary> sensors = Sets.newTreeSet(SummaryComparators.nameComparator());
         Set<EffectorSummary> effectors = Sets.newTreeSet(SummaryComparators.nameComparator());
 
+        EntitySpec<?> spec = null;
+
         try {
-            @SuppressWarnings({ "unchecked", "rawtypes" })
-            EntitySpec<?> spec = (EntitySpec<?>) b.getCatalog().createSpec((CatalogItem) item);
+            spec = (EntitySpec<?>) b.getCatalog().createSpec((CatalogItem) item);
```
neykov: Causes `Type safety: Unchecked invocation createSpec(CatalogItem) of the generic method createSpec(CatalogItem<T,SpecT>) of type BrooklynCatalog` in Eclipse. That's the reason it was like this before.

usage/rest-server/src/test/java/org/apache/brooklyn/rest/resources/CatalogResourceTest.java
```diff
@@ -21,8 +21,7 @@
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertTrue;
 
-import java.awt.Image;
-import java.awt.Toolkit;
+import java.awt.*;
```
neykov: Shouldn't collapse imports

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
```diff
@@ -51,9 +56,59 @@ public String getContents() {
             return contents;
         }
     }
-    
-    public static NamedStringTag newYamlSpecTag(String contents) { return new NamedStringTag(YAML_SPEC_KIND, contents); }
-    public static NamedStringTag newNotesTag(String contents) { return new NamedStringTag(NOTES_KIND, contents); }
+
+    public static class ListTag<T> {
```
neykov: I think we shouldn't reuse this, if there are multiple items each should be inserted in a separate tag (i.e. have flat tags). How does adding a trait look like with this type of tag? Should it be inserted in a new collection of traits? Just thinking out loud.

PR #1066, opened 2015-11-24
Updates to Test framework assertions to support multiple assertion failures.

geomacy:
At the moment in the test framework, as noted in [#1030](https://github.com/apache/incubator-brooklyn/pull/1030#issuecomment-158995925), when assertions fail it's not easy to see from the error text in the Activity log what went wrong. In particular if there are multiple assertions that failed this isn't apparent.

This update merges the assertion code from SimpleShellCommand and TestFrameworkAssertions to give:

- Updates test framework assertions so that multiple failures are described in the assertion failure message, such as
```
    Assertions failed: exit code equals 1 stdout equals bogus stdout contains porkpie
```
- Supports generic 'actual' parameter, so we can avoid doing type coercions to String in
order to check assertions.

- Adds "hasTruthValue" condition.

### DIFF
--- a/usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/BaseTest.java
+++ b/usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/BaseTest.java
@@ -47,16 +47,16 @@ public interface BaseTest extends Entity, Startable {
     ConfigKey<String> TARGET_ID = ConfigKeys.newStringConfigKey("targetId", "Id of the entity under test");
 
     /**
-     * The assertions to be made
+     * The assertions to be made.
      */
-    ConfigKey<List<Map<String, Object>>> ASSERTIONS = ConfigKeys.newConfigKey(
-            new TypeToken<List<Map<String, Object>>>() {},
-            "assert", "Assertions to be evaluated", new ArrayList<Map<String, Object>>());
+    ConfigKey<Object> ASSERTIONS = ConfigKeys.newConfigKey(Object.class, "assert", "Assertions to be evaluated",
+        new ArrayList<Map<String, Object>>());
 
     /**
-     * THe duration to wait
+     * THe duration to wait for an assertion to succeed or fail before throwing an exception.
      */
-    ConfigKey<Duration> TIMEOUT = ConfigKeys.newConfigKey(Duration.class, "timeout", "Time to wait on result", new Duration(1L, TimeUnit.SECONDS));
+    ConfigKey<Duration> TIMEOUT = ConfigKeys.newConfigKey(Duration.class, "timeout", "Time to wait on result",
+        new Duration(1L, TimeUnit.SECONDS));
 
     /**
      * Get the target of the test.
--- a/usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTest.java
+++ b/usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTest.java
@@ -18,7 +18,8 @@
  */
 package org.apache.brooklyn.test.framework;
 
-import com.google.common.collect.Maps;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.reflect.TypeToken;
 import org.apache.brooklyn.api.entity.ImplementedBy;
 import org.apache.brooklyn.config.ConfigKey;
@@ -39,27 +40,10 @@ import static org.apache.brooklyn.core.config.ConfigKeys.newConfigKey;
 @ImplementedBy(SimpleShellCommandTestImpl.class)
 public interface SimpleShellCommandTest extends BaseTest {
 
+    /**
+     * Default location for temporary files.
+     */
     String TMP_DEFAULT = "/tmp";
-
-    /**
-     * Equals assertion on command result.
-     */
-    String EQUALS = "equals";
-
-    /**
-     * String contains assertion on command result.
-     */
-    String CONTAINS = "contains";
-
-    /**
-     * Regex match assertion on command result.
-     */
-    String MATCHES = "matches";
-
-    /**
-     * Is-empty match assertion on command result.
-     */
-    String IS_EMPTY = "isEmpty";
 
     /**
      * Supply the command to invoke directly. Cannot be used together with {@link #DOWNLOAD_URL}.
@@ -84,30 +68,35 @@ public interface SimpleShellCommandTest extends BaseTest {
      */
     @SetFromFlag("runDir")
     ConfigKey<String> RUN_DIR = newConfigKey(String.class, "run.dir", "directory where downloaded scripts should be run from");
+
+
+    /**
+     * If no assertions are configured in the test then the default is this assertion that exit status of the command
+     * is zero (successful).
+     */
+    Map<String, Object> DEFAULT_ASSERTION = ImmutableMap.<String,Object>of(TestFrameworkAssertions.EQUALS, 0);
+
     /**
      * Assertions on the exit code of the simple command.
      *
      * If not explicitly configured, the default assertion is a non-zero exit code.
      */
     @SetFromFlag("assertStatus")
-    ConfigKey<List<Map<String, Object>>> ASSERT_STATUS = ConfigKeys.newConfigKey(
-        new TypeToken<List<Map<String, Object>>>() {},
-        "assert.status", "Assertions on command exit code", new ArrayList<Map<String, Object>>());
+    ConfigKey<Object> ASSERT_STATUS = ConfigKeys.newConfigKey(Object.class, "assert.status", "Assertions on command exit code",
+        new ArrayList<Map<String, Object>>());
 
     /**
      * Assertions on the standard output of the command as a String.
      */
     @SetFromFlag("assertOut")
-    ConfigKey<List<Map<String, Object>>> ASSERT_OUT = ConfigKeys.newConfigKey(
-        new TypeToken<List<Map<String, Object>>>() {},
-        "assert.out", "Assertions on command standard output", new ArrayList<Map<String, Object>>());
+    ConfigKey<Object> ASSERT_OUT = ConfigKeys.newConfigKey(Object.class, "assert.out", "Assertions on command standard output",
+        new ArrayList<Map<String, Object>>());
 
     /**
      * Assertions on the standard error of the command as a String.
      */
     @SetFromFlag("assertErr")
-    ConfigKey<List<Map<String, Object>>> ASSERT_ERR = ConfigKeys.newConfigKey(
-        new TypeToken<List<Map<String, Object>>>() {},
-        "assert.err", "Assertions on command standard error", new ArrayList<Map<String, Object>>());
+    ConfigKey<Object> ASSERT_ERR = ConfigKeys.newConfigKey(Object.class, "assert.err", "Assertions on command standard error",
+        new ArrayList<Map<String, Object>>());
 
 }
--- a/usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTest.java
+++ b/usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTest.java
@@ -19,6 +19,7 @@
 package org.apache.brooklyn.test.framework;
 
 import com.google.common.collect.Maps;
+import com.google.common.reflect.TypeToken;
 import org.apache.brooklyn.api.entity.ImplementedBy;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
@@ -26,6 +27,8 @@ import org.apache.brooklyn.core.sensor.AttributeSensorAndConfigKey;
 import org.apache.brooklyn.entity.software.base.SoftwareProcess;
 import org.apache.brooklyn.util.core.flags.SetFromFlag;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 
 import static org.apache.brooklyn.core.config.ConfigKeys.newConfigKey;
@@ -87,21 +90,24 @@ public interface SimpleShellCommandTest extends BaseTest {
      * If not explicitly configured, the default assertion is a non-zero exit code.
      */
     @SetFromFlag("assertStatus")
-    ConfigKey<Map> ASSERT_STATUS = ConfigKeys.newConfigKey(Map.class, "assert.status",
-            "Assertions on command exit code", Maps.newLinkedHashMap());
+    ConfigKey<List<Map<String, Object>>> ASSERT_STATUS = ConfigKeys.newConfigKey(
+        new TypeToken<List<Map<String, Object>>>() {},
+        "assert.status", "Assertions on command exit code", new ArrayList<Map<String, Object>>());
 
     /**
      * Assertions on the standard output of the command as a String.
      */
     @SetFromFlag("assertOut")
-    ConfigKey<Map> ASSERT_OUT = ConfigKeys.newConfigKey(Map.class, "assert.out",
-            "Assertions on command standard output", Maps.newLinkedHashMap());
+    ConfigKey<List<Map<String, Object>>> ASSERT_OUT = ConfigKeys.newConfigKey(
+        new TypeToken<List<Map<String, Object>>>() {},
+        "assert.out", "Assertions on command standard output", new ArrayList<Map<String, Object>>());
 
     /**
      * Assertions on the standard error of the command as a String.
      */
     @SetFromFlag("assertErr")
-    ConfigKey<Map> ASSERT_ERR = ConfigKeys.newConfigKey(Map.class, "assert.err",
-            "Assertions on command standard error", Maps.newLinkedHashMap());
+    ConfigKey<List<Map<String, Object>>> ASSERT_ERR = ConfigKeys.newConfigKey(
+        new TypeToken<List<Map<String, Object>>>() {},
+        "assert.err", "Assertions on command standard error", new ArrayList<Map<String, Object>>());
 
 }
--- a/usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTestImpl.java
+++ b/usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTestImpl.java
@@ -21,6 +21,7 @@ package org.apache.brooklyn.test.framework;
 
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
+import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableMap;
 import org.apache.brooklyn.api.location.Location;
 import org.apache.brooklyn.api.mgmt.TaskFactory;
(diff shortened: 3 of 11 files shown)


usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTest.java
```diff
@@ -19,6 +19,7 @@
 package org.apache.brooklyn.test.framework;
 
 import com.google.common.collect.Maps;
```
mikezaccardo: Unused import

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTest.java
```diff
@@ -87,21 +90,24 @@
      * If not explicitly configured, the default assertion is a non-zero exit code.
      */
     @SetFromFlag("assertStatus")
-    ConfigKey<Map> ASSERT_STATUS = ConfigKeys.newConfigKey(Map.class, "assert.status",
-            "Assertions on command exit code", Maps.newLinkedHashMap());
+    ConfigKey<List<Map<String, Object>>> ASSERT_STATUS = ConfigKeys.newConfigKey(
+        new TypeToken<List<Map<String, Object>>>() {},
+        "assert.status", "Assertions on command exit code", new ArrayList<Map<String, Object>>());
```
mikezaccardo: Perhaps use `Collections.emptyList()` for an immutable empty list as the default value here and below.

It looks the list is replaced and never added to directly so this should be fine.

mikezaccardo:
Looks good overall but I will finish a closer review on Wednesday!

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTest.java
```diff
@@ -87,21 +90,24 @@
      * If not explicitly configured, the default assertion is a non-zero exit code.
      */
     @SetFromFlag("assertStatus")
-    ConfigKey<Map> ASSERT_STATUS = ConfigKeys.newConfigKey(Map.class, "assert.status",
-            "Assertions on command exit code", Maps.newLinkedHashMap());
+    ConfigKey<List<Map<String, Object>>> ASSERT_STATUS = ConfigKeys.newConfigKey(
```
ahgittin: do the old tests still work?  and do we want to be this strict or make best effort to understand -- e.g. if this is `ConfigKey<Object>` we can have a method e.g. `getAssertions(ASSERT_STATUS)` which retrieves it and creates a list of maps, but which would be more forgiving, e.g. if it's a single map we could wrap it in a list.

this is new so i don't mind either way.  i prefer making best-efforts to parse people's intentions and warning but still working if they're wrong (especially if we're changing a syntax which people are already using, as we are here).  however other people in the community prefer fail fast even on user input so either is fine...

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTestImpl.java
```diff
@@ -107,13 +107,23 @@ protected void handle(Result result) {
...
-        checkAssertions(support, exitCodeAssertions(), "exit code", result.getExitCode());
-        checkAssertions(support, getConfig(ASSERT_OUT), "stdout", result.getStdout());
-        checkAssertions(support, getConfig(ASSERT_ERR), "stderr", result.getStderr());
+        TestFrameworkAssertions.checkAssertions(support, flags, exitCodeAssertions(), "exit code", supply(result.getExitCode()));
+        TestFrameworkAssertions.checkAssertions(support, flags, getConfig(ASSERT_OUT), "stdout", supply(result.getStdout()));
+        TestFrameworkAssertions.checkAssertions(support, flags, getConfig(ASSERT_ERR), "stderr", supply(result.getStderr()));
         support.validate();
     }
 
+    private static <T> Supplier<T> supply(final T t) {
```
ahgittin: see `Suppliers.ofInstance`.  in fact for most such helpers you'll probably find it exists.  worth asking.

also, avoid anonymous classes except for very transient operations.  if they get in to config values and get persisted the persistence gets ugly!

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCallImpl.java
```diff
@@ -66,31 +68,35 @@ public void start(Collection<? extends Location> locations) {
...
 
-    private Supplier<String> buildDataSupplier(final HttpAssertionTarget httpAssertionTarget, final String url) {
-
-        switch (httpAssertionTarget) {
+    private void doRequestAndCheckAssertions(Map<String, Duration> flags, List<Map<String, Object>> assertions,
+                                             HttpAssertionTarget target, final String url) {
+        switch (target) {
             case body:
-                return new Supplier<String>() {
+                Supplier<String> getBody = new Supplier<String>() {
```
ahgittin: for checks like this which will be immediate this (anonymous class) is how we'd do it (just FYI).  if we need this in more than one or two places then I'd create a `ResourceUtils.getResourceAsStringSupplier()` to do this.

btw we normally use `ResourceUtils` for reading URL's because it will use classpath URL's and can be context aware.  not important in this case but again just FYI.

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCallImpl.java
```diff
@@ -66,31 +68,35 @@ public void start(Collection<? extends Location> locations) {
...
                     @Override
-                    public String get() {
+                    public Integer get() {
                         try {
-                            return String.valueOf(HttpTool.getHttpStatusCode(url));
+                            return HttpTool.getHttpStatusCode(url);
                         } catch (Exception e) {
-                            LOG.error("HTTP call to [{}] failed due to [{}] ... returning Status code [0 - Unreachable]", url, e.getMessage());
-                            return "0";
+                            LOG.error("HTTP call to [{}] failed due to [{}] ... returning Status code [500]",
```
ahgittin: definitely not ERROR level, since here we're just in a status check.

normally in catch blocks where we swallow generic exceptions we include

        Exceptions.propagateIfFatal(e);

in case it's a severe exception.  but here' i'd simply propagate it regardless (`Exceptions.propagate(e)') and make sure the wrapper code treats errors appropriately.  returning `500` is dishonest :) .

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestFrameworkAssertions.java
```diff
@@ -39,122 +35,179 @@
...
  */
 public class TestFrameworkAssertions {
-    private static final Logger LOG = LoggerFactory.getLogger(TestFrameworkAssertions.class);
+
+    public static final String IS_NULL = "isNull";
+    public static final String NOT_NULL = "notNull";
+    public static final String IS_EQUAL_TO = "isEqualTo";
+    public static final String EQUAL_TO = "equalTo";
+    public static final String EQUALS = "equals";
+    public static final String MATCHES = "matches";
```
ahgittin: in time (not now) i'd like to introduce aliases across brooklyn, and then allow `matches_regex` and `regex` here.  `matches` is too vague, is it literal, is it glob, is it regex.  but okay for now (and i see you've not changed that in this PR anyway).

ahgittin:
a few minor comments from me

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTest.java
geomacy: This change doesn't alter the syntax of the assertions in the Yaml, (except for the new SimpleShellCommand which it just brings in to line with the other tests), i.e.
```
assert:
- equals: true
```
But in fact any tests that were written before the new test assertions class were put in wouldn't work now (actually since https://github.com/apache/incubator-brooklyn/pull/1049) as they read
```
assert:
   equals: true
```
However I think there are very few such tests, as people had only started using the framework when 1049 was done.

At the same time it does seem a good improvement to make it work with a single map and wrap that in a list if need be.  (Is there code already somewhere that does that sort of thing?  How can you have a config key that might be a map or might be a list?)

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCallImpl.java
geomacy: Thanks, those are good to know, especially about the risk of the anonymous classes getting into persistence.

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTest.java
ahgittin: sadly you can't do `ConfigKey<Map | List>` :) so the way is just `ConfigKey<Object>` and use a static method to coerce the result.

geomacy: Ok, thanks!

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTestImpl.java
```diff
@@ -228,89 +238,14 @@ private static String getLastPartOfPath(Iterable<String> path, String defaultNam
...
-    }
 
-    private Map<?, ?> exitCodeAssertions() {
-        Map<?, ?> assertStatus = getConfig(ASSERT_STATUS);
-        if (isEmpty(assertStatus)) {
-            assertStatus = ImmutableMap.of(EQUALS, SUCCESS);
+    private List<Map<String, Object>> exitCodeAssertions() {
+        List<Map<String, Object>> assertStatus = getConfig(ASSERT_STATUS);
+        if (assertStatus.isEmpty()) {
+            Map<String, Object> shouldSucceed = ImmutableMap.<String,Object>of(EQUALS, SUCCESS);
```
mikezaccardo: Might be nice to define this as a member variable `DEFAULT_ASSERTION` or the like, for clarity.

usage/test-framework/src/test/java/org/apache/brooklyn/test/framework/SimpleShellCommandIntegrationTest.java
```diff
@@ -28,7 +28,9 @@
 import org.apache.brooklyn.location.localhost.LocalhostMachineProvisioningLocation;
 import org.apache.brooklyn.location.ssh.SshMachineLocation;
 import org.apache.brooklyn.test.Asserts;
+import org.apache.brooklyn.util.collections.MutableMap;
```
mikezaccardo: Not in the diff, but `import com.google.common.collect.ImmutableMap;` is unused.

usage/test-framework/src/test/java/org/apache/brooklyn/test/framework/TestFrameworkAssertionsTest.java
```diff
@@ -19,6 +19,7 @@
 package org.apache.brooklyn.test.framework;
 
 import com.google.common.base.Supplier;
+import org.apache.brooklyn.test.Asserts;
```
mikezaccardo: Not in the diff, but `import static org.assertj.core.api.Assertions.assertThat;` is unused.

mikezaccardo:
Tests all pass, LGTM after @ahgittin's comments are resolved (mine are minor and could be merged without being addressed).

geomacy:
Thanks @ahgittin, @mikezaccardo, code updated as per comments above.

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/BaseTest.java
```diff
@@ -47,16 +47,16 @@
...
 
     /**
-     * The assertions to be made
+     * The assertions to be made.
      */
-    ConfigKey<List<Map<String, Object>>> ASSERTIONS = ConfigKeys.newConfigKey(
-            new TypeToken<List<Map<String, Object>>>() {},
-            "assert", "Assertions to be evaluated", new ArrayList<Map<String, Object>>());
+    ConfigKey<Object> ASSERTIONS = ConfigKeys.newConfigKey(Object.class, "assert", "Assertions to be evaluated",
+        new ArrayList<Map<String, Object>>());
```
aledsage: Personal preference for the default value of the config key to be immutable. Otherwise if someone modifies it, things get very strange (it changes the default for all other entities that use it, until Brooklyn is restarted!)

I'd therefore use `ImmutableList.of()` (presumably the generics don't matter, now that it's change to be a config key of type `Object`).

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTest.java
```diff
@@ -36,27 +40,10 @@
...
-    /**
-     * Regex match assertion on command result.
-     */
-    String MATCHES = "matches";
-
-    /**
-     * Is-empty match assertion on command result.
-     */
-    String IS_EMPTY = "isEmpty";
+    String TMP_DEFAULT = "/tmp";
```
aledsage: You can use `org.apache.brooklyn.util.os.Os.tmp()`. But it doesn't matter given that this is just for tests. So consider this just information, rather than needing changed.

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTest.java
```diff
@@ -81,27 +68,35 @@
...
     /**
      * Assertions on the exit code of the simple command.
      *
      * If not explicitly configured, the default assertion is a non-zero exit code.
      */
     @SetFromFlag("assertStatus")
-    ConfigKey<Map> ASSERT_STATUS = ConfigKeys.newConfigKey(Map.class, "assert.status",
-            "Assertions on command exit code", Maps.newLinkedHashMap());
+    ConfigKey<Object> ASSERT_STATUS = ConfigKeys.newConfigKey(Object.class, "assert.status", "Assertions on command exit code",
+        new ArrayList<Map<String, Object>>());
```
aledsage: Again, it's good if config key default values are immutable. Same below.

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/SimpleShellCommandTest.java
aledsage: Actually, I misread the context - this is in main rather than test. But it's the path on the *target* machine rather than the Brooklyn machine, so using `/tmp` is right.

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestFrameworkAssertions.java
```diff
@@ -39,122 +42,216 @@
...
-        for (final Map.Entry<String, Object> assertion : assertionsMap.entrySet()) {
-            final Maybe<Predicate<String>> optionalPredicate = getPredicate(assertion.getKey(), assertion.getValue());
-            Asserts.succeedsEventually(flags, new PredicateChecker(dataSupplier, optionalPredicate.get()));
+        Maybe<List<Map<String, Object>>> maybeList = TypeCoercions.tryCoerce(config,
+            new TypeToken<List<Map<String, Object>>>() {});
+        if (maybeList.isPresent()) {
+            return maybeList.get();
         }
+
+        throw new FatalConfigurationRuntimeException(key.getDescription() + " is not a map or list of maps");
```
aledsage: I'd include in error as much context as possible (as long as it doesn't risk logging security info) - e.g. `key.getDescription() + " of " + entity + " must be map or list of maps, but is " + (config == null ? null : config.getClass().getName())`

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestFrameworkAssertions.java
```diff
@@ -39,122 +42,216 @@
...
-                || StringUtils.equalsIgnoreCase("equalTo", predicateKey)
-                || StringUtils.equalsIgnoreCase("equals", predicateKey)) {
-            return Maybe.of(Predicates.equalTo(TypeCoercions.coerce(predicateTarget.toString(), String.class)));
-        } else if (StringUtils.equalsIgnoreCase("matches", predicateKey)) {
-            return Maybe.of(buildMatchesPredicate(TypeCoercions.coerce(predicateTarget, String.class)));
-        } else if (StringUtils.equalsIgnoreCase("contains", predicateKey)) {
-            return Maybe.of(buildContainsPredicate(TypeCoercions.coerce(predicateTarget, String.class)));
+
+    public static <T> void checkAssertions(Map<String,?> flags,
+                                           Map<String, Object> assertions,
```
aledsage: General convention in Brooklyn is to indent this 8 spaces, rather than aligning with the previous param. No particularly strong feelings, but I do like consistency.

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestFrameworkAssertions.java
```diff
@@ -39,122 +42,216 @@
...
+                                           String target,
+                                           final Supplier<T> actualSupplier) {
+
+        AssertionSupport support = new AssertionSupport();
```
aledsage: Minor: perhaps just call `checkAssertions(flags, ImmutableList.of(assertions), target, actualSupplier)`

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestFrameworkAssertions.java
```diff
@@ -39,122 +42,216 @@
...
+        }
+        try {
+            Asserts.succeedsEventually(flags, new Runnable() {
+                @Override
+                public void run() {
+                    T actual = actualSupplier.get();
+                    checkActualAgainstAssertions(support, assertions, target, actual);
+                }
+            });
+        } catch (Throwable t) {
```
aledsage: I'm very wary of swallowing the wrong kind of exception - e.g. if it was an `InterruptedException`, then important the thread is left in an interrupted state (see http://www.ibm.com/developerworks/library/j-jtp05236/).

Perhaps do `support.fail(t); Exceptions.propagateIfFatal(t);`?

(but I haven't checked what support.fail does).

usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestFrameworkAssertions.java
```diff
@@ -39,122 +42,216 @@
  * @author m4rkmckenna on 11/11/2015.
  */
 public class TestFrameworkAssertions {
-    private static final Logger LOG = LoggerFactory.getLogger(TestFrameworkAssertions.class);
+
+    public static final String IS_NULL = "isNull";
+    public static final String NOT_NULL = "notNull";
+    public static final String IS_EQUAL_TO = "isEqualTo";
```
aledsage: I have a personal preference of not supporting lots of variants. It feels nice sometimes, but then when someone starts looking at different examples and sees different text in each, it looks confusing (*). But no strong feelings.

(*) there are exceptions to that, such as in Duration where we accept `1h` and `1hr` etc.

usage/test-framework/src/test/java/org/apache/brooklyn/test/framework/SimpleShellCommandIntegrationTest.java
```diff
@@ -96,15 +102,70 @@ public void shouldInsistOnJustOneOfCommandAndScript(String command, String scrip
...
+    @Test(groups = "Integration")
+    public void shouldFailUsingSuccessfulExitAsDefaultCondition() {
+        TestEntity testEntity = app.createAndManageChild(EntitySpec.create(TestEntity.class).location(machine));
+
+        SimpleShellCommandTest uptime = app.createAndManageChild(EntitySpec.create(SimpleShellCommandTest.class)
+            .configure(TARGET_ENTITY, testEntity)
+            .configure(COMMAND, "ls /tmp/bogus-" + Identifiers.randomLong()));
+
+        try {
+            app.start(ImmutableList.of(loc));
```
aledsage: Do we need to fail if the exception was not thrown?

aledsage:
Only extremely minor comments from me - the only one that needs addressed is the mutable values for config key defaults, but that wasn't introduced in this PR.

Good to merge now. Comments can be reviewed and selectively addressed later.

PR #1071, opened 2015-11-26
SoftwareProcess.dontRequireTtyForSudo - sudoers file verification

ygy:
- it doesn't directly edit /etc/sudoers
- uses visudo to verify the prepared file
- added integration test
- excluded test sudoers file from rat plugin checks


ahgittin:
looks good.  is there a situation where in-place `sed` doesn't work?  or is the `visudo` only used to give fail-fast if we edit wrongly?  worth a comment in code to address.

(btw you comment that you exclude test sudoers file from rat but i don't see that; it looks like it has the appropriate license header which is actually better.)

nice tests.  good enough to merge, so merging.

aledsage:
@ahgittin problems were reported when using `BashCommands.dontRequireTtyForSudo()` on a RHEL environment - the sed command had left `/etc/sudoers` wrong, so it was then impossible for anyone to sudo - i.e. we'd entirely broken the VM.

Unfortunately I'm not sure what the original `/etc/sudoers` file looked like, such that our `sed` command would break it.

But it highlighted that we were following *very* bad practices by modifying the file directly; switching to `visudo` is definitely the right thing to do.

PR #1072, opened 2015-11-26
Add an overview guide for using the web based GUI to the Operations Guide

lloyddave:
With the introduction of a CLI tool for Brooklyn, the Getting Started Guide will
be updated to document using the CLI instead of the GUI.  This change is to copy
the content related to using the existing web based GUI from the Getting Started
Guide to the Operations Guide.  It incorporates some re-wording and changing of links
to reflect the new location.

**NOTE:** This PR has no impacts on the Getting Started Guide itself.  That work will be
done in parallel to the CLI tool.


ahgittin:
nice!  merging.  @shartzel you might want to take a look from a pedagogy point of view.  you can continue to comment here and we can improve it as a follow-up PR.

PR #1074, opened 2015-11-26
Add documentation and additional tests to SshToolAbstractIntegrationTest

bostko:


core/src/test/java/org/apache/brooklyn/util/core/internal/ssh/SshToolAbstractIntegrationTest.java
```diff
@@ -264,8 +271,44 @@ public void testSshKeyWithPassphrase() throws Exception {
...
+        assertEquals(localtool.execScript(MutableMap.<String,Object>of(), ImmutableList.of("date")), 0);
+
+        // Also needs the negative test to prove that we're really using an ssh-key with a passphrase
+        try {
+            final SshTool localtool2 = newTool(ImmutableMap.<String,Object>builder()
+                    .put(SshTool.PROP_HOST.getName(), "localhost")
+                    .put(SshTool.PROP_PRIVATE_KEY_DATA.getName(), "invalid data")
+                    .build());
+            localtool2.connect();
+            localtool2.execScript(MutableMap.<String,Object>of(), ImmutableList.of("date"));
```
bostko: @aledsage this was succeeding for me. I found that I have to use "IdentitiesOnly=yes" in order to deny ssh to try keys other than the specified keys as a parameter or from the config file.
Check http://linux.die.net/man/5/ssh_config
Do you think it is reasonable make a configurable IdentitiesOnly parameter in the SshTool in order to use it for such tests?

ahgittin: @bostko i've never encountered a situation where the default `identitiesOnly=no` causes a problem.  can you tell what settings (presumably in the `sshd` config on your localhost) make this necessary?

for integration tests on localhost you should be able to undo those.

we could allow arbitrary tool-specific ssh-tool configuration to be specified in `brooklyn.properties` in the first instance, only introducing an `identitiesOnly=yes` parameter if the issue recurs.

definitely open a jira issue in case other people hit this, including info on the ssh client and server config, and the errors which result.

core/src/test/java/org/apache/brooklyn/util/core/internal/ssh/SshToolAbstractIntegrationTest.java
```diff
@@ -264,8 +271,44 @@ public void testSshKeyWithPassphrase() throws Exception {
...
+            // Notice that executing a command may succeed for SshCliToolIntegrationTest.testSshKeyWithPrivateKeyData if you already have valid keys loaded in the ssh-agent
+            fail();
```
ahgittin: you can use `Asserts.shouldHaveFailedPreviously()` then `Asserts.expectedFailureContains(SshException.class.getSimpleName())` for a more readable test.

ahgittin:
looks good, merging (but please open the jira issue)

PR #1077, opened 2015-11-30
Fix itest/pom: include BROOKLYN_VERSION marker

aledsage:
Also follows convention used in other poms, to include relativePath
for parent, and to declare the artifactId+name before the parent.

The `BROOKLYN_VERSION` fix is very important for our release process, so we can easily update the version number everywhere.

Out of interest @CMoH why does kara/pom.xml depend on brooklyn rather than brooklyn-parent? I have not changed that here.


CMoH:
Thanks for your input, @aledsage . LGTM. Merging.

PR #1080, opened 2015-12-01
Fix poms: include BROOKLYN_VERSION marker

aledsage:
Similar idea to https://github.com/apache/incubator-brooklyn/pull/1077


sjcorbett:
Looks good.

PR #1081, opened 2015-12-02
Make default values for assertions config be immutable.

geomacy:
This is in response to the comment on 1066, https://github.com/apache/incubator-brooklyn/pull/1066/#issuecomment-160699587.


sjcorbett:
Looks good.

PR #1082, opened 2015-12-02
Jclouds machines: do not persist NodeMetadata or Template

aledsage:
Work in progress:
* requires the same to be done for `JcloudsWinrmMachineLocation`
* requires more testing for backwards compatibility (see the unit test included here; but needs tested in the wild).

Previously JcloudsSshMachineLocation would persist the `NodeMetadata` and the `Template`, which pulled in lots of jclouds stuff. This was very brittle: if the NodeMetadata fields changed, or something like `AzureComputeTemplateOptions` fields changed, then rebind could fail.

This changes `JcloudsSshMachineLocation` to not persist node and template. Instead it persists the nodeId + imageId + privateAddresses + publicAddresses + hostname (i.e. the things in NodeMetadata that are most important).

For backwards compatibility, if on rebind it finds the node in the persisted state then it extracts the fields we need, sets those, and then sets the `node` field to null. Same for the `template` field.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsSshMachineLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsSshMachineLocation.java
@@ -21,7 +21,6 @@ package org.apache.brooklyn.location.jclouds;
 import static org.apache.brooklyn.util.JavaGroovyEquivalents.groovyTruth;
 
 import java.net.InetAddress;
-import java.net.UnknownHostException;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -46,6 +45,7 @@ import org.jclouds.compute.ComputeServiceContext;
 import org.jclouds.compute.callables.RunScriptOnNode;
 import org.jclouds.compute.domain.ExecResponse;
 import org.jclouds.compute.domain.Hardware;
+import org.jclouds.compute.domain.Image;
 import org.jclouds.compute.domain.NodeMetadata;
 import org.jclouds.compute.domain.OperatingSystem;
 import org.jclouds.compute.domain.OsFamily;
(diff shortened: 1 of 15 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsSshMachineLocation.java
```diff
@@ -146,24 +264,48 @@ public JcloudsLocation getParent() {
...
+                }
+                if (groovyTruth(getPublicAddresses())) {
+                    _privateHostname = getPublicAddresses().iterator().next();
+                } else if (groovyTruth(getHostname())) {
+                    _privateHostname = getHostname();
+                } else {
+                    return null;
+                }
+            }
+        }
```
neykov: I think having a single code path might be better here, i.e. have only the fallback? Otherwise could be surprised by different behaviour in some edge cases.

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/RebindJcloudsLocationLiveTest.java
```diff
@@ -94,6 +101,100 @@ public void testRebindsToJcloudsMachine() throws Exception {
...
+        assertEquals(machine2.getHostname(), "ip-10-144-66-5", errmsg); // TODO would prefer the hostname that works inside and out
+        assertEquals(machine2.getOsDetails().isWindows(), false, errmsg);
+        assertEquals(machine2.getOsDetails().isLinux(), true, errmsg);
+        assertEquals(machine2.getOsDetails().isMac(), false, errmsg);
+        assertEquals(machine2.getOsDetails().getName(), "centos", errmsg);
+        assertEquals(machine2.getOsDetails().getArch(), "x86_64", errmsg);
+        assertEquals(machine2.getOsDetails().getVersion(), "6.5", errmsg);
+        assertEquals(machine2.getOsDetails().is64bit(), true, errmsg);
+
+        // Force it to be persisted again. Expect to pesist without the NodeMetadata and Tempalte.
```
neykov: typo `Tempalte`

neykov:
Really happy to see so much brittle state removed from persistence. Looks good so far.

ahgittin:
+1 ... look forward to having this merged!

aledsage:
@neykov @ahgittin I've done various updates (including getting this working for `JcloudsWinRmMachineLocation`) and fixed some live tests. Can you take another look please?

ahgittin:
looks good

PR #1086, opened 2015-12-03
Fix: winrm and jclouds to private-VMs, and miscelaneous

aledsage:


bostko:
Looks good.

aledsage:
Thanks @bostko - merging.

PR #1087, opened 2015-12-03
JcloudsLocation: fix waitForWinRm, for alternate WinRmTool

aledsage:
Previously JcloudsLocation was hard-coded to use the winrm4j directly,
in waitForWinrmAvailable. Now it follows the same pattern as for 
ssh: it calls createTemporaryWinRmMachineLocation, and uses that for
the health check. This means any special config for an alternative
WinRmTool implementation will be respected.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -2743,7 +2743,7 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
         Stopwatch stopwatch = Stopwatch.createStarted();
 
         ReferenceWithError<Boolean> reachable = new Repeater()
-                .every(1,SECONDS)
+                .backoff(Duration.ONE_SECOND, 2, Duration.TEN_SECONDS) // exponential backoff, to 10 seconds
                 .until(checker)
                 .limitTimeTo(timeout)
                 .runKeepingError();
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -24,8 +24,6 @@ import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.apache.brooklyn.util.JavaGroovyEquivalents.elvis;
 import static org.apache.brooklyn.util.JavaGroovyEquivalents.groovyTruth;
 import static org.apache.brooklyn.util.ssh.BashCommands.sbinPath;
-import io.cloudsoft.winrm4j.pywinrm.Session;
-import io.cloudsoft.winrm4j.pywinrm.WinRMFactory;
 
 import java.io.ByteArrayOutputStream;
 import java.io.File;
@@ -47,7 +45,6 @@ import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import javax.annotation.Nullable;
@@ -96,6 +93,8 @@ import org.apache.brooklyn.util.core.flags.SetFromFlag;
 import org.apache.brooklyn.util.core.flags.TypeCoercions;
 import org.apache.brooklyn.util.core.internal.ssh.ShellTool;
 import org.apache.brooklyn.util.core.internal.ssh.SshTool;
+import org.apache.brooklyn.util.core.internal.winrm.WinRmTool;
+import org.apache.brooklyn.util.core.internal.winrm.WinRmToolResponse;
 import org.apache.brooklyn.util.core.text.TemplateProcessor;
 import org.apache.brooklyn.util.exceptions.CompoundRuntimeException;
 import org.apache.brooklyn.util.exceptions.Exceptions;
@@ -179,7 +178,6 @@ import com.google.common.collect.Sets;
 import com.google.common.collect.Sets.SetView;
 import com.google.common.io.Files;
 import com.google.common.net.HostAndPort;
-import com.google.common.primitives.Ints;
 
 /**
  * For provisioning and managing VMs in a particular provider/region, using jclouds.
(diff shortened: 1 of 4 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2537,28 +2563,53 @@ protected LoginCredentials waitForWinRmAvailable(final ComputeService computeSer
...
+                        boolean success = (response.getStatusCode() == 0);
+                        if (success) {
+                            credsSuccessful.set(entry.getValue());
+                            return true;
+                        }
+                    }
+                    return false;
+                }};
+    
+            waitForReachable(checker, connectionDetails, credentialsToTry, setup, timeout);
```
bostko: I suggest to reduce the checker interval (reduce the time in the Repeater at the end of `waitForReachable`) to 10 seconds.

bostko:
Looks good, with one minor comment.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
aledsage: We want to wait a *very long time* because we might be waiting for the Windows machine to boot up, download security patches, etc. We start calling this method when the cloud provider says the VM is "ready", but that mean the OS has booted up.

We wait for 30 minutes by default, with it being configurable using `waitForWinRmAvailable` (it is 5 minutes for  `waitForSshable`).

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1608,7 +1606,35 @@ protected SshMachineLocation createTemporarySshMachineLocation(HostAndPort hostA
         }
     }
 
-    
+    /**
+     * Creates a temporary WinRM machine location (i.e. will not be persisted), which uses the given credentials.
```
neykov: Is this what `TEMPORARY_LOCATION` really does (doesn't persist)? There's `LocalLocationManager.CREATE_UNMANAGED` which doesn't persist for sure, but looks like this one just doesn't log.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2692,7 +2743,7 @@ protected void waitForReachable(Callable<Boolean> checker, String hostAndPort, L
         Stopwatch stopwatch = Stopwatch.createStarted();
 
         ReferenceWithError<Boolean> reachable = new Repeater()
-                .every(1,SECONDS)
+                .backoff(Duration.ONE_SECOND, 2, Duration.TEN_SECONDS) // exponential backoff, to 10 seconds
```
bostko: Great

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2537,28 +2563,53 @@ protected LoginCredentials waitForWinRmAvailable(final ComputeService computeSer
...
-        final Session session = WinRMFactory.INSTANCE.createSession(vmIp+":"+vmPort, user, password);
-
+        Set<String> users = Sets.newLinkedHashSet();
+        for (LoginCredentials creds : credentialsToTry) {
+            users.add(creds.getUser());
+        }
+        String user = (users.size() == 1) ? Iterables.getOnlyElement(users) : "{" + Joiner.on(",").join(users) + "}";
+        String vmIp = hostAndPortOverride.isPresent() ? hostAndPortOverride.get().getHostText() : getFirstReachableAddress(node, setup);
         if (vmIp==null) LOG.warn("Unable to extract IP for "+node+" ("+setup.getDescription()+"): subsequent connection attempt will likely fail");
+        int vmPort = hostAndPortOverride.isPresent() ? hostAndPortOverride.get().getPortOrDefault(5985) : 5985;
```
neykov: This would deserve a constant in a subsequent PR, repeated too much throughout the code.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
bostko: I ment increase, sorry

bostko:
Looks good!

neykov:
+1, apart from the potentially misleading comment.
Will merge after Jenkins is happy with the build.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
aledsage: Good catch. I'll add that as well. @sjcorbett Might explain why we see various extra persisted location files, as we do the same for temporary `SshMachineLocation` files!

bostko: @sjcorbett could you explain?

bostko: Keeping "various extra persisted location" files is definitely a bug.
Such extra files are called "dangling references" in Apache Brooklyn and we have validation about it called minRequiredHealthy which checks a healthy quorum check. If the minRequiredHealthy validation doesn't have healthy objects quorum, Brooklyn fails to start.
These failures are reported several times by clients.
We may want to keep some of the extra locations for history and durability reasons, but then we have to modify the minRequiredHealthy validation accordingly.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1608,7 +1608,36 @@ protected SshMachineLocation createTemporarySshMachineLocation(HostAndPort hostA
...
+        winrmProps.remove("password");
+        winrmProps.remove("privateKeyData");
+        winrmProps.remove("privateKeyFile");
+        winrmProps.remove("privateKeyPassphrase");
+
+        if (initialPassword.isPresent()) winrmProps.put("password", initialPassword.get());
+        if (initialPrivateKey.isPresent()) winrmProps.put("privateKeyData", initialPrivateKey.get());
+
+        if (isManaged()) {
+            return getManagementContext().getLocationManager().createLocation(winrmProps, WinRmMachineLocation.class);
```
bostko: This location is temporary. Why do we have to manage it (i.e. persist it)?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1608,7 +1608,36 @@ protected SshMachineLocation createTemporarySshMachineLocation(HostAndPort hostA
...
+        String initialUser = creds.getUser();
+        Optional<String> initialPassword = creds.getOptionalPassword();
+        Optional<String> initialPrivateKey = creds.getOptionalPrivateKey();
+
+        Map<String,Object> winrmProps = Maps.newLinkedHashMap(config.getAllConfig());
+        winrmProps.put("user", initialUser);
+        winrmProps.put("address", hostAndPort.getHostText());
+        winrmProps.put("port", hostAndPort.getPort());
+        winrmProps.put(AbstractLocation.TEMPORARY_LOCATION.getName(), true);
+        winrmProps.put(LocalLocationManager.CREATE_UNMANAGED.getName(), true);
```
bostko: I see you flagged it as unmanaged but unfortunately this setting was overridden in 
https://github.com/apache/incubator-brooklyn/blob/master/brooklyn-server/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalLocationManager.java#L95
@ahgittin What is the motivation behind  LocalLocationManager.java#L95 ?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
bostko: Please ignore my last comments.

PR #1088, opened 2015-12-04
[DISCUSSION] Sequential Entity Stop Order

mikezaccardo:
This adds a `stop.sequence.position` config value which enables YAML blueprints to explicitly specify a stop order.  This can prevent the problem of entities hanging infinitely on stop when they attempt to resolve dependent values of already-stopped entities.

Discussion points:
-Does this seem like a valid approach overall?
-Is the config value named intuitively?

Areas of improvement:
-Validation of config values (ensure values start at 1 and proceed sequentially)
-Allow multiple entities to share a value (e.g. 1, 2, 2, 3, 4 -- the 2s would stop in parallel)

IGNORE commit `Add all.members.up sensor to DynamicCluster` -- this was needed for testing.


googlielmo:
Excellent idea Mike!
For the sake of discussion, what happens if you don't provide a value? Is that an implicit 1?
Generally speaking, I'd prefer to introduce a partial ordering with a "stops-after" property, so that you can specify that entity B "stops-after" entity A.
WDYT?

PR #1089, opened 2015-12-04
tryLoadClass: return optional.absent on NoClassDefFoundError

aledsage:
We encountered this error when there were a Java entity in `lib/dropins` that still referenced the old `brooklyn.entity.basic.StartableApplication" (i.e. it had not been recompiled, with its source code updated).

The consequence was that `GET /v1/catalog/applications` failed, so no catalog items were displayed in the web-console. Also, the exception gives no indication of which catalog item the error came from (we unpacked all the jars, and did a string search across them to find it!).

With this fix, we'll get the log.debug that says which catalog item the error came from. The catalog item will be excluded from the list, and the other catalog items returned.

Ideally I'd like something better than just a `log.debug` to tell the user of this serious problem. Not convinced that `JavaBrooklynClassLoadingContext` is the right place to log.warn etc though.

I suggest we merge this, and then think about how to log better in a subsequent PR.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/classloading/JavaBrooklynClassLoadingContext.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/classloading/JavaBrooklynClassLoadingContext.java
@@ -27,13 +27,18 @@ import java.util.Collections;
 import java.util.Enumeration;
 
 import org.apache.brooklyn.api.mgmt.ManagementContext;
+import org.apache.brooklyn.core.entity.AbstractEntity;
 import org.apache.brooklyn.core.mgmt.persist.DeserializingClassRenamesProvider;
 import org.apache.brooklyn.util.exceptions.Exceptions;
 import org.apache.brooklyn.util.guava.Maybe;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Objects;
 
 public class JavaBrooklynClassLoadingContext extends AbstractBrooklynClassLoadingContext {
+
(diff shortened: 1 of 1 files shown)


ygy:
LGTM. Tested in a production setup scenario.

core/src/main/java/org/apache/brooklyn/core/mgmt/classloading/JavaBrooklynClassLoadingContext.java
```diff
@@ -82,6 +87,10 @@ private ClassLoader getClassLoader() {
         try {
             className = DeserializingClassRenamesProvider.findMappedName(className);
             return (Maybe) Maybe.of(getClassLoader().loadClass(className));
+        } catch (NoClassDefFoundError e) {
+            String msg = "Invalid linkage in (transitive dependencies of) class "+className+": "+e.toString();
+            LOG.debug(msg);
```
bostko: As you said in the comment I prefer to be warning.
Users are not supposed to use invalid classes in their apps.

neykov: This can be called multiple times a second due to refreshes from the browser, doesn't it flood the logs?

bostko: It is, you are right.

aledsage: I'll live with flooding the debug log, as folk should really fix he underlying problem.
Longer term, we could guard to only log.warn once about it for a given class.

aledsage:
Merging now.

PR #1090, opened 2015-12-05
Add timeout to attributeWhenReady if entity is in destructive state

mikezaccardo:
Previously, `attributeWhenReady` would only abort if the entity was on fire.

Now `attributeWhenReady` will timeout after 1 minute if the entity is in the `STOPPING`, `STOPPED`, or `DESTROYED` state.

The call does not abort straight away because there is a chance that the dependent values may be needed in the `stop` phase, so this timeout gives such values a fair chance to be resolved before the resolution is canceled by way of a timeout.


ahgittin:
nice fix!  merged, with a tweak in the api 1c80fb4d1a8dc06138246d2558c30c6dff41db68 .  does that sound reasonable @mikezaccardo ?

PR #1093, opened 2015-12-08
Fix/deferred location config

alasdairhodge:
Resolves issues with `$brooklyn:external()` in location config:
* `TypeCoercions` resolves `DeferredSupplier` values
* locations constructed within a new execution context on `LocalLocationManager` so that relevant context objects can be retrieved in a consistent way

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalLocationManager.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalLocationManager.java
@@ -23,6 +23,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import java.io.Closeable;
 import java.util.Collection;
 import java.util.Map;
+import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.brooklyn.api.location.Location;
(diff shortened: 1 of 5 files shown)


alasdairhodge:
Once this is reviewed/merged into `master`, I'll issue the same fix against `0.8.x`.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalLocationManager.java
```diff
@@ -78,9 +85,13 @@
     public LocalLocationManager(LocalManagementContext managementContext) {
         this.managementContext = checkNotNull(managementContext, "managementContext");
         this.locationFactory = new InternalLocationFactory(managementContext);
-        
+
         this.storage = managementContext.getStorage();
-        locationTypes = storage.getMap("locations");
+        this.locationTypes = storage.getMap("locations");
+
+        BasicExecutionManager executionManager = new BasicExecutionManager(managementContext.getManagementNodeId());
```
aledsage: Do we really need to create our own `BasicExecutionManager`? Can we not reuse `managementContext.getExecutionManager()`?

core/src/test/java/org/apache/brooklyn/core/location/LocationExtensionsTest.java
```diff
@@ -164,21 +166,24 @@ public void testAddExtensionThroughLocationSpecIllegally() {
         try {
             Location loc = createConcrete(MyExtension.class, "not an extension");
             fail("loc="+loc);
-        } catch (IllegalArgumentException e) {
+        } catch (Exception e) {
+            assertTrue(Exceptions.getFirstInteresting(e) instanceof IllegalArgumentException);
```
aledsage: I prefer that `e` is rethrown if it doesn't match, rather than losing the exception in the error report. Alex (I think it was) added:

    Asserts.expectedFailureOfType(e, IllegalArgumentException.class)

But not sure if that deals with looking at the `causedBy` properly.

aledsage:
@alasdairhodge Looks good. Only one important comment about reusing the `ManagementContext`s ExecutionManager, rather than creating a new one.

aledsage:
Actually, can we have a test that externalised configuration now works in locations? As I recall, it was only in one situation that it failed - when the config was accessed during the `init` of the location object? For example in `JcloudsLocation.setCreationString()`, which is called by `JcloudsLocation.configure()`?

aledsage:
How about a test like this in `ExternalConfigYamlTest`:

    @Test
    public void testExternalisedLocationConfigAccessedDuringLocationInit() throws Exception {
        String yaml = Joiner.on("\n").join(
            "services:",
            "- type: org.apache.brooklyn.core.test.entity.TestApplication",
            "location:",
            "  aws-ec2:",
            "    identity: myidentity",
            "    credential: mycredential",
            "    region: $brooklyn:external(\"myprovider\", \"mykey\")");

        TestApplication app = (TestApplication) createAndStartApplication(new StringReader(yaml));
        waitForApplicationTasks(app);
        LocationInternal loc = (LocationInternal) Iterables.getOnlyElement( app.getLocations() );
        assertEquals(loc.config().getLocalBag().getDescription(), "aws-ec2:myHardcodedVal");
    }

Does that now pass for you? It fails for me in master:

    org.apache.brooklyn.util.exceptions.CompoundRuntimeException: Unable to instantiate item, 2 errors including: Transformer for brooklyn-camp gave an error creating this plan: Illegal parameter for 'location' (aws-ec2); not resolvable: java.lang.ClassCastException: org.apache.brooklyn.camp.brooklyn.spi.dsl.methods.BrooklynDslCommon$DslExternal cannot be cast to java.lang.String

aledsage:
@alasdairhodge any progress on including that test and getting it to pass? Ping me if you want to discuss or pair.

ahgittin:
@alasdairhodge ping

ahgittin:
@alasdairhodge ping

PR #1094, opened 2015-12-08
Fix assertions for 'succeeds eventually' behaviour.

geomacy:
The following test case will not work with the code as currently defined:

```yaml
name: testcase1
location:
  mylocation
services:
- type: org.apache.brooklyn.entity.webapp.tomcat.TomcatServer
  id: tomcatServer
  timeout: 15m
- type: org.apache.brooklyn.test.framework.TestCase
  name: TomServTest
  brooklyn.children:
  - type: org.apache.brooklyn.test.framework.TestSensor
    target: $brooklyn:component("tomcatServer")
    sensor: service.isUp
    timeout: 1m
    assert:
    - equals: true
```

With the two entities (server and test) deployed independently instead of
within the same test case, the "checkAssertions" code will fail as it collects
all failures in AssertionSupport, when it should only collect those that
occur from the "Asserts.succeedsEventually".

Changing this code in this commit to allow tests to work.  The effect will
be that only one failed assertion will be reported on, even if many
assertions might fail. We can come back to this in order to see if multiple
assertion failures may be collected, but this is not necessary now.

### DIFF
--- a/usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestFrameworkAssertions.java
+++ b/usage/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestFrameworkAssertions.java
@@ -144,7 +144,7 @@ public class TestFrameworkAssertions {
                 @Override
                 public void run() {
                     T actual = actualSupplier.get();
-                    checkActualAgainstAssertions(support, assertions, target, actual);
+                    checkActualAgainstAssertions(assertions, target, actual);
                 }
             });
         } catch (Throwable t) {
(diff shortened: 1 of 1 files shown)


johnmccabe:
:+1: my positive and negative tests are now working as expected with the timeout in the `Asserts#succeedsEventually` being honoured.

sjcorbett:
@geomacy If you add \`\`\` (three backticks) on the lines before and after code blocks they will be formatted
```
like this
```

sjcorbett:
You can also indicate it's yaml by starting the block with \`\`\`yaml to get a bit of syntax highlighting:

```yaml
location: foo
services:
- baz: bar
  
```

geomacy:
Thanks @sjcorbett have updated commit as suggested.

aledsage:
Build failure looks unrelated, and an infrastructure thing ("Internal error: java.lang.IllegalStateException: Invalid object ID 19 iota=66").

Note that Sam's comments about formatting yaml refer to the PR description, rather than changes within the code.

Merging.

PR #1096, opened 2015-12-09
GeoDns improvements

sjcorbett:
A variety of improvements to the geo-DNS entities. In particular:
* Entities whose service state is not `RUNNING` are automatically excluded from the pool of targets. This can be disabled to maintain existing behaviour.
* The group to be tracked by `AbstractGeoDnsService` can be given in a config key, which means users don't need to call `setTargetEntityProvider` and thus that the service can be used from yaml.

### DIFF
--- a/software/webapp/src/test/java/org/apache/brooklyn/entity/dns/geoscaling/GeoscalingWebClientTest.java
+++ b/software/webapp/src/test/java/org/apache/brooklyn/entity/dns/geoscaling/GeoscalingWebClientTest.java
@@ -26,11 +26,17 @@ import static org.apache.brooklyn.entity.dns.geoscaling.GeoscalingWebClient.PROV
 import static org.testng.Assert.assertNotNull;
 import static org.testng.Assert.assertNull;
 
+import org.apache.brooklyn.api.mgmt.ManagementContext;
+import org.apache.brooklyn.core.internal.BrooklynProperties;
+import org.apache.brooklyn.core.mgmt.internal.ManagementContextInternal;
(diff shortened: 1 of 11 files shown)


software/webapp/src/test/java/org/apache/brooklyn/entity/dns/geoscaling/GeoscalingWebClientTest.java
```diff
@@ -100,12 +106,9 @@
 ... 31 more
  */
 @Test(groups="Broken", enabled=false)
```
ahgittin: still broken?

sjcorbett: Yes, for reasons described here: https://issues.apache.org/jira/browse/BROOKLYN-200

usage/camp/src/test/resources/org/apache/brooklyn/camp/brooklyn/geodns.yaml
```diff
@@ -0,0 +1,42 @@
...
+# specific language governing permissions and limitations
+# under the License.
+#
+services:
+
+- name: Web cluster
+  type: org.apache.brooklyn.entity.group.DynamicRegionsFabric
+  id: web-fabric
+
+  # Location required but test should not do any provisioning.
```
ahgittin: would be nice if locations weren't required -- but not needed for this PR

ahgittin:
looks good.  minor comments, can be addressed subsequently.  merging.

PR #1098, opened 2015-12-10
Misc MySQL 5.5.x and 5.6.x fixes

alasdairhodge:
Remaining issues:
* 5.7.x introduces breaking changes in certain command/script behaviours, notably `mysql_install_db`, which is invoked during `customize`
* 5.5.x and 5.6.x working on Ubuntu but failing on RHEL / Centos / Amazon Linux due to missing perl dependencies

Will address these in subsequent PR.

### DIFF
--- a/software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlNodeImpl.java
+++ b/software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlNodeImpl.java
@@ -135,8 +135,8 @@ public class MySqlNodeImpl extends SoftwareProcessImpl implements MySqlNode {
     public int getPort() {
         return getAttribute(MYSQL_PORT);
     }
-    
-    public String getSocketUid() {
+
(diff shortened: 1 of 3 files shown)


sjcorbett:
Looks good.

software/database/src/main/java/org/apache/brooklyn/entity/database/mysql/MySqlNodeImpl.java
```diff
@@ -135,8 +135,8 @@ protected void disconnectSensors() {
     public int getPort() {
         return getAttribute(MYSQL_PORT);
     }
-    
-    public String getSocketUid() {
+
+    public synchronized String getSocketUid() {
```
aledsage: Post merge comment... I'd have preferred if we'd declared a `private final Object mutex` to synchronize on. This is probably ok, but in general I don't like synchronizing on `this` in case other threads/callers have synchronized on this in other situations. No strong feelings.

PR #1099, opened 2015-12-10
replaced jython imports with guava equivalents

johnmccabe:
encountered the following apparently unnecessary jython imports while looking at building the proposed split brooklyn-server repo, have replaced with guava equivalents.

### DIFF
--- a/usage/camp/pom.xml
+++ b/usage/camp/pom.xml
@@ -195,9 +195,6 @@
                     </supportedProjectTypes>
                     <instructions>
                         <Export-Package>org.apache.brooklyn.*</Export-Package>
-                        <Import-Package>
-                            *
-                        </Import-Package>
                     </instructions>
                 </configuration>
             </plugin>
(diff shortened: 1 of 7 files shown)


usage/camp/pom.xml
```diff
@@ -196,7 +196,6 @@
                     <instructions>
                         <Export-Package>org.apache.brooklyn.*</Export-Package>
                         <Import-Package>
```
neykov: You can remove the whole element, no longer needed.

neykov:
Looks great. Worth adding the package to the `IllegalImport` list in `parent/pom.xml` so we get build failures.

johnmccabe:
@neykov does that checkstyle ever run? I've included the jython package and added a duff import to a class but I can't seem to get it to fail the build or generate a checkstyle report?
```
<property name="illegalPkgs" value="com.google.api.client.repackaged,org.python.google"/>
```

sjcorbett:
@johnmccabe Seems to be working:
```
[INFO] There are 1 checkstyle errors.
[ERROR] BrooklynFeatureEnablement.java[21:1] (imports) IllegalImport: Import from illegal package - com.google.api.client.repackaged.com.google.common.base.*.
```
The check runs in the `process-classes` phase, after `compile`.

johnmccabe:
Pushed @neykov suggestions

neykov:
thanks @johnmccabe

PR #1100, opened 2015-12-10
Generate Swagger API Specification using a maven plugin

andreaturli:
- combined with https://github.com/swagger-api/swagger-codegen, it can generate rest bindings for multiple languages

### DIFF
--- a/usage/rest-api/pom.xml
+++ b/usage/rest-api/pom.xml
@@ -7,9 +7,9 @@
     to you under the Apache License, Version 2.0 (the
     "License"); you may not use this file except in compliance
     with the License.  You may obtain a copy of the License at
-    
+
      http://www.apache.org/licenses/LICENSE-2.0
-    
+
     Unless required by applicable law or agreed to in writing,
     software distributed under the License is distributed on an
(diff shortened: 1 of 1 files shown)


usage/rest-api/pom.xml
```diff
@@ -98,4 +98,48 @@
...
+                <groupId>com.github.kongchen</groupId>
+                <artifactId>swagger-maven-plugin</artifactId>
+                <version>3.1.0</version>
+                <configuration>
+                    <apiSources>
+                        <apiSource>
+                            <springmvc>false</springmvc>
+                            <locations>org.apache.brooklyn.rest.api</locations>
+                            <schemes>http,https</schemes>
+                            <host>brooklyn.io</host>
```
sjcorbett: Why include `host`? brooklyn.io doesn't serve this API.

usage/rest-api/pom.xml
```diff
@@ -98,4 +98,48 @@
...
+                            <basePath>/api</basePath>
```
sjcorbett: Isn't this dependent on whoever has deployed the API?

andreaturli:
thanks @sjcorbett, you are right, I was misled by the reference example of that maven plugin. I'll remove those configurations as not required.

PR #1101, opened 2015-12-10
Ported some Clocker changes back to brooklyn.

Graeme-Miller:
It is now possible to remove security groups permissions

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupCustomizer.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupCustomizer.java
@@ -72,6 +72,7 @@ import com.google.common.collect.FluentIterable;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
+import com.google.common.util.concurrent.UncheckedExecutionException;
 
 /**
  * Configures custom security groups on Jclouds locations.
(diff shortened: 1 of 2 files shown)


sjcorbett:
Would be good to add tests for the new additions to `JcloudsLocationSecurityGroupCustomizerTest`.

Graeme-Miller:
@sjcorbett thanks Sam, have added some tests now.
Also, I amended removePermissionsFromLocation so that is is synchronized on the class, to keep it in line with addPermissionsToLocation

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupCustomizerTest.java
```diff
@@ -147,6 +149,59 @@ public void testAddPermissionsToNode() {
...
+    public void testAddPermissionWhenNoExtension() {
+        IpPermission ssh = newPermission(22);
+        IpPermission jmx = newPermission(31001);
+        String nodeId = "node";
+
+        when(securityApi.listSecurityGroupsForNode(nodeId)).thenReturn(Collections.<SecurityGroup>emptySet());
+
+        RuntimeException exception = null;
+        try {
+            customizer.addPermissionsToLocation(ImmutableList.of(ssh, jmx), nodeId, computeService);
```
sjcorbett: The approach we generally take is to `fail("should have thrown some exception")` after the exceptional statement and then to assert on the exception in the catch block. You can also use `Exceptions.getFirstInteresting(e)` and the recent additions of `shouldHaveFailedPreviously()` and `expectedFailure()` to the `Asserts` class.

sjcorbett:
@Graeme-Miller Thanks. One minor comment but am happy for this to be merged as is.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupCustomizer.java
```diff
@@ -201,6 +229,47 @@ public JcloudsLocationSecurityGroupCustomizer addPermissionsToLocation(final Jcl
...
         }
     }
+    /**
+     * Removes the given security group permissions from the given node with the given compute service.
+     * <p>
+     * Takes no action if the compute service does not have a security group extension.
+     * @param permissions The set of permissions to be removed from the location
+     * @param location Location to remove permissions from
+     */
+    public void removePermissionsFromLocation(final JcloudsMachineLocation location, final Iterable<IpPermission> permissions) {
```
aledsage: @Graeme-Miller should include `@since 0.9.0` when adding public methods like this.

PR #1104, opened 2015-12-11
Brooklyn properties files may be symlinks

sjcorbett:
The `?00` permissions check is enforced on the link's target rather than the link itself.

### DIFF
--- a/usage/launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynLauncher.java
+++ b/usage/launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynLauncher.java
@@ -22,14 +22,15 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import java.io.Closeable;
 import java.io.File;
+import java.io.IOException;
 import java.net.InetAddress;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


sjcorbett:
`Build timed out (after 40 minutes). Marking the build as aborted.`

usage/launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynLauncher.java
```diff
@@ -710,32 +719,42 @@ private void initManagementContext() {
...
+     */
+    private File resolveSymbolicLink(File f) {
+        File f2 = f;
+        try {
+            f2 = f.getCanonicalFile();
+            if (Files.isSymbolicLink(f.toPath())) {
+                LOG.debug("Resolved symbolic link: {} -> {}", f, f2);
+            }
+        } catch (IOException e) {
+            LOG.warn("Could not determine canonical name of global properties file", e);
```
aledsage: Perhaps change log message to not say `global properties file`, as this method is more generic (even if currently only called in that one case)?

How about `Could not determine canonical path of file "+f, e`

aledsage:
@sjcorbett Looks good - one very minor comment about an improvement to a log message. Good to merge.

ahgittin:
agree re log message, i've applied that and merged

PR #1108, opened 2015-12-14
Hazelcast Cluster support

ygy:
- Upgrade Hazelcast to v.3.5.4
- Small refactoring
- Adding integration tests


pom.xml
```diff
@@ -126,7 +126,7 @@
         <mockwebserver.version>20121111</mockwebserver.version>
         <freemarker.version>2.3.22</freemarker.version>
         <commons-io.version>2.4</commons-io.version>
-        <hazelcast.version>3.0</hazelcast.version>
+        <hazelcast.version>3.5.4</hazelcast.version>
```
ahgittin: this probably belongs in the hazelcast project rather than here

ahgittin:
looks good.  one comment which i'll address.  this will force a bit of cleanup on the repo migration side but i'll take care of that.

pom.xml
ahgittin: scratch that, i see other projects mongo/qpid/etc used in tests for one subproject follow the smae pattern (and you just changed it) so that's a separate issue

ahgittin:
merging

PR #1110, opened 2015-12-15
Fix concurrent use of entity.setAttribute()

aledsage:
Previously we were using a vanilla LinkedHashMap for storing attributes
(ever since the FEATURE_USE_BROOKLYN_LIVE_OBJECTS_DATAGRID_STORAGE
was disabled by default!)

Now uses synchronized set / synchronized map (rather than `ConcurrentMap`, as need to accept null values), with underlying `LinkedHashSet` to preserve order.

Adds `EntityConcurrencyTest` for concurrently:
- setting attributes (previously failed fairly often)
- setting config
- adding tags
- adding groups (previously failed sometimes)
- adding children
- adding locations
- adding policies
- adding enrichers
- adding feeds

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/sensor/AttributeMap.java
+++ b/core/src/main/java/org/apache/brooklyn/core/sensor/AttributeMap.java
@@ -21,6 +21,7 @@ package org.apache.brooklyn.core.sensor;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Map;
 
 import org.apache.brooklyn.api.entity.Entity;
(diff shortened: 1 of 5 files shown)


ahgittin:
looks good.  presumably the objective is to prevent concurrent mod exceptions, and caller is still not granted access to the concurrency model.  so you can't e.g. *increment* a sensor.  but this would allow us to add a `modify(Sensor<T>, Function<T,T> modification)` method which would modify atomically.

core/src/main/java/org/apache/brooklyn/core/sensor/AttributeMap.java
```diff
@@ -59,7 +60,20 @@
...
+        // Not using ConcurrentMap, because want to (continue to) allow null values.
+        // Could use ConcurrentMapAcceptingNullVals (with the associated performance hit on entrySet() etc).
+        this(entity, Collections.synchronizedMap(Maps.<Collection<String>, Object>newLinkedHashMap()));
+    }
+
+    /**
+     * Creates a new AttributeMap.
+     *
+     * @param entity the EntityLocal this AttributeMap belongs to.
+     * @param values the Map in which to store the values - should be concurrent or synchronized.
```
sjcorbett: values should be storage.

sjcorbett:
+1. I've left one trivial comment but otherwise looks good to me.

aledsage:
@ahgittin thanks. Interesting we weren't getting any `ConcurrentModificationException`s that I noticed. Instead we just had some subset of the 100 setAttribute calls actually having worked! I encountered this when writing tests for https://github.com/brooklyncentral/advanced-networking/pull/86. It was most confusing and frustrating when assuming the bug was in advanced-networking, until I left the advanced-networking and tried a simple tests in brooklyn!

@sjcorbett thanks; will update that javadoc and then merge.

ahgittin:
@aledsage yikes -- great this is fixed!

PR #1111, opened 2015-12-16
Proper ordering of catalog config

neykov:
Increment the priority because the config container is a `Set`. Server-side we are using an ordered set which results in correctly ordered items on the wire (as a list). Clients which use the java bindings though will push the items in an unordered set - so give them means to recover the correct order.


aledsage:
Thanks @neykov - looks straightforward; merging.

PR #1112, opened 2015-12-16
support for external config in BrooklynProperties

aledsage:
An alternative implementation to https://github.com/apache/incubator-brooklyn/pull/1018 (i.e. replaces that PR).

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/DeferredBrooklynProperties.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/DeferredBrooklynProperties.java
@@ -1,0 +1,370 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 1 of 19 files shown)


aledsage:
Test failure is unrelated: `BindException: Address already in use` in `BrooklynApiRestClientTest.setUp`. This is caused by us finding a free port, and then immediately trying to use it - but another process running on the same jenkins test box has claimed that port.

We call `Networking.nextAvailablePort(FAVOURITE_PORT)` in `BrooklynRestApiLauncher.startServer`.  We should really let it choose a port, given that class is just in test code.

But that should be for a different PR.

Closing and re-opening to kick off jenkins again.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/DeferredBrooklynProperties.java
```diff
@@ -0,0 +1,370 @@
...
+    public DeferredBrooklynProperties(BrooklynProperties delegate, ManagementContextInternal mgmt) {
+        this.delegate = checkNotNull(delegate, "delegate");
+        this.mgmt = checkNotNull(mgmt, "mgmt");
+    }
+    
+    private Object transform(ConfigKey<?> key, Object value) {
+        if (value instanceof CharSequence) {
+            String raw = value.toString();
+            if (raw.startsWith(EXTERNAL_CONFIG_PREFIX)) {
+                CampYamlParser parser = mgmt.getConfig().getConfig(CampYamlParser.YAML_PARSER_KEY);
```
ahgittin: why not `startsWith("$brooklyn:)` ?

ahgittin:
nice work.  one substantive comment -- if not addressing, put a TODO in the doc.  but for instance i could see it useful to set an entity etc.  also pathological case but if user wanted to say

    external_prefix=$brooklyn:external

this would break it.  if we supported `literal` we could say:

    external_prefix=$brooklyn:literal("$brooklyn:external")

also docs need update:
* to say what is permitted in properties (ie currently only `$brooklyn:external(...)`)
* to say that the method for constructing `BrooklynProperties` has changed (release notes / api breakages)

aledsage:
Thanks @ahgittin - great comments, and thanks for pointing out docs (too often ignored by me and others!). I've incorporated those changes and written some docs. Merging now.

PR #1113, opened 2015-12-16
Rebind NPE fixes

neykov:
The logic in some of the places assumes that the field is always non-null, initialized in the constructor. The assumption is broken when rebinding from an old state, because constructors are not called.


neykov:
Failures look related, don't merge yet.

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigKey.java
```diff
@@ -217,6 +217,10 @@ public ConfigInheritance getInheritance() {
     /** @see ConfigKey#getConstraint() */
     @Override @Nonnull
     public Predicate<? super T> getConstraint() {
+        // Could be null after rebinding
+        if (constraint != null) {
```
aledsage: Did you mean this to be `if (constraint == null)`?

neykov:
@aledsage obviously it was wrong, fixed it now.

aledsage:
Merging now.

PR #1114, opened 2015-12-17
Downstream-parent / OSGi

sjcorbett:
Removes the `Export-Package` manifest configuration from `brooklyn-downstream-parent` and adds a number of excludes of packages that are not bundles to `Import-Package`. 

Projects using downstream-parent can be loaded into catalogues via bundles by default. This can be tested by running `DownstreamParentTest` in the QA module.


ahgittin:
looks good, testing then will merge

PR #1117, opened 2015-12-19
Fix cassandra on aws

aledsage:
And some other minor improvements as well.


iyovcheva:
Looks good, small comment only.

ahgittin:
good spot @iyovcheva .  i'll fix that and merge.

PR #1120, opened 2015-12-22
Test correct persistence of attributeWhenReady

googlielmo:
An entity with an instance of `attributeWhenReady` is serialised and then deserialised

### DIFF
--- a/brooklyn-server/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/DslAndRebindYamlTest.java
+++ b/brooklyn-server/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/DslAndRebindYamlTest.java
@@ -95,7 +95,6 @@ public class DslAndRebindYamlTest extends AbstractYamlTest {
 
     public Application rebind(Application app) throws Exception {
         RebindTestUtils.waitForPersisted(app);
-        // Removed because of issues in some tests: // RebindTestUtils.checkCurrentMementoSerializable(app);
         Application result = RebindTestUtils.rebind(mementoDir, getClass().getClassLoader());
         mgmtContexts.add(result.getManagementContext());
         return result;


usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/DslAndRebindYamlTest.java
```diff
@@ -124,6 +126,19 @@ public void testDslAttributeWhenReady() throws Exception {
...
+        Entity testEntity = entityWithAttributeWhenReady();
+        Application app2 = rebind(testEntity.getApplication());
+        Entity e2 = Iterables.getOnlyElement( app2.getChildren() );
+
+        Maybe<Object> maybe = ((EntityInternal)e2).config().getLocalRaw(TestEntity.CONF_NAME);
+        Assert.assertTrue(maybe.isPresentAndNonNull());
+        Assert.assertTrue(BrooklynDslDeferredSupplier.class.isInstance(maybe.get()));
+        BrooklynDslDeferredSupplier deferredSupplier = (BrooklynDslDeferredSupplier) maybe.get();
+        Assert.assertEquals(deferredSupplier.toString(), "$brooklyn:entity(\"x\").attributeWhenReady(\"foo\")");
+    }
```
ahgittin: suggest also adding something like this:

```
        BrooklynMementoRawData raw = BrooklynPersistenceUtils.newStateMemento(app2.getManagementContext(), MementoCopyMode.LOCAL);
        String persistedStateForE2 = raw.getEntities().get(e2.getId());
        Matcher matcher = Pattern.compile(".*\\<test.confName\\>(.*)\\<\\/test.confName\\>.*", Pattern.DOTALL)
        		.matcher(persistedStateForE2);
        Assert.assertTrue(matcher.find());
        String testConfNamePersistedState = matcher.group(1);
        
        Assert.assertNotNull(testConfNamePersistedState);
        // should be about 200 chars long...
        Assert.assertTrue(testConfNamePersistedState.length() < 400, "persisted state too long: "+testConfNamePersistedState);
```

this asserts the persisted state is as expected, and not too big.  btw this is what it looks like:

```
//        <test.confName>
//        <org.apache.brooklyn.camp.brooklyn.spi.dsl.methods.DslComponent_-AttributeWhenReady>
//          <component>
//            <componentId>x</componentId>
//            <scope>GLOBAL</scope>
//          </component>
//          <sensorName>foo</sensorName>
//        </org.apache.brooklyn.camp.brooklyn.spi.dsl.methods.DslComponent_-AttributeWhenReady>
//      </test.confName>
```

usage/camp/src/test/java/org/apache/brooklyn/camp/brooklyn/DslAndRebindYamlTest.java
```diff
@@ -124,6 +126,19 @@ public void testDslAttributeWhenReady() throws Exception {
     }
 
     @Test
+    public void testDslAttributeWhenReadyPersisted() throws Exception {
```
ahgittin: as discussed we want to test where `attributeWhenReady` is used in an entity spec, e.g. something like

```
name: Test Cluster
services:
- type: org.apache.brooklyn.entity.stock.DynamicCluster
  id: test-cluster
  initialSize: 1
  memberSpec:
    $brooklyn:entitySpec:
      type: org.apache.brooklyn.entity.test.TestEntity
      brooklyn.config:
        test.confName: $brooklyn:component("test-cluster").attributeWhenReady("sensor")
```

with this you could then in your test set `test-cluster` sensor `sensor` e.g. to `foo` and then ensure that `foo` is *not* being written (ie not in the `testConfNamePersistedState`)

googlielmo: I set up a test with that (see below), except that I now get `java.io.NotSerializableException: org.apache.brooklyn.entity.group.DynamicClusterImpl$NextClusterMemberIdSupplier`
Will look into it.

aledsage: For the `java.io.NotSerializableException`: that's surprising. We shouldn't be relying on normal java serialization. We use xstream for persistence, which doesn't need `Serializable`.

Some test assertions can enable a stronger assertion about being able to serialize all of the persisted state using java serialization (in an attempt to fail-fast). We can just disable that, if it's being done. It's too strict.

Worth looking down the stacktrace to see what is trying to do the java serialization, to see if it is the test assertion or something else.

googlielmo: @aledsage you are right! it was due to a stronger test assertion -- I disabled it for now.

googlielmo:
@ahgittin I turned it into a slightly more complicated test that serialises and then deserialises an entity **spec** while a `Task` is waiting for the attribute value to be available. This is to simulate a situation that may happen frequently at runtime. Previously we only tested AWR persistence with no other threads in wait.

brooklyn-server/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/DslAndRebindYamlTest.java
```diff
@@ -83,15 +95,14 @@ protected Logger getLogger() {
 
     public Application rebind(Application app) throws Exception {
         RebindTestUtils.waitForPersisted(app);
-        // not strictly needed, but for good measure:
-        RebindTestUtils.checkCurrentMementoSerializable(app);
+        // Removed because of issues in some tests: // RebindTestUtils.checkCurrentMementoSerializable(app);
```
aledsage: Fine to just delete this line entirely, I think.

brooklyn-server/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/DslAndRebindYamlTest.java
```diff
@@ -115,34 +125,207 @@ public T call() throws Exception {
...
+        Assert.assertTrue(maybe.isPresentAndNonNull());
+        Assert.assertTrue(BrooklynDslDeferredSupplier.class.isInstance(maybe.get()));
+        BrooklynDslDeferredSupplier deferredSupplier = (BrooklynDslDeferredSupplier) maybe.get();
+        Assert.assertEquals(deferredSupplier.toString(), "$brooklyn:entity(\"test-cluster\").attributeWhenReady(\"sensor\")");
+
+        // Check that the Task is still waiting for attribute to be ready
+        Assert.assertFalse(stringFuture.isDone());
+
+        // Now set sensor value
+        ((EntityInternal) clusterEntity1).sensors().set(Sensors.newStringSensor("sensor"), "bar");
```
aledsage: This seems to be confusing modifying the entities from before rebind, but doing so after rebind. The call to `rebind()` is to simulate us having restarted the Brooklyn server. It's configurable whether we actually terminate the original management context. I guess this test is passing because the original management context was not terminated.

I'm looking at these tests in more detail. I'll make a few changes and submit a pull request against your branch @googlielmo (either this evening or tomorrow).

aledsage:
See suggested additional tests, and alterations to tests, in https://github.com/googlielmo/incubator-brooklyn/pull/2

brooklyn-server/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/DslAndRebindYamlTest.java
```diff
@@ -19,58 +19,82 @@
...
         mementoDir.deleteOnExit();
         LocalManagementContext mgmt = RebindTestUtils.newPersistingManagementContext(mementoDir, classLoader, 1);
         mgmtContexts.add(mgmt);
         return mgmt;
     }
+
+    @BeforeMethod(alwaysRun = true)
+    @Override
+    public void setUp() {
+    	super.setUp();
```
aledsage: Are these tabs? Or 3 spaces? Can you change them to 4 spaces please?

aledsage:
Thanks @googlielmo - merging now (and will check about tabs versus spaces after that).

PR #1123, opened 2015-12-30
RabbitMQ upgrade

ygy:
RabbitMQ to v.3.6.0
Erlang to v.18.2

RabbitMQ recommends Erlang 18. Release notes state:
Minimum required Erlang version is R16B03 for plain ("just TCP")
connections for all protocols 
and 17.5 for TLS ones (18.x is recommended for both).

The recommended provider for up to date Erlang versions is Erlang
Solutions now.
Supported platforms by Erlang Solutions are: CentOS, RHEL, Ubuntu and
Debian.

New Erlang versions for SUSE are provided via the openSUSE repositories

### DIFF
--- a/brooklyn-library/software/messaging/src/main/java/org/apache/brooklyn/entity/messaging/rabbit/RabbitSshDriver.java
+++ b/brooklyn-library/software/messaging/src/main/java/org/apache/brooklyn/entity/messaging/rabbit/RabbitSshDriver.java
@@ -27,7 +27,6 @@ import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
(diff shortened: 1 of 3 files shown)


brooklyn-library/software/messaging/src/main/java/org/apache/brooklyn/entity/messaging/rabbit/RabbitSshDriver.java
```diff
@@ -205,4 +195,69 @@ public void kill() {
...
+        return chainGroup(
+                INSTALL_WGET,
+                format("wget --quiet -O %s %s", debFileName, entity.getConfig(RabbitBroker.ERLANG_DEB_REPO_URL)),
+                sudo(format("dpkg -i %s", debFileName))
+            );
+    }
+
+    private String getZypperRepository() {
+        return chainGroup(
+                ok(sudo("zypper --non-interactive addrepo http://download.opensuse.org/repositories/devel:/languages:/erlang/SLE_11_SP3 erlang_sles_11")),
```
bostko: I see this is a moved code from previously. However do we need `ok` here?

ahgittin: @bostko You're probably right, but for another PR...

PR #1124, opened 2015-12-30
On stop Application, unmanage location provisioner

bostko:


brooklyn-server/core/src/main/java/org/apache/brooklyn/core/entity/AbstractApplication.java
```diff
@@ -220,13 +221,34 @@ public void stop() {
                 //but that could be handled by the impl at management
                 //(keeping recently unmanaged things)  
                 //  however unmanaging must be done last, _after_ we stop children and set attributes 
+                Location location = Iterables.getFirst(getLocations(), null);
```
Graeme-Miller: Could getLocations return an empty Collection? If so, we will use the default value of null and throw a null pointer exception in unmanageProvisionerIfNotUsed

brooklyn-server/core/src/main/java/org/apache/brooklyn/core/entity/AbstractApplication.java
```diff
@@ -220,13 +221,34 @@ public void stop() {
...
                 getEntityManager().unmanage(this);
+                unmanageProvisionerIfNotUsed(location);
             }
         }
 
         logApplicationLifecycle("Stopped");
     }
 
+
+    private boolean unmanageProvisionerIfNotUsed(/*MachineProvisioningLocation<MachineLocation>*/Location provisioner) {
```
Graeme-Miller: This method returns a boolean value, representing if anything was unmanaged. However, this isn't used by the calling code. Is it necessary?

brooklyn-server/core/src/main/java/org/apache/brooklyn/core/entity/AbstractApplication.java
```diff
@@ -220,13 +221,34 @@ public void stop() {
...
+        if (provisioner.getChildren().isEmpty()) {
+            Entity parentEntity = getParent();
+            if (parentEntity == null) {
+                getManagementContext().getLocationManager().unmanage(provisioner);
+                return true;
+            } else {
+                /*
```
Graeme-Miller: Else block code is commented out. What is the purpose of this section? Is it complete?

bostko:
@Graeme-Miller thank you for your comments!
I put a commented code which be a possible general approach for unmanaging instances of `MachineProvisioningLocation`.

brooklyn-server/core/src/main/java/org/apache/brooklyn/core/entity/AbstractApplication.java
```diff
@@ -220,13 +221,37 @@ public void stop() {
...
+                }
             }
         }
 
         logApplicationLifecycle("Stopped");
     }
 
+    //TODO Think of general approach for expunging location provisioners.
+    // I left a code which could be a possible general approach. However it needs tests.
+    protected boolean unmanageProvisionerIfNotUsed(/*MachineProvisioningLocation<MachineLocation>*/Location machineProvisioningLocation) {
```
aledsage: Can you write a simple test that demonstrates that the location was being left as managed (i.e. a failing test before this change, and passes after this change).

brooklyn-server/core/src/main/java/org/apache/brooklyn/core/entity/AbstractApplication.java
```diff
@@ -220,13 +221,37 @@ public void stop() {
...
+        if (machineProvisioningLocation.getChildren().isEmpty()) {
+            Entity parentEntity = getParent();
+            if (parentEntity == null) {
+                getManagementContext().getLocationManager().unmanage(machineProvisioningLocation);
```
aledsage: This doesn't feel right - it wasn't the application that created the location, so should it be the application that unmanages the location? If we do this, it would be good to have some of indicating that the location is "owned" by the app, rather than it just having been passed to the new application when calling `start(location)`.

bostko: For `MachineLocation` the entity is responsible for starting and managing the location
and the entity is responsible for stopping the `MachineLocation` instances (see 
https://github.com/apache/incubator-brooklyn/blob/master/brooklyn-server/software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java#L926 ).

Unlike the `MachineLocation` instance a `MachineProvisioningLocation` instance is created from `InternalEntityFactory.initEntityAndDescendants` when `entitySpec` is being processed
and respective to that instead of putting the expunge in `MachineLifecycleEffectorTasks.suspendAnyProvisionedMachines` I putted it directly in the `AbstractApplication.stop`.

Is there a case where the `MachineProvisioningLocation` instance could be owned by another application? Do we have cross application usage of `MachineProvisioningLocation` instances?

ahgittin:
@bostko yes -- i think with Clocker there can be `MachineProvisioningLocation` *instances* passed to an application which applications should not unmanage

is the problem here that in some cases a `MPL` instance is being created but never unmanaged, and so is leaking?  if so can you make a test case and we can discuss how best to clean up.  we may need reference counts, or maybe relationships; it could help to fix by making `Location` a subclass of `Entity` and then they are explicitly managed somewhere.

ahgittin:
@bostko is this problem fixed by https://github.com/apache/brooklyn-server/pull/70 ?

even if not let's continue discussion there

ahgittin:
(and please close this PR)

PR #1125, opened 2016-01-06
Update commons-lang3 to 3.3.2

sjcorbett:
Matches Alien4Cloud version, used by brooklyn-tosca.


aledsage:
Not sure this change is enough. Looking at `mvn dependency:tree`, it also has:

    [INFO] +- io.swagger:swagger-core:jar:1.5.3:compile
    [INFO] |  +- org.apache.commons:commons-lang3:jar:3.1:compile

Do we need an excludes in the swagger-core dependency (from brooklyn-utils-rest-swagger)? Will this break swagger?

sjcorbett:
@aledsage Thanks for checking this. I've excluded swagger-core's dependency on commons-lang3. As far as I can tell - manual experimentation with various GET/POST/DELETE requests - Swagger continues to function.

aledsage:
Thanks @sjcorbett - merging now.

PR #1127, opened 2016-01-06
improvements to rest-api module

andreaturli:
Initial work to simplify the rest server:

- remove `JsonNode` from API interfaces
- refactor consistently `toString`and `hashCode` and `equals` for the domain objects
- remove deprecated code, although LocationSpec needs still some work
- simplified pom dependencies

Interestingly, with these changes, the `swagger.json`, obtained from the Swagger Annotation, can be used to generate automatically a java client  using something like [that](https://github.com/swagger-api/swagger-codegen#to-generate-a-sample-client-library)

@sjcorbett could you have a look?

### DIFF
--- a/brooklyn-server/karaf/features/src/main/history/dependencies.xml
+++ b/brooklyn-server/karaf/features/src/main/history/dependencies.xml
@@ -1,0 +1,103 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<features xmlns="http://karaf.apache.org/xmlns/features/v1.3.0" name="org.apache.brooklyn-0.9.0-SNAPSHOT">
+    <feature version="0.0.0">
+        <feature prerequisite="false" dependency="false">brooklyn-api</feature>
+        <feature prerequisite="false" dependency="false">brooklyn-api</feature>
+        <feature prerequisite="false" dependency="false">brooklyn-camp-base</feature>
+        <feature prerequisite="false" dependency="false">brooklyn-camp-base</feature>
+        <feature prerequisite="false" dependency="false">brooklyn-camp-base</feature>
+        <feature prerequisite="false" dependency="false">brooklyn-camp-brooklyn</feature>
+        <feature prerequisite="false" dependency="false">brooklyn-core</feature>
+        <feature prerequisite="false" dependency="false">brooklyn-core</feature>
+        <feature prerequisite="false" dependency="false">brooklyn-rest-api</feature>
+        <feature prerequisite="false" dependency="false">brooklyn-utils-common</feature>
+        <feature prerequisite="false" dependency="false">brooklyn-utils-common</feature>
+        <feature prerequisite="false" dependency="false">brooklyn-utils-common</feature>
+        <feature prerequisite="false" dependency="false">brooklyn-utils-common</feature>
+        <feature prerequisite="false" dependency="false">brooklyn-utils-rest-swagger</feature>
+        <feature prerequisite="false" dependency="false">brooklyn-utils-rest-swagger</feature>
+        <feature prerequisite="false" dependency="false">jetty</feature>
+        <feature prerequisite="false" dependency="false">swagger-crippled</feature>
+        <feature prerequisite="false" dependency="false">war</feature>
+        <feature prerequisite="false" dependency="false">war</feature>
+        <bundle>mvn:ch.qos.logback/logback-classic/1.0.7</bundle>
+        <bundle>mvn:ch.qos.logback/logback-core/1.0.7</bundle>
+        <bundle>mvn:com.fasterxml.jackson.core/jackson-annotations/2.4.5</bundle>
+        <bundle>mvn:com.fasterxml.jackson.core/jackson-annotations/2.4.5</bundle>
+        <bundle>mvn:com.fasterxml.jackson.core/jackson-core/2.4.5</bundle>
+        <bundle>mvn:com.fasterxml.jackson.core/jackson-core/2.4.5</bundle>
+        <bundle>mvn:com.fasterxml.jackson.core/jackson-databind/2.4.5</bundle>
+        <bundle>mvn:com.fasterxml.jackson.core/jackson-databind/2.4.5</bundle>
+        <bundle>mvn:com.fasterxml.jackson.dataformat/jackson-dataformat-yaml/2.4.5</bundle>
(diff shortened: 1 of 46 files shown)


sjcorbett:
@andreaturli I'll take a look. The Jenkins failure is valid:
```
java.lang.AssertionError: expected / found:
[{"id":"123","name":"localhost","spec":"localhost","links":{"self":"/v1/locations/123"}}] 
[{"id":"123","name":"localhost","spec":"localhost","config":{},"links":{"self":"/v1/locations/123"}}]
```

sjcorbett:
In org.apache.brooklyn.rest.domain.LocationSummaryTest.testSerializeToJSON.

andreaturli:
thanks @sjcorbett I think I fixed it now, sorry.
Looks like jenkins has not restarted ...

sjcorbett:
The diff is too large to review on GitHub. I'll leave comments here instead.

sjcorbett:
brooklyn-server/karaf/features/src/main/history/dependencies.xml:
Do you really want to remove `BROOKLYN_VERSION` from

sjcorbett:
brooklyn-server/rest/rest-api/generated/swagger-api-spec/retrofit2.json:
What is this file for? Why does it refer to "petstore"?
```
{
  "library": "retrofit2",
  "artifactId": "swagger-petstore-retrofit2"
}
```

sjcorbett:
brooklyn-server/rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/AccessSummary.java
I think it would be better to continue using `Objects.hashCode`. It does the same thing.
```java
     @Override
     public int hashCode() {
-        return Objects.hashCode(locationProvisioningAllowed);
+        int result = (locationProvisioningAllowed ? 1 : 0);
+        result = 31 * result + (links != null ? links.hashCode() : 0);
+        return result;
     }
```
Just write `Objects.hashCode(locationProvisioningAllowed, links)`.

sjcorbett:
Likewise with AccessSummary.equals could do the type checks and then `Objects.equals(this.locationProvisioningAllowed, that.locationProvisioningAllowed) && Objects.equal(this.links, that.links)`

sjcorbett:
`toString` methods would be cleaner using `ToStringHelper` rather than IntelliJ's automated mess.

ahgittin:
> brooklyn-server/karaf/features/src/main/history/dependencies.xml:
> Do you really want to remove BROOKLYN_VERSION from all of the entries?

i've noticed the build sometimes replaces this.  @CMoH @hzbarcea is that file auto-generated?  if so should it be removed from version control?

ahgittin:
using https://github.com/apache/incubator-brooklyn/pull/1127/files?w=1 you can at least see the changes in github.  (@andreaturli worth checking what your IDE does to cause so many whitespace changes!)

you can't comment on that page though so i'll also make comments here

ahgittin:
are you sure the `LocationApi` methods aren't used in the current UI ?

ahgittin:
i prefer `Objects.hashCode` over autogenerated and use of `Objects.equal(.., ..)` over three-part check but no strong feeling.  i have no preference on the `toString` impls.

ahgittin:
why remove `@JsonSerialize(include=Inclusion.NON_EMPTY)` in places?

ahgittin:
`public` methods (e.g. `displayNameComparator()`) shouldn't be removed without a deprecation cycle unless there is a good reason

ahgittin:
why move `HighAvailabilityMode` and `ManagementNodeState` to different packages?  that will break backwards compatibility (and so should be in README if we do it).  in this case i don't think it gives us anything.

andreaturli:
thanks @ahgittin for comments
- `Object.hashCode` or `Objects.equals` from Guava are nicer but I wanted to avoid to have a direct dependency on Guava just for that. I know Guava is pulled in anyway, but we may in time be Guava-free at least for domain objects.
- @JsonSerialize introduces another deps on jackson (databind) which I think it doesn't need to be there as it really depends on the ObjectMapper settings how to ser/deser the domain objects. In fact, `ObjectMapper` can deal with `NON_NULL` or `NON_EMPTY` using something like [1](https://github.com/apache/incubator-brooklyn/pull/1127/files#diff-0686331603044ae35702b44e20abdbb6R42)
- re `LocationApi` I removed the deprecated methods and tests are passing, but I will have another look at the UI.
- re `HighAvailabilityMode` and `ManagementNodeState` I thought that having a dependency on `brooklyn-api` just for those enums was a bit too much, that's why I moved to `brooklyn-utils-common`.
- `displayNameComparator` was not used by any project and looked like an util method, but I agree, I should have deprecated first.

CMoH:
@sjcorbett @ahgittin Yes, the ```dependencies''' file is generated for dependency control across versions. I think for now is not useful and I'll remove it in a separate PR.

CMoH:
See #1139 for a solution for dependencies.xml

sjcorbett:
@andreaturli Could you resolve the merge conflicts?

ahgittin:
> @JsonSerialize introduces another deps on jackson (databind) which I think it doesn't need to be there as it really depends on the ObjectMapper settings how to ser/deser the domain objects. In fact, ObjectMapper can deal with NON_NULL or NON_EMPTY using something like 1

your link "1" is huge when i click it (maybe due to the latest push); can you be more specific?  i have a slight pref for the current behaviour (where nulls/empties are often dropped giving more readable JSON) and don't mind the dep on jackson, but happy to defer to you.

> re HighAvailabilityMode and ManagementNodeState I thought that having a dependency on brooklyn-api just for those enums was a bit too much, that's why I moved to brooklyn-utils-common.

-1 to moving these to `brooklyn-utils-common`; that's an abuse of project structure for visibility.  either keep the dependency on `brooklyn-api`, which seems reasonable (as it is relatively dependency-free isn't it?) or make copies of the enums here along with tests that assert they are real copies (with `brooklyn-api` added in test scope).

andreaturli:
@ahgittin thanks for your review, I think this PR clashes with https://github.com/apache/incubator-brooklyn/pull/1140 
Happy to close this and work on #1140

neykov:
@andreaturli The plan for #1140 (at least currently) is to try to move the changes in the existing modules instead of forking them. It will also ease merging. Feel free to update this PR and merge it.

andreaturli:
I'll close this PR and open a simplified version of it where I will
- remove JsonNode from API interfaces
- refactor consistently `toString`, `hashCode` and `equals` for the domain objects in rest-api

PR #1128, opened 2016-01-06
Parameterize the start effector

ygy:
- Customize phase will be skipped (as the install phase), if 
  skipInstallation is true


iyovcheva:
LGTM

brooklyn-server/software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -140,7 +141,10 @@ public void attachLifecycleEffectors(Entity entity) {
      * the behaviour in this lifecycle class instance.
      */
     public Effector<Void> newStartEffector() {
-        return Effectors.effector(Startable.START).impl(newStartEffectorTask()).build();
+        return Effectors.effector(Startable.START)
+                .parameter(StartSoftwareParameters.SKIP_INSTALLATION)
```
neykov: A better place for this would be the `SoftwareProcessDriverLifecycleEffectorTasks` (overriding the method there), but given that the rest of the lifecycle effectors set their process related parameters here ok to stay here for consistency.

brooklyn-server/software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -203,6 +213,12 @@ public Void call(ConfigBag parameters) {
         public Void call(ConfigBag parameters) {
             Collection<? extends Location> locations = null;
 
+            Boolean isSkipInstallation = parameters.get(StartSoftwareParameters.SKIP_INSTALLATION);
+            if (isSkipInstallation != null && isSkipInstallation) {
+                entity().config().set(BrooklynConfigKeys.SKIP_ENTITY_INSTALLATION, isSkipInstallation);
+                entity().config().set(BrooklynConfigKeys.SKIP_ENTITY_CUSTOMIZATION, isSkipInstallation);
```
neykov: It's dangerous to set the parameter as config value as it would be (unexpectedly) used for future calls as well. I'd suggest using curl to change the config value if needed and then do a regular `start`.

neykov:
Overall it's a very useful thing to have when doing support. Don't like how the effector has side effects which will influence how future calls behave. Would be nice if the SKIP_CUSTOMIZE config is kept, but set out of the effector - either through curl of groovy.
nitpick - having the SKIP_CUSTOMIZE with identical behaviour to SKIP_INSTALL.

PR #1129, opened 2016-01-06
Fix getRequiredOpenPorts in SameServerDriverLifecycleEffectorTasks

sjcorbett:
Previously any entity config with a value that TypeCoercions could coerce to PortRange was included in the list of ports.

### DIFF
--- a/brooklyn-server/software/base/src/main/java/org/apache/brooklyn/entity/software/base/SameServerDriverLifecycleEffectorTasks.java
+++ b/brooklyn-server/software/base/src/main/java/org/apache/brooklyn/entity/software/base/SameServerDriverLifecycleEffectorTasks.java
@@ -73,23 +73,17 @@ public class SameServerDriverLifecycleEffectorTasks extends MachineLifecycleEffe
            firewall will open the initial port instead. Mostly a problem for SameServerEntity, localhost location.
         */
         // TODO: Remove duplication between this and SoftwareProcessImpl.getRequiredOpenPorts
-        for (ConfigKey<?> k: entity.getEntityType().getConfigKeys()) {
-            Object value;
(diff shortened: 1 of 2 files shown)


brooklyn-server/software/base/src/main/java/org/apache/brooklyn/entity/software/base/SameServerDriverLifecycleEffectorTasks.java
```diff
@@ -73,23 +73,17 @@ protected SameServerEntityImpl entity() {
            firewall will open the initial port instead. Mostly a problem for SameServerEntity, localhost location.
         */
         // TODO: Remove duplication between this and SoftwareProcessImpl.getRequiredOpenPorts
-        for (ConfigKey<?> k: entity.getEntityType().getConfigKeys()) {
-            Object value;
+        final Set<ConfigKey<?>> configKeys = entity.getEntityType().getConfigKeys();
```
nakomis: Minor point, but is there any reason for this to be final?

nakomis:
One minor point (more of a clarification), but other than that, LGTM

brooklyn-server/software/base/src/main/java/org/apache/brooklyn/entity/software/base/SameServerDriverLifecycleEffectorTasks.java
sjcorbett: Just good practice. In this case I think IntelliJ automatically added it.

sjcorbett: Flip it round - is there any reason for it _not_ to be final?

sjcorbett:
@nakomis Thanks.

PR #1130, opened 2016-01-07
Support "externalised config" in catalog definition

iyovcheva:


brooklyn-server/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ExternalConfigYamlTest.java
```diff
@@ -108,6 +110,34 @@ public void testExternalisedLocationConfigReferencedFromYaml() throws Exception
...
+        String yaml = Joiner.on("\n").join(
+                "brooklyn.catalog:",
+                "    id: osgi.test",
+                "    itemType: template",
+                "    version: 1.3",
+                "    description: CentOS 6.6 With GUI - 1.3",
+                "    displayName: CentOS 6.6",
+                "    iconUrl: classpath:///centos.png",
+                "    brooklyn.libraries:",
+                "    - $brooklyn:external(\"myprovider\", \"osgikey\")",
```
aledsage: Can you also test with the format:

    brooklyn.libraries: [ { name: symbolic-name, version: 1.0, url: http://... }, ... ]

See https://github.com/apache/incubator-brooklyn/blob/master/brooklyn-docs/guide/ops/catalog/index.md for what is supported in a library value.

brooklyn-server/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -391,18 +396,61 @@ else for (String key: otherKeys) {
...
+            itemKey = itemMetadata.get("item");
+        }
+        if (itemMetadata.containsKey("items")) {
+            itemsKey = itemMetadata.get("items");
+        }
+
+        itemMetadata = mgmt.getConfig().getConfig(CampYamlParser.YAML_PARSER_KEY).parse(itemEvaluatedMetadata);
+
+        ExecutionContext ec = mgmt.getServerExecutionContext();
+        for (Map.Entry<String, Object> item: itemMetadata.entrySet()) {
```
aledsage: Can we use `ValueResolver.deep(true)` to get it to check if the value is a list/map, so you just have to call something like (untested code below):

    ...  = Tasks.resolveDeepValue(itemMetadata, Object.class, ec)

brooklyn-server/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -391,18 +396,61 @@ else for (String key: otherKeys) {
     }
 
     @SuppressWarnings("unchecked")
-    private void collectCatalogItems(String sourceYaml, Map<?,?> itemMetadata, List<CatalogItemDtoAbstract<?, ?>> result, Map<?,?> parentMetadata) {
+    private void collectCatalogItems(String sourceYaml, Map<String,Object> itemMetadata, List<CatalogItemDtoAbstract<?, ?>> result, Map<String,?> parentMetadata) {
```
aledsage: I'm not clear why we need `Map<String, Object>` for itemMetadata. Are we directly modify that map in this method? If so, why?

brooklyn-server/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ExternalConfigYamlTest.java
iyovcheva: I see a problem using the name/version in test, because we need a real one that match the url and if they are changed in future, someone would wonder why the test fails.

There is one more test without external config to verify that the behaviour doesn't change then.

If you think it's not that bad to add matching for now name/version, no problem to add them.

aledsage:
@iyovcheva please see https://github.com/iyovcheva/incubator-brooklyn/pull/1, for some suggested improvements to the tests (and some additional fixes I needed before `mvn clean install` would work for me).

brooklyn-server/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -376,7 +378,7 @@ else for (String key: otherKeys) {
         catalogMetadata.remove("items");
         if (!itemDef.isEmpty()) {
             log.debug("Reading brooklyn.catalog peer keys as item ('top-level syntax')");
-            Map<String,?> rootItem = MutableMap.of("item", itemDef);
+            Map<String,?> rootItem = MutableMap.of("item", (Object) itemDef);
```
aledsage: Surprised you have to cast this to an Object. But will merge as-is.

aledsage:
Thanks @iyovcheva - merging now.

PR #1132, opened 2016-01-07
In skipInstall include also pre-install-command

bostko:
copy-pre-install-resources, pre-install, pre-install-command

It makes much more sense to include those in the skip install family.
These phases are barely used but still I think it simplifies the lifecycle if it is like this.


grkvlt:
@bostko If I remember right, I wanted the option to run some commands even if we were skipping the normal entity startup. However, you're probably right that it's barely used, but I'd like to be sure nothing uses this before merging.

ahgittin:
AFAIK very little uses pre-install, and none of those should be surprised if it's skipped on "skipInstall".  will go ahead and merge.  @grkvlt @bostko if you discover anything give a shout.  maybe it's worth a call out in the release notes?  (again i tend to think not necessary but no objections.)

PR #1133, opened 2016-01-08
BROOKLYN-213 Fix localhost external IP loader

mattunderscorechampion:
Remove shutdown IP address lookup service from list of resolvers.
Block the lookup methods until either all providers have been checked or the IP address has been resolved.

https://issues.apache.org/jira/browse/BROOKLYN-213


sjcorbett:
@mattunderscorechampion Thanks. Looks good.

PR #1134, opened 2016-01-08
Expose BrooklynFeatureEnablement.setDefault

sjcorbett:
I want the activation of [brooklyn-tosca](https://github.com/cloudsoft/brooklyn-tosca/) to be configurable and to be enabled by default. At the moment any property beginning `brooklyn.experimental.feature` that is not defined in `BrooklynFeatureEnablement` defaults to `false` (because `Boolean.parseBoolean(null) == false`). This change allows downstream projects to explicitly configure defaults.


aledsage:
Looks sensible; merging

PR #1135, opened 2016-01-08
fix bash web cluster template in default catalog

johnmccabe:
the template `4-resilient-bash-web-cluster-template` did not deploy successfully in either `0.8.0-incubating` or `0.9.0-SNAPSHOT` (not checked earlier releases), the following problems are encountered if you try to deploy:

- the cluster member Bash Web Server instances fail to deploy as they try to use the placeholder location defined in the `2-bash-web-server-template` catalog entry. Have commented out the location in template 2.

The observed result for the next two points were that the balancer would never have any Bash Web Server instances added to `proxy.serverpool.targets` and you would just get 404s when browsing to the balancer endpoint.

- the `app.port` sensor isn't populated as its trying to create a sensor on the `BasicApplication` while the config exists on the `VanillaSoftwareProcess`.
Added an id `bash-web-server` to the `vanilla-bash-server` in template 2 to allow us to address the config when creating the `my-app` StaticSensor.

- the balancer is unable to populate `member.sensor.hostname` as its attempting to pull `host.subnet.hostname` from the  `BasicApplication` while this sensor is on the `VanillaSoftwareProcess`. Added a `Propagator`.

- moved the `SshCommandSensor` to template 2 so that it could access the `output.txt` and added `enricher.producer` to the `YamlTimeWeightedDeltaEnricher` so it could pull from the `VanillaSoftwareProcess` (id `bash-web-server`)
- minor usability tweaks turning off sticky session and reducing the `maxPoolSize`

Would appreciate someone having a look at the changes here, as its quite a change in the structure of the template app and may be indicative of issues elsewhere?

**Not** resolved - the `my.message` config in template 4 is not overriding that set in template 2.

### DIFF
--- a/brooklyn-server/server-cli/src/main/resources/brooklyn/default.catalog.bom
+++ b/brooklyn-server/server-cli/src/main/resources/brooklyn/default.catalog.bom
@@ -1,4 +1,3 @@
-
 # this catalog bom is an illustration supplying a few useful sample items
 # and templates to get started using Brooklyn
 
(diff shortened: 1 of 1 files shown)


johnmccabe:
@ahgittin would appreciate your opinion on this one, specifically about whether you think this might be papering over an issue elsewhere?

johnmccabe:
obsoleted by #1138

PR #1137, opened 2016-01-12
BROOKLYN-214: fix cancelling of AttributeWhenReady task

aledsage:


sjcorbett:
@aledsage Nice solution to an interesting problem.

PR #1138, opened 2016-01-12
YAML: don’t wrap sub-entities in BasicApplication

aledsage:
Previously, any catalog item of type “template” would always be wrapped
in a BasicApplication. This was being done in three places. The first
is ok; the second two are bad:

1. When deploying a top-level “template” from the catalog.
2. When referring to a “template” catalog item in a member-spec (e.g.
   for a DynamicCluster)
3. When referring to a “template” catalog item in brooklyn.children.

@johnmccabe @ahgittin @neykov you'll be interested in this.

I don't like the approach (duplicating the fix in three places, to remove the wrapping app, for the different contexts that it might be used).

It would feel better for the YAML-parsing to *not* wrap the entity at all, unless there was a really good reason to (e.g. the catalog item consists of multiple top-level services defined in its yaml). The caller could then decide if it needs to wrap it (i.e. if deliberately creating a top-level app), or leave as-is.

### DIFF
--- a/brooklyn-server/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynAssemblyTemplateInstantiator.java
+++ b/brooklyn-server/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynAssemblyTemplateInstantiator.java
@@ -40,6 +40,7 @@ import org.apache.brooklyn.util.core.flags.TypeCoercions;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 
 public class BrooklynAssemblyTemplateInstantiator implements AssemblyTemplateSpecInstantiator {
@@ -89,6 +90,16 @@ public class BrooklynAssemblyTemplateInstantiator implements AssemblyTemplateSpe
         // first build the children into an empty shell app
         List<EntitySpec<?>> childSpecs = createServiceSpecs(template, platform, loader, encounteredTypeSymbolicNames);
         for (EntitySpec<?> childSpec : childSpecs) {
+            
+            if (Application.class.isAssignableFrom(childSpec.getType())) {
+                EntitySpec<? extends Application> appSpec = (EntitySpec<? extends Application>) childSpec;
+                if (EntityManagementUtils.canPromoteChildrenInWrappedApplication(appSpec)) {
+                    EntitySpec<?> appChildSpec = Iterables.getOnlyElement(appSpec.getChildren());
+                    EntityManagementUtils.mergeWrapperParentSpecToChildEntity(appSpec, appChildSpec);
+                    childSpec = appChildSpec;
+                }
+            }
+
             app.child(childSpec);
         }
 
--- a/brooklyn-server/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
+++ b/brooklyn-server/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
@@ -29,6 +29,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 
 import javax.annotation.Nullable;
 
+import org.apache.brooklyn.api.entity.Application;
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.entity.EntitySpec;
 import org.apache.brooklyn.api.location.Location;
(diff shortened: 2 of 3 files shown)


johnmccabe:
fyi @aledsage, deployed `4-resilient-bash-web-cluster-template` and observe the wrapping occur for the VanillaSoftwareProcess under the DynamicCluster (case 2 above) (deploying the same template in 0.7.0-incubating does not result in the VSP being wrapped).

johnmccabe:
as an fyi we're still observing a number of issues related to this.
- wrapper parent spec isn't being merged with the child entity (for example location)
- VanillaSoftwareApplication is still wrapped when the parent spec includes policies/enrichers (template 4 for example)

ahgittin:
well, we should definitely refactor so the added lines are shared code.  i will look at cleaning up those pieces as a PR which builds on this.

i don't think it will work nicely to defer wrapping because plans might contain multiple entities or they might attach policies etc at the root.  it might be nice to have a more elegant solution wrt the `Application` v `Entity` split and being clear what the output of the plan parse is -- we expect that the *output* of that parse is a single `Application` but the inputs in yaml are a *list* of `Entity` objects.  this is a consequence of the camp yaml spec which came after the groovy-led brooklyn approach where `Application` was more natural.  however wrapping and unwrapping is proving a pretty good way to resolve the mismatch, provided we can resolve the issues john raises and minimise code duplication.

@johnmccabe regarding your point 2 about adjuncts (policies/enrichers) added at the root, i think this necessarily blocks unwrapping because the enricher might run over children in which case merging would break it.  i'll see if template 4 allows us to put the adjuncts on the child.

brooklyn-server/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynAssemblyTemplateInstantiator.java
```diff
@@ -89,6 +90,16 @@ private Application create(AssemblyTemplate template, CampPlatform platform) {
         // first build the children into an empty shell app
         List<EntitySpec<?>> childSpecs = createServiceSpecs(template, platform, loader, encounteredTypeSymbolicNames);
         for (EntitySpec<?> childSpec : childSpecs) {
+            
+            if (Application.class.isAssignableFrom(childSpec.getType())) {
```
ahgittin: i don't think we want this check here do we?

ahgittin: ah nvm -- it's harmless as it checks if the child might be a wrapper (not what the child wraps), although it's unnecessary as it won't matter if the generic types are wrong and the `canPromote` call will do the check

johnmccabe:
@ahgittin regarding the location/config/policies/enrichers in template 4, prior to 0.8.0-incubating they were getting merged

aledsage:
@ahgittin @johnmccabe I definitely agree with you in the case where there are multiple entities within the plan - the most sensible thing is to compose all of those under one entity - a `BasicApplication` is a reasonable choice for that.

Where there is just one type under services, it's less clear what the user would expect. In the case of using `2-bash-web-server-template` as the type of the `memberSpec` for the `DynamicCluster` of `4-resilient-bash-web-cluster-template`... it looked most logical to me that we'd get the *entity* rather than the wrapping application. The wrapping app doesn't have all the sensors propagated, so is not usable as a member of the cluster directly.

The initializers, enrichers etc are coming from additional config on the memberSpec within `4-resilient-bash-web-cluster-template`. Again it felt natural to me that this was being applied to the entity that I see in the definition of `2-bash-web-server-template` (rather than an extra wrapping app).

---

My (current) take is that we should expect the output of the parse to be a single entity - if we had to wrap it (e.g. because there were multiple children), then that top level entity will be an app; otherwise it will be the entity itself.

This requires that the caller knows whether they expected an app. And has the ability to wrap it in an app if it really required that. However, that is fiddly: by the time we hit the camp code, we don't necessarily know what the caller really wanted. When we're inside the brooklyn-core code (e.g. `BasicBrooklynTypeRegistry.createSpecFromPlan`) then we don't know how to look at the template plan to know how best to wrap it.

Maybe there is a way inside the camp code to understand what the caller really wanted. But we need to be really careful, because there aren't very clear entry points into the camp code (that are not alos used by internal methods).

ahgittin:
see my improvements to this in #1143 but note i'm still working on that; it should merge this PR automatically btw.

@aledsage yes case where we're doing a camp yaml parse of yaml with a single service and we don't need an application is ugly as we wrap and unwrap, but at least it gives a consistent contract to the camp stuff.  with the type registry i think we can expose a better contract where we pass in only the content of an entry in the services list (i.e. the map including `type: SomeEntity`) and we get back just an entity which should reduce the frequency of this ugly case.

for now at least we've got less duplication and it's easy to see where things are and aren't wrapped.  however it needs more work as i note in #1143 ... i should have it soon

PR #1139, opened 2016-01-12
Remove karaf-features auto-generated dependencies.xml

CMoH:
May be re-enabled in the future if its purpose is necessary, but for now it's more of a headache.


sjcorbett:
@CMoH Thanks.

PR #1140, opened 2016-01-12
[WIP] CXF implementation of rest api

CMoH:
I've worked on a separate REST API project for Brooklyn using CXF, since changing the current brooklyn-rest-server in a manner that's compatible both with karaf and the current launcher seems impossible.

However, the main downside is that for the time being there are two parallel implementations of brooklyn REST resources, and the same for the rest-api.  Below are the causes for this:
1. While the first duplication is to be expected, duplication of the rest-api project is a bad idea. Unfortunately CXF-3.1.4 is already based on jax-rs-2.x, while brooklyn is using jax-rs-1.1 because of jersey-1.19.
2. Another problem is that `CatalogApi.createFromMultipart()` uses HTTP forms, which do not seem to have standard jax-rs annotations. This means custom jersey annotations in brooklyn-rest-api and the reverse in brooklyn-rest-api-cxf.

Does anyone have any idea on how to create a single rest-api project with annotations usable both by CXF and Jersey1?

I've also moved tests that rely on a having a fully-blown brooklyn instance running to karaf integration tests project. For the time being they fail - I'll try to address them and update the PR, but your input is already valuable.

Thanks


neykov:
@CMoH thanks for figuring this out so far.
A quick observation from scanning through the code. The `rest-api` project is used for a client implementation in addition to the server (see https://github.com/apache/incubator-brooklyn/tree/master/brooklyn-server/rest/rest-client). This means that there is actual java code compiling against the rest API. Adding a `@Context UriInfo ui` parameter changes the signatures and also doesn't make sense for a client-based API. Can we use a class field instead of a parameter as done in some of the classes? This will have two benefits:
  * Remove the bulk of the changes on the API (getting us closer to reusing the same classes)
  * Keeping the client API clean of server-side "stuff".

wdyt?

CMoH:
@neykov Thanks for taking the time to review this (rather large) PR.

I have tried to keep changes out of ```rest-api```, and now I'm trying to factor the ```rest-api-cxf``` project back into the original. Doing a diff between folders revealed the following types of changes:

1. To get the simple matter out of the way: the ```@Context UriInfo ui``` parameter. I completely agree with your suggestion and will implement it. However, I was wondering if using swagger to generate the ```rest-client``` would not be easier (allowing us to collapse the annotations directly into the resource classes). #1100 generates the swagger spec during build, so this should be quite feasible now. I might experiment and come back to you, but I'm curious if you think this endeavour is worth the time.

2. ```CatalogApi.createFromMultipart()``` uses non-standard annotations for injecting form parameters. I can experiment using the standard JAX-RS ```@FormParam``` annotation as a common denominator.

3. The new CXF resources are no longer annotated with the full URI path (e.g. ```/catalog``` instead of ```/v1/catalog```), which can help relocate the REST API independent of other WARs - including ```brooklyn-jsgui```. To complete the picture: CXF is exposed by default in karaf at ```/cxf```, and under that each service has its own context path (see ```jaxrs:server``` in ```src/main/resources/service.xml```). For the purpose of backward compatibility I moved the entire CXF servlet from the default ```/cxf``` to ```/v1``` (see ```karaf/apache-brooklyn/src/main/resources/etc/org.apache.cxf.osgi.cfg```) and left the ```rest-resources-cxf``` path at ```/```, therefore yielding ```/v1/catalog```. However, I can't think of this as a permanent solution.
The end-plan is to have JSGUI configurable with the REST base URI, therefore leave CXF with its default servlet context (or whatever the karaf administrator may choose) and have the brooklyn REST APIs at ```/brooklyn/v1```, ```/brooklyn/v2``` etc. Would that be doable in the far future?

4. The hardest problem: Jersey-1 uses JAX-RS-1.x (```jsr311-api-1.1.1.jar```), while CXF-3.1.4 relies on JAX-RS-2.0 (```javax.ws.rs-api-2.0.1.jar```). For this reason CXF is unable to find the annotations on resources, in spite of its advertised JAX-RS-1.1 backward compatibility. Actually this was the main reason for forking the ```rest-api``` project. If we can find a solution to this, we can merge back the two API projects.

Sorry for the lengthy reply. Basically points 1 and 2 are more informative, while points 3 and 4 are more important. Thanks!

neykov:
1. I think using the swagger spec to build a client is definitely worth chasing on its own right, but it would be much more work in the short term than sticking to the exiting API. Definitely worth exploring in a separate piece of work though - even better if we can choose dependencies or just use lean libraries with no (notable) transitive dependencies. A major issue with the current client is that its transitive dependencies are too intrusive for downstream projects.
2. I have never seen the multipart method used so decided to test it and see how it behaves with jersey vs generic annotations. Turns out it doesn't work at all, meaning that the call is dispatched to the non-multipart alternative. This means that if there's no clear `JAX-RS` supported method to implement it, then it can just be removed. Even better if we can fix it, but it's not something that should be preventing us from moving forward.
3. Maybe we can configure jersey servlet so it serves the resources in the same way and the paths are identical? A related thing regarding pt 1. If we are serving the resources under `/v1` in both jersey and cxf then the `UriInfo` changes are not required for the first iteration  (although would be nice to have them eventually).
4. Can you give a specific example? I didn't see differences in the annotations between projects.

andreaturli:
@neykov @CmHo notice that #1127 should complement #1100 to make the client generation based on Swagger possible. Main problem was the usage of JsonNode and similar jackson classes in rest-api interfaces. Happy to apply the same changes on top of #1140, if you think they are still relevant.

CMoH:
@andreaturli It seems the duplication of the rest-api is already proven a bad idea by the fact that you are working only on a copy of it. I hope that I can undo that duplication.

Back to the issues at hand:

2. I have tried using ```@FormParam``` for ```multipart/form-data```, but that does not work. I'll remove the method for now. I guess we'll find out if anyone thinks its absolutely necessary, and we'll take it from there.

4. JAX-RS 2.0 should be backward compatible syntactically with JAX-RS 1.1. However the classes themselves are from different jars, so CXF was unable to find resource methods when looking for annotations. I'll try replacing ```jsr311-api.jar``` with ```javax.ws.rs-api.jar``` and hope that Jersey-1.x will still work.

CMoH:
On point 4: the bad news is that Jersey-1.19 cannot use ```javax.ws.rs-api.jar``` because it throws a ```java.lang.AbstractMethodError``` when using the ```javax.ws.rs.core.UriBuilder.uri()```. Any ideas on how to go past this? See CMoH@c0f0bbed215b1a74bf39d5b469fb691372137285

neykov:
@CMoH, here are my thoughts on 4. The good news is that there is source code level compatibility between the libraries as far as annotations/resources are involved so we can compile with one or the other. Having this in mind, I suggest the following plan:
  1. `rest-api`: create a maven profile to conditionally compile with either jersey or CXF
  2. `rest-server`: move all jersey specific code to a `src/jersey/` folder (hopefully just a handful of files)
  3. `rest-server`: re-implement identical functionality in `src/cxf/` folder but depending on CXF
  4. `rest-server`: create a maven profile to conditionally compile one or the other
  5. `rest-client`: same as `rest-server`
  6. mail the dev list, warning of the transient dependency change and possible API changes
  7. flip the default profile so we use CXF now, leaving jersey still there, but not in the official build (may be after next release?)
  8. after one more release remove jersey code and profiles

brooklyn-server/cxf/rest-resources-cxf/pom.xml
```diff
@@ -0,0 +1,213 @@
...
+        <artifactId>brooklyn-parent</artifactId>
+        <version>0.9.0-SNAPSHOT</version><!-- BROOKLYN_VERSION -->
+        <relativePath>../../parent/pom.xml</relativePath>
+    </parent>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.brooklyn</groupId>
+            <artifactId>brooklyn-rest-api-cxf</artifactId>
+            <version>${project.version}</version>
```
neykov: Should be `cxf.version`, drop the version altogether as it's already set in parent's `dependencyManagement`.

neykov: Ooops still sleepy it seems, please ignore.

CMoH:
@neykov I have applied your changes here, and with some minor changes the default maven build succeeds. However, running ```mvn clean install -Djax-rs-2.0=cxf``` fails. Did you encounter this error?

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:testCompile (default-testCompile) on project brooklyn-launcher: Compilation failure: Compilation failure:
[ERROR] /home/cipi/src/incubator-brooklyn/brooklyn-server/launcher/src/test/java/org/apache/brooklyn/rest/jsgui/BrooklynJavascriptGuiLauncher.java:[27,32] cannot find symbol
[ERROR] symbol:   class BrooklynRestApiLauncher
[ERROR] location: package org.apache.brooklyn.rest
[ERROR] /home/cipi/src/incubator-brooklyn/brooklyn-server/launcher/src/test/java/org/apache/brooklyn/rest/jsgui/BrooklynJavascriptGuiLauncherTest.java:[23,32] cannot find symbol
[ERROR] symbol:   class BrooklynRestApiLauncherTestFixture
[ERROR] location: package org.apache.brooklyn.rest
[ERROR] /home/cipi/src/incubator-brooklyn/brooklyn-server/launcher/src/test/java/org/apache/brooklyn/rest/jsgui/BrooklynJavascriptGuiLauncher.java:[65,16] cannot find symbol
[ERROR] symbol:   variable BrooklynRestApiLauncher
[ERROR] location: class org.apache.brooklyn.rest.jsgui.BrooklynJavascriptGuiLauncher
[ERROR] /home/cipi/src/incubator-brooklyn/brooklyn-server/launcher/src/test/java/org/apache/brooklyn/rest/jsgui/BrooklynJavascriptGuiLauncher.java:[76,25] cannot find symbol
[ERROR] symbol:   variable BrooklynRestApiLauncher
[ERROR] location: class org.apache.brooklyn.rest.jsgui.BrooklynJavascriptGuiLauncher
[ERROR] /home/cipi/src/incubator-brooklyn/brooklyn-server/launcher/src/test/java/org/apache/brooklyn/rest/jsgui/BrooklynJavascriptGuiLauncherTest.java:[56,9] cannot find symbol
[ERROR] symbol:   variable BrooklynRestApiLauncherTestFixture
[ERROR] location: class org.apache.brooklyn.rest.jsgui.BrooklynJavascriptGuiLauncherTest
[ERROR] /home/cipi/src/incubator-brooklyn/brooklyn-server/launcher/src/test/java/org/apache/brooklyn/rest/jsgui/BrooklynJavascriptGuiLauncherTest.java:[57,9] cannot find symbol
[ERROR] symbol:   variable BrooklynRestApiLauncherTestFixture
[ERROR] location: class org.apache.brooklyn.rest.jsgui.BrooklynJavascriptGuiLauncherTest
```

CMoH:
@neykov I'm having the following issues doing an `mvn clean install`:

```
ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:compile (default-compile) on project brooklyn-rest-resources: Compilation failure: Compilation failure:
[ERROR] /home/cipi/src/incubator-brooklyn/brooklyn-server/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/DefaultExceptionMapper.java:[41,28] package org.eclipse.jetty.io does not exist
[ERROR] /home/cipi/src/incubator-brooklyn/brooklyn-server/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/DefaultExceptionMapper.java:[61,35] cannot find symbol
[ERROR] symbol:   class EofException
[ERROR] location: class org.apache.brooklyn.rest.util.DefaultExceptionMapper
[ERROR] -> [Help 1]
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:compile (default-compile) on project brooklyn-rest-server-jersey: Compilation failure
[ERROR] /home/cipi/src/incubator-brooklyn/brooklyn-server/rest/rest-server-jersey/target/generated-sources/rest-deps/org/apache/brooklyn/rest/BrooklynRestApp.java:[46,5] method does not override or implement a method from a supertype
[ERROR] -> [Help 1]
```

I'll start rebasing the branch onto master so we can move this PR onto the new repositories and continue there.

neykov:
No change is minor it seems :) , fixed in https://github.com/CMoH/incubator-brooklyn/pull/7

PR #1141, opened 2016-01-13
Fixed issue with exceptions in Rest Client

Graeme-Miller:
Calling getEntity on BuiltResponsePreservingError would propagate a previously caught and ignored exception. Because of this the output did not mention that getEntity was the line propagating the exception. We now throw an IllegalStateException to make it clear where the exception is coming from


sjcorbett:
Looks good.

PR #1142, opened 2016-01-13
deprecate group.members.delegate and prevent delegate creation when adding to quarantine

johnmccabe:
Use of delegate to denote quarantine group membership in the gui is no longer required following #929


grkvlt:
@johnmccabe LGTM happy to merge

PR #1143, opened 2016-01-13
clean up app spec unwrapping

ahgittin:
builds on #1138 cleaning it up, and work towards setting the correct plan yaml on entity specs coming from catalog template items

note this is WIP, as the test for the latter currently fails, and the former is untested, and i've not yet looked at example 4 -- but i wanted to put it out for early review given the attention on #1138


ahgittin:
still WIP, but getting closer; the right-unwrapping fix also helped get the correctly YAML source attached to items (unexpanded).  however still working on template-4:  there is some weirdness in when the locations defined in a referenced type (e.g. template-2) are overwritten vs added to.  template-3 references template-2 as a `type` in the `services` list, where 2 defines aws with `identity: <REPLACE>` but it is overwritten and everything works, whereas template-4 references template-2 using `entitySpec` and there the `<REPLACE>` identity is *not* overwritten thus the failure.

i've added logic which says if a *referencer* defines a `location: ...` or `locations` block (including `locations: []`) then it overwrites locations in the referenced type.  a bit inconsistent but what people want almost always i think.  at least as a stopgap.  but the difference between template-3 and template-4 is worrying.

@johnmccabe thanks for the comments, pls keep them coming!

ahgittin:
as posted on the mailing list, two significant changes:

FIRST:  Locations inherited rather than passed to start(...)

* Previously when a location was set on an app in YAML, it was passed by the app as a parameter to "start"
* Entities tend to find their locations by looking 1) at what is passed in to start (if supported), then
    2) what is defined as location/locations on the entity in yaml, then 3) what ancestors have defined as location/locations

This combination meant that if I have this YAML:

services:
- type: A
  location: L2
- type: B
location: L1

Then A would normally take location L1 because the app passed it and (1) above applied.  This meant L2 could be ignored.  With a change in #1143 we will now prefer L2 for A.  The more common case, "B", will still get L1 because it should look in its ancestry as it has no locations available locally.  Note that this means that calls to "Entity.getLocations()" might return a smaller set than previously.  "Entities.getAllInheritedLocations(Entity)" has been introduced for when that behaviour is desired.

This moves us away on relying on start(Location) which while it was handy in java, is awkward in a YAML-based world.  It also makes behaviour consistent with when things are declared as entity-specs (when a locally-defined location, eg L2, was preferred).


SECOND:  Can override locations when extending a type

Now let's consider a type T1 defined in the type registry (fka catalog) which includes a "location: L1" block intended as a default or an example (as in templates 2 and 4 in our default catalog).  A caller defines a block with

    type: T1
    location: L2

Previously these were *combined* and T1 would have L1 and L2.  But we didn't normally notice because of the FIRST problem.  Now however we need a way for the caller to override L1.  And combining locations is an odd thing in most cases.  So these semantics have changed, L2 replaces L1.  The same is true if the caller supplies a list "locations".  In particular, if the caller wants his block to inherit from the outer location, he can do this:

services:
- type: T1
  locations: []   # means inherit
location: L2

It would be good to find clearer ways to express this, but this shift in semantics solves all known use cases (in particular templates 2,3,4 in the examples!) without needing new concepts.  It passes the "does what I expect in most cases" test, although edge cases are more complicated than we'd like!

ahgittin:
@aledsage @johnmccabe this should now be ready for review.  looks like templates 3 and 4 are again healthy (assuming i haven't broken anything since testing them).

@neykov this is the PR i mentioned on IRC, minor correction to how parameters are unwrapped, and lots of TODO comments about what more might be expected beyond the simple case where it works well

neykov:
LGTM. Agree with the discussion in #1138 that unwrapping in the caller is clunky. My 2c - not allowing root adjuncts will make catalog items easier to write, understand and parse. Only allow them in app markup.

brooklyn-server/server-cli/src/main/resources/brooklyn/default.catalog.bom
```diff
@@ -240,6 +242,8 @@ brooklyn.catalog:
               # and a lot of sensors defined
               type:           2-bash-web-server-template
               name:           My Bash Web Server VM with Sensors
+              # and clear the location defined there so it is taken from this template
+              locations:      []
```
johnmccabe: on first sight I'm bit torn on this, its unambiguous but the 0.7.0-incubating behaviour felt very slick..

johnmccabe:
@ahgittin looks like its still not playing nicely with template 4, I'm seeing VanillaSoftwareProcess is wrapped with a BasicApplication. The enrichers/sensors that are being set in the memberSpec then get applied to the BA rather than the VSP - all deploys ok, but if you try to connect to the Nginx Proxy endpoint you'll get a 404 as this is whats getting added to the `proxy.serverpool.targets`:

```
{"BasicApplicationImpl{id=z1rYauyy}":null}
```

Also the `my.message` config doesn't get overwritten (still `my.message` rather than `part of the cluster`).

This is what we'd hacked around in #1135 - having to add cluster specific sensors etc to the template 2 app felt like a step backwards - the ability to layer on cluster-smarts to a non-cluster app was awesome

ahgittin:
when i did it nothing was wrapped in a `BasicApplication`.  i'll have another look.

also i see a couple of unit tests are failing; i thought i had a clean run but i'll fix those.

re overriding with `[]` -- yes it was slick when it worked without that, *if* ignoring the referenced type's location was your intention.  but often i think you *won't* want to ignore the referenced type's location so it's not a good general solution.

johnmccabe:
I'll reclone and double check here as well

johnmccabe:
Seeing this when I deploy template 4 (via the Create Application wizard and just changing the placeholder location to a valid one - bluebox-lon in my case)
```
BasicApplication - Resilient Load-Balanced Bash Web Cluster (Brooklyn Example)
  - DynamicCluster - Cluster of Bash Web Nodes
    - BasicApplication - My Bash Web Server VM with Sensors **
      - VanillaSoftwareProcess - My Bash Web Server VM
  - NginxController - Load Balancer (nginx)
```

ahgittin:
Aha @johnmccabe that was very helpful; the nice names lulled me in to thinking it was working.

There was one more place unwrapping was needed.  Try with latest.

I think this fixes the failing tests too -- running on my box now (and jenkins is doing the same).

johnmccabe:
its alive !! looks bang on..
- nginx `proxy.serverpool.targets` populated as expected
- config `my.message` set using value from memberSpec in the cluster
- Policies and enrichers getting layered on top of the template 2 app
- CreateUserPolicy working again
- BA wrapping only at the top level, no extraneous wrapping of the VSP
:+1: 
I'll have a look through the changes tomorrow

ahgittin:
merging, given it works for you @johnmccabe and other review comments here and on ML, and tests happy

but further comments on changes welcome @johnmccabe

PR #1144, opened 2016-01-14
Getting Started with Vagrant updates - for review

drigodwin:


brooklyn-docs/guide/start/blueprints.md
```diff
@@ -6,44 +6,83 @@ children:
...
+hover your mouse over the right side of the text box below to get a Javascript "copy" button).
 
 {% highlight yaml %}
 name: Tomcat
 location:
-  jclouds:aws-ec2:
-    identity: ABCDEFGHIJKLMNOPQRST
-    credential: s3cr3tsq1rr3ls3cr3tsq1rr3ls3cr3tsq1rr3l
 services:
 - serviceType: brooklyn.entity.webapp.tomcat.TomcatServer
```
ahgittin: this should be `type` not `serviceType` (both work but the former is now preferred)

ahgittin: also this class is no longer valid, it should be `org.apache.brooklyn.entity.webapp.tomcat.TomcatServer`

brooklyn-docs/guide/start/blueprints.md
```diff
@@ -6,44 +6,83 @@ children:
...
 
-Copy the blueprint below into a text file, "myapp.yaml", in your workspace, but *before* you create an application with 
-it, modify the YAML to specify the location where the application will be deployed.  (Note, to copy the file you can
-hover your mouse over the right side of the text box below to get a Javascript "copy" button.)
+Copy the blueprint below into a text file, "myapp.yaml", in your workspace (Note, to copy the file you can
+hover your mouse over the right side of the text box below to get a Javascript "copy" button).
 
 {% highlight yaml %}
 name: Tomcat
 location:
```
ahgittin: maybe move `location:` to the end, and leave a placeholder

```
services:
- type: org.apache.brooklyn.entity.webapp.tomcat.TomcatServer
location:
  # YOUR LOCATION SPEC HERE (see below)

File: brooklyn-docs/guide/start/blueprints.md [hunk 3]
```diff
@@ -6,44 +6,83 @@ children:
...
+
+## Locations
+
+Before you can create an application with this configuration, you need to modify the YAML to specify a location. Locations in Apache Brooklyn are server resources which Brooklyn can use to deploy applications. These locations may be servers or cloud providers which provide access to servers. 
+
+In order to configure the location in which Apache Brooklyn launches an application, replace the ```location:``` element with values for your chosen target environment. Here are some examples of the various location types:
+
+{::options parse_block_html="true" /}
+
+<ul class="nav nav-tabs">
```
ahgittin: tabs!  love it!

brooklyn-docs/guide/start/blueprints.md
```diff
@@ -56,30 +95,36 @@ location:
...
+## Deploying the Application
+
+First, log in to brooklyn with the command line interface (CLI) tool by typing:
 {% highlight bash %}
 $ br login http://localhost:8081/
 {% endhighlight %}
 
-To secure the server you can add a username and password in Brooklyn's properties file, as described in the User Guide. 
-Then the login command will require the additional parameters of the userid and password.
+To secure the Apache Brooklyn instance, you can add a username and password to Brooklyn's properties file, as described in the User Guide [here](../ops/brooklyn_properties.html){:target="_blank"}.
```
ahgittin: i've not seen this style of `{:target...}` markup.  i assume it works?  :)

brooklyn-docs/guide/start/blueprints.md
```diff
@@ -6,44 +6,83 @@ children:
...
+
+The Vagrant configuration described in [Running Apache Brooklyn](./running.html), on the previous page is the recommended way of running this tutorial. This configuration comes with four blank vagrant configurations called byon1 to byon4.
+
+These can be launched by entering the following command into the terminal in the vagrant configuration directory.
+
+{% highlight bash %}
+ $ vagrant up byon1 byon2 byon3 byon4
+{% endhighlight %}
+
+The location in "myapp.yaml" can then be replaced with the following YAML to launch using these.
```
ahgittin: `s/then/now/` ?

ahgittin: `using these vagrant servers.` ?

brooklyn-docs/guide/start/running.md
```diff
@@ -7,33 +7,64 @@ menu_parent: index.md
...
+
+[Vagrant](https://www.vagrantup.com/){:target="_blank"} is a software package which automates the process of setting up virtual environments such as [Oracle VirtualBox](https://www.virtualbox.org){:target="_blank"}. We recommend it as
+the easiest way of getting started with Apache Brooklyn.
+
+Firstly, download and install:
+
+ * [Vagrant](http://www.vagrantup.com/downloads){:target="_blank"}
+ * [Oracle VirtualBox](https://www.virtualbox.org/wiki/Downloads){:target="_blank"}
+ 
+Then download the default Apache Brooklyn vagrant configuration from [here](https://github.com/johnmccabe/vagrant-brooklyn-getting-started/archive/master.tar.gz){:target="_blank"}. This configuration contains everything you need to get started using Apache Brooklyn.
```
ahgittin: can we move this to be a brooklyn project rather than in @johnmccabe 's repo?

not essential as part of this PR but i think it is important for apache that our primary getting-started path is pure apache -- or other well-known sources -- not some random guy, no matter how swell he is!

ahgittin:
great stuff.  mostly minor comments, but we should pull in @johnmccabe 's work to brooklyn either as part of this or very soon after.  preference for doing it as part of this.

brooklyn-docs/guide/start/blueprints.md
drigodwin: Aled pointed out that if there was something invalid in the location such as a <placeholder> then the error message from Brooklyn would be more invalid location rather than invalid YAML which is more sensible

johnmccabe:
@ahgittin We'd talked about possibly pulling the vagrant env itself into the repo, but in retrospect I'd prefer this be kept separate - I like having the option to pull the env on its own rather than as part of brooklyn or docs.

wdyt?

I could move the existing repo to brooklyncentral?

rdowner:
@johnmccabe If our official docs recommend using Vagrant then those files should be part of the Brooklyn repo IMO. It looks bad if our own documentation recommends going to a third party for the files needed to get started...

johnmccabe:
@rdowner I'm talking about a separate repo similar to how we'll end up with brooklyn-docs, brooklyn-server etc, I don't have an opinion on the root organisation.
For example Mesos has `mesosphere/playa-mesos`, RIAK has `basho-labs/vagrant-riak-cs-cluster` etc

rdowner:
Ah I see what you mean. I'm OK with that (slightly resistant to further proliferating the number of repos our project has, but there is justification here). However just to be clear it should not be a brooklyncentral nor cloudsoft repo, but an Apache one.

johnmccabe:
How about something completely unimaginative like `apache/brooklyn-vagrant`

ahgittin:
could it live in the `brooklyn-dist` e.g. `brooklyn-dist/vagrant/` and make a tarball for it as part of the official build?

that feels like the right emphasis (as opposed to a top-level project)

johnmccabe:
@ahgittin makes sense, I've got some enhancements/fixes to add and will raise a PR tonight

ahgittin:
like the concertina.  would be nice if i could click anywhere in the title bar, not just on text.

not sure about the glossary.  cute at first but obnoxious after a while.  :)

i'll overwrite the license when i merge, and store the info you provided somewhere handy (new `_licensing/` folder in docs)

overall great stuff -- merging

PR #1145, opened 2016-01-14
Adds all interfaces to proxy if deploying concrete application

nakomis:
Previously the archetype SingleWebServerSample was failing to deploy as the generated proxy did not include the 'Application' interface


ahgittin:
+1

PR #1147, opened 2016-01-14
Fix catalog XML deserialization

neykov:
Fixes the serialized XML format so it can be deserialized later.

There's an explicit alias for `List.class`, so the actual object type is not dumped. Any non-ArrayList will fail deserialization as that's what the default for `Collection` is. 
Another positive side of the change is that we won't have function class references serialized.

### DIFF
--- a/brooklyn-server/core/src/test/java/org/apache/brooklyn/core/catalog/internal/CatalogDtoTest.java
+++ b/brooklyn-server/core/src/test/java/org/apache/brooklyn/core/catalog/internal/CatalogDtoTest.java
@@ -39,12 +39,14 @@ import org.apache.brooklyn.core.catalog.internal.CatalogUtils;
 import org.apache.brooklyn.core.catalog.internal.CatalogXmlSerializer;
 import org.apache.brooklyn.core.catalog.internal.CatalogClasspathDo.CatalogScanningModes;
 import org.apache.brooklyn.core.entity.Entities;
+import org.apache.brooklyn.core.mgmt.BrooklynTags;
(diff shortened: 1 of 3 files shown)


brooklyn-server/core/src/test/java/org/apache/brooklyn/core/catalog/internal/CatalogDtoTest.java
```diff
@@ -111,6 +113,9 @@ public CatalogDto buildExampleCatalog() {
         testEntitiesJavaCatalog.addEntry(CatalogItemBuilder.newTemplate(TestApplication.class.getCanonicalName(), BasicBrooklynCatalog.NO_VERSION)
                 .displayName("Test App from JAR")
                 .javaType(TestApplication.class.getCanonicalName())
+                .tag(BrooklynTags.newNotesTag("Some notes for catalog testing"))
```
aledsage: Should we test in this method that we can get these tags back out again?
And explicitly test rebind, to confirm the tags are still there?

aledsage:
Looks good. One comment about testing. But will merge this now @neykov

brooklyn-server/core/src/test/java/org/apache/brooklyn/core/catalog/internal/CatalogDtoTest.java
neykov: For the particular case this is fixing the deserialization will fail. But in general agree that it would be nice to add those checks, will do.

PR #1149, opened 2016-01-14
turn off sticky sessions to allow easy balancing validation

johnmccabe:
turning off sticky sessions in template 4 allows the user to very easily validate balancing in action by refreshing the browser where they'll see a different VanillaSoftwareProcess entity id and IP address as requests round robin across balancer members.
```
I am VanillaSoftwareProcessImpl{id=sqyei7aO}, part of the cluster.
```
```
I am running at 10.220.0.230, with on-box IP configuration:
```


ahgittin:
good idea

PR #1151, opened 2016-01-15
License generation

ahgittin:
update for new project structure, plus more recent dependencies


ahgittin:
This is trivial and other people who want to update licenses need this so I'm merging.

@m4rkmckenna @azbarcea @aledsage @neykov could one of you do a review post-commit pls?

PR #1152, opened 2016-01-16
[BROOKLYN-183] Add karaf features

kcris:
Add brooklyn-software-base, brooklyn-software-winrm and brooklyn-policy features.
Ignore for now python imports in brooklyn-software-base.


CMoH:
Tested OK. LGTM.

PR #1153, opened 2016-01-18
WinRM docs for correct preparation of windows templates

ygy:
- Short troubleshooting section that incorrect windows template can
  cause timeouts or fail some deployments
- Settings to Use for an Unattended Installation in known limitations
  section


sjcorbett:
Looks good.

PR #1155, opened 2016-01-18
Set error logging for MySqlNode

bostko:
Logs the login attempts as well.
@iyovcheva can you review


ahgittin:
i tested with `general_log` and `general_log_file` which gave lots of info.

curious whether this PR will give enough output.  should we add `general_log*` also, maybe with a comment that it should be enabled/disabled in prod?

bostko:
It gives the login attempts. I think this is enough for now.
general_log  is for the sql queries.

iyovcheva:
Added the `general_log` here: https://github.com/apache/incubator-brooklyn/pull/1156

aledsage:
@bostko can you elaborate? If we have the login attempt, then we might see a username/ip of the hacker. But we wouldn't see *how* they got access. For that, it would be useful to see the sql query(ies) they execute, to see what exploit(s) they take advantage of.

bostko:
`general_log` has a lot of output and it is really not needed in a general production environment.
I made it optional as @ahgittin suggested. I tested and really logs gives every single sql query.
@aledsage I agree in a possible abuse attempt `general_log_file` will be helpful.

ahgittin:
good resolution.  the docs say to use `0` or `OFF` -- can i confirm you checked it's case-insensitive (ie `off` and `on` in lower case work as expected) ?

bostko:
@ahgittin I tested it and it works.

PR #1157, opened 2016-01-18
Additional security for MySQL

ygy:
- it changes the default of sql_mode to include "NO_AUTO_CREATE_USER"
- it will prevent the GRANT statement from automatically creating new
user accounts unless authentication information is specified.

### DIFF
--- a/brooklyn-library/software/database/src/main/resources/org/apache/brooklyn/entity/database/mysql/mysql.conf
+++ b/brooklyn-library/software/database/src/main/resources/org/apache/brooklyn/entity/database/mysql/mysql.conf
@@ -15,5 +15,9 @@ datadir         = ${driver.dataDir}
 bind-address    = 0.0.0.0
 # skip-networking
 
+#Prevent the GRANT statement from automatically creating new user accounts if it would otherwise do so,
+#unless authentication information is specified
+sql_mode = NO_AUTO_CREATE_USER
+
 # Custom configuration options
 ${driver.mySqlServerOptionsString}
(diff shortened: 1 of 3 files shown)


ygy:
The expected behavior after applying it is:
```
mysql> GRANT ALL ON *.* TO user1@localhost;
ERROR 1133 (42000): Can't find any matching row in the user table
```

iyovcheva:
Looks good

grkvlt:
@ygy LGTM

aledsage:
Looks good. I confirmed that the example still works at https://brooklyn.apache.org/v/latest/start/blueprints.html (which uses the script at https://bit.ly/brooklyn-visitors-creation-script).

@ygy @grkvlt do we also want to include `NO_ENGINE_SUBSTITUTION`, which according to http://dev.mysql.com/doc/refman/5.6/en/sql-mode.html is the default in MySQL 5.6.6 and later? Looking at the default file in `mysql-5.6.26-osx10.9-x86_64/my.cnf`, it has `sql_mode=NO_ENGINE_SUBSTITUTION,STRICT_TRANS_TABLES`. I think we produced the templated version of the configuration file before those options were added to the defaults.

Merging now.

ygy:
I think that we don't need to explicitly include ```NO_ENGINE_SUBSTITUTION``` and ```STRICT_TRANS_TABLES``` for now.
Maybe, we can think of upgrading MySQL to 5.7
@aledsage @grkvlt wdyt?

PR #1159, opened 2016-01-19
Fix DynamicFabric locations

aledsage:
- When locations are defined at the YAML top-level, then ensure the
  DynamicFabric finds those.

### DIFF
--- a/brooklyn-server/core/src/main/java/org/apache/brooklyn/entity/group/DynamicFabricImpl.java
+++ b/brooklyn-server/core/src/main/java/org/apache/brooklyn/entity/group/DynamicFabricImpl.java
@@ -18,11 +18,13 @@
  */
 package org.apache.brooklyn.entity.group;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static org.apache.brooklyn.util.groovy.GroovyJavaMethods.elvis;
 import static org.apache.brooklyn.util.groovy.GroovyJavaMethods.truth;
 
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


ahgittin:
looks good; will merge, but also have some related follow-on tidies

brooklyn-server/core/src/main/java/org/apache/brooklyn/entity/group/DynamicFabricImpl.java
```diff
@@ -112,9 +115,10 @@ public void start(Collection<? extends Location> locsO) {
         if (locsO!=null) {
             addLocations(locsO);
         }
-        
+        Collection<Location> locs = Collections.unmodifiableCollection(Locations.getLocationsCheckingAncestors(getLocations(), this));
```
johnmccabe: move this into the `newLocations.isEmpty()` condition we only need it it `locs0` is empty

brooklyn-server/core/src/main/java/org/apache/brooklyn/entity/group/DynamicFabricImpl.java
```diff
@@ -18,11 +18,13 @@
  */
 package org.apache.brooklyn.entity.group;
 
+import static com.google.common.base.Preconditions.checkArgument;
```
johnmccabe: we're already importing `com.google.common.base.Preconditions`

ahgittin:
@johnmccabe i saw the same things, have applied them at #1161

PR #1160, opened 2016-01-19
Tasks cancellation

ahgittin:
clearer semantics for cancel, and cancel dependent submitted tasks by default
   
prevents leaks where dependent tasks (eg resolveValue) are submitted in the background,
then the caller is cancelled; the interruption was not propagated.  now by default it is,
to children tasks and to submitted transients, with options for other (weaker and stronger) cancellations.
see TaskInternal.cancel(TaskCancellationOptions), and new tests for cancelling children in DynamicSequentialTaskTest.

also remove deprecated loose-typing ExecutionUtils.invoke,
more TRACE logging for activities,
and wrap batch config in a task so more resolutions are nested


brooklyn-server/core/src/main/java/org/apache/brooklyn/util/core/task/BasicExecutionManager.java
```diff
@@ -526,19 +538,63 @@ public String toString() {
...
+                            log.trace("Cancelling "+child+" on recursive cancellation of "+task);
+                        }
+                        subtasksFound++;
+                        if (((TaskInternal<?>)child).cancel(mode)) {
+                            result = true;
+                            subtasksReallyCancelled++;
+                        }
+                    }
+                }
+                for (Task<?> t: execMgmt.getAllTasks()) {
```
aledsage: This feels like it could get very inefficient - perhaps would be good to index the tasks by "submitted by", so can retrieve them easily.

brooklyn-server/core/src/main/java/org/apache/brooklyn/util/core/task/BasicExecutionManager.java
```diff
@@ -665,9 +728,27 @@ protected void internalBeforeStart(Map<?,?> flags, Task<?> task) {
...
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    // not ideal, such loose typing on the callback -- should prefer Function<Task,Object>
+    // but at least it's package-private
+    static Object invokeCallback(Object callable, Task<?> task) {
+        if (callable instanceof Closure) return ((Closure<?>)callable).call(task);
+        if (callable instanceof Callable) {
+            try {
+                return ((Callable<?>)callable).call();
+            } catch (Throwable t) {
+                throw Throwables.propagate(t);
```
aledsage: Should use `Exceptions.propagate(t)`, so we properly handle `InterruptedException`.

brooklyn-server/core/src/main/java/org/apache/brooklyn/util/core/task/BasicTask.java
```diff
@@ -288,15 +288,33 @@ public synchronized boolean uncancel() {
...
     
     @Override
-    public synchronized boolean cancel(boolean mayInterruptIfRunning) {
+    public final synchronized boolean cancel(boolean mayInterruptIfRunning) {
+        // semantics changed in 2016-01, previously "true" was INTERRUPT_TASK_BUT_NOT_SUBMITTED_TASKS
+        return cancel(mayInterruptIfRunning ? TaskCancellationMode.INTERRUPT_TASK_AND_DEPENDENT_SUBMITTED_TASKS
+            : TaskCancellationMode.DO_NOT_INTERRUPT);
+    }
+    
+    public synchronized boolean cancel(TaskCancellationMode mode) {
```
aledsage: Let's mark this `@Beta` (note that `TaskCancellationMode` is already `@Beta`).

brooklyn-server/core/src/main/java/org/apache/brooklyn/util/core/task/DynamicSequentialTask.java
```diff
@@ -309,7 +326,7 @@ public T call() throws Exception {
                                         }
 
                                         if (!primaryFinished && failureHandlingConfig.cancelPrimaryOnSecondaryFailure) {
-                                            cancel(true, false, false);
+                                            cancel(TaskCancellationMode.INTERRUPT_TASK_BUT_NOT_SUBMITTED_TASKS, false);
```
aledsage: Wow, this code is winning a prize for one of the biggest indents, which is a bit of a code smell!

But nothing to do with this PR, so ignore here.

brooklyn-server/core/src/test/java/org/apache/brooklyn/util/core/task/DynamicSequentialTaskTest.java
```diff
@@ -207,6 +222,85 @@ public void testCancelled() throws Exception {
...
+        waitForMessages(Predicates.compose(MathPredicates.greaterThanOrEqual(2), CollectionFunctionals.sizeFunction()), TIMEOUT);
+        Asserts.assertEquals(MutableSet.copyOf(messages), MutableSet.of("1-wait", "2-wait"));
+
+        Time.sleep(Duration.millis(400));
+        
+        if (cancellationMode==null) {
+            ((TaskInternal<?>)t).cancel();
+        } else if (cancellationMode instanceof Boolean) {
+            t.cancel((Boolean)cancellationMode);
+        } else if (cancellationMode instanceof TaskCancellationMode) {
```
aledsage: Include an:

    } else {
        throw new IllegalStateException("Invalid cancellationMode: "+cancellationMode);
    }

(even though it's a test, simple to add and makes clear to readers/extenders that we don't expect anything else.)

brooklyn-server/core/src/test/java/org/apache/brooklyn/util/core/task/DynamicSequentialTaskTest.java
```diff
@@ -207,6 +222,85 @@ public void testCancelled() throws Exception {
...
+        
+        final Task<List<?>> t = Tasks.parallel(
+                submitting(t1),
+                sayTask("2-wait", Duration.minutes(10), "2-done"));
+        ec.submit(t);
+        
+        waitForMessages(Predicates.compose(MathPredicates.greaterThanOrEqual(2), CollectionFunctionals.sizeFunction()), TIMEOUT);
+        Asserts.assertEquals(MutableSet.copyOf(messages), MutableSet.of("1-wait", "2-wait"));
+
+        Time.sleep(Duration.millis(400));
```
aledsage: Why do we need the sleep? I thought that was what your assert for the messages was for (that the task was executing)?

aledsage:
@ahgittin looks good - only very minor comments.

Happy for you to merge once you've looked over those.

brooklyn-server/core/src/main/java/org/apache/brooklyn/util/core/task/BasicExecutionManager.java
ahgittin: have added a TODO -- i agree with your intuition, but would rather work on optimising tasks as a separate exercise

brooklyn-server/core/src/main/java/org/apache/brooklyn/util/core/task/BasicExecutionManager.java
ahgittin: good spot -- this code was moved but worth improving

brooklyn-server/core/src/main/java/org/apache/brooklyn/util/core/task/BasicTask.java
ahgittin: the super is marked `@Beta` -- have added `@Override @Beta`

brooklyn-server/core/src/test/java/org/apache/brooklyn/util/core/task/DynamicSequentialTaskTest.java
ahgittin: you're right, this shouldn't have been there

ahgittin:
Thanks @aledsage -- all good comments, all addressed.  Merging.

PR #1161, opened 2016-01-19
various fixes for start(Collection<Location>) implementations

ahgittin:
building on #1159 /cc @grkvlt @aledsage

### DIFF
--- a/brooklyn-library/software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/ControlledDynamicWebAppClusterImpl.java
+++ b/brooklyn-library/software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/ControlledDynamicWebAppClusterImpl.java
@@ -38,6 +38,7 @@ import org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogic;
 import org.apache.brooklyn.core.entity.trait.Startable;
 import org.apache.brooklyn.core.entity.trait.StartableMethods;
 import org.apache.brooklyn.core.feed.ConfigToAttributes;
+import org.apache.brooklyn.core.location.Locations;
 import org.apache.brooklyn.enricher.stock.Enrichers;
 import org.apache.brooklyn.entity.group.DynamicGroupImpl;
 import org.apache.brooklyn.entity.proxy.LoadBalancer;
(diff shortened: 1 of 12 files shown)


brooklyn-server/core/src/main/java/org/apache/brooklyn/entity/group/DynamicFabricImpl.java
```diff
@@ -112,16 +110,11 @@ public void setFactory(EntityFactory<?> factory) {
     
     @Override
     public void start(Collection<? extends Location> locsO) {
```
ahgittin: /cc @johnmccabe these are the tidies i've made

Graeme-Miller:
lgtm

brooklyn-server/core/src/main/java/org/apache/brooklyn/entity/group/DynamicFabricImpl.java
johnmccabe: looks good

brooklyn-server/core/src/main/java/org/apache/brooklyn/core/location/BasicLocationRegistry.java
```diff
@@ -430,7 +430,9 @@ public static boolean isResolverPrefixForSpec(LocationResolver resolver, String
     public List<Location> resolve(Iterable<?> spec) {
         List<Location> result = new ArrayList<Location>();
         for (Object id : spec) {
-            if (id instanceof String) {
+            if (id==null) {
+                // drop a null entry
+            } if (id instanceof String) {
```
neykov: Missing else?

brooklyn-library/software/webapp/src/main/java/org/apache/brooklyn/entity/webapp/ControlledDynamicWebAppClusterImpl.java
```diff
@@ -229,7 +231,8 @@ public void start(Collection<? extends Location> locations) {
                 }
             }
 
-            Entities.invokeEffectorList(this, childrenToStart, Startable.START, ImmutableMap.of("locations", locations)).get();
+            // don't propagate start locations
+            Entities.invokeEffectorList(this, childrenToStart, Startable.START, ImmutableMap.of("locations", MutableList.of())).get();
```
neykov: Just curious, would passing an empty map instead of empty locations would be any different?

PR #1163, opened 2016-01-20
Feature: EmptyWindowsProcess

aledsage:
- Includes EmptyWindowsProcess.USE_WINRM_MONITORING, for disabling checking of WinRM connectivity.
- Adds EmptySoftwareProcess.USE_SSH_MONITORING, for disabling checking of ssh connectivity.
    
With EmptyWindowsProcess, one can provision an entity that will not execute anything on the box. That makes it usable even if there is no direct WinRM connectivity to the box.


bostko:
Looks good!

aledsage:
Thanks @bostko - merging.

PR #1164, opened 2016-01-20
better shutdown

ahgittin:
call Main.terminate() in the main thread, rather than relying on shutdown hooks

fixes rest-initiated shutdown when using BrooklynJavascriptGuiLauncher
(looks like that has been broken since #771)

/cc @neykov does this look like a reasonable enhancement of your work in #771 ?

### DIFF
--- a/brooklyn-server/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/AbstractManagementContext.java
+++ b/brooklyn-server/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/AbstractManagementContext.java
@@ -372,10 +372,15 @@ public abstract class AbstractManagementContext implements ManagementContextInte
         return configMap;
     }
 
+    private Object locationRegistrySemaphore = new Object();
+    
     @Override
-    public synchronized LocationRegistry getLocationRegistry() {
-        if (locationRegistry==null) locationRegistry = new BasicLocationRegistry(this);
-        return locationRegistry;
+    public LocationRegistry getLocationRegistry() {
+        // NB: can deadlock if synched on whole LMC
+        synchronized (locationRegistrySemaphore) {
+            if (locationRegistry==null) locationRegistry = new BasicLocationRegistry(this);
+            return locationRegistry;
+        }
     }
 
     @Override
--- a/brooklyn-server/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/AbstractManagementContext.java
+++ b/brooklyn-server/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/AbstractManagementContext.java
@@ -372,7 +372,7 @@ public abstract class AbstractManagementContext implements ManagementContextInte
         return configMap;
     }
 
-    private Object locationRegistrySemaphore = new Object();
+    private final Object locationRegistrySemaphore = new Object();
     
     @Override
     public LocationRegistry getLocationRegistry() {
--- a/brooklyn-server/rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/ServerResource.java
+++ b/brooklyn-server/rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/ServerResource.java
@@ -218,6 +218,7 @@ public class ServerResource extends AbstractBrooklynRestResource implements Serv
                         if (shutdownHandler != null) {
                             shutdownHandler.onShutdownRequest();
                         } else {
+                            // should always be set as it is required by jersey injection?
                             log.warn("ShutdownHandler not set, exiting process");
                             System.exit(0);
                         }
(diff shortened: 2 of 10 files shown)


brooklyn-server/server-cli/src/main/java/org/apache/brooklyn/cli/Main.java
```diff
@@ -474,17 +472,22 @@ public Void apply(CatalogInitialization catInit) {
...
-                waitAfterLaunch(ctx, shutdownHandler);
+                waitAfterLaunch(mgmt, shutdownHandler);
             }
 
-            // will call mgmt.terminate() in BrooklynShutdownHookJob
+            // BrooklynShutdownHookJob will invoke terminate() to do mgmt.terminate() and BrooklynWebServer.stop();
+            // and System.exit is invoked immediately after ...
+            // but seems better to do it explicitly here. 
+            // ('launcher' is local to us so the caller *cannot* do it, and no harm in doing it twice.) 
+            launcher.terminate();
```
neykov: Won't this get called before `BrooklynShutdownHookJob`, causing an non-graceful exit?

brooklyn-server/rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/ServerResource.java
```diff
@@ -218,6 +218,7 @@ public void run() {
                         if (shutdownHandler != null) {
                             shutdownHandler.onShutdownRequest();
                         } else {
+                            // should always be set as it is required by jersey injection?
```
neykov: I think some tests could inject null value if the hook is of no interest to them. Certainly shouldn't get to this piece of code in this case.
Probably safe to remove now.

brooklyn-server/rest/rest-server/src/test/java/org/apache/brooklyn/rest/util/ServerStoppingShutdownHandler.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public void onShutdownRequest() {
+        log.info("Shutting down (when running in rest-api dev mode)...");
+
+        // essentially same as BrooklynLauncher.terminate() but cut down as this is only used in dev mode
+        
+        if (server!=null) {
+            try {
+                server.stop();
```
neykov: Stop the server **during** the request? It would result in a client error.

ahgittin: good shout, i'll background this

brooklyn-server/server-cli/src/main/java/org/apache/brooklyn/cli/Main.java
ahgittin: thanks!  yes there is stuff in `BrooklynShutdownHookJob` we need to tab into.  i still think we should trigger it manually here, because we can rather than rely on the shutdown hook, but what i've done is definitely NOT good enough.

ahgittin:
@neykov have addressed most of these, just need to fix in-process `terminate` to instead refer to `BrooklynShutdownHookJob`

ahgittin:
@neykov ready to go; see last commit, I've just backed out the `terminate` in main thread

ahgittin:
jenkins failure due to jmx test setup --

```
java.rmi.server.ExportException: Port already in use: 40125; nested exception is: 
```

/cc @aledsage

rebasing and pushing to trigger new build

brooklyn-server/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/AbstractManagementContext.java
```diff
@@ -372,10 +372,15 @@ public BrooklynProperties getBrooklynProperties() {
         return configMap;
     }
 
+    private Object locationRegistrySemaphore = new Object();
```
aledsage: declare as `final`

brooklyn-server/rest/rest-server/src/test/java/org/apache/brooklyn/rest/util/ServerStoppingShutdownHandler.java
```diff
@@ -0,0 +1,75 @@
...
+                }
+
+                if (mgmt instanceof ManagementContextInternal) {
+                    ((ManagementContextInternal)mgmt).terminate();
+                }
+            }
+        }).start();
+    }
+
+    /** Expect this to be injeted; typically it is not known when this is created, but we need it to trigger shutdown. */
```
aledsage: typo: "injeted"

aledsage:
LGTM

ahgittin:
thx @aledsage -- addressed and merged

PR #1165, opened 2016-01-21
mark ConnectionFailureDetectorTest.java as Integration because it requires ports accessible

ahgittin:
the failure log is recorded at:
https://gist.github.com/davidx/1a21649a5e637b7b2a6a


brooklyn-server/policy/src/test/java/org/apache/brooklyn/policy/ha/ConnectionFailureDetectorTest.java
```diff
@@ -37,19 +37,24 @@
...
 import org.testng.annotations.Test;
-import org.apache.brooklyn.policy.ha.HASensors.FailureDescriptor;
 
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.net.HostAndPort;
 
+@Test(groups="Integration")
+// can fail if a firewall is in place on high-numbered ephemeral ports;
```
aledsage: If this test fails because we can't open a serverSocket on a high-numbered port, then I'm guessing quite a few more tests would also fail.

Happy to merge this. But longer term, might be worth artificially adding such a firewall and seeing the full set of tests that fail as a result.

PR #1166, opened 2016-01-21
Port parameters docs and fixes

ahgittin:
Expand the "custom-entities" yaml guide (and its `netcat` example) to cover port inferencing.  Also add some other useful things:  parameters, catalog, and cluster.

Several fixes/features to enable this:
* a spec parameter can imply the creation of sensors, so port inferencing works from parameters
* a `resizeMode` can easily be set for a dynamic cluster (`off`, `sequential`, or `parallel`)
* a bug is fixed where driver initialization would hang if shell env vars had unresolved dependent config


ahgittin:
/cc @murdoaird @shartzel @Graeme-Miller please feel free to extend / adapt the extended netcat example.  it's quite a good intro now i think to what's at the heart of brooklyn.

see https://github.com/ahgittin/incubator-brooklyn/commit/411d3f8a3536de6e321dea14d72cdffeb9938310

ahgittin:
@googlielmo FYI this provides some documentation for #1148

googlielmo:
@ahgittin seen that, thanks.

ahgittin:
triggering retest

ahgittin:
merging

@shartzel i'd like you to have a look at this -- feels like a few more things like this (examples and ease-of-use tweaks) in the right places would make brooklyn much easier to get started with

PR #1167, opened 2016-01-21
improve the start task tree shape

ahgittin:
previously start showed a long list of subtasks; now they are grouped a bit more nicely i think.  see commit message for details.

this is quite a simple change.


johnmccabe:
LGTM - much tidier !!

ahgittin:
failure unrelated, should be fixed in #1171.  merging.

PR #1168, opened 2016-01-21
Uses externally accessible address for main uri of controller

nakomis:
Previously, if deploying to BYON AWS instances, the internal IP address was being displayed in the main.uri for the ControlledDynamicWebAppCluster

### DIFF
--- a/brooklyn-library/software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractControllerImpl.java
+++ b/brooklyn-library/software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractControllerImpl.java
@@ -312,7 +312,7 @@ public abstract class AbstractControllerImpl extends SoftwareProcessImpl impleme
         ConfigToAttributes.apply(this);
 
         sensors().set(PROTOCOL, inferProtocol());
-        sensors().set(MAIN_URI, URI.create(inferUrl()));
+        sensors().set(MAIN_URI, URI.create(inferUrl(true)));
         sensors().set(ROOT_URL, inferUrl());
  
         checkNotNull(getPortNumberSensor(), "no sensor configured to infer port number");


brooklyn-library/software/webapp/src/main/java/org/apache/brooklyn/entity/proxy/AbstractControllerImpl.java
```diff
@@ -283,7 +283,7 @@ protected String inferUrl(boolean requireManagementAccessible) {
     }
 
     protected String inferUrl() {
-        return inferUrl(false);
+        return inferUrl();
```
sjcorbett: Infinite loop?

nakomis:
@sjcorbett Good catch, had pushed some code I hadn't intended to

nakomis:
Test failure appears unrelated, and JmxFeedTest passes locally

ahgittin:
@nakomis agree failure unrelated, have snuck an attempt to fix this in at #1171 btw

this behaviour change seems reasonable as we normally want a load-balancer to be accessible.  however this might not always be the case and it does set up further inconsistencies in when `main_uri` is the subnet address vs main address.  @grkvlt in particular in clocker is this going to cause an issue, i.e. might it combine the public IP with the private port?  /cc @aledsage 

we should set up some guidelines about when public v private ip's are used, and how to configure

as a quick win if we need to the `inferUrl(true/false)` call in this PR could be configurable via a config key, e.g. `PREFER_ADVERTISING_PUBLIC_IP`.  but that feels ugly.

if it doesn't cause an issue in clocker i'm happy to merge and live with inconsistency for now, but let's be thinking about how to cleanly handle the public/private ambiguity!

ahgittin:
(plus another test failure fix at #1174)

ahgittin:
ugly failure this time.  not one i've come across.  probably nondet.  @nakomix you're not having luck!

```
org.apache.brooklyn.AssemblyTest.checkBrooklynCoreFeature

Failing for the past 1 build (Since Unstable#2472 )
Took 5.5 sec.
Error Message

gave up waiting for service org.apache.karaf.features.BootFinished
Stacktrace

org.ops4j.pax.swissbox.tracker.ServiceLookupException: gave up waiting for service org.apache.karaf.features.BootFinished
	at org.ops4j.pax.swissbox.tracker.ServiceLookup.getService(ServiceLookup.java:199)
	at org.ops4j.pax.swissbox.tracker.ServiceLookup.getService(ServiceLookup.java:136)
	at org.ops4j.pax.exam.inject.internal.ServiceInjector.injectField(ServiceInjector.java:89)
	at org.ops4j.pax.exam.inject.internal.ServiceInjector.injectDeclaredFields(ServiceInjector.java:69)
	at org.ops4j.pax.exam.inject.internal.ServiceInjector.injectFields(ServiceInjector.java:61)
	at org.ops4j.pax.exam.invoker.junit.internal.ContainerTestRunner.createTest(ContainerTestRunner.java:61)
	at org.junit.runners.BlockJUnit4ClassRunner$1.runReflectiveCall(BlockJUnit4ClassRunner.java:266)
```

ahgittin:
@nakomis could you spearhead the others incl @sjcorbett @aledsage @grkvlt to determine what we want to do wrt public/private?

i don't think we can merge this until we've got some guidelines.

nakomis:
/bump @sjcorbett @grkvlt @aledsage Any input on @ahgittin's comments?

PR #1169, opened 2016-01-21
Create Password Sensor

duncangrant:
A simple initialiser for creating passwords in yaml as a sensor

Added a password generating function in the Identifiers class which uses
SecureRandom

### DIFF
--- a/brooklyn-server/core/src/main/java/org/apache/brooklyn/core/sensor/password/CreatePasswordSensor.java
+++ b/brooklyn-server/core/src/main/java/org/apache/brooklyn/core/sensor/password/CreatePasswordSensor.java
@@ -34,7 +34,7 @@ public class CreatePasswordSensor extends AddSensor<String> {
     public static final ConfigKey<String> ACCEPTABLE_CHARS = ConfigKeys.newStringConfigKey("password.chars", "The characters allowed in password");
 
     private Integer passwordLength;
-    private String accecptableChars;
+    private String acceptableChars;
 
     public CreatePasswordSensor(Map<String, String> params) {
         this(ConfigBag.newInstance(params));
@@ -43,16 +43,16 @@ public class CreatePasswordSensor extends AddSensor<String> {
     public CreatePasswordSensor(ConfigBag params) {
         super(params);
         passwordLength = params.get(PASSWORD_LENGTH);
-        accecptableChars = params.get(ACCEPTABLE_CHARS);
+        acceptableChars = params.get(ACCEPTABLE_CHARS);
     }
 
     @Override
     public void apply(EntityLocal entity) {
         super.apply(entity);
 
-        String password = accecptableChars == null
+        String password = acceptableChars == null
                 ? Identifiers.makeRandomPassword(passwordLength)
-                : Identifiers.makeRandomPassword(passwordLength, accecptableChars);
+                : Identifiers.makeRandomPassword(passwordLength, acceptableChars);
         
         entity.sensors().set(sensor, password);
     }
--- a/brooklyn-server/core/src/main/java/org/apache/brooklyn/core/sensor/password/CreatePasswordSensor.java
+++ b/brooklyn-server/core/src/main/java/org/apache/brooklyn/core/sensor/password/CreatePasswordSensor.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.core.sensor.password;
+
+import java.util.Map;
+
+import org.apache.brooklyn.api.entity.EntityLocal;
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.effector.AddSensor;
+import org.apache.brooklyn.util.core.config.ConfigBag;
+import org.apache.brooklyn.util.text.Identifiers;
+
+public class CreatePasswordSensor extends AddSensor<String> {
+
+    public static final ConfigKey<Integer> PASSWORD_LENGTH = ConfigKeys.newIntegerConfigKey("password.length", "The length of the password to be created", 12);
+
+    public static final ConfigKey<String> ACCEPTABLE_CHARS = ConfigKeys.newStringConfigKey("password.chars", "The characters allowed in password");
+
+    private Integer passwordLength;
+    private String accecptableChars;
+
+    public CreatePasswordSensor(Map<String, String> params) {
+        this(ConfigBag.newInstance(params));
+    }
+
+    public CreatePasswordSensor(ConfigBag params) {
+        super(params);
+        passwordLength = params.get(PASSWORD_LENGTH);
+        accecptableChars = params.get(ACCEPTABLE_CHARS);
+    }
+
+    @Override
+    public void apply(EntityLocal entity) {
+        super.apply(entity);
+
+        String password = accecptableChars == null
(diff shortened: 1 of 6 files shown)


brooklyn-server/utils/common/src/test/java/org/apache/brooklyn/util/text/IdentifiersTest.java
```diff
@@ -98,5 +96,17 @@ public void testJavaClassRegex() {
         Assert.assertTrue("foo".matches(Identifiers.JAVA_GOOD_BINARY_REGEX));
         Assert.assertTrue("foo.bar.Baz$1".matches(Identifiers.JAVA_GOOD_BINARY_REGEX));
     }
+
+    @Test
+    public void testPassword() {
+        String upper = "ABC";
+        String numbers = "123";
+        String symbols = "!£$";
+        String id1 = Identifiers.makeRandomPassword(4, upper, numbers, symbols, Identifiers.PASSWORD_VALID_CHARS);
```
ahgittin: anyone inspecting the code will be in a position to quite easily reconstruct passwords.
the `makeRandomPassword` method could simply randomly reorder the result, and the test here check for containment rather than position.

brooklyn-server/utils/common/src/main/java/org/apache/brooklyn/util/text/Identifiers.java
```diff
@@ -91,15 +105,53 @@ public static String makeRandomId(int l) {
...
+     *
+     * @param length length of password
+     * @param passwordValidCharsPool pool of acceptable character sets
+     * @return a randomly generated password
+     */
+    public static String makeRandomPassword(final int length, String... passwordValidCharsPool) {
+        Preconditions.checkState(length >= passwordValidCharsPool.length);
+        int l = 0;
+
+        char[] id = new char[length];
```
richardcloudsoft: Minor: call it `password` instead of `id`

brooklyn-server/core/src/main/java/org/apache/brooklyn/core/sensor/password/CreatePasswordSensor.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.brooklyn.util.text.Identifiers;
+
+public class CreatePasswordSensor extends AddSensor<String> {
+
+    public static final ConfigKey<Integer> PASSWORD_LENGTH = ConfigKeys.newIntegerConfigKey("password.length", "The length of the password to be created", 12);
+
+    public static final ConfigKey<String> ACCEPTABLE_CHARS = ConfigKeys.newStringConfigKey("password.chars", "The characters allowed in password");
+
+    private Integer passwordLength;
+    private String accecptableChars;
```
ahgittin: typo "acceCptable"

brooklyn-server/utils/common/src/main/java/org/apache/brooklyn/util/text/Identifiers.java
```diff
@@ -18,16 +18,24 @@
...
     private static Random random = new Random();
 
-    /** @see #JAVA_GOOD_PACKAGE_OR_CLASS_REGEX */ 
-    public static final String JAVA_GOOD_START_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz_";
-    /** @see #JAVA_GOOD_PACKAGE_OR_CLASS_REGEX */ 
-    public static final String JAVA_GOOD_NONSTART_CHARS = JAVA_GOOD_START_CHARS+"1234567890";
+    public static final String UPPER_CASE_ALPHA = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+    public static final String LOWER_CASE_ALPHA = "abcdefghijklmnopqrstuvwxyz";
+    public static final String NUMERIC = "1234567890";
+    public static final String NON_ALPHA_NUMERIC = "!$%@#";
```
richardcloudsoft: Can we add some more here?
`!@$%^&*()-_=+[]{};:\|/?,.<>~`

ahgittin:
looks good.  minor comments.

one idea:  a yaml test would be a nice addition, esp to illustrate how it should be used -- based e.g. on `ObjectsYamlTest` reading a simple yaml which creates a password, then the code reads the sensor and asserts the password is non-empty and contains at least one letter and number ?

richardcloudsoft:
Agree with Alex's command that the password should be jumbled up a bit to stop the first few characters being predictable. Apart from that this is a very good step forward. It would be nice to have an example of how to use this - could you add to the docs, or modify an example to use it, or at the very least embed an example into the javadoc of `CreatePasswordSensor`?

ahgittin:
BTW the test failure is unrelated, seems #1160 introduced a non-deterministic test failure in `DSTTest.testCancel`, which i'll fix /cc @aledsage

brooklyn-server/utils/common/src/main/java/org/apache/brooklyn/util/text/Identifiers.java
duncangrant: I'd prefer to add more but went with the default set used by [lastpass](https://lastpass.com/generatepassword.php) in case others were problematic - possibly this isn't a problem anymore.

duncangrant:
@ahgittin I'm a bit confused about brooklyn's new structure.  If I want to extend abstractyamltest so that I can write a yaml test for this then either I need to add a dependency from core to brooklyn-camp or move the test out of core (or move the initializer out of core).  Is there a sensible place for a test like this to sit?

brooklyn-server/utils/common/src/main/java/org/apache/brooklyn/util/text/Identifiers.java
```diff
@@ -91,15 +105,53 @@ public static String makeRandomId(int l) {
...
+        Preconditions.checkState(length >= passwordValidCharsPool.length);
+        int l = 0;
+
+        char[] id = new char[length];
+
+        for(int i = 0; i < passwordValidCharsPool.length; i++){
+            id[l++] = pickRandomCharFrom(passwordValidCharsPool[i]);
+        }
+
+        String remainingValidChars = passwordValidCharsPool[passwordValidCharsPool.length - 1];
```
aledsage: I'd concatenate all of the `passwordValidCharsPool` (optionally removing duplicate chars!), and choose the rest of the letters from that. But no strong feelings.

brooklyn-server/core/src/test/java/org/apache/brooklyn/core/sensor/password/CreatePasswordSensorTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+import com.google.common.collect.ImmutableList;
+
+public class CreatePasswordSensorTest {
+
+    final static AttributeSensor<String> SENSOR_STRING = Sensors.newStringSensor("aString");
+    private EntityInternal entity;
+    private TestApplication app;
+
+    @BeforeMethod
```
aledsage: personal preference: always include `@BeforeMethod(alwaysRun=true)`. Otherwise if someone marks one of your tests as `groups="Integeration"` then the before-method won't be run anymore in mvn (but it will be in Eclipse!)

brooklyn-server/core/src/test/java/org/apache/brooklyn/core/sensor/password/CreatePasswordSensorTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.brooklyn.core.test.entity.TestEntity;
+import org.apache.brooklyn.test.Asserts;
+import org.apache.brooklyn.util.core.config.ConfigBag;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import com.google.common.collect.ImmutableList;
+
+public class CreatePasswordSensorTest {
```
aledsage: Minor: I'd extend `BrooklynAppUnitTestSupport` - then your setup can be a little simpler and you don't need the teardown (and delete the `app` field from here).

aledsage:
@duncangrant Looks good. For your question about yaml tests, you need to put the test in `brooklyn-camp` (e.g. near `brooklyn-server/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/`).

Unfortunately that test package is getting too big - we'll need to split out that package at some point. You could create a new `org.apache.brooklyn.camp.brooklyn.policy` package (no strong feelings at this point).

neykov:
@duncangrant I'd create an `EntitySpec` instead of going to yaml. Yaml is converted to a spec anyway and that's not what you want to test here.

duncangrant:
@neykov The nice thing about yaml is it becomes an example of how to use something that you know works because it has been tested so I'll probably keep it like that unless anyone disagrees.

ahgittin:
+1 @duncangrant -- i prefer to see YAML as it doubles as an example, and ultimately a java-free framework for running YAML tests

duncangrant:
I think I've made all the changes that I've been asked for.  If everything is ok then please let me know if I should squash my commits before merging.

PR #1170, opened 2016-01-21
add getting started vagrant env to release artifacts

johnmccabe:
- added following inclusion of Vagrant as a target in #1144
- previously at https://github.com/johnmccabe/vagrant-brooklyn-getting-started
- update release process docs to match current repo structure
- update make-release-artifacts to match current repo structure
- builds/signs/chksums `apache-brooklyn-<VER>-vagrant.tar.gz|zip`


johnmccabe:
@ahgittin this includes updates to the release scripts (surprised at how little was required.. )
@rdowner fyi

johnmccabe:
Expect that we should be able to use the following URL to get at the archive (for inclusion in #1144?) (pending inclusion on the mirror)

https://www.apache.org/dyn/closer.lua?action=download&filename=brooklyn/apache-brooklyn-${BROOKLYN_VERSION}/apache-brooklyn-${BROOKLYN_VERSION}-vagrant.tar.gz

edit - changed closer mirror script to the lua version to match that used in the vagrant env.

johnmccabe:
RAT check failed... I'll sort that out

brooklyn-dist/vagrant/src/main/vagrant/files/brooklyn.properties
```diff
@@ -0,0 +1,325 @@
...
+# specific language governing permissions and limitations
+# under the License.
+#
+# This is Brooklyn's dot-properties file.
+# It should be located at "~/.brooklyn/brooklyn.properties" for automatic loading,
+# or can be specified as a CLI option with --localProperties /path/to/these.properties.
+
+##################################  Welcome!  ############################################
+
+# It's great to have you here.
```
ahgittin: if this is an exact/near copy of another file can we have the build copy it in rather than maintain two copies?

brooklyn-dist/vagrant/src/main/vagrant/servers.yaml
```diff
@@ -0,0 +1,47 @@
...
+    box: ubuntu/vivid64
+    ram: 2048
+    cpus: 4
+    ip: 10.10.10.100
+    shell:
+      env:
+        BROOKLYN_VERSION: 0.9.0-SNAPSHOT
+      cmd:
+        - sudo sh -c 'export DEBIAN_FRONTEND=noninteractive; apt-get install --yes openjdk-8-jre-headless'
+        - curl -s -S -J -O -L "https://www.apache.org/dyn/closer.cgi?action=download&filename=brooklyn/apache-brooklyn-${BROOKLYN_VERSION}/apache-brooklyn-${BROOKLYN_VERSION}-bin.tar.gz"
```
ahgittin: i think links are different for snapshot and release -- not sure how to resolve that

brooklyn-dist/vagrant/src/main/vagrant/files/brooklyn.properties
ahgittin: or we make this one a small file just setting a user/pass and maybe defining the BYON location
(or even better, the BYON location is in a catalog.bom which gets read in?)
and referring people to the docs page if they want to configure clouds etc?

ahgittin:
Great job fixing the release scripts @johnmccabe and really nice to have vagrant added.

Question esp for @rdowner @aledsage @grkvlt @shartzel :  is a new release for `*-vagrant` alongside `*-bin` and `*-src` the standard @apache way to create packaging for different environments?  In particular would we do the same thing for docker and RPM and DEB packages?  Or would those be unofficial artifacts published via some other mechanism, e.g. as part of the `-bin` with the `-bin` itself being half-a-dozen items instead of one artifact?

brooklyn-dist/vagrant/src/main/vagrant/servers.yaml
johnmccabe: I'd spoken to @rdowner about this and since there isn't a consistent source for snapshots the suggestion was that we instead prompt the user to specify their own location in the server.yaml.

ie. we don't let BROOKLYN_VERSION take a snapshot value, and the user will have to update the  curl/unzip/ln steps with the snapshot they explicitly want to check out

feels a bit like sidestepping the issue tho... at least for now

rdowner:
Hi @ahgittin in response to your naming query. I don't think we're bound by any Apache standard here. `NAME-src.tar.gz` and `NAME-bin.tar.gz` are common for Java-based projects (e.g. see [Maven](https://maven.apache.org/download.cgi), [Commons](https://commons.apache.org/proper/commons-collections/download_collections.cgi), etc.) but not exclusive - e.g. Tomcat omits the `-bin` part. I think we can continue to add more sensibly-named artifacts to the release, including Docker/.deb/.rpm, without any issue.

As far as Apache is concerned, the main thing a project releases is a source code bundle. We are allowed to add optional extras for the convenience of our users, such as binary bundles and distribution packages.

johnmccabe:
Last Jenkins failure is unrelated.

brooklyn-dist/vagrant/src/main/vagrant/servers.yaml
ahgittin: mmm -- don't much like that.  what about something like this:

    if [ ! -z "${BROOKLYN_VERSION##*-SNAPSHOT}" ] ; then 
        # url for official release versions
        BROOKLYN_URL=https://https://repository.apache.org/content/repositories/releases/org/apache/brooklyn/brooklyn-dist/${BROOKLYN_VERSION}/brooklyn-dist-${BROOKLYN_VERSION}-dist.tar.gz
    else
        # url for community-managed snapshots
        BROOKLYN_URL='https://repository.apache.org/service/local/artifact/maven/redirect?r=snapshots&g=org.apache.brooklyn&a=brooklyn-dist&v='${BROOKLYN_VERSION}'&c=dist&e=tar.gz'
    fi
    # alternatively if you want to use your local build, uncomment this line:
    # BROOKLYN_URL=~/.m2/repository/org/apache/brooklyn/brooklyn-dist/${BROOKLYN_VERSION}/brooklyn-dist-${BROOKLYN_VERSION}-dist.tar.gz

reminder you'll need to update the `curl` so it does a `-o brooklyn-dist-${BROOKLYN_VERSION}-dist.tar.gz` instead of `-O` for the latter command, as well as reading from "${BROOKLYN_URL}" of course.  (looks like it already follows links and ignores bad certs which is good.)

ahgittin: it will need testing but we should be able to do the same to script the vagrant download, just replace the right `dist` mention with `vagrant` (ie not the `brooklyn-dist`!).

BTW i recall problems in the past with `closer.cgi` which prompted the use of `closer.lua` instead -- not sure how good that advice is but we've had no issues with the latter since the switch.  it seems to take the same arguments, which i didn't know about (i only knew about the redirect to the mirror, which our download page points you to ... probably worth updating the download page not to redirect you there unless you want to, and we can remove the "Sorry" message!); i think you're right it's better to use that, so for official releases instead do:

    https://www.apache.org/dyn/closer.lua?action=download&filename=brooklyn/apache-brooklyn-${BROOKLYN_VERSION}/apache-brooklyn-${BROOKLYN_VERSION}-bin.tar.gz

and `s/bin/vagrant` of course elsewhere.  (i think the `-bin` is the same as the `-dist` and vagrant would be the `-vagrant` in both places ... but not entirely sure.)

johnmccabe:
@ahgittin its a bit messy to pull the -dist artifact in from a directory outside of the Vagrantfile directory, hence the copy step. I changed INSTALL_FROM_LOCAL_DIST to take a path to the file to deploy (ie to your .m2 example earlier) and added the steps to copy it to the ruby Vagrantfile - what's messy is detecting *when* to run this, it should only occur on an `up`, or `provision` is specified, I don't see a way to get at that context from the Vagrantfile however - so you end up copying for all vagrant commands (up, destroy etc).

johnmccabe:
Note that I'm trying to avoid having the user install any plugins - [vagrant-triggers](https://github.com/emyl/vagrant-triggers) for example.

ahgittin:
great stuff.  may need more review /cc @drigodwin @aledsage but i'm going to merge so it doesn't have to be ported to new repos

johnmccabe:
@drigodwin - note that the download URL _may_ change as a result of the new structure, @rdowner will we still have all the artifacts under a `/brooklyn/apache-brooklyn-<VERSION>/` dir?

rdowner:
@johnmccabe yes our release artifacts are a different repository which is NOT moving.

PR #1171, opened 2016-01-22
fix race in DSTTest.testCancel

ahgittin:
seeing errors on jenkins server though it runs fine locally;
have patched what looks like one race, but not sure it explains the failure
at https://builds.apache.org/job/incubator-brooklyn-pull-requests/org.apache.brooklyn$brooklyn-core/2440/testReport/junit/org.apache.brooklyn.util.core.task/DynamicSequentialTaskTest/testCancelled/
where we are seeing an extra semaphore at the end of the test;
have added comments and logging in case it is seen again.
(locally all is well.)


ahgittin:
yes! tests passed. closing and re-opening to check it's not a fluke, but ready for merging, asap to prevent other PR's from having intermittent failures /cc @aledsage

ahgittin:
merged as fairly trivial changes and failures in PR's is so disruptive!

PR #1173, opened 2016-01-26
Fix: SshFeed duplicate cmds

aledsage:
Fix the avoidance of executing the same command repeatedly, if an `SshFeed` is configured with multiple sensors that use the output of the same command.


ahgittin:
nice, merging

PR #1174, opened 2016-01-26
fix intermittent failing test on constraint validation

ahgittin:


ahgittin:
doing CTR as this is trivial and it has failed the last several pull requests

this has always been a time-sensitive test, dropping some of the waits internally has made it more likely

/cc @aledsage or @sjcorbett mind giving a quick glance?

brooklyn-server/core/src/test/java/org/apache/brooklyn/core/config/ConfigKeyConstraintTest.java
```diff
@@ -297,12 +299,17 @@ public void testContextAwarePredicateInformedOfEntity() {
...
+            // may or may not fail above, depending on speed, but should fail if assert after forcing resolution
+            Object value = child.getConfig(EntityRequiringConfigKeyInRange.RANGE);
+            // NB the call above does not currently/necessarily apply validation
+            log.debug(JavaClassNames.niceClassAndMethod()+" got "+value+" for "+EntityRequiringConfigKeyInRange.RANGE+", now explicitly validating");
+            ConfigConstraints.assertValid(child);
+            fail("Expected exception when managing entity with incorrect config; instead passed assertion and got: "+value);
         } catch (Exception e) {
-            Throwable t = Exceptions.getFirstThrowableOfType(e, ConstraintViolationException.class);
-            assertNotNull(t, "Original exception was: " + Exceptions.collapseText(e));
+            e.printStackTrace();
```
sjcorbett: Did you intend to leave this in place?

brooklyn-server/core/src/test/java/org/apache/brooklyn/core/config/ConfigKeyConstraintTest.java
```diff
@@ -297,12 +299,17 @@ public void testContextAwarePredicateInformedOfEntity() {
     public void testQuickFutureResolved() {
         // Result of task is -1, outside of the range specified by the config key.
         try {
-            app.createAndManageChild(EntitySpec.create(EntityRequiringConfigKeyInRange.class)
+            EntityRequiringConfigKeyInRange child = app.createAndManageChild(EntitySpec.create(EntityRequiringConfigKeyInRange.class)
                     .configure(EntityRequiringConfigKeyInRange.RANGE, sleepingTask(Duration.ZERO, -1)));
-            fail("Expected exception when managing entity with incorrect config");
+            // may or may not fail above, depending on speed, but should fail if assert after forcing resolution
```
sjcorbett: I infer that the semantics of the behaviour under test should be improved.

ahgittin: I don't think it's that bad right now -- basically it asserts that any blocking config which resolves very quickly is valid, and those which don't resolve in time don't cause an error.  So understandable it would be sensitive to the speed of the environment.

However I can think of two improvements:
* Dependent configuration / subtasks have a "transitive-no-wait" mode where they are not permitted to block or sleep *except* if they are waiting on another "transitive-no-wait" task (similar to if they run such sub-tasks in the same thread but probably easier to leave the threading as is but somehow prevent waits, e.g. in transitive-no-wait mode we interrupt the thread at start, and temporarily clearing the thread if it is about to wait on another transitive-no-wait task) -- wdyt? /cc @aledsage
* When `getConfig` returns a coerced value it again asserts the validity of that value -- again wdyt @sjcorbett ?

brooklyn-server/core/src/test/java/org/apache/brooklyn/core/config/ConfigKeyConstraintTest.java
ahgittin: ouch now good spot i'll remove this

brooklyn-server/core/src/test/java/org/apache/brooklyn/core/config/ConfigKeyConstraintTest.java
sjcorbett: I think the second option would be too surprising. It may not be the caller's fault if the value is invalid but it's always the case when setting values.

PR #1176, opened 2016-01-27
 Generate config/effector descriptions for yaml-only blueprints

ivanayov:


tbouron:
LGTM.

One comment though. This generate a JS file which encapsulates the JSON data into a JS var. While this is ok for the documentation as it imports the file directly, it's a bit tedious to share the info with other services (i.e. community catalog) as we will need to `eval` this file to get the data.
I know it was not introduced by this PR but would it be better to get this opportunity to change the format from JS to JSON?

brooklyn-server/server-cli/src/main/java/org/apache/brooklyn/cli/ItemLister.java
```diff
@@ -98,88 +115,150 @@ public Void call() throws Exception {
...
+                result = ImmutableMap.<String, Object>builder()
+                        .put("entities", ItemDescriptors.toItemDescriptors(entityTypes, headingsOnly, "name"))
+                        .put("policies", ItemDescriptors.toItemDescriptors(policyTypes, headingsOnly, "name"))
+                        .put("enrichers", ItemDescriptors.toItemDescriptors(enricherTypes, headingsOnly, "name"))
+                        .put("locations", ItemDescriptors.toItemDescriptors(locationTypes, headingsOnly, "type"))
+                        .put("locationResolvers", ItemDescriptors.toItemDescriptors(ImmutableList.copyOf(ServiceLoader.load(LocationResolver.class)), true))
+                        .build();
+                jsonList.add(toJson(result));
+            } else if (!yaml.isEmpty()) {
+                LocalManagementContext lmgmt = new LocalManagementContext();
```
neykov: This will read config from the user's `brooklyn.properties`. Better initialize it with an empty state using `new LocalManagementContext(BrooklynProperties.Factory.newEmpty());`

brooklyn-server/server-cli/src/main/java/org/apache/brooklyn/cli/ItemLister.java
```diff
@@ -198,17 +277,36 @@ private void mkdir(String rootDir, String dirName) {
...
+                            for (File file : fileList) {
+                                if (file.isFile() && (file.getName().contains(".yaml") || file.getName().contains(".bom"))) {
+                                    urls.add(file.toURI().toURL());
+                                }
+                            }
+                        } else {
+                            urls.add(new File(yaml).toURI().toURL());
                         }
+                    } catch (IOException e) {
+                        e.printStackTrace();
```
neykov: Better not to catch it at all, add it to the method signature instead.

brooklyn-server/server-cli/src/main/java/org/apache/brooklyn/cli/ItemLister.java
```diff
@@ -198,17 +277,36 @@ private void mkdir(String rootDir, String dirName) {
...
         protected List<URL> getUrls() throws MalformedURLException {
             List<URL> urls = Lists.newArrayList();
-            if (jars.isEmpty()) {
-                String classpath = System.getenv("INITIAL_CLASSPATH");
-                if (Strings.isNonBlank(classpath)) {
-                    List<String> entries = Splitter.on(":").omitEmptyStrings().trimResults().splitToList(classpath);
-                    for (String entry : entries) {
-                        if (entry.endsWith(".jar") || entry.endsWith("/*")) {
-                            urls.addAll(ClassFinder.toJarUrls(entry.replace("/*", "")));
+            if (jars.isEmpty()) { //TODO: where is `jars` value write - AbstractMain.execCli(Cli<BrooklynCommand> parser, String ...args):line 250 BrooklynCommand command = parser.parse(args)
```
neykov: TODO probably not needed any more?

brooklyn-server/server-cli/src/main/java/org/apache/brooklyn/cli/ItemLister.java
```diff
@@ -98,88 +115,150 @@ public Void call() throws Exception {
...
-                    String enricherHtml = TemplateProcessor.processTemplateContents(enricherTemplateHtml, ImmutableMap.of("type", type, "name", name));
-                    Files.write(enricherHtml, new File(Os.mergePaths(outputFolder, "enrichers", type + ".html")), Charsets.UTF_8);
+                for (URL url: urls) {
+                    List<Map<?,?>> entities = new ArrayList<>();
+                    List<Map<?,?>> policies = new ArrayList<>();
+                    List<Map<?,?>> enrichers = new ArrayList<>();
+                    List<Map<?,?>> locations = new ArrayList<>();
+                    List<Map<?,?>> locationResolvers = new ArrayList<>();
+
+                    String yamlContent = Files.toString(new File(url.getFile()), Charset.forName("UTF-8"));
```
neykov: You could treat `url` as a generic URL, so the app can fetch remote catalogs as well. Changing it to something like `Streams.readFullyString(url.openStream)` will be independent of the location.

brooklyn-server/server-cli/src/main/java/org/apache/brooklyn/cli/ItemLister.java
```diff
@@ -98,88 +115,150 @@ public Void call() throws Exception {
...
+
+                    catalog.addItems(yamlContent);
+                    for (CatalogItem item: catalog.getCatalogItems()) {
+                        AbstractBrooklynObjectSpec<?,?> spec = catalog.createSpec(item);
+                        Map<String,Object> itemDescriptor = ItemDescriptors.toItemDescriptor((Class<? extends BrooklynObject>) spec.getType(), false);
+                        List<EntityConfigSummary> config = new ArrayList<>();
+
+                        if (!itemDescriptor.containsKey("name")) {
+                            itemDescriptor.put("name", item.getDisplayName());
```
neykov, line 166: The catalog item info should override the Java info, reverse the logic here.

brooklyn-server/server-cli/src/main/java/org/apache/brooklyn/cli/ItemLister.java
```diff
@@ -98,88 +115,150 @@ public Void call() throws Exception {
...
+                        }
+                        if (!itemDescriptor.containsKey("description")) {
+                            itemDescriptor.put("description", item.getDescription());
+                        }
+                        if (!itemDescriptor.containsKey("iconUrl")) {
+                            itemDescriptor.put("iconUrl", item.getIconUrl());
+                        }
+                        for (SpecParameter<?> param: spec.getParameters()) {
+                            EntityConfigSummary entityConfigSummary = EntityTransformer.entityConfigSummary(param.getType(),
+                                    param.getLabel(), null, MutableMap.<String,URI>of());
```
neykov: Apart for ordering `priority` can be used to decide on whether an item should be shown to the user.
Create a counter here, passing its value when `param.isPinned()`, `null` otherwise.

brooklyn-server/server-cli/src/main/java/org/apache/brooklyn/cli/ItemLister.java
```diff
@@ -98,88 +115,150 @@ public Void call() throws Exception {
...
+                            config.add(entityConfigSummary);
+                        }
+                        itemDescriptor.put("config", config);
+                        if (item.getCatalogItemType().name().equals("ENTITY")) {
```
neykov: `getCatalogItemType()` returns an enum - compare against it instead of using strings.

neykov: `TEMPLATE`s also fall in this case (they are entities).

brooklyn-server/server-cli/src/main/java/org/apache/brooklyn/cli/ItemLister.java
```diff
@@ -98,88 +115,150 @@ public Void call() throws Exception {
...
+                    mkdir(parentDir, "policies");
+                    mkdir(parentDir, "enrichers");
+                    mkdir(parentDir, "locations");
+                    mkdir(parentDir, "locationResolvers"); //TODO nothing written here yet...
+
+                    mkdir(parentDir, "style");
+                    mkdir(Os.mergePaths(parentDir, "style"), "js");
+                    mkdir(Os.mergePaths(parentDir, "style", "js"), "catalog");
+
+                    Files.write("var items = " + json, new File(Os.mergePaths(outputFolder, "items.js")), Charsets.UTF_8);
```
neykov: Wouldn't this overwrite the file for each catalog entry? Also no need to re-generate the whole doc structure for each item as it's identical. 

Maybe better to have a separate case for generating the output if coming from yaml, creating just the .js files? @tbouron wdyt?

brooklyn-server/server-cli/src/main/java/org/apache/brooklyn/cli/ItemLister.java
```diff
@@ -98,88 +115,150 @@ public Void call() throws Exception {
...
+                    result = ImmutableMap.<String, Object>builder()
+                            .put("entities", entities)
+                            .put("policies", policies)
+                            .put("enrichers", enrichers)
+                            .put("locations", locations)
+                            .put("locationResolvers", locationResolvers)
+                            .build();
+
+                    jsonList.add(toJson(result));
                 }
```
neykov, line 216: Remove the catalog items after processing them, so the next `addItems` call works with a clean catalog - avoiding conflicts.

brooklyn-server/server-cli/src/main/java/org/apache/brooklyn/cli/ItemLister.java
tbouron: I don't think it matters in this case @neykov as a blueprint author will use the `br list-objects` command to generate this file for his YAML blueprint then copy it onto his blueprint git repository

brooklyn-server/server-cli/src/main/java/org/apache/brooklyn/cli/ItemLister.java
ivanayov: It is not allowed now to parse jars and yaml at the same time. The idea was to prevent it and `jars` to have higher priority. Do you think it's better to throw exception after parsing the arguments, instead of making it impossible from the code?

neykov: Yes, could print an error and exit, but I'm fine with current behaviour. What I was referring to though is the "TODO" note on the right.

iyovcheva:
Jenkins fails due to `DynamicSequentialTaskTest.testCancelled` which I think is not related to the issue.  The test is successful locally.

bostko:
@tbouron @neykov if you had noticed docs were loading from a json file for a while.
However I noticed the experience wasn't so smooth on production and reverted it back.
When it is js file in the html it is automatically cached by the browser with no extra code.
https://github.com/apache/incubator-brooklyn/pull/622

brooklyn-server/server-cli/src/main/java/org/apache/brooklyn/cli/ItemLister.java
ivanayov, line 216: Fixed that. It's on another row and guess that's why it stays unchanged.

tbouron:
@bostko Didn't notice that. Let's not reintroduce a slow process then.

However, I quite like the @iyovcheva's idea of generate the `JS` **and** ` JSON` file. That will be much more safer and easier for the community catalog :)

ahgittin:
test failure is clearly unrelated, maybe close and reopen to run again.

meanwhile we should look at this /cc @aledsage :

```
org.apache.brooklyn.entity.software.base.test.autoscaling.AutoScalerPolicyNoMoreMachinesTest.testResizeDirectly (from TestSuite)

java.lang.AssertionError: removed=[EmptySoftwareProcessImpl{id=NhzxGta6}] expected [2] but found [1]
	at org.testng.Assert.fail(Assert.java:94)
	at org.testng.Assert.failNotEquals(Assert.java:494)
	at org.testng.Assert.assertEquals(Assert.java:123)
	at org.testng.Assert.assertEquals(Assert.java:370)
	at org.apache.brooklyn.entity.software.base.test.autoscaling.AutoScalerPolicyNoMoreMachinesTest.assertSize(AutoScalerPolicyNoMoreMachinesTest.java:182)
	at org.apache.brooklyn.entity.software.base.test.autoscaling.AutoScalerPolicyNoMoreMachinesTest.testResizeDirectly(AutoScalerPolicyNoMoreMachinesTest.java:105)
```

ahgittin:
(easily fixed - the test was doing an immediate assert on something happening in another thread; PR to follow; @iyovcheva give it an hour or so for the other fix to land then please close+reopen this to run the jenkins tests again)

PR #1178, opened 2016-01-30
fix failing (time sensitive) test

ahgittin:
solves the nondet test failure observed in #1176 .

and improve the "Asserts.eventually" routines which that used,
adding a new simpler eventuallyOnNotify(...) and having it used elsewhere.

also add convenience methods for CountdownTimer so its usage is more readable.


ahgittin:
seems to fix nondet test and checks pass so i'm going to merge; @aledsage or @geomacy (since you did a lot of the work on `Asserts`) could you review after my commit pls?

geomacy:
Looks good to me. One minor query on a comment.

PR #1179, opened 2016-01-30
fix the nondet cancellation test race observed previously

ahgittin:
the "fix" in #1171 didn't do everything, but provided enough clues to see the other culprit which this fixes, i think

### DIFF
--- a/brooklyn-server/core/src/test/java/org/apache/brooklyn/util/core/task/DynamicSequentialTaskTest.java
+++ b/brooklyn-server/core/src/test/java/org/apache/brooklyn/util/core/task/DynamicSequentialTaskTest.java
@@ -88,7 +88,11 @@ public class DynamicSequentialTaskTest {
     
     @AfterMethod(alwaysRun=true)
     public void tearDown() throws Exception {
-        if (em != null) em.shutdownNow();
+        if (em != null) {
+            // need to await termination, otherwise interrupted-but-still-running threads 
+            // may update the cancellations/messages and interfere with subsequent tests
+            Assert.assertTrue(em.shutdownNow(Duration.FIVE_SECONDS));
+        }
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


ahgittin:
CTR as this is simple test-only and it's interfering with other PRs.  

@aledsage this is one i'd really like you to be aware of, it's quite subtle and quite possibly could bite us elsewhere

brooklyn-server/core/src/test/java/org/apache/brooklyn/util/core/task/DynamicSequentialTaskTest.java
```diff
@@ -88,7 +88,11 @@ public void setUp() {
     
     @AfterMethod(alwaysRun=true)
     public void tearDown() throws Exception {
-        if (em != null) em.shutdownNow();
+        if (em != null) {
+            // need to await termination, otherwise interrupted-but-still-running threads 
+            // may update the cancellations/messages and interfere with subsequent tests
+            Assert.assertTrue(em.shutdownNow(Duration.FIVE_SECONDS));
```
aledsage: Personal preference to use `Asserts.DEFAULT_TIMEOUT` pretty much everywhere, unless a test is genuinely time-sensitive. I don't see any reason to decide on different length of time to wait when doing functional testing (i.e. when waiting for something to complete asyc).

aledsage:
@ahgittin the change looks sensible, but I don't understand exactly what it's fixing. I can see that it avoids interleaved logging from the old test's tearDown and then new test's run.

    // need to await termination, otherwise interrupted-but-still-running threads 
    // may update the cancellations/messages and interfere with subsequent tests

But this is in `DynamicSequentialTaskTest.tearDown`; the setUp + tearDown are marked `alwaysRun=true`; and there's only one beforeMethod/afterMethod. Therefore subsequent tests should be using an entirely new `BasicExecutionManager`. How could the previous test interfere with it (unless we're doing things with statics somewhere?!).

